PR #13, opened 2021-07-19
Fix flaky util test

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
This fixes a flaky test caused by the fact that python's `email.utils.formataddr` _conditionally_ quotes the address' name component if there's a character in there that needs to be escaped. This means that this test would fail a small portion of the time if the data generated for `User.name` contained a character that needed to be escaped, for example a period (`.`) in a person's salutation (e.g. "Mrs. Smith"). I updated the tests to instead parse the result and make sure they match the expected fields on `User`.

Tests should pass, if you want to try to reproduce this on `main` you can run `pytest users/utils_test.py -k test_format_recipient
` a few times and you should eventually hit a failure case. Running it repeatedly on this branch should always succeed.

### DIFF
--- a/users/utils_test.py
+++ b/users/utils_test.py
@@ -1,4 +1,5 @@
 """User utils tests"""
+from email.utils import parseaddr
 from unittest.mock import patch
 
 import pytest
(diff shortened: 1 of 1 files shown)


gsidebo (approved):
LGTM 👍

Issue #16, opened 2021-07-19
Implement home page product section


gsidebo:
Related: https://github.com/mitodl/mitxonline/issues/14
Desktop design: https://projects.invisionapp.com/share/KH11D3ZN53GC#/screens/455419669
Mobile design: https://projects.invisionapp.com/share/KH11D3ZN53GC#/screens/455488637

PR #11 added Wagtail and some initial barebones models for the home page. This issue covers the work to implement the product/courses section with the individual cards (headed by the "Duis aute" text in the designs)

**Acceptance criteria:**
- Products can be added to home page product section, and any product type can be chosen (to start out, the only valid product type will be `CoursePage`)
- When fully configured in Wagtail, the feature product section of the home page matches the designs in desktop and mobile screen widths

**Not needed:**
- Courses (products) do not need to link to anything  
- Header section with MITx Online logo
- Footer

**Implementation details:**
- Updated product page models (`ProductPage`, `CoursePage`) were added in PR #28 (still open as of 7/26)
- We will not be using Wagtails views and template tags to serve images. We're going to expect that images are optimized and correctly sized when they're uploaded, and the `wagtail_img_src` template tag can be used to get the image source URL
  - You can use services like [img.pizza](https://img.pizza/) or [placekitten](https://placekitten.com/) to create exactly-sized images if you don't want to chop up your own images
- The `HomePage` field that stores the "Duis aute" text in the designs can be called `product_section_title 
- In Wagtail, the field name of this section in the home page should be something like `featured_products`
- We should be able to add any number of these product cards  to the home page, and to add one you should only need to select a `ProductPage` object 
  - In terms of UX, this should work like FAQ page in mitxpro ([RC link](https://xpro-rc.herokuapp.com/cms/pages/151/edit/)) using an inline panel
  - [This Wagtail doc](https://docs.wagtail.io/en/stable/reference/pages/panels.html#inline-panels-and-model-clusters) should help guide the implementation of this inline panel so that we can select any object of any model that inherits from `ProductPage`
- Taking the "Microeconomics" card as an example: "Microeconomics" is the `Course.title` field, the image is the `ProductPage.feature_image` field, the "Use economic model..." text is the `ProductPage.description` field, and the start date should be `CourseRun.start_date` of the related course run with the closest future start date. If there is no course run with a future start date, it should be blank
- If the screen is too narrow to display all of the courses/products, they should flow to a new line and the cards should all be vertically and horizontally aligned in a grid

[...]

PR #35, opened 2021-07-26
add homepage header hero contents

arslanashraf7:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [x] Mobile width screenshots
- [x] Migrations
  - [x] Migration is backwards-compatible with current production code
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
fixes #14 

#### What's this PR do?
- Adds the design for the HomePage hero image
- Adds title and subtitle fields in HomePage
- Fixes a crash when there is no Hero Image added in CMS
- Adds design for title and subtitles on HomePage

**Note:**
- Font Used: Roboto, I think we will need a purchased `Avenir Next` font?
- Titles left spacing is based on the container to keep it responsive with the overall website
- Kept title/subtitle fields optional so if they aren't added in CMS there will be no title/subtitle section over Hero image.

#### How should this be manually tested?
- Set up the mix online
- Run `configure_wagtail` command and make sure that Hero title/subtitle is added to Home Page
- Check that the Home Page loads fine and doesn't crash if there is no Hero image set up in CMS
- Check that the design matches

#### Where should the reviewer start?
- Home Page

#### Screenshots (if appropriate)

**Desktop**
Screenshot 2021-07-29 at 11 39 32 AM

**Mobile**
Screenshot 2021-07-29 at 11 40 50 AM

**Wagtail UI & Image dimensions helper message**
Screenshot 2021-07-29 at 11 39 12 AM

**Wagtail UI Image upload dialog**

Screenshot 2021-07-29 at 11 41 11 AM

PR #38, opened 2021-07-28
home page product section

HamzaIbnFarooq:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested

Closes https://github.com/mitodl/mitxonline/issues/16

#### What's this PR do?
Implements home-page product cards section according to https://projects.invisionapp.com/d/main/#/console/21535972/455419669/preview

**Desktop view**:
image

**Desktop View with Hover**:
image

**Mobile view**:
image


cms/models.py
```diff
@@ -30,10 +30,62 @@ class HomePage(Page):
     ]
 
     def get_context(self, request, *args, **kwargs):
-        return {
-            **super().get_context(request, *args, **kwargs),
-            "hero": self.hero,
-        }
+        context = super().get_context(request)
+        context.update({
```
HamzaIbnFarooq, line 34: This is temporary data until we decide a specific way of implementing product cards backend.

asadiqbal08:
@gsidebo can you review the PR, I jumped over the Hamza's work and Tried to update it as per Invision Design ? 



screencapture-localhost-8013-data-analysis-for-social-scientist-2021-08-05-18_11_52

main/templates/base.html
```diff
@@ -9,10 +9,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="icon" href="{% static 'images/favicon.ico' %}" />
     {% include "partials/gtm_head.html" %}
-        <link rel="stylesheet" type="text/css"
-              href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,600,700|Montserrat:300,400,500,600,700&display=swap"
-              media="all"/>
-        <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons&display=swap"
+    <link rel="stylesheet" type="text/css" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,600,700|Rajdhani:300,400,500,600,700|Montserrat:300,400,500,600,700:&display=swap"
+          media="all" />
+    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons&display=swap"
```
gsidebo, line 14: Not needed

static/scss/course-cards.scss
```diff
@@ -0,0 +1,106 @@
+.course-section-header {
```
gsidebo, line 1: 1. I don't think we need this class. You can add a class to the `h1` and apply any necessary spacing to that class
2. Please refer to my [comment on the other PR](https://github.com/mitodl/mitxonline/pull/45#discussion_r681984316) about header tags (`h1`, `h2`, etc). Please make absolutely sure that the custom styles are necessary. If the styling of this h1 in the designs is anywhere close to the standard h1/h2 styles that we have for attributes like size, weight, color, font, etc., please remove those styles here.
3. Please get rid of all `font-family: Montserrat` definitions, and make sure that all these `font-size` definitions are necessary. Just in general I think a lot of these styles might not need to be here

static/scss/layout.scss
```diff
@@ -16,11 +16,13 @@
...
 @import "expandable";
 @import "dashboard";
+@import "course-cards";
 @import "home";
 @import "product-details";
 
+
 body {
-  font-family: Montserrat, "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-family: Muli, "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif, Montserrat;
```
gsidebo, line 25: Not needed

cms/models.py
```diff
@@ -47,8 +81,94 @@ class HomePage(Page):
...
+        return {
+            **super().get_context(request),
+            "product_section_title": self.product.product_section_title
+            if self.product
+            else "Product Section Title",
+            "product_cards": self.product.product_cards if self.product else None,
+        }
+
+
+class HomeProductPage(Page):
```
gsidebo, line 111: Not sure if you tried this already and it just didn't work out, but it makes much more sense and will be much more user friendly if this is not a child page of the home page, but a part of the home page itself. If we can find a way to have `product_section_title` as a field in `HomePage` and the `featured_products` inline panel also defined as a part of `HomePage`. Users are going to think of these elements as part of the home page and it would be best for them to see these inputs when they're editing the home page itself

cms/models.py
```diff
@@ -47,8 +81,94 @@ class HomePage(Page):
...
+        Gets the "HomeProductPage" section sub page.
+        """
+        return self._get_child_page_of_type(HomeProductPage)
+
+    def get_context(self, request, *args, **kwargs):
+        return {
+            **super().get_context(request),
+            "product_section_title": self.product.product_section_title
+            if self.product
+            else "Product Section Title",
```
gsidebo, line 106: This should be removed. If the section title hasn't been set, we shouldn't show a title at all. We don't get any value from this default

gsidebo (changes requested):

cms/models.py
asadiqbal08: screencapture-localhost-8013-cms-pages-6-edit-2021-08-06-13_00_53

asadiqbal08: @gsidebo is it good now ?

cms/templates/home_page.html
```diff
@@ -10,16 +11,29 @@
...
-          {% endif %}
-          {% if page.hero_subtitle %}
-            <h2>{{ page.hero_subtitle }}</h2>
-          {% endif %}
-        </div>
-      </div>
-    </div>
-{% endblock %}
\ No newline at end of file
+<div class="course-section">
```
gsidebo, line 14: You're completely removing a section of the page here (see `hero-container` in the removed part of the diff). This is a section that was added recently, it's outlined in the designs, and it needs to be there.

static/scss/layout.scss
```diff
@@ -16,11 +16,13 @@
...
 @import "expandable";
 @import "dashboard";
+@import "course-cards";
 @import "home";
 @import "product-details";
 
+
 body {
-  font-family: Montserrat, "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-family: "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif, Montserrat;
```
gsidebo, line 25: Not needed

static/scss/course-cards.scss
```diff
@@ -0,0 +1,105 @@
...
+      height: 223px;
+      display: block;
+      object-fit: cover;
+    }
+  }
+}
+
+.course-section {
+  background: #f8f8f8;
+}
```
gsidebo, line 105:
```suggestion
```

This isn't needed. The background is that color in the course section of the home page in the designs because there is a whole-page gradient that fades down to that color.

I just tested this locally and it seems to work. Change the `background` style of `body` in `layout.scss` to this, and we should get roughly the correct gradient, and it should fade to the correct color by the time the course section is rendered: 

`background: linear-gradient(to bottom, #ffffff, #f5f5f5 200px);`

static/scss/course-cards.scss
```diff
@@ -0,0 +1,105 @@
...
+    @include media-breakpoint-down(sm) {
+      width: calc(100% - 44px);
+    }
+    letter-spacing: normal;
+    display: inline-block;
+    vertical-align: top;
+    width: calc(33.33% - 44px);
+    margin: 0 22px 44px;
+
+    a {
```
gsidebo, line 46:
```suggestion
    .course-card {
```

static/scss/course-cards.scss
```diff
@@ -0,0 +1,105 @@
...
+      background: white;
+      border-bottom: 3px solid transparent;
+      box-shadow: 0 1px 0 0 rgba(0, 0, 0, 0.23);
+      display: block;
+      text-decoration: none;
```
gsidebo, line 51: Not needed
```suggestion
```

static/scss/course-cards.scss
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.course-listing {
+  padding: 0;
+  margin: 0 -22px;
+  list-style: none;
+  letter-spacing: -0.32em;
+
+  .course-card-description {
+    overflow-y: scroll;
```
gsidebo, line 28: I don't know if we want internal scrollbars here to begin with, but this is making an empty scrollbar show up in Mac Chrome when the contents do not overflow. To get rid of that, this needs be be `overflow-y: auto;`, but I think it would be better just to set a `min-height` for the text section and let it grow vertically if the text is too long.

static/scss/course-cards.scss
```diff
@@ -0,0 +1,105 @@
...
+  padding-bottom: 54px;
+  margin: 0 auto;
+  max-width: 760px;
+
+  @include media-breakpoint-down(sm) {
+    font-size: 30px;
+    line-height: 40px;
+    padding-top: 20px;
+    padding-bottom: 22px;
+  }
```
gsidebo, line 18: This can be simplified. Our first thought should always be to match existing styles even if the design suggests something slightly different. It gives us less css to maintain, less chance of UI bugs, and we can easily change after a designer review

```suggestion
  text-align: center;
  padding: 50px 10%;
  margin: 0;

  @include media-breakpoint-down(sm) {
    font-size: 30px;
    padding: 20px 0;
  }
```

cms/models.py
```diff
@@ -40,17 +72,77 @@ class HomePage(Page):
...
+        on_delete=models.SET_NULL,
+        related_name="+",
+    )
+
+    panels = [
+        PageChooserPanel("course_product_page", "cms.CoursePage"),
+    ]
+
+
+class CourseIndexPage(CourseObjectIndexPage):
```
gsidebo, line 138: We're probably going to introduce this change in a future PR, but is there any reason it's necessary here? When we do introduce it, we're probably going to add this index page if it doesn't exist, and introduce a data migration to move all existing course pages to be children of it, just like we did in xpro. If this isn't necessary for this PR, please remove it

cms/models.py
```diff
@@ -40,17 +72,77 @@ class HomePage(Page):
...
+            if page.course_product_page:
+                pages.append(page.course_product_page.specific)
+        return pages
+
+    def get_context(self, request, *args, **kwargs):
+        return {
+            **super().get_context(request),
+            "product_cards_section_title": self.product_section_title,
+            "product_cards": self.product_cards,
+        }
```
gsidebo, line 113: Like I mentioned in the product detail page PR, it's strongly preferred to do calculation and deep variable access in the model and pass it down to the template whenever we can. Suggestions here also have some small variable name changes:

```python
    @property
    def products(self):
        page_data = []
        for page in self.featured_products.all():
            if page.course_product_page:
                product_page = page.course_product_page.specific
                page_data.append({
                    "title": product_page.title,
                    "description": product_page.description,
                    "feature_image": product_page.feature_image,
                    "start_date": product_page.product.first_unexpired_run
                })
        return page_data

        return {
            **super().get_context(request),
            "product_section_title": self.product_section_title,
            "products": self.products,
        }
```

cms/templates/home_page.html
```diff
@@ -10,16 +11,29 @@
...
+              {{ card.description |richtext }}
+            </div>  
+            <span class="date">
+              Start date: {{ card.product.first_unexpired_run.start_date |date:"F j, Y" }}
+            </span>
+          </div>
+        </a>
+      </li>
+      {% endfor %}
+    </ul>
```
gsidebo, line 36: Corresponding change to the suggestion I made for the model `get_context` method above. A couple extra changes I made:
- I changed the `<a>` to `<div class="course-info">` because this element isn't linking to anything, so it shouldn't be a link. You can update the css selector accordingly
- I removed the "start date" section if there is no start date to show, which is something we should _always_ assume is the desired behavior

```suggestion
    <ul class="course-listing">
      {% for product in products %}
      <li>
        <div class="course-card">
          <img src="{% wagtail_img_src product.feature_image %}" alt="Featured course image">
          <div class="course-info">
            <h2 class="course-card-title">{{ product.title }}</h2>
            <div class="course-card-description">
              {{ product.description |richtext }}
            </div>
            {% if product.start_date %}
              <span class="date">
                Start date: {{ product.start_date | date:"F j, Y" }}
              </span>
            {% endif %}
          </div>
        </div>
      </li>
      {% endfor %}
    </ul>
```

gsidebo (changes requested):

cms/models.py
```diff
@@ -40,17 +72,77 @@ class HomePage(Page):
         help_text="The subtitle text to display in the hero section of the home page.",
     )
 
+    product_section_title = models.CharField(
+        max_length=255,
+        null=True,
+        blank=True,
+        help_text="The tile text to display in the product cards section of the home page.",
```
gsidebo, line 79:
```suggestion
        help_text="The title text to display in the product cards section of the home page.",
```

static/scss/course-cards.scss
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.course-listing {
+  padding: 0;
+  margin: 0 -22px;
```
gsidebo, line 23: This PR introduces a space at the right side of the page in the mobile view, and I think it is a result of this negative margin. Can you look into this?

ss 2021-08-06 at 15 44 18

gsidebo (changes requested):

cms/models.py
asadiqbal08: I can remove it but at the moment it seems me mess when I saw a lot of `CoursePage` under `HomePage`. Ok make sense to do it separate PR but it should be.

static/scss/course-cards.scss
asadiqbal08: So we are now using `div` instead of `<a>`, let me know, is it not be clickable to redirect it to specific product ?

cms/templates/home_page.html
asadiqbal08: If it is not the link then How the user be able to get the product detail page ?

static/scss/course-cards.scss
asadiqbal08: space seems same spacing on both side. I did not notice the difference. 

screencapture-localhost-8013-homepage-2021-08-09-14_30_42

cms/templates/home_page.html
asadiqbal08: Did not get this point.

gsidebo: @asadiqbal08 This PR: https://github.com/mitodl/mitxonline/pull/35/files#diff-153077c90d135769287ff8c6025f219227d11fa426cc1da35b348a14050fe3d7

I know this PR was handed off to you and it might not have been you that rebased this on the main branch. Basically in the process of merging this with the code in master, all of the hero section code has been deleted from this template. Everything you see in that PR needs to remain in this template

gsidebo (changes requested):

cms/templates/home_page.html
gsidebo: @asadiqbal08 Sorry, I got this confused with the faculty section which will not have links. We have the product detail page now, so you should link to it from these cards. I think you should be able to use Wagtail's `pageurl` for this (https://docs.wagtail.io/en/stable/topics/writing_templates.html#pageurl)

static/scss/course-cards.scss
gsidebo: @asadiqbal08 I figured out what was happening. For some reason, when that negative margin is applied, I'm able to scroll very slightly to the right in the mobile view, and that's what causes the slight space to the right of the header and the main body of the page. I don't think it's likely that users will see this, so I'm fine leaving it as it is for now, and revisiting later if we see any issues during testing

cms/models.py
```diff
@@ -40,17 +48,76 @@ class HomePage(Page):
...
+        page_data = []
+        for page in self.featured_products.all():
+            if page.course_product_page:
+                product_page = page.course_product_page.specific
+                page_data.append(
+                    {
+                        "title": product_page.title,
+                        "description": product_page.description,
+                        "feature_image": product_page.feature_image,
+                        "start_date": product_page.product.first_unexpired_run.start_date,
```
gsidebo, line 87: Sorry, I made a mistake in my suggestions here earlier. `first_unexpired_run` can return None, so this needs to be updated to handle that possiblity:

```suggestion
                product_page = page.course_product_page.specific
                run = product_page.product.first_unexpired_run
                page_data.append(
                    {
                        "title": product_page.title,
                        "description": product_page.description,
                        "feature_image": product_page.feature_image,
                        "start_date": run.start_date if run is not None else None,
```

cms/models.py
```diff
@@ -115,7 +182,7 @@ class Meta:
         FieldPanel("what_you_learn"),
         FieldPanel("feature_image"),
     ]
-    parent_page_types = ["HomePage"]
+
```
gsidebo, line 185: I don't see the need for this change. We can keep `ProductPage`/`CoursePage` as a child page of `HomePage` until we implement the course index page properly. If there's no immediate need to change that hierarchy, we don't need this change as part of this PR

gsidebo (changes requested):
4 last things and I think this should be ready to go:

1. Rebase on `main`
2. Add the home page hero section back in
3. Address the comments I made in `cms/models.py`
4. Implement the links to the product detail page (https://github.com/mitodl/mitxonline/pull/38#discussion_r685439738
5)

cms/models.py
asadiqbal08: @gavin if we move the product page to child of just  homepage then while creating Homepage in CMS you cannot linked products then so good to keep on page level right now.

umarmughal824:
@gsidebo this is how it looks like now

Screenshot 2021-08-10 at 18 59 18
Screenshot 2021-08-10 at 18 59 21

gsidebo (approved):
Looks good! 👍

Issue #32, opened 2021-07-23
Implement basic product detail page


gsidebo:
**Depends on #16**
Design (desktop): https://projects.invisionapp.com/d/main/#/console/21535972/455477372/preview
Design (mobile): https://projects.invisionapp.com/d/main/#/console/21535972/455488639/preview

PR #11 added Wagtail and some initial barebones models for the home page. We want implement the product detail page in Wagtail and render it in a way that matches the designs (for our first version of this, course runs will be the only product, and a single page will exist for each course no matter how many runs it has, so practically this can just be referred to as the course detail page for now)

**Acceptance criteria:**
- [ ] Fields added to the `ProductPage` (implemented in #28) to store new data for course pages: `length` (text), `effort`, (text), `prerequisites` (text), ~~`long_description` (rich text)~~ about (text), `what_you_learn` (text)
  - [ ] Unlike xPRO, `price` should be a wagtail field where the CMS user can enter text or a link (as it appears in [InVision](https://invis.io/RN11DSJOGXFD#/455477372_Course_Detail)). This will give a MicroMasters program like DEDP a place to explain their personal pricing scheme. 
- [ ] `CoursePage`s can be created that match the designs in desktop and mobile views
- [ ] Courses featured on the home page link to the relevant product detail page 

**Out of scope:**
- Video (a simple placeholder div can be used instead of a video in the template)
- Faculty section (separate issue)
- The "Enroll now" button does not need to do anything (separate issue – https://github.com/mitodl/mitxonline/issues/33)
- ~~Price field. Our first courses in mitxonline will be free and we do not need to add any db models/fields to keep track of price~~
- Expand/collapse functionality for sections of text (indicated by "More..." in the designs)

**Implementation details:**
- Using the [design](https://projects.invisionapp.com/d/main/#/console/21535972/455477372/preview) as an example, the "About this course" and "What you'll learn" contents will be represented by the `about` and `what_you_learn` fields. The titles of those sections can be hard-coded into the template.

HamzaIbnFarooq:
Need view permissions on invisionapp previews

pdpinch:
The course team has requested that instead of omitting Price, we include a link under Price, with the link text and the link configurable in the CMS. (The text will probably be `Varies` but I'm not sure the link will go yet). 

To be prepared for future courses with actual prices, the CMS model should also allow values like "$150". 

I've updated the issue description to reflect this.

PR #45, opened 2021-07-30
asadiqbal08/Basic Product Detail Page

asadiqbal08:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested

fixes: #32 

#### What's this PR do?
Design the basic product detail page.

screencapture-localhost-8013-homepage-data-analysis-for-social-scientist-2021-08-04-17_26_56

Screen Shot 2021-08-03 at 4 52 40 PM

screencapture-localhost-8013-homepage-data-analysis-for-social-scientist-2021-08-04-17_27_40

## When price link is not there:

screencapture-localhost-8013-homepage-seed-digital-learning-course-2021-08-03-18_46_28


cms/migrations/0003_product_detail_page.py
```diff
@@ -0,0 +1,58 @@
...
+                blank=True,
+                help_text="A short description indicating how long it takes to complete (e.g. '4 weeks').",
+                max_length=50,
+                null=True,
+            ),
+        ),
+        migrations.AddField(
+            model_name="coursepage",
+            name="prerequisites",
+            field=models.CharField(
```
pdpinch, line 44: Can we make this a rich text field? It seems like courses on edX have bullet lists and links in their prerequisites. (I can't remember if rich text fields support links).

from https://www.edx.org/course/microeconomics:

image

pdpinch (changes requested):
I think the price field is missing. Here it is in the (current) design:

image

Unlike xPRO, price should be a wagtail field where the CMS user can enter text or a link (as it appears in InVision).

cms/migrations/0003_product_detail_page.py
asadiqbal08: I am going to add the price but I did not get what `Various` is pointing out and how to show the price if not having field in CMS  ?

asadiqbal08: Field is richtext so can be used something like that. 


screencapture-localhost-8013-homepage-data-analysis-for-social-scientist-2021-08-02-15_43_46

cms/templates/product_page.html
```diff
@@ -10,7 +11,72 @@
...
+              </div>
+            </div>
+            <div class="stat-col">
+              <span class="title">EFFORT</span>
+              <div class="text">
+                <strong>{{ page.effort }}</strong>
+              </div>
+            </div>
+            <div class="stat-col small">
+              <span class="title">PRICE</span>
```
pdpinch, line 55: Can we capitalize the text in CSS instead of in the HTML?

cms/templates/product_page.html
```diff
@@ -10,7 +11,72 @@
...
+              <div class="text">
+                <strong>Varies</strong>
```
pdpinch, line 57: This text should be managed in the CMS

static/scss/product-details.scss
```diff
@@ -0,0 +1,257 @@
+.container {
+  max-width: 1260px;
+  padding: 0 15px;
+  box-sizing: border-box;
+  margin: 0 auto;
+}
+
+.hero-block {
+  font-family: 'Roboto', sans-serif;
```
pdpinch, line 9: I thought we were moving to Montserrat? (not sure I spelled that right)

cms/templates/product_page.html
asadiqbal08: Thanks for answering this https://github.com/mitodl/mitxonline/pull/45#discussion_r680813560

static/scss/product-details.scss
asadiqbal08: Ok, I need to remove it from here , I guess, in one of the Arslan's PR it is added in base level.

asadiqbal08:
@pdpinch I have updated the screenshots. Let me know if do you think to change anything ? 
Sorry, I have to squash the commits after rebased as many things need to push.

pdpinch:
The price link should be optional. For MVP we will have a link, but after MVP some courses will have simple string for price like "$150." I think it's best to set it to optional now.

cms/templates/product_page.html
```diff
@@ -10,7 +11,72 @@
...
+              </div>
+            </div>
+            <div class="stat-col small">
+              <span class="title">price</span>
+              <div class="text">
+                <strong><a target="_blank" href={{page.price.0.value.link}}>{{page.price.0.value.text}}</a></strong>
+              </div>
+            </div>  
+            <div class="stat-col large">
+              <span class="title">prerequisite</span>
```
pdpinch, line 61:
```suggestion
              <span class="title">prerequisites</span>
```

should match all the design

pdpinch:
Small thing, but in the design, the width of the "prerequisites" section is the same as the width of the video player. Is there a reasonable way to get them to line up? If it's too hard, we can leave it for a follow up.

pdpinch (changes requested):

asadiqbal08:
> Small thing, but in the design, the width of the "prerequisites" section is the same as the width of the video player. Is there a reasonable way to get them to line up? If it's too hard, we can leave it for a follow up.

> Out of scope:
    Video (a simple placeholder div can be used instead of a video in the template)


@pdpinch , In the screenshot, There is not any video but just an image holder. What do you think if we update it when start working on the video area there ?

pdpinch:
>What do you think if we update it when start working on the video area there ?

sure

static/scss/product-details.scss
```diff
@@ -0,0 +1,260 @@
...
+    }
+  }
+
+  .text-info {
+    @media (min-width: 1024px) {
+      max-width: 625px;
+    }
+  }
+
+  h1 {
```
gsidebo, line 69: We should try to avoid specialized styles for different header elements if we can. Does this really need a different font-size, line-height, or font-weight definition from the base styles that we have defined for `h1`? As far as I can tell, this is a header that should look like all other headers of the same size/type

static/scss/product-details.scss
```diff
@@ -0,0 +1,260 @@
...
+    font-size: 42px;
+    line-height: 50px;
+    margin: 0 0 13px;
+    font-weight: 700;
+
+    @media (min-width: 768px) {
+      font-size: 48px;
```
gsidebo, line 76: Adding to my comment above: if the designs call for the most prominent headers to be a certain size in mobile or desktop widths, those styles should be applied to the `h1` definition in `common.scss` unless this header is truly exceptional for some reason

static/scss/product-details.scss
```diff
@@ -0,0 +1,260 @@
...
+    transition: linear 0.4s;
+
+    &:hover {
+      opacity: 0.8;
+    }
+  }
+}
+
+.info-block {
+  font-family: 'Montserrat', sans-serif;
```
gsidebo, line 109: Not needed

static/scss/product-details.scss
```diff
@@ -0,0 +1,260 @@
...
+  font: 16px/28px 'Montserrat', sans-serif;
+  color: $black;
+  padding: 20px 0;
+
+  @media (min-width: 1024px) {
+    padding: 30px 0;
+  }
+
+  h2 {
+    font-size: 28px;
```
gsidebo, line 215: Please refer to my comment above about styling consistency with these header elements

static/scss/product-details.scss
```diff
@@ -0,0 +1,260 @@
...
+      }
+
+      &.large {
+        flex-grow: 1;
+        text-align: left;
+        }
+      }
+    }
+
+  .text {
```
gsidebo, line 163: Can you wrap these `.text` and `.title` definition in a `.stats-col` selector? "text" and "title" are very general terms and the extra selector would help to understand what they're actually styling

static/scss/product-details.scss
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    a:link {
+      color: $link-blue;
+      text-decoration: underline;
+    }
+  }
+
+  .title {
+    display: block;
```
gsidebo, line 184: If we're using `display: block` and setting width to 100%, any reason we wouldn't just use a `div` instead of a `span`?

static/scss/product-details.scss
```diff
@@ -0,0 +1,260 @@
...
+    color: $black;
+
+    &:hover {
+      text-decoration: none;
+    }
+  }
+}
+
+.text-block {
+  font: 16px/28px 'Montserrat', sans-serif;
```
gsidebo, line 206: Font family definition is not needed

gsidebo (changes requested):
To add to the code comments: the `small` style applied to these elements is making the font sizes different between the titles. I don't think we want that

ss 2021-08-03 at 13 09 21

static/scss/product-details.scss
```diff
@@ -0,0 +1,260 @@
...
+      font-size: 48px;
+      line-height: 60px;
+    }
+  }
+
+  p {
+    margin: 0 0 28px;
+  }
+
+  .btn.btn-default {
```
gsidebo, line 85: One more comment: wherever possible we should be looking for styles that can be reused in other parts of the app. The "Enroll Now" button in the design has the same styling as many other buttons in other pages. You would expect any website to have some consistent styling for commonly-used elements like buttons, headers, borders, etc. This selector will style one single button in the product detail page, but looking at the designs, we know that this look and feel will be needed for other buttons. We should move these styles to a css class that any other button element can use (`.action-btn` or something like that)

gsidebo (changes requested):

asadiqbal08:
Updated the Screenshots

static/scss/product-details.scss
```diff
@@ -0,0 +1,218 @@
+// sass-lint:disable mixins-before-declarations
+.container {
+  max-width: 1260px;
+  padding: 0 15px;
+  box-sizing: border-box;
+  margin: 0 auto;
```
gsidebo, line 6:
```suggestion
  box-sizing: border-box;
```

1. I don't think the `padding` or `margin` rules are necessary here. I disabled both in mobile and desktop widths and it doesn't seem to have an effect
2. If this page needs to have a max-width that is greater than the standard width for bootstrap `container` elements, then that should be applied to _all_ top-level `container` elements. I think we should hold off on setting a new max-width for all elements and see what Ferdi, Peter, and the designer have to say. That can be done in a separate PR

gsidebo (approved):
I made one final [comment](https://github.com/mitodl/mitxonline/pull/45#discussion_r682738693) suggesting a simple css change. After you apply that, feel free to merge! 👍

PR #66, opened 2021-08-06
fix wagtail media upload error

umarmughal824:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

NA

#### What's this PR do?
fix 502 error while uploading images on wagtail

I have tested it by adding AWS and S3 keys locally


requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.8
+# This file is autogenerated by pip-compile with python 3.9
```
pdpinch, line 2: Just checking -- this app _is_ on python 3.9, right?

umarmughal824: yeah, it's running on python 3.9

pdpinch (approved):
Seems necessary, and what harm can it do?

requirements.txt
umarmughal824: https://github.com/mitodl/mitxonline/blob/main/Dockerfile#L1

umarmughal824:
> Seems necessary, and what harm can it do?

no harm at all

PR #71, opened 2021-08-07
Added styling to pin footer to the bottom of the page

### LINKED ISSUE #70: Ensure that footer renders at the bottom of page
When there is little content in the body of the page, the footer is rendered directly beneath that content and leaves a significant space underneath. The footer should render at the bottom of the page when there is not enough content to fill the screen, but should not be pinned to the bottom regardless of scroll position

ss 2021-08-06 at 16 51 35

gsidebo:
#### Pre-Flight checklist

  - [x] Desktop screenshots

Closes #70

#### What's this PR do?
Adds styling to pin footer to the bottom of the page, whether there is enough content to push the footer down or not

View any page in the site and fill in/remove content to make sure that the footer is always at the bottom of the page

Before:
ss 2021-08-06 at 16 51 35

After:
ss 2021-08-06 at 22 03 11

### DIFF
--- a/static/scss/layout.scss
+++ b/static/scss/layout.scss
@@ -23,7 +23,10 @@ body {
   font-family: Montserrat, "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
   margin: 0;
   color: #000000;
-  background: linear-gradient(to bottom, #ffffff, #f5f5f5 200px);
+  background: linear-gradient(to bottom, #ffffff, #f5f5f5 250px);
+  display: flex;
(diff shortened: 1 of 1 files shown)


arslanashraf7 (approved):
Looks good 👍 

I also tested it within multiple pages with less/more content and the footer works as expected.

There is something I noticed which definitely isn't related to this PR or the associated issue but there is a difference of `2px` in address font size and `1px` in copyright and links. I just added this comment for you to decide if we should take care of it in a new issue or maybe or not to worry about it.

PR #81, opened 2021-08-11
Added course index page

gsidebo:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [x] Mobile width screenshots
- [x] Migrations
  - [x] Migration is backwards-compatible with current production code
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Closes #79

#### What's this PR do?
Adds a course index page, adds a data migration that moves existing course pages under that index page, and changes the URL rules such that course pages are now accessed at `/courses/<readable_id>/`

#### How should this be manually tested?
- Run migrations, and make sure all of your course pages are children of the course index page, which is a child of the home page
- Add a new course page under the course index page in Wagtail to make sure that works
- Confirm that the URLs for course pages are as described

#### Screenshots (if appropriate)
ss 2021-08-10 at 20 40 46

ss 2021-08-10 at 22 45 31

PR #88, opened 2021-08-12
fix home page feature products section

umarmughal824:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

The course page was not loading so it's just a quick fix of it

#### What's this PR do?
Fix course page access (direct link or through home page feature section)

Add courses and choose the course as a featured product on the home page

Screenshot 2021-08-12 at 14 50 20
Screenshot 2021-08-12 at 14 50 30
Screenshot 2021-08-12 at 14 57 19

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -65,6 +65,7 @@ class HomePage(Page):
 
     parent_page_types = [Page]
     subpage_types = [
+        "CoursePage",
         "ResourcePage",
     ]
 
@@ -86,7 +87,7 @@ class HomePage(Page):
                         "description": product_page.description,
                         "feature_image": product_page.feature_image,
                         "start_date": run.start_date if run is not None else None,
-                        "url_path": product_page.url_path,
+                        "url_path": product_page.slug,
                     }
                 )
         return page_data
@@ -193,7 +194,7 @@ class CoursePage(ProductPage):
     Detail page for courses
     """
 
-    parent_page_types = [Page]
+    parent_page_types = [HomePage]
 
     course = models.OneToOneField(
         "courses.Course", null=True, on_delete=models.SET_NULL, related_name="page"


umarmughal824:
small fix for https://github.com/mitodl/mitxonline/pull/38

cms/models.py
```diff
@@ -193,7 +194,7 @@ class CoursePage(ProductPage):
     Detail page for courses
     """
 
-    parent_page_types = [Page]
+    parent_page_types = [HomePage]
```
asadiqbal08, line 197: @umarmughal824 By making it child of HomePage, user be not able to associate `ProductDetails` while creating the `HomePage`

asadiqbal08 (changes requested):

cms/models.py
asadiqbal08: I mean, If there is not any product page exist then he cannot create it until not complete the HomePage creation and HomePage will not be created if not any Product is associated.

umarmughal824: the course pages are not working by making it a direct child of Page. You could see the attached screenshots of working attached products with the home page.

gsidebo: @asadiqbal08 @umarmughal824 Please review this PR: https://github.com/mitodl/mitxonline/pull/81. I created this PR because we were going to create this index page anyway, but also because making `CoursePage` a child of `Page` instead of `HomePage` broke those links as Umar described. Merging that PR will fix this issue (at least that's what I observed)

umarmughal824: yeah exactly it breaks the link and neither it loads any course page nor it shows any preview or live button on the wagtail page. I could still attach those course pages to the home page by making it the child of the home page.

asadiqbal08 (approved):

PR #89, opened 2021-08-12
add faculty section in the product page

### LINKED ISSUE #34: Implement product detail page faculty section
**Depends on https://github.com/mitodl/mitxonline/issues/32**
Design (desktop): https://projects.invisionapp.com/d/main/#/console/21535972/455477372/preview
Design (mobile): https://projects.invisionapp.com/d/main/#/console/21535972/455488639/preview

#### Acceptance Criteria:
- New CMS model(s) added to store faculty information and link faculty to any number of product pages
- When a course page has no associated faculty, the section is omitted (including the "Meet your instructors" header)
- When a course page has associated faculty, they appear in a way that matches the desktop and mobile designs

#### Out of scope:
- Faculty cards do not need to link to anything
- Different styling when hovering over faculty cards

#### Implementation details:
- We implemented course/program faculty in xpro, so we can take some inspiration from that (`cms.blocks.FacultyBlock`, etc)

umarmughal824:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested

fixes #34 

#### What's this PR do?
add faculty section in the product page

Go to the course page and add faculty members

Screenshot 2021-08-13 at 17 56 00
Screenshot 2021-08-13 at 17 56 49
Screenshot 2021-08-13 at 17 46 18
Screenshot 2021-08-13 at 17 46 48
Screenshot 2021-08-13 at 17 46 58
Screenshot 2021-08-13 at 17 47 09
Screenshot 2021-08-13 at 17 47 20
Screenshot 2021-08-13 at 17 47 42
Screenshot 2021-08-13 at 17 47 54


pdpinch:
There were a few things I should have caught in the design. Sorry for these late changes: 

1. We don't need a designation field. Description is sufficient (same as xPRO, also same as edx.org) 
2. I think we want a carousel UI for mobile, again same as xPRO. I'm going to check on this.

umarmughal824:
@pdpinch what about the carousel on tablet or desktop as on desktop we are only showing one faculty per row on tablet?

pdpinch:
Sorry for the hold-up @umarmughal824. We can go ahead with the "stacked" layout on mobile. If we want a carousel, we'll add it later.

umarmughal824:
> Sorry for the hold-up @umarmughal824. We can go ahead with the "stacked" layout on mobile. If we want a carousel, we'll add it later.

sure, thanks

umarmughal824:
@gsidebo could you please review it as it's ready for review again.

cms/models.py
```diff
@@ -191,7 +191,35 @@ def get_child_by_readable_id(self, readable_id):
...
+    content_panels = Page.content_panels + [
+        FieldPanel("faculty_section_title"),
+        StreamFieldPanel("faculty_members"),
+    ]
+
+    class Meta:
+        abstract = True
+
+
+class ProductPage(FactultyPage):
```
gsidebo, line 222: I'm not sure about this pattern. If we keep implementing possibly-reusable sections of pages like this, then we'll have page classes with lots of different parent classes. I don't think that's sustainable. I think we should get rid of the `FactultyPage` class and move its fields into `ProductPage`. If another page ends up having a faculty section, it would just need to define those same 2 fields, which I think is completely reasonable

static/scss/faculty-members.scss
```diff
@@ -0,0 +1,111 @@
+// sass-lint:disable mixins-before-declarations
```
gsidebo, line 1: Given the amount of styles in `product-detail.scss`, I think it's fine to define these in a separate scss file, but I think we'd benefit from naming this `product-faculty-members.scss` to at least give a clearer indication that this is part of the product detail page

static/scss/faculty-members.scss
```diff
@@ -0,0 +1,111 @@
...
+.faculty-members {
+  padding: 0;
+  margin: 0 -22px;
+  list-style: none;
+  letter-spacing: -0.32em;
+
+  .course-card-description {
+    overflow-y: auto;
+    margin-bottom: 30px;
+  }
```
gsidebo, line 25: This was probably pasted in by accident 

```suggestion
```

static/scss/faculty-members.scss
```diff
@@ -0,0 +1,111 @@
...
+
+      .member-info {
+        overflow: hidden;
+        margin-top: 7px;
+
+        @include media-breakpoint-down(md) {
+          margin-top: 20px;
+        }
+
+        h2 {
```
gsidebo, line 53: Why is this an `h2` and resized to be smaller? Why not just make it an `h4`? The existing size that we have for the `h4` is perfectly good for this

static/scss/faculty-members.scss
```diff
@@ -0,0 +1,111 @@
...
+          color: #000000;
+          font-size: 18px;
+          font-weight: 600;
+          letter-spacing: 0;
+          line-height: 22px;
```
gsidebo, line 58: If you change to `h4` you may be able to get rid of all of this
```suggestion
```

static/scss/faculty-members.scss
```diff
@@ -0,0 +1,111 @@
...
+          margin: 0 0 5px;
+        }
+
+        p {
+          color: #333333;
+          font-size: 14px;
+          font-weight: 400;
+          letter-spacing: 0;
+          line-height: 22px;
+          margin-bottom: 0;
```
gsidebo, line 68: I don't think any of these styles are necessary besides the `font-size`. Please make an effort to reduce the amount of styling code you're adding to the codebase

```suggestion
          font-size: 14px;
```

cms/templates/product_page.html
```diff
@@ -93,5 +93,26 @@ <h2>What you’ll learn</h2>
...
+        <div class="container">
+          <h2>{{ page.faculty_section_title }}</h2>
+          <ul class="faculty-members">
+            {% for member in page.faculty_members %}
+              <li>
+                <div class="member-card">
+                  <img src="{% wagtail_img_src member.value.image %}" alt="{{ member.name }}">
+                  <div class="member-info">
+                    <h2 class="name">{{ member.value.name }}</h2>
+                    <div class="designation">{{ member.value.designation }}</div>
```
gsidebo, line 107: Unless I'm missing something, this field does not exist and users cannot enter anything for it

static/scss/faculty-members.scss
```diff
@@ -0,0 +1,111 @@
...
+          p {
+            margin: 0;
+            font-weight: 500;
+          }
+        }
+
+        .description {
+          p {
+            margin: 0;
+            color: #767676;
```
gsidebo, line 81: Please use `$light-gray` from `variables.scss` for this. We should always try to use existing colors unless there is something really noticeably different from what we already have defined, or if the existing color definition is clearly being used for a different purpose (for example, we wouldn't want to use a variable called something like `$border-color-blue` for text)

static/scss/faculty-members.scss
```diff
@@ -0,0 +1,111 @@
...
+          }
+        }
+      }
+
+      &:hover {
+        border-color: #a41e34;
+        box-shadow: 0 2px 15px 4px rgba(0, 0, 0, 0.16);
+
+        h2 {
+          color: #a41e34;
```
gsidebo, line 91: `#a41e34` is defined twice here, and the `box-shadow` styling is basically the same as the course cards on the home page. That makes sense, of course. In both cases, they are cards that have a box shadow and a colored bottom border when you hover over them. Obviously we do not want to define the same set of styles for elements that have the exact same look and feel. 

Please define a class in `common.scss` or somewhere like that, call it something like `highlight-card`, apply it to both the home page cards and these faculty cards, and get rid of the repeated styles. Like I said above, we need to make a greater effort to reuse existing styles and reduce the amount of styling code we're adding to the codebase

gsidebo (changes requested):

gsidebo (approved):
Looks great! 👍

PR #110, opened 2021-08-19
Added enrollment sync when dashboard loads

### LINKED ISSUE #74: Update enrollments when user opens dashboard
As a MicroMasters learner, I want my mitxonline dashboard to reflect any course enrollments I've made through the MicroMasters site. 

This is complicated by the fact that enrollments initiated from the MM site will be created in open edX directly. 

#### Acceptance Criteria:

- [ ] When a user views their dashboard, make a synchronous request to open edX to update all their enrollments. 
- [ ] Put this feature behind a feature flag, so we can turn it off if it causes performance issues

#### Related issues

- periodic task to update course enrollments asynchronously

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [x] Settings
  - [x] Opened issue for DevOps regarding necessary configuration changes to deployed environments

Closes #74

#### What's this PR do?
- Syncs a user's local enrollment records with edX every time that user's full enrollment data is requested (if a feature flag is turned on)
- Adds a management command to sync enrollment data

- Add `FEATURE_SYNC_ON_DASHBOARD_LOAD=True` to your `.env`
- Run the app alongside edX, and make sure you have at least 1 course run that exists both locally and in edX
- Load a user's dashboard. It should load successfully
- Enroll that user in a course run that has a matching run in edX
- Navigate to the course on edX and unenroll (should be doable via the gear icon)
- Check your enrollment in mitxonline Django admin. It should still exist and should still be active
- Load the dashboard. You should see that the enrollment no longer shows up in the dashboard, and in Django admin you should see the enrollment set to inactive

You can also test the management command (`sync_enrollments`) by leaving off the last step and running the management command instead.

- There's a noticeable "blink" when the dashboard data is loading now. We need to settle on a loading UI. Mentioning it here just to say that it's a known issue and is out of scope for this PR

ss 2021-08-19 at 13 04 11


openedx/api.py
```diff
@@ -598,3 +601,65 @@ def update_edx_user_name(user):
...
+            results.created.append(local_enrollment)
+    # Deactivate enrollment records for any enrollments that exist locally but not in edX
+    if local_only_ids:
+        enrollments_to_deactivate = [
+            enrollment
+            for courseware_id, enrollment in local_enrollments_map.items()
+            if courseware_id in local_only_ids and enrollment.active is True
+        ]
+        for local_enrollment in enrollments_to_deactivate:
+            local_enrollment.deactivate_and_save(
```
rhysyngsun, line 661: So one thing I think we need to watch out for here is that since this deactivates enrollments that don't show up in edx, regardless of whether they've been synced or not, meaning that it will deactivate enrollments in situations such as when there's an intermittent openedx outage. I can definitely see an enrollment failing when a learner create it, then the learner going to the dashboard and the enrollment gets deactivated. Particularly because most operations use `CourseRunEnrollments.objects` manager. I'm wondering if we actually need to delete enrollments though? I think MITx Online is supposed to be the source of truth on state, so we only really need to pull in missing enrollments to cover those that MITx Online didn't create itself, because we now have MicroMasters as a secondary source of truth.

Related to this, is the periodic task `retry_failed_edx_enrollments`, which we disabled at one point with a hard-coded return statement at the function start, but never circled back. This uses the `CourseRunEnrollments.objects` manager, so it'd be subjected to not seeing (and thus not repairing) these enrollments. I'm not proposing you fix that particular function here necessarily, although it would be a good way to test some edge cases.

rhysyngsun (changes requested):

openedx/api.py
gsidebo: @rhysyngsun  The original reason that I thought we'd need something like this is that, at least in my local edX instance, it's possible to unenroll from a course run in the edX dashboard:

ss 2021-09-03 at 11 38 24

Maybe there are other ways that a user could end up being unenrolled on the edX side as well. If this happens and we don't sync, then we'd still have an enrollment record on our end that says it's active, and if they follow the link they'll see something like this (again, maybe it's different from my local instance on production):

ss 2021-09-03 at 11 46 07

If the enrollment is deactivated in edX, it seems to me that we should reflect that in our own records to avoid confusion.

Your comment did make me realize a _major_ oversight that I had a note to implement but forgot about. If that sync function fails, we should log then return local enrollments as they are. Right now the API call just fails and doesn't return anything. This function is already not going to sync anything if there's an intermittent openedx outage, but I do need to add exception handling in the enrollment API.

Let me know what you think about that scenario I described above.

pdpinch:
Good discussion. I really want mitxonline to be the source of truth, but we know it can't be due to the MicroMaster enrollment path. 

However, we are planning (if we haven't already) to block access to the edX dashboard, so users should not be unenrolling via that path. It's just too confusing to have two (or three) dashboards with conflicting functionality. Is users want to unenroll, it will have to be through the mitxonline dashboard (maybe we need an issue for this).

gsidebo:
@rhysyngsun I changed the logic to just log an error and continue if we find local enrollments that do not exist in edX. I know it shouldn't happen in practice but I figure a log message is a simple and cheap just-in-case measure. Let me know what you think

rhysyngsun (approved):
LGTM

PR #113, opened 2021-08-20
Fixed issues with register API and recaptcha (#111)

### LINKED ISSUE #111: Recaptcha checks break registration if enabled
If you enable recaptcha (by setting the appropriate settings), registration breaks because the frontend isn't displaying the recaptcha, and hence not sending the required parameter.

This is happening because `SETTINGS.recaptchaKey` isn't present.

See: https://github.com/mitodl/mitxpro/blob/a137e9e5066b377f8ba5fc634e02e6f1977d63a6/mitxpro/utils.py#L586

rhysyngsun:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes #111 

#### What's this PR do?
This updates `get_js_settings` to return the recaptcha key since all of the other code expects it to be there.

- Configure a V2 recaptch key (https://www.google.com/recaptcha/admin/create)
- In `.env`, set `RECAPTCHA_SITE_KEY` and `RECAPTCHA_SECRET_KEY` based on the values the admin console gives you.
- Got to the create account page, the recaptcha should render and you should be able to submit the form and receive a confirmation email.

### DIFF
--- a/main/utils.py
+++ b/main/utils.py
@@ -35,6 +35,7 @@ def get_js_settings(request: HttpRequest):
         "environment": settings.ENVIRONMENT,
         "public_path": webpack_public_path(request),
         "release_version": settings.VERSION,
+        "recaptchaKey": settings.RECAPTCHA_SITE_KEY,
         "sentry_dsn": remove_password_from_url(settings.SENTRY_DSN),
         "support_email": settings.EMAIL_SUPPORT,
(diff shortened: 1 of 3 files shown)


gsidebo (approved):
We might want to add those settings to `.env.example` as well. Feel free to merge with or without! 👍

Issue #103, opened 2021-08-18
Home: make entire course card clickable


pdpinch:
The entire course card on the home page should be clickable, not just the course title. The clickable area should be the same as the hover state.

PR #114, opened 2021-08-20
Made entire course card clickable

Wassaf-Shahzad:
#### Pre-Flight checklist

- [x] Testing
 
Ticket #103 

#### What's this PR do?
Makes entire course card clickable.

- Add course to home page through cms
- Verify that it redirects to detail page on click


cms/templates/home_page.html
```diff
@@ -31,7 +31,7 @@ <h1>{{ product_cards_section_title }}</h1>
     <ul class="course-listing card-listing">
       {% for product in products %}
       <li>
-        <div class="course-card highlight-card">
+        <div onclick="location='{{ product.url_path }}'" class="course-card highlight-card">
```
gsidebo, line 34: For accessibility reasons and a few others, we should implement this with a normal `<a>` tag. You can either replace this `<div>` with an `<a>`, or add the `<a>` as the first child of this div

gsidebo (changes requested):

static/scss/course-cards.scss
```diff
@@ -46,6 +46,25 @@
...
+          &:hover {
+            color: #63131f;
+            text-decoration: underline;
+            cursor: pointer;
+          }
+
+        }
+      }
+
+      .course-link {
```
gsidebo, line 62: `course-link` is a little ambiguous. Changing this to `course-card-link` would make it clearer that this style applies to the link that wraps the whole card

static/scss/course-cards.scss
```diff
@@ -46,6 +46,25 @@
...
+        display: block;
+        width: 100%;
+        height: 100%;
```
gsidebo, line 65: I don't think these are necessary

```suggestion
```

static/scss/course-cards.scss
```diff
@@ -46,6 +46,25 @@
...
+        cursor: inherit;
```
gsidebo, line 66: Personally I think it makes more sense for the cursor to be a pointer if it's linking to a new page. If you were told that this should just be a normal cursor, or if you have a specific reason for using the default cursor here, you can ignore this suggestion.

gsidebo (approved):
3 very minor comments. Feel free to merge after taking care of them. Looks great! 👍

PR #119, opened 2021-08-20
Added setting to avoid name collisions in Wagtail

### LINKED ISSUE #108: CMS: S3 backend is allowing filename collisions
### Steps to Reproduce

1. On RC CMS (where the S3 storage is turned on), upload an image with the name "course_image.jpg"
2. upload a different image, with the same file name
3. View the thumbnails for the two images in the CMS or
4. View the two image in published pages

For example, see 

### Expected Behavior

- in both cases, you should see the two different images that you uploaded

### Actual Behavior

- In both cases, you will see the same image.  

### Screenshot

from https://rc.mitxonline.mit.edu/cms/images/?q=14.

image

gsidebo:
#### Pre-Flight checklist

- [x] Testing
- [x] Settings
  - [x] ~~Opened issue for DevOps regarding necessary configuration changes to deployed environments~~ N/A - defaulting this setting to `False` which what we want for all envs

Fixes #108

#### What's this PR do?
Adds a setting to prevent name collisions for uploaded files (especially critical for files uploaded in Wagtail)

*NOTE:* To get this working I couldn't use my existing database, which had a bunch of CMS data from 

- Copy `AWS_` settings from RC and set `MITX_ONLINE_USE_S3=True` in your `.env`
- Run the app, and upload an image, then upload the same image again
- Navigate to the detail page for each of those images in Wagtail. Under "File:" the first image should have the same filename as the file you uploaded, and the second one should have an alphanumeric suffix attached

ss 2021-08-20 at 17 53 46

### DIFF
--- a/app.json
+++ b/app.json
@@ -35,6 +35,10 @@
       "description": "Enables querystring auth for S3 urls",
       "required": false
     },
+    "AWS_S3_FILE_OVERWRITE": {
+      "description": "Django Storages setting. By default files with the same name will overwrite each other. Set this to False to have extra characters appended.",
+      "required": false
+    },
     "AWS_SECRET_ACCESS_KEY": {
       "description": "AWS Secret Key for S3 storage.",
       "required": false
--- a/main/settings.py
+++ b/main/settings.py
@@ -642,6 +642,16 @@ AWS_QUERYSTRING_AUTH = get_bool(
     default=False,
     description="Enables querystring auth for S3 urls",
 )
+AWS_S3_FILE_OVERWRITE = get_bool(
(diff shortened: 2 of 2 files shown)


arslanashraf7 (approved):
Looks good 👍 

Also, note that I was able to test it locally with AWS_ keys from RC and the default of `AWS_S3_FILE_OVERWRITE` which is `False`. Upon uploading the same image twice I could verify the added character in the name.

Screenshot 2021-08-23 at 5 11 52 PM
Screenshot 2021-08-23 at 5 11 44 PM

PR #121, opened 2021-08-23
Update openedx configuration docs

### LINKED ISSUE #31: MicroMasters Auth: update documentation to describe how to configure the system
Update documentation to describe how to configure mitX Online + MicroMasters to work together.

_Note:_
I suspect locally we might hit some issues here with devstack - specifically for the case that the user already has an account created in devstack for MicroMasters (or xPro for that matter), and they try to add an mitX Online authentication they will probably see an error back from OpenedX and their account creation will fail. This might require some hackery to get around but we should be clear in distinguishing that this is a local dev issue because in deployed environments MicroMasters and mitX Online won’t be running against the same (Open)edX deployment.

rhysyngsun:
Closes #31 

#### What's this PR do?
This updates the documentation with a setting that was missing. The MM configuration itself already exists over in the MM repo.

N/A, documentation only

### DIFF
--- a/docs/configure_open_edx.md
+++ b/docs/configure_open_edx.md
@@ -114,6 +114,7 @@ In Open edX (derived from instructions [here](https://edx.readthedocs.io/project
         "ENABLE_COMBINED_LOGIN_REGISTRATION": true,
         "ENABLE_THIRD_PARTY_AUTH": true,
         "ENABLE_OAUTH2_PROVIDER": true,
+        "SKIP_EMAIL_VALIDATION": true,
         ...
       },
       ...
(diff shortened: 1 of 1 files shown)


gsidebo (changes requested):
Since you're fixing up this doc already, there are a couple other things you can change:
- There's a typo above where `EXTERNAL_XMITX_ONLINE_HOST` should be `EXTERNAL_MITX_ONLINE_HOST`
- All instances of `8053` need to be changed to `8013`

gsidebo (approved):
Looks great! 👍

Issue #84, opened 2021-08-11
Update styling for account creation pages


pdpinch:
#### Designs and Mockups

See https://projects.invisionapp.com/d/main?origin=v7#/console/21535972/455982825/preview?scrollOffset=1558

#### Acceptance Criteria:

- [ ] /signin
- [ ] /signin/forgot-password
- [ ] /create-account
- [ ] /create-account/confirm

PR #127, opened 2021-08-26
Wassaf/update account creation style

Wassaf-Shahzad:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing  

Relevant Ticket #84

#### What's this PR do?
Updated Styles for the following pages.
 - [x] /signin
 - [x] /signin/forgot-password
 - [x] /create-account
 - [x] /create-account/confirm

This PR can be manually tested by following these steps.
 - Visit /signin URL and verify updated styles.
 - Visit /signin/forgot-password URL and verify updated styles.
 - Visit /create-account URL and verify updated styles.
 - Visit /create-account/confirm URL and verify updated styles.
 
During the implementation of this feature. We identified some improvements, which can be resolved in a separate PR.
- The auth-card class could be a separate react component rather then html elements sharing styles.
- The submit button could also be separate react component as well.
- The LoginForgotPasswordPage.js contains a Reset Email Confirm part which is developed standard html elements while incase of RegisterEmail form we use a react component. We should use the react component for both cases as they share all the styles, content and differ only in headings. 
- The notification on Register User Account has a blue colour due to it getting the class of info and not success.
  
I also wanted to add that I might be missing some context or technical knowledge and could have miss identified. 

Sign In page

Desktop View:

Screenshot 2021-08-26 at 5 25 01 PM

Screenshot 2021-08-26 at 5 29 11 PM

Screenshot 2021-08-26 at 5 26 27 PM

Mobile View:

Screenshot 2021-08-26 at 5 25 31 PM

Screenshot 2021-08-26 at 5 29 44 PM

Screenshot 2021-08-26 at 5 26 57 PM

Password reset Page:

Desktop View:

Screenshot 2021-08-26 at 5 30 54 PM

Mobile View:

Screenshot 2021-08-26 at 5 31 08 PM

Register User Page : 

Desktop View:

Screenshot 2021-08-26 at 5 31 08 PM

Screenshot 2021-08-26 at 5 01 28 PM

Mobile View:

Screenshot 2021-08-26 at 4 58 52 PM

Screenshot 2021-08-26 at 5 02 11 PM

Register User Confirm Page:

Desktop View:

Screenshot 2021-08-26 at 4 59 26 PM

Mobile View:

Screenshot 2021-08-26 at 4 59 57 PM

Change Password Page:

Screenshot 2021-09-02 at 6 05 09 PM

### DIFF
--- a/static/js/containers/pages/login/LoginEmailPage.js
+++ b/static/js/containers/pages/login/LoginEmailPage.js
@@ -60,15 +60,15 @@ export class LoginEmailPage extends React.Component<Props> {
         title={`${SETTINGS.site_name} | ${LOGIN_EMAIL_PAGE_TITLE}`}
       >
         <div className="container auth-page">
-          <div className="row auth-header">
-            <h1 className="col-12">Sign in</h1>
-          </div>
           <div className="row auth-card card-shadow auth-form">
             <div className="col-12">
+              <div className="row auth-header">
+                <h1 className="col-12">Sign in</h1>
+              </div>
               <EmailForm onSubmit={this.onSubmit.bind(this)}>
                 <React.Fragment>
                   <span>Don't have an account? </span>
-                  <Link to={link} className="link-light-blue">
+                  <Link to={link} className="link-black">
                     Create Account
                   </Link>
                 </React.Fragment>
--- a/static/js/containers/pages/login/LoginForgotPasswordPage.js
+++ b/static/js/containers/pages/login/LoginForgotPasswordPage.js
@@ -32,8 +32,8 @@ type State = {
 
 const passwordResetText = (email: string) => (
   <p>
-    If an account with the email <span className="email">{email}</span> <br />{" "}
-    exists, an email has been sent with a password reset link.
+    If an account with the email <b>{email}</b> exists, an email has been sent
+    with a password reset link.
   </p>
 )
 
@@ -45,7 +45,10 @@ const getCustomerSupportLayout = (isError: boolean) => {
           Sorry, there was an error sending the email.
         </h3>
       ) : (
-        <b>Still no password reset email? </b>
+        <>
+          <hr />
+          <b>Still no password reset email? </b>
+        </>
       )}
       Please {isError ? "try again or" : ""} contact our{" "}
       {` ${SETTINGS.site_name} `}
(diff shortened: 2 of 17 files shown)


pdpinch (approved):
screenshots look good to me. Please have someone do a code review.

static/js/containers/pages/login/LoginForgotPasswordPage.js
```diff
@@ -95,50 +98,63 @@ export class LoginForgotPasswordPage extends React.Component<Props, State> {
...
+                    <b>Check your spam folder.</b> It might be there.
+                  </li>
+                  <li>
+                    <b>Is your email correct?</b> If you made a typo, no
+                    problem,{" "}
+                    <Link
+                      to={routes.login.forgot.begin}
+                      onClick={this.resetEmailLinkSent.bind(this)}
+                    >
+                      {`create an account`}
```
gsidebo, line 139: This was only in brackets before so we could add spaces on both sides of the string. You can remove the brackets and backticks, or keep them and remove the `{" "}` spaces on both sides of the `Link`. Same comment applies to other pages in this PR

static/scss/auth.scss
```diff
@@ -160,6 +196,34 @@
...
+    div {
+      font-weight: inherit !important;
+    }
+
+    .subtitle {
+      color: $label-subtitle;
+      font-size: 14px;
+    }
+
+    .red {
```
gsidebo, line 218: Can you change this class name from `red` to `required`? What we're really styling here is a character that indicates the field is required, so that's a more useful class name, and it won't be a problem if the actual color changes at some point

static/js/containers/pages/register/RegisterConfirmSentPage_test.js
```diff
@@ -47,13 +47,22 @@ describe("RegisterConfirmSentPage", () => {
...
+      routes.register.begin
+    )
   })
 
   it("displays user's email on the page", async () => {
     const { inner } = await renderPage()
     assert.equal(
-      inner.find(".confirm-sent-page > p > span").text("href"),
+      inner
+        .find(".auth-card > .col-12 > p > b")
```
gsidebo, line 63: This highly specific selector could cause unnecessary problems if the HTML changes at all. I think it's better to just check that a string is included in the overall text in the card. Something like:

```es6
assert.include(
  inner.find(".auth-card").text(),
  `We sent an email to ${email}`
)

**Review by gsidebo — CHANGES_REQUESTED** [review]

File: static/js/containers/pages/login/LoginEmailPage.js [hunk 4]
```diff
@@ -60,15 +60,15 @@ export class LoginEmailPage extends React.Component<Props> {
...
       >
         <div className="container auth-page">
-          <div className="row auth-header">
-            <h1 className="col-12">Sign in</h1>
-          </div>
           <div className="row auth-card card-shadow auth-form">
             <div className="col-12">
+              <div className="row auth-header">
+                <h1 className="col-12">Sign in</h1>
+              </div>
```
gsidebo, line 67:
```suggestion
              <h1>Sign in</h1>
```

Sorry I missed this on the first round. This extra div is not necessary. All it appears to be doing is adding some spacing below the `h1`. If we need spacing below the `h1`, we can just define it here with no added class and style it with `.auth-card h1`. This comment applies to all other pages where there is a `<div className="row">` with children that all have `className="col-12"`. I think I'm seeing that this pattern was used in some other parts of the codebase but unless I'm missing some necessary styling property it's adding, we should get rid of it

gsidebo (changes requested):

static/js/containers/pages/login/LoginEmailPage.js
Wassaf-Shahzad: The spacing under the `h1` is actually conditional. Whenever a sub-heading exists for example "Signing in as {username}", the spacing removed. Also pages like `RegisterConfirmSentPage.js` and `LoginForgotPasswordPage.js` do not have a `div` separating heading from the content which stops us from applying the spacing based on number of siblings.

gsidebo (approved):
Looks good 👍

PR #134, opened 2021-08-31
Fixed CMS migrations, added startup command to configure Wagtail

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
- Changes the existing CMS data migration for setting up course index page etc. to be a no-op in favor of using the `configure_wagtail` command exclusively.
- Changes our app configuration so that the `configure_wagtail` command is run immediately after the `migrate` command, both for local development (via `docker-compose.yml`) and deployment (via [`app.json`](https://devcenter.heroku.com/articles/app-json-schema)).

Running the management command and making sure it completes successfully is a good start. 

Beyond that, the purpose of this command is to ensure that all of the Wagtail objects we care about are in the database, so the best way to test is with a fresh database. You can run the command with a fresh db and ensure that it succeeds – our custom home page is set up as the site root, a course index page exists as a child of that home page, and basic resource pages have been added.

- I was made aware of the need for this change when my database got stuck in a bad state w/ cms migrations. Hacking my way through that and remembering the hardships we had with cms data migrations in xpro, it was pretty clear that we needed an alternate solution for getting Wagtail configured correctly and automatically.
- For any future steps that we need to take to ensure that Wagtail is configured correctly, we will not use data migrations anymore once this merges. We will just add to the `configure_wagtail` management command.

### DIFF
--- a/cms/api.py
+++ b/cms/api.py
@@ -1,14 +1,17 @@
 """API functionality for the CMS app"""
 import logging
+from typing import Tuple
 from urllib.parse import urljoin, urlencode
 
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
+
 from django.utils.text import slugify
 from wagtail.core.blocks import StreamValue
 from wagtail.core.models import Page, Site
 
 from cms import models as cms_models
+from cms.exceptions import WagtailSpecificPageError
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


umarmughal824 (approved):
Looks good 👍 , just consider that one comment

PR #143, opened 2021-09-02
Added welcome message for users that complete first authentication

### LINKED ISSUE #69: Show notification after user authenticates for the first time
Related: #50 (PR #68)
Design: https://projects.invisionapp.com/d/main/#/console/21535972/455982830/preview

**Acceptance criteria:**
- After a user authenticates for the first time (and _only_ the first time) show the "Account created!" user notification in a way that matches the design

**Implementation details:**
- Search for `addUserNotification` for other examples of these notifications
- This issue is complicated by the fact that we have 2 react apps (the header app and the main body app). When a user gets a successful final auth response, we use `window.location` instead of `history.push` to take them to the dashboard because _both_ the header app and the main body app need to refresh at that time (using `history.push` does not cause the header app to refresh). If we just execute `addUserNotification` before redirecting, the message will disappear as soon as the page redirects (almost instantly). To accomplish this, we may want to set a cookie or consider some other approach that maintains some state when the entire app refreshes

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #69 

#### What's this PR do?
Adds a welcome message for users that complete authentication for the first time

- Register a new user and complete authentication. You should see a welcome banner that will not show up w/ page reloads after ~20s
- Log into the app with an existing user. You should not be shown the welcome banner

ss 2021-09-01 at 13 50 11

### DIFF
--- a/authentication/views_test.py
+++ b/authentication/views_test.py
@@ -29,7 +29,9 @@ from social_core.backends.email import EmailAuth
 
 from authentication.serializers import PARTIAL_PIPELINE_TOKEN_KEY
 from authentication.utils import SocialAuthState
+from main.constants import USER_MSG_COOKIE_NAME, USER_MSG_TYPE_COMPLETED_AUTH
 from main.test_utils import MockResponse
+from main.utils import encode_json_cookie_value
 from users.factories import UserFactory, UserSocialAuthFactory
 
 pytestmark = [pytest.mark.django_db]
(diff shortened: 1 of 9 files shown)


authentication/views_test.py
```diff
@@ -386,7 +410,7 @@ def login_password_user_inactive(self, auth_state, verify_exports):
     #         "authentication.pipeline.compliance.api.verify_user_with_exports",
     #         side_effect=Exception("register_details_export_temporary_error"),
     #     ):
-    #         assert_api_call(
+    #         assert_api_call_json(
```
arslanashraf7, line 413: I know the commented test is not part of this PR, but if possible for you, I think we should remove it or add a proper TODO/Comment on the test if we really want to keep it like this.

arslanashraf7 (approved):
Looks good/works well 👍.

Just one small comment which you can feel free to address or respond to. Feel free to merge it after that.

PR #151, opened 2021-09-03
Added loading animation component and applied to dashboard

### LINKED ISSUE #138: Loading animation
As a user, I want to be reassured that something is happening background processes take a few seconds. 

#### Designs and Mockups

loader-1

#### Acceptance Criteria:

- [ ] on the dashboard display a loading indicator while we are updating the user's enrollments.

gsidebo:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

Closes #138 

#### What's this PR do?
- Adds the `Loader` component, which renders a delayed spinner when the prop `isLoading` is true, or the children of the component if `isLoading` is false
- Wraps the dashboard page contents in the Loader component so a spinner will be loaded
- Adds some consistency to the spacing of content underneath the header. We had a mish-mash of inconsistent rules governing that spacing and it was straightforward enough to refactor them and use the new rule for this spinner

One good way to test is to add a delay in the API view so the loader is guaranteed to be visible. You can add this to the beginning of `courses.views.v1.UserEnrollmentsApiViewSet.get_queryset`

```python
import time
time.sleep(5)
```

When you load `/dashboard/`, you should see a loading spinner fade in quickly after a very short delay (200ms). When the API call gets a response, the spinner should disappear and the dashboard contents should be shown

- I used a pure CSS loader from [loading.io](https://loading.io/css/) because the gif attached to the issue has a pure white background, which is going to clash with the eventual background color we're going to use and it limits our options anyway, and its 9kb. This was approved by Peter. **NOTE**: The animation uses [CSS `keyframes`](https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes), which appear to be supported by all the browsers we care about.
- I only added the loading animation to the dashboard page to begin with since that is the page that is the most likely to have a noticeable delay. We can add it to other pages as needed.
- I added the delay and the fade-in to avoid a clunky UI where the spinner "blinks" into and out of view when a page loads quickly. With the current logic, the Loader component waits 200ms before rendering the spinner, and fades it into view using CSS

ss 2021-09-03 at 15 02 54
ss 2021-09-03 at 16 41 36
ss 2021-09-03 at 16 41 27


static/scss/layout.scss
```diff
@@ -50,4 +51,12 @@ body {
       }
     }
   }
+
+  .std-page-body {
```
arslanashraf7, line 55: Thanks for adding this common style for padding. ⭐

arslanashraf7 (approved):
Looks/works well. 👍 

Alongside spinner, I was also able to check the common styling that would apply to container sections e.g. Registration, Authentication, Profile, etc.

Issue #147, opened 2021-09-02
Forgot password should be case-insensitive


briangrossman:
### Steps to Reproduce

- Register for an account on https://mitxonline.mit.edu/ (e.g. `youremail@gmail.com`)
- Sign out
- Click `Sign In`
- Enter your email (e.g. `youremail@gmail.com`)
- Click the `Forgot Password?` link
- Enter your email, but change the case for some of the letters (e.g. `YOureMAil@gMAil.com`)

### Expected Behavior

- If the only difference is capital / lowercase emails, forgot password functionality should work and learner will be sent an email with a link to update their password
- If the user's account actually doesn't exist, the learner should receive the same message as when the account does exist:
```
Thank You!
If an account with the email some_email_that_doesnt_exist_in_the_mitxonline_database@gmail.com
exists, an email has been sent with a password reset link.

If you do NOT receive your password reset email, here's what to do:

Wait a few moments. It might take several minutes to receive your password reset email.
Check your spam folder. It might be there.
Is your email correct? If you made a typo, no problem, just try to reset your password again.
Still no password reset email? Please contact our MITx Online Customer Support Center.
```

### Actual Behavior

- Forgot password functionality fails, giving the error: `Sorry, there was an error sending the email`
![Screen Shot 2021-09-02 at 3 34 27 PM](https://user-images.githubusercontent.com/21342568/131909471-7afbbf14-3365-4698-bebc-3dcf743dd7d8.png)

- Note: when registering for the site, email addresses are not case sensitive.

### Related Issues
- https://github.com/mitodl/mitxpro/issues/2304
- https://github.com/mitodl/bootcamp-ecommerce/issues/1295

PR #159, opened 2021-09-08
Make Forgot Password case insensitive

Wassaf-Shahzad:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Relevant Ticket #147 

#### What's this PR do?
This PR makes Forgot Password case insensitive

- Create new User,
- Follow Sign in process until Password Page.
- Click Forgot password and enter email with some characters in Capital form.
- Verify email is sent.

### DIFF
--- a/authentication/serializers.py
+++ b/authentication/serializers.py
@@ -13,7 +13,8 @@ from social_core.utils import (
     user_is_authenticated,
 )
 from social_django.views import _do_login as login
-
+from djoser.serializers import SendEmailResetSerializer
+from djoser.conf import settings
 from authentication.exceptions import (
     EmailBlockedException,
     InvalidPasswordException,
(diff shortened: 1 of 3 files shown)


pdpinch:
@gsidebo @rhysyngsun we've discovered this issue on xPRO and Bootcamps as well. Is there a way to include this fix in ol-django, or should we just patch each repo?

authentication/serializers.py
```diff
@@ -344,3 +345,22 @@ class RegisterExtraDetailsSerializer(SocialAuthSerializer):
     def create(self, validated_data):
         """Try to 'save' the request"""
         return super()._authenticate(SocialAuthState.FLOW_REGISTER)
+
+
+class CustomSendEmailResetSerializer(SendEmailResetSerializer):
+
+    def get_user(self, is_active=True):
```
gsidebo, line 352: Normally when some library functionality is overridden, we would call `super()` somewhere to default to the base implementation, or we would completely replace the logic. This is directly copying most of the logic of the overridden method, so I think it would be useful to note that explicitly 

```suggestion
    # NOTE: This directly copies the implementation of djoser.serializers.UserFunctionsMixin.get_user
    # and only changes the User query. If this method is changed in an updated Djoser
    # release, this method may need to be updated as well.
    def get_user(self, is_active=True):
```

gsidebo (approved):
Looks and works great! Once you add that suggested comment, feel free to merge 👍

pdpinch:
@Wassaf-Shahzad I don't want this to be blocked by my question. Please go ahead and merge and you've responded to @gsidebo's comment.

Issue #155, opened 2021-09-07
Fix accessibility issues in Home Page


arslanashraf7:
This ticket is a part of #150 EPIC.

The fixes that should be done under the banner of this ticket are:

- [x] Fix duplicate H1 tags on home page (there can be only one H1 per page)
- [x] Use empty ALT tags for course cards on home page by setting `alt=""`. The thumbnails are purely decorative and no alt tag would be preferable to the repetitive ones we have now.

arslanashraf7:
Closed via #160 and #168 (Partially)

PR #160, opened 2021-09-08
fix: accessibility issues on homepage

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#155 

#### What's this PR do?
- Removes multiple `h1` tags from the homepage and keeps only now `h1`
- Removed the `alt` text from the course list on the homepage since the cards are now decorative and won't need repetitive alt names.

**Note:**
There are currently 2 `H1` elements on the Home Page. `1) Hero title` `(2 Courses Heading`. To keep only one H1 we had alternate options:
1) We already had a customized style for Hero Title's `h1` on the home page and just changing it to `h2` would let us do the job that we needed.
2) We replace `h1` with any other `h tag` in the courses heading and add a relevant style for the CSS.

I followed the approach in `point#1` since we are trying to keep common styling for all of our `h1 elements` in all the pages and the course list heading is one of them following the style from the common. So instead of updating it, I've updated the hero title's h1 which was already using a custom styling.

- Visit the home page and check there is only one H1
- Visit the home page and check that course cards have empty `alt` value for the images

- Home Page

### DIFF
--- a/cms/templates/home_page.html
+++ b/cms/templates/home_page.html
@@ -15,10 +15,10 @@
   <div class="container hero-content">
     <div class="hero-title">
       {% if page.hero_title %}
-        <h1>{{ page.hero_title }}</h1>
+        <h2>{{ page.hero_title }}</h2>
       {% endif %}
       {% if page.hero_subtitle %}
-        <h2>{{ page.hero_subtitle }}</h2>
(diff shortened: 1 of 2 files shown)


pdpinch (approved):

pdpinch:
Code looks good. You can go ahead and merge, IMO.

Issue #153, opened 2021-09-07
Fix accessibility issues in the forms


arslanashraf7:
This ticket is a part of #150 EPIC.

The fixes that should be done under the banner of this ticket are:

- [ ] sign in form input tag is missing an id attribute, `id="email"` which is necessary for the `<label>` tag to function. 
  - [ ] review any other forms for missing `id` attributes and/or missing `<label>` tags. 
- [ ] When an error message is updated in forms, the container needs a `role="alert"` attribute
- [ ] input tags should have an `aria-desribed-by` attribute, set to the id of the container of any error messages.

PR #161, opened 2021-09-08
Fixed accessibility issues in Forms

Wassaf-Shahzad:
#### Pre-Flight checklist
- [x] Testing

Ticket #153 

#### What's this PR do?
This PR has the following changes
- Adds relevant id to input tags.
- Adds role="alert" to the FormError Component
- Adds property aria-describedby to input tags

- By using light house to test accessibility.

Light house score for local testing 
Screenshot 2021-09-09 at 4 14 31 PM

Screenshot 2021-09-10 at 3 28 50 PM

Screenshot 2021-09-10 at 3 03 24 PM

Screenshot 2021-09-10 at 2 58 03 PM

Screenshot 2021-09-10 at 12 28 53 PM

Screenshot 2021-09-10 at 3 51 51 PM

### DIFF
--- a/static/js/components/forms/ChangeEmailForm.js
+++ b/static/js/components/forms/ChangeEmailForm.js
@@ -49,10 +49,12 @@ const ChangeEmailForm = ({ onSubmit, user }: Props) => (
             </label>
             <Field
               name="email"
+              id="email"
               className="form-control"
               component={EmailInput}
+              aria-describedby="emailError"
             />
-            <ErrorMessage name="email" component={FormError} />
+            <ErrorMessage name="email" id="emailError" component={FormError} />
           </div>
           <div className="form-group">
             <label htmlFor="confirmPassword" className="row">
@@ -64,11 +66,17 @@ const ChangeEmailForm = ({ onSubmit, user }: Props) => (
               </div>
             </label>
             <Field
+              id="confirmPassword"
               name="confirmPassword"
               className="form-control"
               component={PasswordInput}
+              aria-describedby="confirmPasswordError"
             />
-            <ErrorMessage name="confirmPassword" component={FormError} />
+            <ErrorMessage
+              name="confirmPassword"
+              id="confirmPasswordError"
+              component={FormError}
+            />
           </div>
         </section>
 
--- a/static/js/components/forms/ProfileFormFields.js
+++ b/static/js/components/forms/ProfileFormFields.js
@@ -116,10 +116,16 @@ export const LegalAddressFields = ({
       <Field
         type="text"
         name="legal_address.first_name"
+        id="legal_address.first_name"
         className="form-control"
         autoComplete="given-name"
+        aria-describedby="legal_address.first_name_error"
       />
-      <ErrorMessage name="legal_address.first_name" component={FormError} />
+      <ErrorMessage
+        name="legal_address.first_name"
+        id="legal_address.first_name_error"
+        component={FormError}
+      />
     </div>
     <div className="form-group">
       <label htmlFor="legal_address.last_name" className="font-weight-bold">
@@ -128,10 +134,16 @@ export const LegalAddressFields = ({
       <Field
         type="text"
         name="legal_address.last_name"
+        id="legal_address.last_name"
         className="form-control"
         autoComplete="family-name"
+        aria-describedby="legal_address.last_name_error"
       />
-      <ErrorMessage name="legal_address.last_name" component={FormError} />
+      <ErrorMessage
+        name="legal_address.last_name"
+        id="legal_address.last_name_error"
+        component={FormError}
+      />
     </div>
     <div className="form-group">
       <label htmlFor="name" className="row">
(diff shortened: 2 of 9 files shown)


pdpinch (commented):
I had one question, which came up in many, many places. 

You should have someone more familiar with react review this as well.

static/js/components/forms/ChangeEmailForm.js
```diff
@@ -47,19 +47,27 @@ const ChangeEmailForm = ({ onSubmit, user }: Props) => (
...
             <label htmlFor="confirmPassword">Confirm Password</label>
             <Field
+              id="confirmPassword"
               name="confirmPassword"
               className="form-control"
               component={PasswordInput}
+              aria-describedby="confirmPasswordError"
+            />
+            <ErrorMessage
+              name="confirmPassword"
```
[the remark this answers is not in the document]
Wassaf-Shahzad: (According to my knowledge)
The name attribute is not passed  to our custom component and is used by Formik internally for associating the fields name in Formik state.
The name attribute is also required in case for ErrorMessage component.
Reference: https://formik.org/docs/api/errormessage

static/js/containers/pages/settings/AccountSettingsPage.js
```diff
@@ -123,7 +123,7 @@ export class AccountSettingsPage extends React.Component<Props> {
       >
         <div className="container auth-page account-settings-page">
           <div className="auth-header">
-            <h1>User Settings</h1>
```
gsidebo, line 126: I don't think this change is necessary, and I have an open PR that updates this page layout anyway (#185). I think you can revert this change

static/js/components/forms/ProfileFormFields.js
```diff
@@ -108,10 +108,16 @@ export const LegalAddressFields = ({
       <Field
         type="text"
         name="legal_address.first_name"
+        id="legal_address.first_name"
         className="form-control"
         autoComplete="given-name"
+        aria-describedby="legal_address.first_nameError"
```
gsidebo, line 114: Our original mistake here was naming these form fields in a way that matches the API results, instead of giving them camel case names like our other forms and mapping the API results to those fields. This field should have `name="firstName"` and the error should have `id="firstNameError"`. That issue existed long before this PR though, so you definitely don't need to change all of these values here. In the meantime, this mixture of snake case and camel case is awkward. Can you change the naming scheme so that this value is `legal_address.first_name.error` or `legal_address.first_name_error`, whichever one you prefer? After that, please apply the same naming scheme to these other attributes that were added in this PR.

gsidebo (changes requested):

pdpinch (commented):
I saw that we didn't get 100 on the profile and user settings forms. Can we improve those scores?

Wassaf-Shahzad:
> I saw that we didn't get 100 on the profile and user settings forms. Can we improve those scores?

Regarding the User settings,  I believe @gsidebo PR will improve the score 
and Incase of Profile form @arslanashraf7 PR will improve the score.

Issues in Profile Settings
- ARIA ID's are not unique
Screenshot 2021-09-14 at 5 42 10 PM

This issue is handled by Arslan's header PR,

Issue in User Settings

- ARIA ID's are not unique
- Heading Elements are not in sequential order 

Screenshot 2021-09-14 at 5 45 50 PM

This will be handled in Gavin's PR

static/js/components/forms/ChangeEmailForm.js
```diff
@@ -47,19 +47,27 @@ const ChangeEmailForm = ({ onSubmit, user }: Props) => (
...
+            <ErrorMessage name="email" id="email_error" component={FormError} />
           </div>
           <div className="form-group">
             <label htmlFor="confirmPassword">Confirm Password</label>
             <Field
+              id="confirmPassword"
               name="confirmPassword"
               className="form-control"
               component={PasswordInput}
+              aria-describedby="confirm_password_error"
```
gsidebo, line 64: Sorry, I don't think I was clear enough in my last review. I was suggesting that only the fields that already have the anti-pattern snake case attribute values (like `name="profile.job_title"`) should have ids that are also snake case. I know it seems weird to have different patterns for the same attributes in different files, but eventually we're going to refactor those snake case attribute values (#186) and the id's/aria attributes can be updated as well. If you notice here, `confirm_password_error` introduces **three** different conventions for attribute values in a single component. We have camel case for `id` and `name`, lowercase + dashes for `className`, and now snake case for `aria-describedby`. Dashes in CSS class names are already an understood convention that is specific to CSS. For other attributes, we should try to stick to camel case. So for example:

- ~~`aria-describedby="confirm_password_error"`~~ --> `aria-describedby="confirmPasswordError"`
- ~~`aria-describedby="email_error"`~~ --> `aria-describedby="emailError"`

Please apply that logic to all `id` and `aria-describedby` attributes in this PR _except for_ the ones that are associated with a field that already has snake case attribute values (like `legal_address.first_name`)

gsidebo (changes requested):

static/js/components/forms/ChangeEmailForm.js
Wassaf-Shahzad: This was my first thought as well However `please apply the same naming scheme to these other attributes that were added in this PR.` this line threw me off a bit and I assumed I had to apply the style to all attributes to maintain consistency.
I should have consulted you before hand, that was an oversight on my end and as requested above I have completed the required change.

gsidebo (approved):
Looks great! 👍

PR #166, opened 2021-09-09
Pull courserun title, dates from studio

### LINKED ISSUE #106: Pull courserun title, dates from studio
same as xPRO, originally implemented in https://github.com/mitodl/mitxpro/pull/1314:

As a course team member, I'd like to update the title of a course, or its dates in Studio, and see the update reflected in the mitxonline catalog.

As an admin, I'd like studio to be the source of truth for courserun titles and dates, so I don't have to update the mitxonline admin on a regular basis.

#### Acceptance Criteria:

- [x] daily task, configurable by environment variable 
- [x] management command for ad hoc syncing
- [x] use edx-api-client library 
- [x] pull course titles and dates from open edX LMS api using the course_detail API

umarmughal824:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings

fixes #106 

#### What's this PR do?
Pull courserun title, dates from studio

- make a course / course run with same courseware_id both on MIT xOnline and edx-platform  e.g `course-v1:edX+DemoX+Demo_Course`
- then run management command / celery tasks to run the syncs of data
- `docker-compose exec web python manage.py sync_courserun --run=course-v1:edX+DemoX+Demo_Course`
- `courses.tasks.sync_course_runs_data`
- for further details see the attached screenshots

**THROUGH MANAGEMENT COMMAND**

Screenshot 2021-09-10 at 19 27 04

**THROUGH CELERY TASK (CALLING MANUALLY / SCHEDULED)**

Screenshot 2021-09-10 at 19 22 21

**BEFORE SYNCING**

Screenshot 2021-09-10 at 19 18 33

**AFTER SYNCING**

Screenshot 2021-09-10 at 19 18 50

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -719,12 +719,12 @@ CELERY_TIMEZONE = "UTC"
 CRON_COURSERUN_SYNC_HOURS = get_string(
     name="CRON_COURSERUN_SYNC_HOURS",
     default=0,
-    description="'hours' value for the 'sync-courseruns-data' scheduled task (defaults to midnight)",
+    description="'hours' value for scheduled task to sync course run data (by default, it will run at midnight",
 )
 CRON_COURSERUN_SYNC_DAYS = get_string(
     name="CRON_COURSERUN_SYNC_DAYS",
-    default=None,
-    description="'day_of_week' value for 'sync-courseruns-data' scheduled task (default will run once a day).",
+    default="*",
+    description="day_of_week' value for scheduled task to sync course run data (by default, it will run each day of the week).",
 )
 RETRY_FAILED_EDX_ENROLLMENT_FREQUENCY = get_int(
     name="RETRY_FAILED_EDX_ENROLLMENT_FREQUENCY",
@@ -755,7 +755,7 @@ CELERY_BEAT_SCHEDULE = {
         "schedule": crontab(
             minute=0,
             hour=CRON_COURSERUN_SYNC_HOURS,
-            day_of_week=CRON_COURSERUN_SYNC_DAYS or "*",
+            day_of_week=CRON_COURSERUN_SYNC_DAYS,
             day_of_month="*",
             month_of_year="*",
         ),
--- a/main/settings.py
+++ b/main/settings.py
@@ -716,6 +716,16 @@ CELERY_TASK_SERIALIZER = "json"
 CELERY_RESULT_SERIALIZER = "json"
 CELERY_ACCEPT_CONTENT = ["json"]
 CELERY_TIMEZONE = "UTC"
+CRON_COURSERUN_SYNC_HOURS = get_string(
+    name="CRON_COURSERUN_SYNC_HOURS",
+    default=0,
+    description="'hours' value for the 'sync-courseruns-data' scheduled task (defaults to midnight)",
+)
+CRON_COURSERUN_SYNC_DAYS = get_string(
+    name="CRON_COURSERUN_SYNC_DAYS",
+    default=None,
+    description="'day_of_week' value for 'sync-courseruns-data' scheduled task (default will run once a day).",
+)
 RETRY_FAILED_EDX_ENROLLMENT_FREQUENCY = get_int(
     name="RETRY_FAILED_EDX_ENROLLMENT_FREQUENCY",
     default=60 * 30,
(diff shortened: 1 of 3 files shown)


rhysyngsun:
If the title is always going to be pulled from openedx, then we should make it readonly in the django admin and put some help text in there noting this.

umarmughal824:
> If the title is always going to be pulled from openedx, then we should make it readonly in the django admin and put some help text in there noting this.

@pdpinch what's your opinion on that?

gsidebo:
If nothing else, I think it would be good to leave the title field alone for local development. It's helpful just to be able to create/update courses locally no matter if they have equivalents in edX or not. I think the only thing we'd be protecting against by setting it to read-only would be a mistaken edit in Django admin et al, which I'd consider an unlikely event, and it wouldn't even make much of a difference in the app since we use the `CoursePage` title, not the `Course` title, in the product detail page

main/settings.py
```diff
@@ -716,6 +716,16 @@
...
 CELERY_ACCEPT_CONTENT = ["json"]
 CELERY_TIMEZONE = "UTC"
+CRON_COURSERUN_SYNC_HOURS = get_string(
+    name="CRON_COURSERUN_SYNC_HOURS",
+    default=0,
+    description="'hours' value for the 'sync-courseruns-data' scheduled task (defaults to midnight)",
+)
+CRON_COURSERUN_SYNC_DAYS = get_string(
+    name="CRON_COURSERUN_SYNC_DAYS",
+    default=None,
```
gsidebo, line 726:
```suggestion
    default="*",
```

If we're defaulting this to run once every day, why not make `"*"` the actual default value for the setting? If you do this, you can replace `CRON_COURSERUN_SYNC_DAYS or "*"` with just `CRON_COURSERUN_SYNC_DAYS`

main/settings.py
```diff
@@ -716,6 +716,16 @@
...
+    description="'day_of_week' value for 'sync-courseruns-data' scheduled task (default will run once a day).",
```
gsidebo, line 727:
```suggestion
    description="'day_of_week' value for scheduled task to sync course run data (by default, it will run each day of the week).",
```

main/settings.py
```diff
@@ -716,6 +716,16 @@
 CELERY_RESULT_SERIALIZER = "json"
 CELERY_ACCEPT_CONTENT = ["json"]
 CELERY_TIMEZONE = "UTC"
+CRON_COURSERUN_SYNC_HOURS = get_string(
+    name="CRON_COURSERUN_SYNC_HOURS",
+    default=0,
+    description="'hours' value for the 'sync-courseruns-data' scheduled task (defaults to midnight)",
```
gsidebo, line 722:
```suggestion
    description="'hours' value for scheduled task to sync course run data (by default, it will run at midnight)",
```

gsidebo (changes requested):
Just a few very minor suggestions

pdpinch:
>>  If the title is always going to be pulled from openedx, then we should make it readonly in the django admin and put some help text in there noting this.

> @pdpinch what's your opinion on that?

There's a case where the title isn't pulled from open edX, and that's when the course exists in the catalog, but not yet on open edX. This could happen when we choose to start marketing a course before dates have been chosen. 

I think the help text is a good idea -- esp. if it's not too hard to mention the update the schedule. But I disagree with making it read-only.

umarmughal824:
> If nothing else, I think it would be good to leave the title field alone for local development. It's helpful just to be able to create/update courses locally no matter if they have equivalents in edX or not. I think the only thing we'd be protecting against by setting it to read-only would be a mistaken edit in Django admin et al, which I'd consider an unlikely event, and it wouldn't even make much of a difference in the app since we use the `CoursePage` title, not the `Course` title, in the product detail page

I had the pretty much same feeling because everytime we require some changes in the title we will have to go to shell and then edit it. For us (developers) it's not an issue but for the marketing team or so, they will feel it tedious work.

umarmughal824:
> > > If the title is always going to be pulled from openedx, then we should make it readonly in the django admin and put some help text in there noting this.
> 
> > @pdpinch what's your opinion on that?
> 
> There's a case where the title isn't pulled from open edX, and that's when the course exists in the catalog, but not yet on open edX. This could happen when we choose to start marketing a course before dates have been chosen.
> 
> I think the help text is a good idea -- esp. if it's not too hard to mention the update the schedule. But I disagree with making it read-only.

In my opinion, in that case we should provide help_text for all the fields (title and dates) and I would put up a separate PR for that.

gsidebo (approved):
Looks good! 👍

PR #173, opened 2021-09-09
Implemented country blocklist at the course level

### LINKED ISSUE #105: course admin improvements
As an admin, I would be more efficient if the course data were displayed differently

#### Acceptance Criteria:

- [ ] show readable ID in list view
- [ ] remove program ID from list view, but keep program filter

### LINKED ISSUE #163: Implement country blocklist at the course level
As a mitxonline administrator or course team member, I'd like to be able to define a set of blocked countries for a course.

We want to be able to prevent course run enrollment for any 

#### Acceptance Criteria:

- [ ] Countries can be added/updated/removed from the blocklist for a `Course` in Django admin
  - All ISO-3166-1 countries should be valid entries
- [ ] If a user attempts to enroll in a course run, and their country is among the blocked countries for that course, the enrollment is not created, they are redirected to the course detail page, and an error message is shown to the user (see: #93)

#### Out of scope:

- Allow blocklists to be updated in Wagtail – This can almost definitely be implemented in a way that lets users update these lists [in Wagtail](https://docs.wagtail.io/en/stable/reference/contrib/modeladmin/index.html), but to keep things simple to start out, this will be done in Django admin only, and permission controls will be introduced so course teams will be unlikely to alter any data they shouldn't have access to.
- "Global" blocklist – We don't need to worry about setting some regions that will be blocked for all courses. Per-course is fine for now.
- Interacting with Cybersource export control API or any equivalent 3rd party service

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #163 
Closes #105 

#### What's this PR do?
- Adds `django-countries` as a dependency
- Adds a model for keeping track of countries that are blocked from enrollment for a given course
- Shows an alert to users who attempt to enroll in a course if their country is blocked
- Fixes some Django admin issues in the courses app

- Pick a user and note their country (`user.legal_address.country`)
- Add a `BlockedCountry` for that user's country and the parent course of a course run you want to enroll in
- Visit the detail page of that course run and click "Enroll Now"

You should see an error message indicating that your country is blocked from enrollment. If you remove that `BlockedCountry` record, it should succeed

ss 2021-09-10 at 24 06 17


arslanashraf7:
There are some conflicts in the PR which we might need to resolve.

gsidebo:
@arslanashraf7 Resolved

static/js/lib/notificationsApi.js
```diff
@@ -44,6 +45,11 @@ export function parseStoredUserMessage(
       SETTINGS.support_email
     }.`
     break
+  case USER_MSG_TYPE_ENROLL_BLOCKED:
+    alertType = ALERT_TYPE_DANGER
+    msgText =
```
arslanashraf7, line 50: Once this message is shown to the user upon clicking the `Enroll now` button, it doesn't fade away automatically nor does it hide when the user navigates to some other page e.g. `HomePage` until after a few seconds which suppose is the default interval for the message.

I think we should reduce the interval in this case.

courses/models.py
```diff
@@ -427,6 +426,18 @@ def save(
         )
 
 
+class BlockedCountry(TimestampedModel):
+    """Represents a country that is blocked from enrollment for a course"""
+
+    course = models.ForeignKey(
+        Course, on_delete=models.CASCADE, related_name="blocked_countries"
+    )
+    country = CountryField()
```
arslanashraf7, line 435: Although it doesn't affect the logic we added to check the blocked country upon enrollment but we can still create multiple entries with the same data for this Model. My opinion is that we make them `unique_together`, not allowing multiple entries.

arslanashraf7 (changes requested):

courses/models.py
gsidebo: Yup, I agree

static/js/lib/notificationsApi.js
gsidebo: I set the expiration to 20 sec to account for the potential of a slow API response. I'm going to create a separate issue to hide the message if it has already been viewed

arslanashraf7 (approved):
Looks good 👍.

PR #202, opened 2021-09-17
Implemented user-supplied usernames

### LINKED ISSUE #191: Allow new users to choose username when creating account
As a user, I'd like to be able to choose my username when I create an account, instead of having one autogenerated using my name. This will allow me to control my identity on open edX discussions forums, and allow course teams to connect users without sharing my PII. 

#### Designs and Mockups

Design (elements): https://projects.invisionapp.com/d/main?origin=v7#/console/21535972/458325008/preview?scrollOffset=1869.5999755859375

#### Acceptance Criteria:

- [ ] Add a field to the account creation form for setting username
- [ ] validate that the username is unique when creating an account. 
  - [ ] for this issue, it's OK if validation is on submit only. We may do a follow up later to validate uniquess on the front-end. 
  - [ ] If it is unique on mitxonline, it should also be unique on edx-platform
- [ ] we should do some manual integration testing with MicroMasters before releasing this, but I would be very surprised if there are any issues

#### Out of Scope

- Auto-generate suggestions for a username. To do this right, we really need a way to do front-end validation on uniqueness, so we're not auto-generating suggestions that are not unique. 
- We should consider allowing users to change their username, but that's out of scope for now. We can handle username changes as a customer service request.

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closes #191 

#### What's this PR do?
Requires users to specify a username during account creation

- Create a new account and follow the link to complete registration
- Try entering a username with special characters (&*^ etc...). You should see an error message after submitting.
- Try entering a username that obeys the validation rules. Your new user should be created with the custom username

- The set of valid username characters comes from edX (https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L1346-L1347)
- Due to some shortcomings with Unicode in JS, it's very much not straightforward to implement front-end validation for this username field in a way that will allow for non-Latin letters, but will not allow special characters (&*^ etc...). For this first implementation, validation is just being done on the back-end

ss 2021-09-17 at 08 10 20
ss 2021-09-17 at 08 32 29
ss 2021-09-17 at 08 32 14
ss 2021-09-17 at 13 03 54


rhysyngsun (changes requested):
I think we need to add a maximum length restriction on the username, if you enter one over 50 characters then I get kicked over to the error page with an ambiguous error message. I think the limit actually needs to be 30 though (per `users.constants.USERNAME_MAX_LEN`), because openedx has a much lower limit.

rhysyngsun (approved):

PR #212, opened 2021-09-22
Removed username from profile edit form

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

N/A - Fixes an oversight from #202 

#### What's this PR do?
Removes the username field from the edit profile form. The username should only be in the form if a new account is being created (just like the password field)

- Create a new account. The register details page should have username and password, and submitting the form should successfully create the user
- Edit a user profile. The username and password fields should be omitted, and saving the form should successfully apply changes to the relevant user fields

ss 2021-09-22 at 11 38 48

### DIFF
--- a/static/js/components/forms/ProfileFormFields.js
+++ b/static/js/components/forms/ProfileFormFields.js
@@ -10,11 +10,13 @@ import {
   EMPLOYMENT_INDUSTRY,
   EMPLOYMENT_LEVEL,
   EMPLOYMENT_SIZE,
-  HIGHEST_EDUCATION_CHOICES,
-  USERNAME_LENGTH
+  HIGHEST_EDUCATION_CHOICES
 } from "../../constants"
 import FormError from "./elements/FormError"
-import { newPasswordFieldValidation } from "../../lib/validation"
+import {
+  newPasswordFieldValidation,
+  usernameFieldValidation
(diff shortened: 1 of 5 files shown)


static/js/components/forms/ProfileFormFields.js
```diff
@@ -66,8 +61,9 @@ export const legalAddressValidation = yup.object().shape({
   })
 })
 
-export const passwordValidation = yup.object().shape({
-  password: newPasswordFieldValidation
+export const newLegalAddressValidation = yup.object().shape({
+  password: newPasswordFieldValidation,
+  username: usernameFieldValidation
 })
```
rhysyngsun, line 67: `newLegalAddressValidation` doesn't seem to match the configuration values, did you mean `newAccountValidation`?

rhysyngsun (changes requested):
We should probably also disallow username updates via the API, I'll leave it to your judgement as to whether you want to address that in this PR or open a ticket and address it later.

static/js/components/forms/ProfileFormFields.js
gsidebo: Good point. I was following the existing naming convention without really thinking about it

rhysyngsun (approved):

PR #231, opened 2021-09-28
Fixed product detail links to in-progress enrolled course runs

### LINKED ISSUE #222: Link users to edX from product detail if run is in progress
### Steps to Reproduce

- Log in
- Enroll in a course run
- Set the course run dates so that the enrollment period is in the past, but the run is in progress according to the start and end dates
- Visit the product detail page for that run

### Expected Behavior

The UI shows the "Enrolled" UI and links the user to the course in edX

### Actual Behavior

If a future course run exists, the "Enroll Now" button is shown for that future course run. Otherwise, the button is omitted completely

### Related Issues

#187 
#207 
#221 

### Background

We migrated a lot of code from mitxpro for this app. In this app, we want the product detail page to link to edX for logged-in users if they are enrolled in an in-progress run, and in mitxpro we either weren't using the product detail page for that purpose, or the bug exists there as well.

gsidebo:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Fixes #222 

#### What's this PR do?
Links users to an edX course run from the product detail page


courses/api.py
```diff
@@ -50,6 +56,48 @@
...
+            ),
+        )
+        run_qset = run_qset.annotate(user_enrollments=user_enrollments).order_by(
+            "-user_enrollments", "enrollment_start"
+        )
+        runs = (
+            run
+            for run in run_qset
+            if run.user_enrollments > 0
+            or (run.enrollment_end is None or run.enrollment_end > now)
```
arslanashraf7, line 89: Probably we want to use `run.enrollments` based on the current user because the scenario is currently failing based on the below scenario:
1. Enroll a user in a Course Run(e.g Run1)
2. Set the start date of the Run1 in past
3. Set the end date of the Run1 in future
4. set enrollment start & end date in past

In the above scenario, the `Enrolled` button or any button doesn't show. However, in this case, the current run of the user is still in progress and `Enrolled` button should show and link the user to the course.

arslanashraf7 (changes requested):
I've added a comment regarding the `Enrolled` button.

courses/api.py
gsidebo: @arslanashraf7 That's what this PR was designed to do. I think your enrollment record might not have `active=True` or `edx_enrolled=True`. In either case, the query on the backend would ignore your enrollment in the in-progress course run since we shouldn't be linking you to edX if your enrollment is inactive or it didn't go through in edX. I just tested locally and it's working as expected. If that doesn't work for you, or if I'm not understanding you correctly, let me know.

arslanashraf7: You are correct, Thanks for responding with details. In my case  I somehow unchecked the `edx_enroled` for the course run and due to this, I was facing the above issue. But correcting that makes the uses cases work as well.

arslanashraf7 (approved):
👍 Looks good.

Issue #219, opened 2021-09-23
update sentry


pdpinch:
This project is on sentry.javascript.browser@v5.5.0
Update to sentry.javascript.browser@v6.13.2

This project is on sentry.python@v1.1.0
Update to sentry.python@v1.4.1

arslanashraf7:
Closed via #232

PR #232, opened 2021-09-28
build: upgrade sentry browser and sdk version + RedisIntegration

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#219 

#### What's this PR do?
- Upgrades `sentry/browser` JS version to `6.13.2`
- Upgrades `sentry-sdk` version to `1.4.2`
- Adds RedisIntegration in sentry login. (Note: This wasn't mentioned in this ticket but i remember sentry once notifying us to add RedisIntegration as well for better login in another PR we did in MM (ref: https://github.com/mitodl/micromasters/pull/4940)) So I added it along

- Verify that sentry is working properly for both backend/frontend and the logs are being sent to sentry DSN.


requirements.txt
```diff
@@ -280,7 +280,7 @@ requests-oauthlib==1.3.0
     # via social-auth-core
 s3transfer==0.3.7
     # via boto3
-sentry-sdk==1.1.0
+sentry-sdk==1.4.2
```
asadiqbal08, line 283: I believe, `.txt` is auto generated file by pip-compile, you need to touch the `.in` one

asadiqbal08 (changes requested):

requirements.txt
arslanashraf7: @asadiqbal08 , Since `sentry-sdk` is not pinned in the `.in` (https://github.com/mitodl/mitxonline/blob/main/requirements.in#L32) and I kept it that way and just upgraded the `sentry-sdk` to the latest with `pip-compile` that's why there is no change in `.in`. 
Let me know if this answers your concern?

asadiqbal08: why is it not pinned in `.in` file ?

arslanashraf7: @asadiqbal08 After the discussion we had on a call, I've pinned sentry's version. You can take a look at the PR now as it's ready for review again.

asadiqbal08 (approved):

PR #236, opened 2021-09-29
Fixed user notifications so they are only seen once

### LINKED ISSUE #193: Prevent showing a user message if it has already been viewed
We use cookies to show banner messages in the React app that originate from the back end (e.g.: #143). The expiration for those cookies is set to 20 seconds to account for a possibly slow API response, but this means that if the response comes back quickly and the user reloads or navigates to other pages in the app, they will see the message after they've already closed it. We should prevent these messages from being seen after they've already been shown for the first time. We may want to look into using localStorage for this

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixed #193

#### What's this PR do?
Prevents repeated displays of a user notification that came from the backend

- Enroll a user in a course run via the product detail page
  - When the enrollment completes, you should be redirected to a new page and see a success banner message
- Navigate to the home page right away (with or without dismissing the notification)

You should not see the success message. On the main branch, this message would be shown every time you loaded a new page within the cookie's life span (20 seconds)


static/js/containers/HeaderApp.js
```diff
@@ -5,11 +5,15 @@ import { connect } from "react-redux"
...
-
 import Header from "../components/Header"
+import { addUserNotification } from "../actions"
+import users, { currentUserSelector } from "../lib/queries/users"
+import {
+  getStoredUserMessage,
+  removeStoredUserMessage
+} from "../lib/notificationsApi"
+import { removeCookie } from "../lib/api"
+import { USER_MSG_COOKIE_NAME } from "../constants"
```
arslanashraf7, line 16: I think we should remove these imports as they are unused.

static/js/lib/notificationsApi.js
```diff
@@ -1,5 +1,8 @@
 /* global SETTINGS: false */
-import { getCookie } from "./api"
+import { isNil } from "ramda"
```
arslanashraf7, line 2: This is an unused import and we should remove it.

static/js/containers/HeaderApp_test.js
```diff
@@ -8,13 +8,17 @@ import * as notificationsApi from "../lib/notificationsApi"
 import { ALERT_TYPE_TEXT } from "../constants"
```
arslanashraf7, line 8: I know it's not a part of this PR but it's also unused.

arslanashraf7 (approved):
Looks good 👍 . Just a few minor comments regarding imports optimization. Feel free to merge after they are addressed.

PR #244, opened 2021-10-04
Added username whitespace trimming and case-insensitive unique validation

### LINKED ISSUE #242: Validate username uniqueness based on trimmed, lowercase version
We've discovered that edX does not consider a username unique if the trimmed, lowercase version of that username already exists. We need to update our validation to ensure that the same logic is applied in mitxonline.

**Acceptance criteria:**
- API function to create a new user fails if there is already a username that matches the one entered by the user after that value has been coerced to lowercase and stripped of whitespace at the ends of the string
- Username value is stripped of whitespace before saving 
- If the user includes capital letters in their username, the username value in the database still has those capital letters

gsidebo:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes #242 

#### What's this PR do?
Trims whitespace around usernames during account creation, and performs a case-insensitive match for existing usernames

- Create an account, and input a unique username with spaces around it. The saved User should have a username with whitespace trimmed
- Create an account, and input a username that matches an existing user, but change the case of some/all of the letters. You should see an error message indicating that the username already exists

### DIFF
--- a/authentication/api_test.py
+++ b/authentication/api_test.py
@@ -16,33 +16,6 @@ def test_create_user_session(user):
     assert session.session_key is not None
 
 
-def test_create_user_with_generated_username(mocker, valid_address_dict):
-    """
-    Integration test to assert that create_user_with_generated_username tries to find an available
-    username and try again to save a User if there was a username collision
-    """
-    username = "testuser"
-    # Create a user with the desired username before calling the function so we get a collision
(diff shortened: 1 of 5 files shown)


gsidebo:
Looking at the test failures now

rhysyngsun (approved):

Issue #237, opened 2021-09-29
Babel/polyfill warning in console


pdpinch:
### Steps to Reproduce

1. Load the home page (or any page?)
2. Open the JavaScript console in Firefox (or any browser?)

### Expected Behavior

- no warnings

### Actual Behavior

```
@babel/polyfill is loaded more than once on this page. This is probably not desirable/intended and may have consequences if different versions of the polyfills are applied sequentially. If you do need to load the polyfill more than once, use @babel/polyfill/noConflict instead to bypass the warning.
```

PR #248, opened 2021-10-06
fix: remove multiple instance loading of polyfill

arslanashraf7:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

#237 

#### What's this PR do?
- It removes the warning on browsers regarding multiple instances loaded of polyfill
- It removes the polyfill package loading in webpack for django since the polyfill will already be loaded through `root.js` that is mentioned above the django.js. And that bundle is rendered in `index.html`

- Check that everything loads fine and there are no warnings/errors regarding webpack/polyfill

Checking the details of the warning in the browser

In an earlier PR(https://github.com/mitodl/mitxpro/pull/2086), we combined two bundles into one and added an initial polyfill load in the webpack configuration. Since MitxOnline code is ported from xPRO so we saw this warning here as well.


umarmughal824 (approved):
👍

PR #255, opened 2021-10-14
Load enrollment status dynamically in product detail page

### LINKED ISSUE #253: Load enrollment status dynamically in product detail page
The product detail page is being cached in deployed environments, which is a problem since it contains user-specific content (the "Enroll Now"/"Enrolled" button). If we change that page to load the enrollment status dynamically after the page is rendered, then we can cache the page without worrying about the user-specific portion being cached.

### Implementation
The naive way to implement this would be to create an entirely new React app and entry point (like `App.js`/`root.js`, or `HeaderApp.js`/`header.js`). Unfortunately this would create a very large additional JS bundle. We should instead look into adding on to `header.js`, which is the entry point for CMS-based pages, and render into an additional element on the page (all of our current entry points in all of our projects render into one element each).

In order to make that work, we'll need to make some changes to the course/course run API and 

**courses/views/v1/__init__.py**
Since we're going to be making a request for a relevant course run, it makes sense to call the existing course runs API and add onto it so we can just fetch the run that is relevant to the current user.

- Add `courses.api.get_user_relevant_course_run_qset` function, and move over all the logic from `get_user_relevant_course_run` except for the code at the end which returns the first element or none. This function can just return the queryset
- Expand the `CourseRunViewSet` so we can make a request to `/api/course_runs?relevant_to=<course readable id>`. 
  - Define a `get_queryset` method instead of setting the `queryset` property. When the `relevant_to` param is passed in, use the queryset returned from `get_user_relevant_course_run_qset`. Otherwise just use `.all()`
  - Add a `get_serializer_context` method, and if `relevant_to` is in the request querystring, pass in a flag to tell the serializer to include enrollment data (`return {"include_enrolled_flag": True}`). You can grep the codebase for other `get_serializer_context` examples
- Update `CourseRunSerializer` to include an `is_enrolled` flag if the context asks for it
  - We can do this by adding a `to_representation` method. Check `courses.serializers.BaseCourseSerializer.to_representation` for another example
  - If the `include_enrolled_flag` is true in the context, include the `is_enrolled` flag with the rest of the data.
    
    ```python
    data = super().to_representation(instance)
    if self.context.get("include_enrolled_flag"):
        return {**data, **{"is_enrolled": getattr(instance, "user_enrollments", 0) > 0}}
    return data
    ```

**product_page.html**
```
{% block content %}
...
              <div class="text">
                <h1>{{ page.title }}</h1>
                {# Description field contents are already rendered wrapped in a <p> tag #}
                {{ page.description | richtext }}
                <input id="courseId" type="hidden" value="{{page.course.readable_id}}" />
                <div id="productDetailEnrollment"></div>
              </div>
...
```

**static/js/entry/header.js**
```tsx
const rootEl = document.getElementById("header")
const enrollSectionEl = document.getElementById("productDetailEnrollment")
const courseIdEl = document.getElementById("courseId")

const renderHeader = () => {
  //...
}

const renderEnrollSection = (courseId) => {
  ReactDOM.render(
    <AppContainer>
      // NOTE: This may also need a <Provider store={store}> component, with a separate store initialized
      <ProductDetailEnrollApp courseId={courseId} />
    </AppContainer>,
    enrollSectionEl
  )
}

renderHeader()

if (enrollSectionEl && courseIdEl) {
  const courseId = courseIdEl.value
  renderEnrollSection(courseId)
}
```

**ProductDetailEnrollApp.js (new component)** 
- Upon loading, make a request to the API to get the relevant run for the logged-in user
  - The response should also indicate whether or not the user is enrolled in that course run (including cases where the user is not logged in)
  - We'll want to add a new type to represent the `CourseRun` with an additional enrollment flag. Something like `EnrollmentFlaggedCourseRun`
- While the API call is in progress, we should render a loading spinner (see `Loader.js`)
- Move the rendering logic for the enrollment button from `product_page.html` to this new component. The response from the API call should have all the necessary data for rendering the button in the correct state

umarmughal824:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

fixes #253 

#### What's this PR do?
Load enrollment status dynamically in product detail page

Just visit the course / product page

Screenshot 2021-10-18 at 17 13 32
Screenshot 2021-10-18 at 17 13 45
Screenshot 2021-10-18 at 17 13 59

### DIFF
--- a/cms/templates/product_page.html
+++ b/cms/templates/product_page.html
@@ -20,33 +20,8 @@
                 <h1>{{ page.title }}</h1>
                 {# Description field contents are already rendered wrapped in a <p> tag #}
                 {{ page.description | richtext }}
-                {% if run %}
-                  {% if is_enrolled %}
-                    {% if run.courseware_url_path and can_access_edx_course %}
-                      <a href="{{ run.courseware_url }}" class="btn btn-primary btn-gradient-red highlight outline" target="_blank" rel="noopener noreferrer">
-                        Enrolled &#10003;
-                      </a>
-                    {% else %}
-                      <div class="btn btn-primary btn-gradient-red highlight outline">
-                        Enrolled &#10003;
-                      </div>
-                    {% endif %}
-                  {% else %}
-                    {% if sign_in_url %}
-                      <a href="{{ sign_in_url }}" class="btn btn-primary btn-gradient-red highlight">
-                        Enroll now
-                      </a>
-                    {% else %}
-                      <form action="/enrollments/" method="post">
-                        {% csrf_token %}
-                        <input type="hidden" name="run" value="{{ run.id }}" />
-                        <button type="submit" class="btn btn-primary btn-gradient-red highlight">
-                          Enroll now
-                        </button>
-                      </form>
-                    {% endif %}
-                  {% endif %}
-                {% endif %}
+                <input id="courseId" type="hidden" value="{{page.course.readable_id}}" />
+                <div id="productDetailEnrollment"></div>
               </div>
             </div>
             {% if page.video_url %}
--- a/courses/api.py
+++ b/courses/api.py
@@ -11,6 +11,7 @@ from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import transaction
 from django.db.models import Q, Count
+from django.db.models.query import QuerySet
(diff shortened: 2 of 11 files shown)


gsidebo (approved):
Looks/works great! 👍  Thanks for your attention to this

Issue #259, opened 2021-10-15
Normalize accented characters when validating username uniqueness


gsidebo:
### Steps to Reproduce

1. Create a user with some vowels in the username
2. Create another user with the same username, but change some/all of the characters to their accented equivalents (e.g.: `online-user` and `önlínê-user`)

### Expected Behavior

The users exist in both mitxonline and Open edX

### Actual Behavior

For unexplained reasons, edX treats both of those usernames as the same, so only the more recent user will have a User record on the Open edX side, and one or both users will experience some of the auth/enrollment issues that we saw before we patched up username creation and validation ()

### Related Issues

#202 
https://github.com/mitodl/mitxonline/pull/244

briangrossman:
Note that when we were looking into some other issues with usernames, we discovered a few users who have a username conflict due to the fact that there are some usernames that differ only by accents. They can be found in the `Accent Issues` section of [this document](https://docs.google.com/document/d/1s9NB-JwE9WTH-jvuTahl5qI4dn91YNy_0qIykzr9C_M/edit#heading=h.o3cr71xfs4l3)

PR #262, opened 2021-10-20
Restrict user to use accented characters in username

asadiqbal08:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

fixes: #259 

#### What's this PR do?
(Required)

Create a user with username that may have accented characters in it e.g. `önlínê-user`. The system should not allow it.

##### NOTE: 

~~- I did not update the message `Username can only contain letters, numbers, spaces, and the following characters: @_+-` , seems of enough but if needs to modify then let me know.~~

### DIFF
--- a/users/serializers.py
+++ b/users/serializers.py
@@ -28,9 +28,9 @@ USER_GIVEN_NAME_RE = re.compile(
     """,
     flags=re.I | re.VERBOSE | re.MULTILINE,
 )
-USERNAME_RE_PARTIAL = r"[\w .@_+-]+"
+USERNAME_RE_PARTIAL = r"[A-Za-z0-9.@_+-]+"
 USERNAME_RE = re.compile(fr"(?P<username>{USERNAME_RE_PARTIAL})")
-USERNAME_ERROR_MSG = "Username can only contain letters, numbers, spaces, and the following characters: @_+-"
+USERNAME_ERROR_MSG = "Username can only contain letters, numbers and the following characters: @_+-"
 
 
 class LegalAddressSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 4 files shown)


users/serializers.py
```diff
@@ -28,7 +28,7 @@
     """,
     flags=re.I | re.VERBOSE | re.MULTILINE,
 )
-USERNAME_RE_PARTIAL = r"[\w .@_+-]+"
+USERNAME_RE_PARTIAL = r"[ A-Za-z0-9.@_+-]+"
```
arslanashraf7, line 31: Although functionality wise it works fine and doesn't allow accented characters, I have two minor things that you might consider here:

1. I'm sure there could be a reference to using the space in the username which is unusual but if we do want to use it we should place that somewhere close to the allowed special characters list e.g. `r"[A-Za-z0-9 .@_+-]+"`
2. Since we are matching the username text through a regex, shouldn't we add this regex validation on the frontend validator(https://github.com/mitodl/mitxonline/blob/main/static/js/lib/validation.js#L20) and we can save us an API call that is just going to tell the users about what characters they can use after hitting the submit button.

arslanashraf7 (changes requested):

users/serializers.py
asadiqbal08: @arslanashraf7 regarding the point.2, 

- It makes sense to me that there should be `frontend validation` earlier before sending an api call. 
- `space` in username does not make sense to me as well. I keep it by considering it may be a requirement earlier but I have removed it and take input from @gsidebo if `space` really makes sense to be part of `username`.

kindly take a look over another commit.

gsidebo:
Closing this so we can re-implement according to the details in the updated issue (#259)

PR #281, opened 2021-11-02
Use SVG for the MIT logo

### LINKED ISSUE #270: Use SVG for the MIT logo
https://upload.wikimedia.org/wikipedia/commons/0/0c/MIT_logo.svg (this probably exists somewhere in the mit.edu domain as well)

The site logo is 3.9kb. Not huge, but it doesn't make much sense to require the browser to load a separate image when we can just use an inline SVG.

umarmughal824:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

fixes #270 

#### What's this PR do?
Use SVG for the MIT log


gsidebo (changes requested):
If we use an [inline SVG](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_In_HTML_Introduction) and add in whatever [accessibility attributes](https://css-tricks.com/accessible-svgs/#2-inline-svg) we need, we can avoid the need for an extra HTTP request. You can just create a new component that returns the SVG contents and call it something like `InstituteLogo`

gsidebo (approved):
Looks great!! Just one nitpick: the `<a>` link that wraps the svg has a little extra space on the top, which makes the logo off-center in mobile. If you add a class to that link (something like `logo-link`) then style that class with `display: inline-flex` in `top-app-bar.scss`, the logo will be perfectly centered in mobile widths.

Once you make that change, feel free to merge 👍

Issue #274, opened 2021-10-29
'Enroll now' button appears when 'Enrollment start' date is in the future


briangrossman:
### Steps to Reproduce

- Create a course run with `enrollment start` date and `enrollment end` date in the future
- Make sure there are no runs for this course that currently have open enrollments
- View the course page (e.g. https://rc.mitxonline.mit.edu/courses/course-v1:MITx+14.100x/)

### Expected Behavior

- The `Enroll now` button should only appear if there is a course that is currently open for enrollment 
- (i.e. `enrollment start < current date < enrollment end`)

### Actual Behavior

- The `Enroll now` button appears
- Testing note: I've seen some caching of the `Enroll now` button in some cases. Be mindful while testing. 

![image](https://user-images.githubusercontent.com/21342568/139478741-08912732-d2b4-47ce-8b15-c9a167f8f803.png)

umarmughal824:
it's pre-requisite is #255

PR #282, opened 2021-11-03
Fixing: 'Enroll now' button appears when 'Enrollment start' date is in the future

asadiqbal08:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

fixes: #274 

#### What's this PR do?
The Enroll now button should only appear if there is a course that is currently open for enrolment and `enrolment start < current date`

- Create course run with start_date in future.
- visit the course page.
- Enroll now button should not be appeared.

### DIFF
--- a/static/js/containers/ProductDetailEnrollApp.js
+++ b/static/js/containers/ProductDetailEnrollApp.js
@@ -16,6 +16,8 @@ import {
   courseRunsQueryKey
 } from "../lib/queries/courseRuns"
 
+import { isWithinEnrollmentPeriod } from "../lib/courseApi"
+
 import { getCookie } from "../lib/api"
 
 type Props = {
@@ -60,7 +62,7 @@ export class ProductDetailEnrollApp extends React.Component<Props> {
               >
                 Enroll now
               </a>
-            ) : run && now.isAfter(moment(run.enrollment_start)) ? (
+            ) : run && isWithinEnrollmentPeriod(run) ? (
               <Fragment>
                 <form action="/enrollments/" method="post">
                   <input
--- a/static/js/containers/ProductDetailEnrollApp.js
+++ b/static/js/containers/ProductDetailEnrollApp.js
@@ -60,8 +60,7 @@ export class ProductDetailEnrollApp extends React.Component<Props> {
               >
                 Enroll now
               </a>
-            ) : run && 
-                now.isAfter(moment(run.enrollment_start)) ? (
+            ) : run && now.isAfter(moment(run.enrollment_start)) ? (
               <Fragment>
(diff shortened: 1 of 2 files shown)


static/js/containers/ProductDetailEnrollApp.js
```diff
@@ -58,7 +60,7 @@ export class ProductDetailEnrollApp extends React.Component<Props> {
               >
                 Enroll now
               </a>
-            ) : (
+            ) : run && now.isAfter(moment(run.enrollment_start)) ? (
```
umarmughal824, line 63: The current date should be in between enrollment start and end date

umarmughal824 (changes requested):

static/js/containers/ProductDetailEnrollApp.js
asadiqbal08: @umarmughal824 I was trying to produce it but as per my understanding if the enrollment end_date is reached then we will get an empty run. 
Can you make sure, you are checking the `Enroll Now` button with the logged in user ?

asadiqbal08: We have an `Enroll Now` labelled on line `61`

gsidebo (changes requested):
@asadiqbal08 I created a utility function in [this PR](https://github.com/mitodl/mitxonline/pull/287) called `isWithinEnrollmentPeriod` and wrote some tests for it (I also have an example of mocking it out for some dashboard tests). I think that function can be used here. This PR is smaller and older so it would probably merge before that other PR. Would you mind bringing that function into this PR and using it for the logic that displays the Enroll Now link?

static/js/containers/ProductDetailEnrollApp_test.js
```diff
@@ -70,6 +80,59 @@ describe("ProductDetailEnrollApp", () => {
     )
   })
 
+  it("checks for enroll now button should not appear if enrollment start in future", async () => {
```
gsidebo, line 83: These two tests can be easily combined into one parametrized test

```javascript
;[[false, false], [true, true]].forEach(([enrollable, expButton]) => {
    it(`enroll now button ${shouldIf(expButton)} appear if run ${isIf(enrollable)} enrollable`, async () => {
      isWithinEnrollmentPeriodStub.returns(enrollable)
      // ...
      const btn = inner.find("button")
      assert.equal(btn.exists(), expButton)
      if (expButton) {
        assert.equal(btn.text(), "Enroll Now")
      }
    })
})
```

gsidebo (changes requested):

gsidebo (approved):
Looks great! 👍

Issue #276, opened 2021-10-29
Accessibility: Created account initial form submission not apparent to screen readers


briangrossman:
### Steps to Reproduce

When a screen reader user submits the first account creation form (where they enter their email), they do not get an alert to know that a message "We've sent you email" appears. They do not know that the text on the screen has been updated.

### Expected Behavior

The focus should shift when the page changes to the new content such that the screen reader is redirected to the correct element to read it. In this case, one could shift focus to the h1 "thank you etc" followed by instructions to check eMail, etc. That should be enough.

From a tester:

> So, what seems to happen is that when you press continue, the view is refreshed via javascript in the browser. This means that the screen reader will not react unless pushed. In normal web interactions (well, what is normal these days is up for debate), a request is made to the server, and the server responds with a bunch of HTML, and the browser erased the current document and refreshed it with the stuff just sent from the server. Often times, a lot of the stuff is the same as what was already on screen, so when you look at the result only part of the page has changed. In reality, the entire thing has been refreshed. In many apps / websites today, instead of making the round trip to the server, the browser itself, via the javascript running as part of the page, responds locally and manipulates only the part of the page that needs to change. This saves bandwidth and time, but the screen reader has no idea that this took place.
> 
> Easiest way to fix that is to shift focus to something on the page. I would just:
> 
> mark the h1 as focusable by adding tabindex="-1" to it;
> shift focus to it once the conformation message has been posted to the view;

### Actual Behavior

Nothing is spoken. No indication is given to the user that the screen has changed.

### Related Issues

https://github.com/mitodl/mitxonline/issues/153 (potentially related)

### Screenshot or Screencast

![image](https://user-images.githubusercontent.com/21342568/139499554-8ad44bcc-5d13-480c-8676-dd73e1d1d9ca.png)

briangrossman:
@asadiqbal08 
I'm reopening this issue. It was just tested for accessibility and there was a request for some improvement. I'm copying the feedback in here to ensure the details aren't lost:

> this is not quite fixed. the thank you page comes up, but focus goes to the body element, rather than the h1.
> 
> Rich shares: The solution that seemed to work most consistently on both Firefox and Chrome with NVDA is:

```
<div autofocus tabindex="0">
<h1>Thank you</h1>
<p>Thank you for creating an account.
If you don't get your confirmation eMail ...
</p></div>

```

[...]

PR #286, opened 2021-11-05
asadiqbal08/ Fix accessibility issue by tabindex to header

asadiqbal08:
@briangrossman can you verify it, Is it good enough as shown in image below.

#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

fixes: #276 

#### What's this PR do?
When a screen reader user submits the first account creation form (where they enter their email), they do not get an alert to know that a message "We've sent you email" appears. They do not know that the text on the screen has been updated.
The focus should shift when the page changes to the new content such that the screen reader is redirected to the correct element to read it. In this case, one could shift focus to the h1 "thank you etc" followed by instructions to check eMail, etc. That should be enough.

Test the above criteria while creating an account.

Screenshot 2021-11-05 at 4 01 42 PM

Screenshot 2021-11-05 at 4 03 07 PM

### DIFF
--- a/static/js/containers/pages/register/RegisterConfirmSentPage.js
+++ b/static/js/containers/pages/register/RegisterConfirmSentPage.js
@@ -27,7 +27,7 @@ export class RegisterConfirmSentPage extends React.Component<Props> {
         <div className="std-page-body container auth-page">
           <div className="auth-card card-shadow auth-form">
             <div className="auth-header">
-              <h1>Thank you!</h1>
+              <h1 id="thanks-msg" tabIndex="0" >Thank you!</h1>
             </div>
             <p>
               We sent an email to <b>{email}</b>, please verify your address to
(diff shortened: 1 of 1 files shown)


briangrossman:
@asadiqbal08 
Based on what I'm seeing in the screenshots, it looks like the focus switches to the `Thank you!` after submitting. If that is the behavior, this looks good to me.

static/js/containers/pages/register/RegisterConfirmSentPage.js
```diff
@@ -27,7 +27,9 @@ export class RegisterConfirmSentPage extends React.Component<Props> {
         <div className="std-page-body container auth-page">
           <div className="auth-card card-shadow auth-form">
             <div className="auth-header">
-              <h1>Thank you!</h1>
+              <h1 id="thanks-msg" tabIndex="0">
```
umarmughal824, line 30: do you really think that id is needed as we don't have a specific use of it neither is CS nor somewhere else?

umarmughal824 (approved):
👍

PR #326, opened 2021-12-07
Added app for ecommerce, Products model, admin bindings

### LINKED ISSUE #298: Product Subsystem: Data Models
The product subsystem is responsible for tracking all product-related data. Purchasable products are typically Programs and Course Runs. Pricing information is tracked as immutable data for the sake of historically accurate pricing for orders.

### Data Models

We'll use [`django-reversion`](https://django-reversion.readthedocs.io/en/stable/) to allow for versioned product data.

```python
import reversion

@reversion.register(
    exclude=("content_type", "object_id", "created_on", "updated_on"
)
class Product(TimestampedModel):
    """
    Representation of a purchasable product. There is a GenericForeignKey to a Course or Program.
    """

    content_type = models.ForeignKey(ContentType)
    object_id = models.PositiveIntegerField()
    purchasable_object = GenericForeignKey("content_type", "object_id")

    price = models.DecimalField()
    description = models.TextField()
```

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

fixes #298

#### What's this PR do?
Adds the Product table (and adds stuff to expose the table through django-admin)
Adds app environment for ecommerce 
Adds dependency on django-reversion

After running migrations, create/update entities through admin.


jkachel:
[James Kachel](https://trello.com/c/ywTx7A7w/8204-james-kachel)

ecommerce/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
rhysyngsun, line 3: You can delete this and `views.py` since they're unused.

ecommerce/models.py
```diff
@@ -0,0 +1,33 @@
...
+    exclude=("content_type", "object_id", "created_on", "updated_on")
+)
+class Product(TimestampedModel):
+    """
+    Representation of a purchasable product. There is a GenericForeignKey to a 
+    Course Run or Program Run.
+    """
+
+    valid_purchasable_objects=models.Q(app_label='courses', model='courserun') | models.Q(app_label='courses', model='programrun')
+    content_type=models.ForeignKey(ContentType, on_delete=models.CASCADE, limit_choices_to=valid_purchasable_objects)
```
rhysyngsun, line 17: Could you change the value passed to `limit_choices_to` to be a function returning the `Q` fragment instead? That has the benefit that if we ever modify this query it doesn't trigger a new migration to be generated since Django will just serialize the function name into the migration instead of the query itself.

rhysyngsun (changes requested):
It looks like our code formatting checks didn't get added to the GitHub Actions workflow when we setup this projects, but could you run the code formatter just on your new code (`black ecommerce/`)? I suspect there's probably other format issues elsewhere so I'll put up a PR to format the whole project and add the check to the tests separately.

ecommerce/admin.py
```diff
@@ -0,0 +1,13 @@
+from django.contrib import admin
+from ecommerce.models import Product
+
+
+class ProductAdmin(admin.ModelAdmin):
```
rhysyngsun, line 5: Could you update this to use django-reversion's admin?

https://django-reversion.readthedocs.io/en/stable/admin.html

rhysyngsun (changes requested):

rhysyngsun:
Functionality is looking good, I think this last comment above will be it.

jkachel:
Last commit adds in the reversion stuff to the dj admin (and adds in some steps I missed when including reversion too, so there's a migration and bootstrapping for the library that'll need to be run).

rhysyngsun (approved):

rhysyngsun:
LGTM :+1:

Issue #305, opened 2021-11-19
Product Subsystem: APIs


rhysyngsun:
The API that would be primarily needed would be one to read back product data. It is presumed that data entry is done through django-admin:

GET /api/v0/products/ -> returns a paginated list of products
GET /api/v0/products/1/ -> returns a single product

PR #337, opened 2021-12-14
Adding product subsystem apis

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#305 

#### What's this PR do?
Adds in the REST API for Product retrieval (and adds the scaffolding for that)

Use the standard DRF API explorer (go to /api/v0/products, etc.) once at least one Product has been created (via dj-admin)
pytest tests should also pass

### DIFF
--- a/ecommerce/views/v0/__init__.py
+++ b/ecommerce/views/v0/__init__.py
@@ -1,0 +1,86 @@
+"""
+MITxOnline ecommerce views
+"""
+
+from rest_framework.response import Response
+from rest_framework.views import APIView
+from rest_framework.viewsets import ReadOnlyModelViewSet
+from rest_framework.pagination import LimitOffsetPagination
+from django.db.models import Q, Count
+from mitol.common.utils import now_in_utc
+
+from courses.models import (
+    CourseRun,
+    Course,
+    Program,
+    ProgramRun,
+)
+
+from ecommerce.serializers import ProductSerializer
+from ecommerce.models import Product
+
+
+class ProductsPagination(LimitOffsetPagination):
+    default_limit = 2
+    limit_query_param = "l"
+    offset_query_param = "o"
+    max_limit = 50
+
+
+class ProductViewSet(ReadOnlyModelViewSet):
+    serializer_class = ProductSerializer
+    pagination_class = ProductsPagination
+
+    def get_queryset(self):
+        now = now_in_utc()
+
+        unenrollable_courserun_ids = CourseRun.objects.filter(
+            enrollment_end__lt=now
+        ).values_list("id", flat=True)
+
+        unenrollable_course_ids = (
+            Course.objects.annotate(
+                num_runs=Count(
+                    "courseruns", filter=~Q(courseruns__in=unenrollable_courserun_ids)
+                )
+            )
+            .filter(num_runs=0)
+            .values_list("id", flat=True)
+        )
+
+        unenrollable_program_ids = (
+            Program.objects.annotate(
+                valid_runs=Count(
+                    "programruns",
+                    filter=Q(programruns__end_date__gt=now)
+                    | Q(programruns__end_date=None),
+                )
+            )
+            .filter(
+                Q(programruns__isnull=True)
+                | Q(valid_runs=0)
+                | Q(courses__in=unenrollable_course_ids)
+            )
+            .values_list("id", flat=True)
+            .distinct()
+        )
+
+        unenrollable_programrun_ids = ProgramRun.objects.filter(
+            Q(program__in=unenrollable_program_ids) | Q(end_date__lt=now)
+        )
+
+        return (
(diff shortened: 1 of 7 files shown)


ecommerce/views/v0/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+
+        expired_courses = (
+            Course.objects.annotate(
+                runs=Count("courseruns", filter=~Q(courseruns__in=expired_courseruns))
+            )
+            .filter(runs=0)
+            .values_list("id", flat=True)
+        )
+
+        expired_programs = (
```
annagav, line 48: I am not sure about the terminology? I feel like expired could mean many different things, like it is not enrollable anymore, or it has finished, or things like that. In MM we used the word 'past' to refer to end date being in the past.

ecommerce/views/v0/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+                Q(programruns__isnull=True)
+                | Q(valid_runs=0)
+                | Q(courses__in=expired_courses)
+            )
+            .values_list("id", flat=True)
+            .distinct()
+        )
+
+        return (
+            Product.objects.exclude(
```
annagav, line 66: Are we sure we want to exclude past runs of a product? If this queryset is returned for users enrollment info I feel like we would want all of them, right? I might be missing on some context though.

annagav (changes requested):
I left a couple comments, but I feel maybe they I more for @rhysyngsun

ecommerce/views/v0/__init__.py
jkachel: Just to add context - my interpretation from the issue was that these API calls would be used to display currently-purchasable items, Pulling enrollment data or order history or things along those lines would be handled elsewhere.

ecommerce/views/v0/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+    limit_query_param = 'l'
+    offset_query_param = 'o'
+    max_limit = 50
+
+class ProductViewSet(ReadOnlyModelViewSet):
+    serializer_class = ProductSerializer
+    pagination_class = ProductsPagination
+
+    def get_queryset(self):
+        now = datetime.datetime.now(tz=pytz.UTC)
```
rhysyngsun, line 34: Can be `mitol.common.utils.datetime.now_in_utc()`

ecommerce/views/v0/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+
+        expired_courseruns = CourseRun.objects.filter(
+            enrollment_end__lt=now
+        ).values_list("id", flat=True)
+
+        expired_courses = (
+            Course.objects.annotate(
+                runs=Count("courseruns", filter=~Q(courseruns__in=expired_courseruns))
```
rhysyngsun, line 42: Nit: `num_runs` might be clearer.

ecommerce/views/v0/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+    max_limit = 50
+
+class ProductViewSet(ReadOnlyModelViewSet):
+    serializer_class = ProductSerializer
+    pagination_class = ProductsPagination
+
+    def get_queryset(self):
+        now = datetime.datetime.now(tz=pytz.UTC)
+
+        expired_courseruns = CourseRun.objects.filter(
```
rhysyngsun, line 36: Nit:
```suggestion
        expired_courserun_ids = CourseRun.objects.filter(
```

ecommerce/views/v0/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+            enrollment_end__lt=now
+        ).values_list("id", flat=True)
+
+        expired_courses = (
```
rhysyngsun, line 40: Nit:
```suggestion
        expired_course_ids = (
```

ecommerce/views/v0/__init__.py
rhysyngsun: `unenrollable_program_ids` is probably what I'd go with

ecommerce/views/v0/__init__.py
rhysyngsun: I think this is ok as-is for now. If we want to add additional filtering based on a user's enrollments I think we'd want to do that with some optional filtering, but I don't think that's worth adding until we have a use case for it.

rhysyngsun (changes requested):

jkachel:
c1c06dd2674f4011825462737faa58a10e46ad39 updates the wording in a handful of places for clarity and fixes some logic errors in the product filtering code (which was trying to filter on programs rather than program runs).

rhysyngsun (approved):
LGTM

PR #344, opened 2021-12-20
Addded Discount models

### LINKED ISSUE #309: Discount Subsystem: Data Models
The discount system would support discounts of multiple types. We’ve done discounts a lot of different ways before so we need to balance out flexibility against keeping complexity down. Each discount would be associated with a certain Product.

A discount may optionally be pre-associated with a User so that it can be automatically applied on checkout.

Discounts should only be computed on the backend, some of our ecommerce implementations have computed the discount on the frontend and we want to avoid this going forward.

#### Discount Types

Discount types would track how the discounted price is computed, some examples/ideas:
- `percent-off`: a percentage off the original price
- `dollars-off`: a fixed dollar reduction in price (e.g. $30 off)
- `fixed-price`: the price is discounted to the fixed price (e.g. a product would cost $100 regardless of what the original price was)

#### Redemption Types
There may be a few different ways we want to track discount usage, for example:

- `one-time`: the discount can only be used once by anyone
- `one-time-per-user`: the discount can be used once per user
- `unlimited`: the discount can be used any number of times

```python
class Discount(TimestampedModel):
    """Discount model"""
    amount = models.DecimalField(
            decimal_places=5,
            max_digits=20,
    )
    automatic = models.BooleanField(default=False)
    discount_type = models.CharField(
          choices=DISCOUNT_TYPES, max_length=30
      )
    redemption_type = models.CharField(
          choices=REDEMPTION_TYPES, max_length=30
      )
    max_redemptions = models.PositiveIntegerField(null=True)

class UserDiscount(TimestampedModel):
    """pre-assignment for a discount to a user"""
    discount = models.ForeignKey(Discount)
    user = models.ForeignKey(User)
```

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

Fixes #309 

#### What's this PR do?
* Adds in models for Discounts, User Discounts, and Redemptions
* Adds bindings to dj-admin so those can be viewed/managed through that

Run migrations, then test via django admin

- This doesn't use reversion but that might not be a bad idea; should be pretty easy to add in.


ecommerce/constants.py
```diff
@@ -0,0 +1,24 @@
...
+ALL_DISCOUNT_TYPES = [
+    DISCOUNT_TYPE_PERCENT_OFF,
+    DISCOUNT_TYPE_DOLLARS_OFF,
+    DISCOUNT_TYPE_FIXED_PRICE,
+]
+DISCOUNT_TYPES = list(zip(ALL_DISCOUNT_TYPES, ALL_DISCOUNT_TYPES))
+
+REDEMPTION_TYPE_ONE_TIME = "onetime"
+REDEMPTION_TYPE_ONE_TIME_PER_USER = "onetimeperuser"
+REDEMPTION_TYPE_UNLIMITED = "unlimited"
```
rhysyngsun, line 16: Nit: could you dash-delimit  the words in the constant values so it's easier to read?

ecommerce/models.py
```diff
@@ -58,3 +59,50 @@ class BasketItem(TimestampedModel):
...
+
+
+class UserDiscount(TimestampedModel):
+    """pre-assignment for a discount to a user"""
+
+    discount = models.ForeignKey(
+        Discount, on_delete=models.CASCADE, related_name="user_discount_discount"
+    )
+    user = models.ForeignKey(
+        User, on_delete=models.CASCADE, related_name="user_discount_user"
```
rhysyngsun, line 87: As a [best practice](https://docs.djangoproject.com/en/4.0/topics/auth/customizing/#referencing-the-user-model),  `settings.AUTH_USER_MODEL` should be used here:
```suggestion
        settings.AUTH_USER_MODEL, on_delete=models.CASCADE, related_name="user_discount_user"
```

ecommerce/models.py
```diff
@@ -58,3 +59,50 @@ class BasketItem(TimestampedModel):
...
+
+
+class DiscountRedemption(TimestampedModel):
+    """
+    Tracks when discounts were redeemed, for discounts that aren't unlimited use
+    """
+
+    redemption_date = models.DateTimeField()
+    redeemed_by = models.ForeignKey(
+        User, on_delete=models.CASCADE, related_name="redeemed_by_user"
```
rhysyngsun, line 101:
```suggestion
        settings.AUTH_USER_MODEL, on_delete=models.CASCADE, related_name="redeemed_by_user"
```

rhysyngsun (changes requested):

ecommerce/constants.py
jkachel: Should be fixed in ad31885b13c9acb18693d2edf90d1778735ecdf9

ecommerce/models.py
jkachel: Should be fixed in ad31885b13c9acb18693d2edf90d1778735ecdf9

ecommerce/models.py
jkachel: Should be fixed in 6255b8b9fa9c4efa830c84a4cb34b7cc068ad195

rhysyngsun (approved):

Issue #351, opened 2021-12-29
Make readable_id searchable in CMS


briangrossman:
### Steps to Reproduce

- In the CMS, search for a readable_id in the search field (e.g. `course-v1:MITxT+14.310x`)

### Expected Behavior

- `course-v1:MITxT+14.310x | Data Analysis for Social Scientists` will be returned in the search results

### Actual Behavior

- No results are returned. 
- The search is not searching the readable_id field

### Related Issues

https://github.com/mitodl/mitxonline/issues/322

### Screenshot or Screencast

![image](https://user-images.githubusercontent.com/21342568/147685510-9b436a65-a5af-4e47-8e17-f05e11b69f48.png)

arslanashraf7:
@briangrossman Can we close this? Just checking in to make sure that we've verified this in production as well since it has a prerequisite, which is to run `./manage.py wagtail_update_index`. However, there a many details on https://github.com/mitodl/mitxonline/pull/352.

briangrossman:
@arslanashraf7 
Thank you for the reminder to run this! I just ran it. The search it working.

PR #336, opened 2021-12-14
formatted course name and ordered them in explorer

Wassaf-Shahzad:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
- [x] Testing
   - [x] Changes have been manually tested

#### What are the relevant tickets?
Closes #322

#### What's this PR do?
This PR  accomplishes the following
- Changes the display name to `{courseReadableId} | {courseTitle}` 
- Sorts the title Alphabetically

#### How should this be manually tested?
This PR can be tested by the following
- Add courses to CMS
- Verify courses title is formatted correctly 
- Verify courses are sorted properly on Explorer Page

#### Screenshots (if appropriate)

Screenshots

Screenshot 2021-12-15 at 12 05 37 AM

Screenshot 2021-12-15 at 12 05 28 AM

##Updated Screenshots

Screenshot 2021-12-15 at 4 04 32 PM

Screenshot 2021-12-15 at 4 04 26 PM

Screenshot 2021-12-15 at 4 04 19 PM

PR #352, opened 2022-01-03
Add readable id in searchable fields of wagtail

arslanashraf7:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations (Wagtail Search App's)
- [x] Testing
  - [x] Code is tested

#351 

**Additional Note - Reviewer Note**
- While working on this, I also noticed that instead of showing the title from `CoursePage` model of wagtail, we were showing the title of the `Course` model from `courses` app. So I fixed that as well. This means updating the title of a course in `/cms/pages` should be displayed everywhere instead of the title from the Course model(Can be changed in Django Admin).
- The wagtail index search doesn't update the index of the related keys automatically, so in the case when you update the readable id of a course manually from admin after the course has been added in CMS, you'll need to run `./manage.py wagtail_update_index` so that the related `readable_id` index updates

#### What's this PR do?
- Adds the readable id to be part of search in wagtail

- Search any course with the readable id instead of title and you should see the matching course in the list
- Search a course by title, it should be returned in the search results

- Readable id was added in `admin_display_title` in one of recent PR(https://github.com/mitodl/mitxonline/pull/336), where we formatted the course pages display title in addition with readable id. 

Screenshot 2022-01-04 at 2 04 46 PM
Screenshot 2022-01-04 at 2 05 03 PM

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -24,6 +24,7 @@ from wagtail.images.models import Image
 from wagtail.images.edit_handlers import ImageChooserPanel
 from wagtail.embeds.embeds import get_embed
 from wagtail.embeds.exceptions import EmbedException
+from wagtail.search import index
 
 from cms.blocks import ResourceBlock, PriceBlock, FacultyBlock
 from cms.constants import COURSE_INDEX_SLUG, CMS_EDITORS_GROUP_NAME
@@ -344,6 +345,12 @@ class CoursePage(ProductPage):
         "courses.Course", null=True, on_delete=models.SET_NULL, related_name="page"
     )
 
+    search_fields = Page.search_fields + [
+        index.RelatedFields('course', [
+            index.SearchField('readable_id', partial_match=True),
+        ])
+    ]
+
     @property
     def product(self):
         """Gets the product associated with this page"""
@@ -351,9 +358,9 @@ class CoursePage(ProductPage):
 
     template = "product_page.html"
 
-    def get_admin_display_title(self): 
-        """Gets the title of the course in the speacified format"""
-        return str(self.course)
+    def get_admin_display_title(self):
+        """Gets the title of the course in the specified format"""
+        return f"{self.course.readable_id} | {self.title}"
 
     def get_context(self, request, *args, **kwargs):
         relevant_run = get_user_relevant_course_run(
--- a/main/settings.py
+++ b/main/settings.py
@@ -141,6 +141,7 @@ INSTALLED_APPS = (
     # WAGTAIL
     "wagtail.contrib.forms",
     "wagtail.contrib.redirects",
+    "wagtail.contrib.postgres_search",
     "wagtail.embeds",
     "wagtail.sites",
     "wagtail.users",
@@ -618,6 +619,13 @@ SITE_NAME = get_string(
     description="Name of the site. e.g MIT mitX Online",
 )
 WAGTAIL_SITE_NAME = SITE_NAME
+
+WAGTAILSEARCH_BACKENDS = {
+    'default': {
+        'BACKEND': 'wagtail.contrib.postgres_search.backend',
+        'ATOMIC_REBUILD': True,
+    },
+}
 
 MEDIA_ROOT = get_string(
     name="MEDIA_ROOT",


arslanashraf7:
@briangrossman Could you please take a look at the screenshots attached to the PR description along with `Reviewer Note` that i added for details that you might want to take a look at.

Extending to the Reviewer note above, we will need to look for the following as well:
-  We need to run `./manage.py wagtail_update_index` just once when this build is deployed (To create the initial index for `readable_id` of the existing courses)
- The index for `readable_id` would update automatically when we create a new course page in CMS.
- We might need to run a cron job occasionally to update the index **_But_** that might not really be necessary as I think the frequency of changing the `readable_id` is quite low and even if we need to change it after the page has been created in CMS, we can just run `./manage.py wagtail_update_index` and the index should rebuild. (Only applies if a course was created in CMS already and its readable id was changed in Django Admin later on.) So I think we can run this manually when needed.


@briangrossman let me know what you think about the above points.

briangrossman:
@arslanashraf7 
The screenshots look good. Thanks for taking care of this--it looks like it was not trivial.

We'll have to remember the notes above for when to run `wagtail_update_index`

arslanashraf7:
>The screenshots look good. Thanks for taking care of this--it looks like it was not trivial.
@briangrossman 

Yes, I reverted my last workaround of using signals and stuff for automatically running the command and instead used the simpler implementation where we can run the command manually. Which made sense based on the low probability of `readable_id` getting changed.

And I hope you had a look at below(Is my below assumption correct?):
>While working on this, I also noticed that instead of showing the title from CoursePage model of wagtail, we were showing the title of the Course model from courses app. So I fixed that as well. This means updating the title of a course in /cms/pages should be displayed everywhere instead of the title from the Course model(Can be changed in Django Admin).

briangrossman:
@arslanashraf7 
> And I hope you had a look at below(Is my below assumption correct?):
> 
> > While working on this, I also noticed that instead of showing the title from CoursePage model of wagtail, we were showing the title of the Course model from courses app. So I fixed that as well. This means updating the title of a course in /cms/pages should be displayed everywhere instead of the title from the Course model(Can be changed in Django Admin).

Yes. This definitely seems like the right implementation. Thanks for that.

Wassaf-Shahzad (approved):

briangrossman:
I reviewed this on RC and it LGTM.

Note that @arslanashraf7  and I had some discussion about the nuances of searching. Just documenting them here:

**Question**
Here’s a couple examples that surprised me:
- Searching for `course` returns all courses with the `readable_id` that begins with `course-v1`. Searching for `cours` returns nothing. I assume this has something to do with the way the search indexes.
- Searching for `+14.750x` returns the course with the `readable_id` `course-v1:MITx+14.750x`, but searching for `14.750x` returns nothing

**Answer**
[...] for the cases you mentioned i think there are a few limitations mentioned for the partial search on the site [documentation](https://docs.wagtail.io/en/v2.13.4/reference/contrib/postgres_search.html#postgres-search). so, what it is doing is because of the how [boost/postgres search](https://www.postgresql.org/docs/current/textsearch.html) is working underneath.

arslanashraf7:
Thanks for adding that comment @briangrossman.

_Just for future references_, I'd add just one additional thing related to when we say `Partial Matching` isn't supported & but `wagtail supports all the text search features of Postgres` as mentioned in [answer here](https://github.com/mitodl/mitxonline/pull/352#issuecomment-1006813221). More clarity on the use-cases below:

>Searching for course returns all courses with the readable_id that begins with course-v1. Searching for cours returns nothing. I assume this has something to do with the way the search indexes.

It still searches the whole text within the fields `title & readable_id`, and the fact that it returns courses on `course-v1` is related to colon (`:` or `<space>`) somehow(It divides the words and index them for search), so whenever there is a colon within the text it divides the text into two and able to search both parts independently that's why we see courses on `course-v1`. (I looked around to find the docs on this but i couldn't but I understood the behavior)

>Searching for +14.750x returns the course with the readable_id course-v1:MITx+14.750x, but searching for 14.750x returns nothing

Same case as mentioned above.

PR #359, opened 2022-01-05
Added discount abstractions and tests

### LINKED ISSUE #310: Discount Subsystem: Discount Abstraction
Rather than codifying the discount logic in a complicated computation function, discount types will be implemented by abstracting the logic around discounts into a registry-driven discount factory like this:

```python
import abc
from dataclasses import dataclass

@dataclass
class DiscountType(abc.ABC):
    _CLASSES = {}

    discount: Discount
    
    # see https://www.python.org/dev/peps/pep-0487/
    def __init_subclass__(cls, *, discount_type, **kwargs):
        super().__init_subclass__(**kwargs)

        if discount_type in _CLASSES:
            raise TypeError(f"{discount_type} already defined for DiscountType")

        cls.discount_type = discount_type
        cls._CLASSES[discount_type] = cls
        
    @classmethod
    def for_discount(cls, discount: Discount):
        DiscountTypeCls = cls._CLASSES[discount.discount_type]

        return DiscountTypeCls(discount)

    def get_product_price(self, product: Product):
        return self.get_product_version_price(product.latest_version)

    @abc.abstractmethod
    def get_product_version_price(self, product_version: ProductVersion): 
          ...


class PercentDiscount(DiscountType, 
          discount_type=Discount.PERCENT_DISCOUNT):

    def get_product_version_price(self, product_version: ProductVersion):
            return product_version.price * self.discount.amount

class FixedPriceDiscount(DiscountType, 
          discount_type=Discount.PERCENT_DISCOUNT):

    def get_product_version_price(self, product_version: ProductVersion):
            return self.discount.amount  # the amount here is the fixed price

```

With this implementation, prices before ordering would use `get_product_price`, whereas the receipt service would use `get_product_version_price` on the purchased versions. This makes it far more scalable to introduce new discount types without having to refactor existing code.

jkachel:
added discount abstraction API and tests for those

#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #310 

#### What's this PR do?
Adds the discount calculation code as an abstract base class, and implementations for four discount types. 

pytest should pass

### DIFF
--- a/ecommerce/discounts_test.py
+++ b/ecommerce/discounts_test.py
@@ -1,0 +1,71 @@
+import pytest
+from decimal import Decimal, getcontext
+
+from ecommerce.factories import ProductFactory, DiscountFactory
+from ecommerce.discounts import (
+    DiscountType,
+    PercentDiscount,
+    FixedPriceDiscount,
+    DollarsOffDiscount,
+)
+
+pytestmark = [pytest.mark.django_db]
+
+
+@pytest.fixture()
+def products():
+    return ProductFactory.create_batch(5)
+
+
+@pytest.fixture()
+def discounts():
+    return DiscountFactory.create_batch(10)
+
+
+def test_discount_factory_generation(discounts):
+    """
+    Runs through discounts and makes sure all the ones that come out of the
+    factory are recognizable by the test suite. (This is a sort of sanity
+    check - if a new discount type gets added and the tests aren't updated, this
+    test will fail.)
+    """
+    for discount in discounts:
+        discount_logic = DiscountType.for_discount(discount)
+
+        what_type = (
+            type(discount_logic) is DollarsOffDiscount,
+            type(discount_logic) is FixedPriceDiscount,
+            type(discount_logic) is PercentDiscount,
+        )
+
+        assert any(what_type)
+
+
+def test_discount_factory_adjustment(discounts, products):
+    """
+    Tests discounting products. This runs through each factory-generated
+    product and applies all of the discounts that have been generated, then
+    compares the discounted price to the discount generated in the test.
+    """
+    for product in products:
+        for discount in discounts:
+            discount_logic = DiscountType.for_discount(discount)
+
+            if type(discount_logic) is DollarsOffDiscount:
+                discounted_price = product.price - discount.amount
+
+                if discounted_price < 0:
+                    discounted_price = 0
+            elif type(discount_logic) is FixedPriceDiscount:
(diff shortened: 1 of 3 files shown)


ecommerce/discounts_test.py
```diff
@@ -0,0 +1,62 @@
...
+        what_type = (
+            type(discount_logic) is DollarsOffDiscount,
+            type(discount_logic) is FixedPriceDiscount,
+            type(discount_logic) is PercentDiscount,
+        )
+
+        assert any(what_type)
+
+
+def test_discout_factory_adjustment(discounts, products):
```
annagav, line 38: typo here

ecommerce/discounts_test.py
```diff
@@ -0,0 +1,62 @@
...
+def products():
+    return ProductFactory.create_batch(5)
+
+
+@pytest.fixture()
+def discounts():
+    return DiscountFactory.create_batch(10)
+
+
+def test_discount_factory_generation(discounts):
```
annagav, line 25: Could you please add comments to each test, mentioning what thei are testing?

ecommerce/discounts_test.py
```diff
@@ -0,0 +1,62 @@
...
+            type(discount_logic) is DollarsOffDiscount,
+            type(discount_logic) is FixedPriceDiscount,
+            type(discount_logic) is PercentDiscount,
+        )
+
+        assert any(what_type)
+
+
+def test_discout_factory_adjustment(discounts, products):
+    # product = products[random.randrange(0, len(products), 1)]
```
annagav, line 39: was this for testing purposes?

annagav (changes requested):
Just couple of minor comments

ecommerce/discounts_test.py
jkachel: Fixed in 20db5b212d945e1d53af7a820d3986145a0a8927

ecommerce/discounts_test.py
jkachel: Added these in 20db5b212d945e1d53af7a820d3986145a0a8927

ecommerce/discounts_test.py
jkachel: Initially I had this just do one product from the factory; I decided later to just run through all of them that were being generated and forgot to remove this. It's gone in 20db5b212d945e1d53af7a820d3986145a0a8927

annagav (approved):
LGTM

PR #370, opened 2022-01-12
Basket Subsystems API

### LINKED ISSUE #307: Basket Subsystem: APIs
`GET  /api/v0/basket/` -> get the current basket state
`POST /api/v0/basket/`  -> update the basket state
#### Notes
The implementation of this would use the discount subsystem to calculate the discounted prices, those values would be returned in the API for the frontend to use for display purposes.

annagav:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested

Fix #307

#### What's this PR do?
Adds REST api to retrieve and update states for Basket and BasketItems.

Screen Shot 2022-01-12 at 1 39 02 PM


ecommerce/urls.py
```diff
@@ -1,14 +1,32 @@
...
+    r'items',
+    BasketItemViewSet,
+    basename='basket-items',
+    parents_query_lookups=['basket_items']
+)
+
 urlpatterns = [
     re_path(r"^api/v0/", include(router.urls)),
     re_path(r"^api/", include(router.urls)),
+    path("add_product/", create_enrollment_view, name="add-product-to-basket"),
```
rhysyngsun, line 31: I don't think this should be here since it's not related to this work

ecommerce/views/v0/__init__.py
```diff
@@ -84,3 +90,49 @@ def get_queryset(self):
...
+class BasketItemViewSet(NestedViewSetMixin, ModelViewSet):
+    """API view set for BasketItem"""
+    queryset = BasketItem.objects.all()
+    serializer_class = BasketItemSerializer
+    permission_classes = [IsAuthenticated]
+
+
+@api_view(["POST"])
+@permission_classes([IsAuthenticated])
+def add_product_to_basket_view(request):
```
rhysyngsun, line 117: I don't think this would require a separate view - it should be part of `BasketItemViewSet` I think.

ecommerce/views/v0/__init__.py
```diff
@@ -84,3 +93,33 @@ def get_queryset(self):
...
+        product_id_str = request.data.get("product_id")
+        product = Product.objects.filter(id=int(product_id_str)).first()
+        item, created = BasketItem.objects.update_or_create(
+            basket=basket, product=product
+        )
+        if created is False:
+            item.quantity += 1
+            item.save()
+
+        return HttpResponseRedirect(reverse("user-dashboard"))
```
rhysyngsun, line 117: This can be removed.

ecommerce/views/v0/__init__.py
```diff
@@ -84,3 +93,33 @@ def get_queryset(self):
             .select_related("content_type")
             .prefetch_related("purchasable_object")
         )
+
+
+class BasketViewSet(NestedViewSetMixin, ModelViewSet):
```
rhysyngsun, line 98: A few things for this view:

- Remove `ModelViewSet` and make it inherit `mixins.ListModelMixin, mixins.RetrieveModelMixin, viewsets.GenericViewSet` instead.
- Add an implementation for `def get_queryset(self, request)` that filters to the current user.
- Set `lookup_field = "user__username"` and `lookup_url_kwarg = "username"` - we'll use `username` as the identifier instead of `Basket.id` because each user has only 1 basket and username is a value the frontend already knows.
- Add an implementation for `def get_object(self,request, username=None)` that does a `Basket.objects.get_or_create(user=request.user)` - we're not using the `queryset` here intentionally

ecommerce/views_test.py
```diff
@@ -35,3 +46,60 @@ def test_get_products(user_drf_client, products):
...
+    )
+
+
+def test_delete_basket_item(user_drf_client, user):
+    """Test the view to delete item from the basket"""
+    basket_item = BasketItemFactory.create(basket__user=user)
+    basket = basket_item.basket
+    assert basket.basket_items.count() == 1
+    user_drf_client.delete(
+        "/api/baskets/{}/items/{}/".format(basket.id, basket_item.id),
```
rhysyngsun, line 87: Should the first part of the url be `username` so that it aligns to the `/api/baskets/:username/` API?

ecommerce/views/v0/__init__.py
```diff
@@ -84,3 +94,41 @@ def get_queryset(self):
...
+
+    def get_queryset(self):
+        return BasketItem.objects.filter(basket__user=self.request.user)
+
+    def create(self, request, *args, **kwargs):
+        basket = Basket.objects.get(user=request.user)
+
+        product_id = request.data.get("product")
+        product = Product.objects.get(id=int(product_id))
+        BasketItem.objects.create(basket=basket, product=product)
```
rhysyngsun, line 133: This might be better as a `get_or_create` - returning a `201` if it's been created and a 200 otherwise - it also ideally would return the serialized basket item so that might mean moving this creation logic into the serializer's `create()` method instead to be cleaner since then the viewset would take care of the serialization as long as `create()` returns an instance.

rhysyngsun (changes requested):

ecommerce/views_test.py
annagav: yeah, I find it interesting that even though we set the look up key to username the lookup by pk is still working.

annagav:
@rhysyngsun I addressed your comments. Thank you for your review

rhysyngsun (approved):

PR #375, opened 2022-01-18
Change unsubscribe UI to email settings

umarmughal824:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [x] Mobile width screenshots
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
fixes #369 

#### What's this PR do?
Change unsubscrib UI to email settings

#### How should this be manually tested?
(Required)

#### Where should the reviewer start?
(Optional)

#### Any background context you want to provide?
(Optional)

#### Screenshots (if appropriate)
Screenshot 2022-01-19 at 18 42 23

Screenshot 2022-01-19 at 18 42 32
Screenshot 2022-01-19 at 18 42 40
Screenshot 2022-01-19 at 18 42 48
Screenshot 2022-01-19 at 18 42 58

PR #381, opened 2022-01-21
Revert "Change unsubscribe UI to email settings"

umarmughal824:
Reverts mitodl/mitxonline#375

### DIFF
--- a/courses/serializers.py
+++ b/courses/serializers.py
@@ -143,7 +143,6 @@ class CourseRunDetailSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.CourseRun
         fields = [
-            "course_number",
             "course",
             "title",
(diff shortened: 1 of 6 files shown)


arslanashraf7 (approved):
👍

@pdpinch FYI this change has been reverted. So, the users will now only have the older view of email subscribe/unsubscribe which was done in (https://github.com/mitodl/mitxonline/pull/329).

@umarmughal824 Please confirm with @pdpinch before merging this.

pdpinch:
Thank you both. 

I was hoping we could release this feature with new UI, but we shouldn’t hold it up any longer. 

Please merge the revert, which will give us some more time to work out the UI, and maybe also answer Umar’s question about the unenrollment UI.

Issue #387, opened 2022-01-24
UI: Checkout page


rhysyngsun:
### Designs

https://projects.invisionapp.com/share/KH11D3ZN53GC#/screens/462885073
https://projects.invisionapp.com/share/KH11D3ZN53GC#/screens/462972188
https://projects.invisionapp.com/share/KH11D3ZN53GC#/screens/462972599

### Requirements

- [ ] The checkout page should be feature flagged, it should result in page not found error if disabled.
- [ ] The checkout page should display the product(s) in the learner's basket with at least the name and price of the product.
- [ ] There should be a button to make the payment, which sends the learner to Cybersource.

### Out of Scope

- The discount/coupon part of this UI will get addressed separately, it should be left out for the MVP.

PR #395, opened 2022-01-31
Add CyberSource integration and checkout APIs

### LINKED ISSUE #341: Cybersource Integration
#### References
- [Existing implementation in MITxPro](https://github.com/mitodl/mitxpro/blob/master/ecommerce/api.py)
- [Cybersource documentation](https://developer.cybersource.com/library/documentation/dev_guides/Secure_Acceptance_Hosted_Checkout/Secure_Acceptance_Hosted_Checkout.pdf)

#### Approach

We'd like to implement this logic in a reusable [`ol-django`](https://github.com/mitodl/ol-django) app (e.g. `mitol.cybersource`). There will be some pieces that can be generalized such as the fulfillment webhook implementation, payload signing, etc. There will be some pieces that are more specific to the application itself (e.g. how to serialize the products to a payload), which will need to be [exposed as configurable functions](https://github.com/mitodl/ol-django/blob/main/src/mitol/digitalcredentials/backend.py#L62-L72). MITx Online would then utilize this library and implement the necessary customizations.

#### Requirements
- [ ] Cybersource payload generation and signing
- [ ] Checkout API (named `CheckoutView` in xPro) that takes the current user's basket and:
  - Transforms it into an unfulfilled order
  - Serializes it into a signed Cybersource payload
  - Returns the payload and the Cybersource url
- [ ] Private fulfillment webhook API (named `OrderFulfillmentView` in xPro)

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested
- [X] Settings
  - [X] New settings are documented and present in `app.json`

Fixes #341 

#### What's this PR do?
- Pulls in the payment_gateway app from ol-django
- Adds in APIs for converting baskets to orders and then submitting them through the payment processor 
- Adds in some stubs to allow you to work the payment processing workflow

Go to /ecommerce-test/checkout to test the checkout process manually. You will need to be logged in as a superuser and the system will need to be set up with the relevant keys for CyberSource.


jkachel:
Created to create the PR - this is waiting on a test failure upstream (in main) and for https://github.com/mitodl/ol-django/pull/53 to be completed (as it depends on stuff there).

ecommerce/views/v0/__init__.py
```diff
@@ -136,3 +169,500 @@ def create(self, request, *args, **kwargs):
...
+        total = 0
+
+        order = PendingOrder(total_price_paid=0, purchaser=request.user)
+        order.save()
+
+        x_forwarded_for = request.META.get("HTTP_X_FORWARDED_FOR")
+        if x_forwarded_for:
+            ip = x_forwarded_for.split(",")[0]
+        else:
+            ip = request.META.get("REMOTE_ADDR")
```
rhysyngsun, line 251: Would be worth pulling in [`django-ipware`](https://pypi.org/project/django-ipware/) to handle determination of the IP address.

rhysyngsun (changes requested):

ecommerce/views/v0/__init__.py
```diff
@@ -136,3 +169,500 @@ def create(self, request, *args, **kwargs):
...
+                "Got this weird thing: {decision}\nMessage: {msg}".format(
+                    decision=processor_response["decision"],
+                    msg=processor_response["message"],
+                )
+            )
+
+        # for now let's just punt this back to the client
+        # cause i have no idea what to do yet
+
+        return Response(internal_response)
```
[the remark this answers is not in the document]
jkachel: I changed this to return just an empty response but really this needs to redirect to somewhere else in the app once it's finished processing the response (so this will all change as other things, notably https://github.com/mitodl/mitxonline/issues/387, get closer to completion).

ecommerce/views/v0/__init__.py
```diff
@@ -136,3 +169,500 @@ def create(self, request, *args, **kwargs):
...
+
+            internal_response = {
+                "state": "It worked!",
+                "full_response": processor_response,
+                "order": converted_order,
+            }
+        else:
+            # something else weird
+            self.logger.debug(
+                "Got this weird thing: {decision}\nMessage: {msg}".format(
```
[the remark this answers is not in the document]
jkachel: Actually removed this - the CyberSource gateway defaults to setting this to error, so the "fall out the bottom" here shouldn't ever happen.

jkachel:
Comments on individual things where needed; everything else I went ahead and just fixed. Did a bit of cleanup to depend on the ol-django payment gateway app instead of decoding/processing stuff internally too.

ecommerce/views/v0/__init__.py
```diff
@@ -136,3 +170,441 @@ def create(self, request, *args, **kwargs):
...
+
+        total = 0
+
+        order = PendingOrder(total_price_paid=0, purchaser=request.user)
+        order.save()
+
+        ip = get_client_ip(request)[0]
+
+        gateway_order = GatewayOrder(
+            username=request.user.username, ip_address=ip, reference=order.id, items=[]
```
rhysyngsun, line 251: We need a more distinguished value for the order reference number, same as over on xpro: https://github.com/mitodl/mitxpro/blob/4fa9c9a401829c4d0946e33763d269d1f09e270b/ecommerce/models.py#L294-L297

rhysyngsun (changes requested):

ecommerce/views/v0/__init__.py
jkachel: Should be fixed in https://github.com/mitodl/mitxonline/pull/395/commits/b6317da2ae294e99f2f0d07deeca57910673ac8f .

ecommerce/views/v0/__init__.py
```diff
@@ -136,3 +170,441 @@ def create(self, request, *args, **kwargs):
...
+            basket = Basket(user=request.user)
+            basket.save()
+        else:
+            basket = Basket.objects.filter(user=request.user).get()
+
+        if basket.basket_items.count() == 0:
+            products = Product.objects.all()
+
+            basket_item = BasketItem(
+                product=products[random.randrange(1, len(products), 1)],
```
rhysyngsun, line 575: Knit, this errors if there's only 1 product.
```suggestion
                product=products[random.randrange(0, len(products), 1)],
```

rhysyngsun (approved):
LGTM, just one small nit, feel free to squash and merge after fixing it.

PR #416, opened 2022-02-16
unsubscribe from course emails after unenroll

### LINKED ISSUE #400: User remains subscribed to course emails after unenrolling
### Steps to Reproduce

- Enroll in a course on MITx Online
- From the dashboard, unenroll from the course

### Expected Behavior

- Learner will be unsubscribed from course emails.

### Actual Behavior

- Learner's enrollment record still shows them subscribed to edX emails for the course (see image below)

### Related Issues

- https://github.com/mitodl/mitxonline/issues/288
- https://github.com/mitodl/mitxonline/pull/329
- https://github.com/mitodl/edx-api-client/pull/83
- Note: We may have to think about this when we implement deferrals, transfers, and refunds in the future

### Screenshot or Screencast

Screen Shot 2022-02-07 at 2 19 01 PM

umarmughal824:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

fixes #400 

#### What's this PR do?
unsubscribe from course emails after unenroll

Try to enroll / unenroll in a course and see the admin record of it

 **Unenroll**

Screenshot 2022-02-16 at 18 33 55

**Enroll**

Screenshot 2022-02-16 at 18 12 58

### DIFF
--- a/courses/views_test.py
+++ b/courses/views_test.py
@@ -414,9 +414,15 @@ def test_user_enrollment_delete(
     enrollment = CourseRunEnrollmentFactory.create(
         user=user, active=True, change_status=None
     )
+    inactive_enrollment = CourseRunEnrollmentFactory.create(
+        user=user,
+        active=False,
+        change_status=ENROLL_CHANGE_STATUS_UNENROLLED,
+        edx_emails_subscription=False,
+    )
     patched_deactivate = mocker.patch(
         "courses.views.v1.deactivate_run_enrollment",
-        return_value=(None if deactivate_fail else enrollment),
+        return_value=(None if deactivate_fail else inactive_enrollment),
     )
     resp = user_drf_client.delete(
         reverse("user-enrollments-api-detail", kwargs={"pk": enrollment.id})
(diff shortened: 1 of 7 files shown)


umarmughal824:
@briangrossman I guess if a user is unenrolled from edx, it would not be able to receive any emails against that course. Anyway, I have put up a PR for it

briangrossman:
@umarmughal824 The screenshots look good.

arslanashraf7 (changes requested):
I have 2 concers in this:

1) The changes themselves look good and work well too but, We haven't reflected these in any tests. Can you please update the existing enrollment, unenrollment tests to reflect these changes too?
2) Although you mentioned that edX emails won't work if user is unenrolled. Even if that's true we might end up with data inconsistency in case:
    - Enroll User, Both MITxOnline and edX would have email subscribed (This is fine)
    - Only unsubscribe from the email through MITxOnline `Dashboard`
    - Unenroll the course and then reenroll the course. In this case, we will have `Email Subscription=True` in MITxOnline but in Edx the email subscription would still be inactive. Let me know if this would not be the case.

umarmughal824:
@arslanashraf7 the part of the test is addressed and for the second point, I am gonna create a separate ticket.

umarmughal824:
https://github.com/mitodl/mitxonline/issues/445

courses/views_test.py
```diff
@@ -414,9 +414,15 @@ def test_user_enrollment_delete(
     enrollment = CourseRunEnrollmentFactory.create(
```
arslanashraf7, line 414: I think we should remove this instance of enrollment since we are now creating an inactive enrollment below. Let me know if it's needed anyway.

arslanashraf7 (approved):
👍 Looks good. Just one minor concern.

courses/views_test.py
umarmughal824: it's being using at other places and being changed to unenroll in mitxonline but we have only mocked the edx enrollment deactivation method.

PR #418, opened 2022-02-17
Add Checkout page UI

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

#387 

#### What's this PR do?
Adds checkout UI. Discount UI is also present but feature-flagged out. 

1. Ensure a product for a course run is set up and test credentials are in place for CyberSource
2. Enable the checkout test UI and go to /ecommerce-test/checkout to establish a basket
3. Go to /cart to view the cart
4. Clicking Place Order should redirect you to CyberSource

There isn't a receipt page here as that's in a different issue. Set ENABLE_DISCOUNT_UI flag to true to enable the discount UI (which is just the UI with placeholder text).

image
image


static/js/containers/pages/checkout/CartPage.js
```diff
@@ -0,0 +1,276 @@
...
+
+  renderOrderSummaryCard() {
+    let totalPrice = 0
+    const { cartItems } = this.props
+
+    if (cartItems !== null) {
+      for (let i = 0; i < cartItems.length; i++) {
+        totalPrice += Number(cartItems[i].product.price)
+      }
+    }
```
rhysyngsun, line 116: We should move this computation to the backend - `BasketWithProductSerializer` should return the cart total so that the backend is the authoritative source of that value and we don't have this computation in both places. That way in the future when we integrate other things that effect the total (e.g. discounts), the frontend just gets that logic incorporated for free.

static/js/containers/pages/checkout/CartPage.js
```diff
@@ -0,0 +1,276 @@
...
+                  Apply
+                </button>
+              </div>
+              <div className="text-primary mt-2 font-weight-bold cart-text-smaller">
+                Adding another coupon will replace the currently applied coupon.
+              </div>
+            </div>
+          </div>
+        ) : (
+          ""
```
rhysyngsun, line 191: Nit: returning `null` for no content is more idiomatic for React.

ecommerce/templates/checkout_interstitial.html
```diff
@@ -0,0 +1,20 @@
...
+<html>
+<head>
+    <title>Checkout Interstitial</title>
+</head>
+<body>
+    <h1 style="font-family: Montserrat, sans-serif">Redirecting to the payment processor...</h1>
+
+    <form id="checkout_form" method="post" action="{{ checkout_payload.url }}">
+        {% for key, value in form.items %}
+        <input style="display: none;" type="text" readonly="readonly" name="{{ key }}" value="{{ value }}" /></p>
```
rhysyngsun, line 11: This can be a hidden input instead, also there's a trailing and unmatched `</p>`at the end of this line.

rhysyngsun (changes requested):
Initial code review feedback.

static/js/containers/pages/checkout/CartPage.js
```diff
@@ -0,0 +1,276 @@
...
+        )}
+
+        <div className="row">
+          <div className="col-12 text-center mt-4 mb-4">
+            <Button
```
rhysyngsun, line 196: This should be disabled if there's nothing in the cart

ecommerce/views/v0/__init__.py
```diff
@@ -450,6 +470,19 @@ def decode_response(self, request):
         return Response({})
 
 
+class CheckoutInterstitialView(LoginRequiredMixin, TemplateView):
+    template_name = "checkout_interstitial.html"
+
+    def get(self, request):
+        checkout_payload = generate_checkout_payload(request)
```
rhysyngsun, line 477: If the cart is empty, this will return a 406 response, and I think there are some other states that also return something other than a `dict`, so in these cases if the return value isn't a `dict`, we probably want to just return that value.

Ideally we'd redirect to a page with an error but for the sake of getting an MVP checkout flow merged, we'll skip that for now.

rhysyngsun (changes requested):
Still testing, but a few more things I found

static/js/containers/pages/checkout/CartPage.js
```diff
@@ -0,0 +1,276 @@
...
+
+type Props = {
+  history: RouterHistory,
+  cartItems: Array<BasketItem>,
+  isLoading: boolean
+}
+
+export class CartPage extends React.Component<Props> {
+  renderCartItemCard(cartItem: BasketItem) {
+    const purchasableObject = cartItem.product.purchasable_object
```
rhysyngsun, line 29: One thing I discovered in my testing is that it's possible for `purchaseable_object` to be `null` because Django's generic object mechanism doesn't enforce referenctial integrity at the DB level, which causes this page to error on render. We probably need to guard against such products being added to the cart, but even with that we probably should have this case.

The options I can think of are:

a) filter out cart items w/ a product that has no `purchasable_object`, the frontend will then treat it like an empty cart
b) the frontend checks for this kind of state and displays an error message

I'm fine with either of these options, but I'm not sure if there are broader implications here too.

ecommerce/views/v0/__init__.py
```diff
@@ -173,6 +175,71 @@ def create(self, request, *args, **kwargs):
...
+                name=product_version.field_dict["description"],
+                quantity=1,
+                sku=f"{product_version.field_dict['content_type_id']}-{product_version.field_dict['object_id']}",
+                unitprice=line_item_price,
+                taxable=0,
+            )
+        )
+
+    order.total_price_paid = total
+    order.save()
```
rhysyngsun, line 226: Since we've now created a pending order, we've consumed the items in the cart - so we should clear the cart out.

rhysyngsun (changes requested):
A few comments from functional testing but otherwise it worked for me.

static/js/containers/pages/checkout/CartPage.js
pdpinch: If I understand this correctly, the user tried to put something in the cart that isn't (currently?) purchasable? In a case like that, I think an error message is called for. 

Could be as simple as "Something went wrong. Contact Customer Support" for now.

static/js/containers/pages/checkout/CartPage.js
```diff
@@ -0,0 +1,276 @@
...
+
+            <div className="row d-none d-md-flex mb-3">
+              <div className="col-md-8">
+                <h4 className="font-weight-normal">
+                  You are about to purchase the following:
+                </h4>
+              </div>
+              <div className="col-md-4 text-right">
+                <h4 className="font-weight-normal">
+                  <a href="#">Help &amp; FAQs</a>
```
rhysyngsun, line 241:
```suggestion
                  <a href="https://mitxonline.zendesk.com/hc/en-us">Help &amp; FAQs</a>
```

Probably in a new tab/window too.

static/js/containers/pages/checkout/CartPage.js
```diff
@@ -0,0 +1,276 @@
...
+              onClick={() => (window.location = "/checkout/to_payment")}
+            >
+              Place your order
+            </Button>
+          </div>
+        </div>
+
+        <div className="row">
+          <div className="col-12 px-3 py-3 py-md-0 cart-text-smaller">
+            By placing my order I agree to the Terms of Service, and Privacy
```
rhysyngsun, line 208: Link TOS and privacy policy to the same urls in the footer

jkachel: New window here too?

rhysyngsun: yeah, for good measure

static/js/containers/pages/checkout/CartPage.js
jkachel: Planning on going with option A here - ideally products with no purchasble_object should get filtered out on the frontend anyway, before the checkout process, since you won't be able to activate the Enroll button for courses that aren't enrollable anyway. 

(There's an edge case here, though: if you enroll on the cusp of the window closing, you could have a cart item that was valid when you start but isn't once you hit the Place Order button. Could take care of this with a grace period but that can maybe be a task for later as it feels like this state wouldn't happen particularly often. If the cart ends up empty for that reason, it can pop up a "please contact support" message.)

ecommerce/templates/checkout_interstitial.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
```
rhysyngsun, line 1: Could you have this view extend the base template as well so that the user doesn't see an unstyled page?

rhysyngsun (changes requested):
Small change

ecommerce/templates/checkout_interstitial.html
jkachel: Fixed in last commit https://github.com/mitodl/mitxonline/pull/418/commits/2f86fae21aa974018012cf31ceca94116eb24f2f

This won't actually perform the redirect (the code for that is commented out) so it's easier to review.

rhysyngsun (approved):
LGTM

PR #423, opened 2022-02-23
View to add products to the cart and proceed directly to checkout

### LINKED ISSUE #422: Ecommerce: when adding a product to the cart, the cart should be cleared and the user directed to checkout
The best way to do this is to have a url we can point the user at and have that view clear the cart and add that product to the cart and do the redirect.

rhysyngsun:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

Fixes #422

#### What's this PR do?
This PR ensures there's only ever one product in the cart when a user tries to pay to avoid situations with multiple products.

- Got to a product page and click to purchase the paid version.
- You should be directed to the checkout page.
- Go to a different product and do the same.
- You should end up on the checkout page with only the second product in your cart.

### DIFF
--- a/ecommerce/urls.py
+++ b/ecommerce/urls.py
@@ -15,6 +15,7 @@ from ecommerce.views.v0 import (
     CheckoutTestView,
     CheckoutTestStepTwoView,
     CheckoutDecodeResponseView,
+    CheckoutProductView,
 )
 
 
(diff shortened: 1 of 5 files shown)


annagav (approved):
LGTM 👍

PR #429, opened 2022-02-24
Updating payment_gateway to 1.2.2, fixing some usage errors

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

none atm

#### What's this PR do?
Updates the mitol-django-payment-gateway dependency to the latest, which fixes a bug in processor response decoding. Also fixes some spots where the app was decoding responses incorrectly.

Run a transaction through - you should be ultimately redirected back into the DRF viewer (since there's no where to go quite yet when the transaction is done) and the appropriate action should be taken for the order that was created.


rhysyngsun (approved):
LGTM

PR #444, opened 2022-02-25
Adding Discount UI

### LINKED ISSUE #431: UI: Apply discount to basket/order
### Designs

https://projects.invisionapp.com/share/KH11D3ZN53GC#/screens/462885073
https://projects.invisionapp.com/share/KH11D3ZN53GC#/screens/462972188
https://projects.invisionapp.com/share/KH11D3ZN53GC#/screens/462972599

Requirements

- [ ] Allow the user to apply at most one discount (we should make sure the backend enforces this if it doesn't yet)
- [ ] Discount computation should occur only on the backend, this might require some API updates.
- [ ] User can remove the discount (I don't think the designs have this, but it really should be allowed)
- [ ] Ensure the discounted price is applied to checkout

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Migrations
- [X] Testing
  - [X] Code is tested
- [x] Settings

Fixes #431

#### What's this PR do?
Adds support for discount redemption:
* Adds UI to apply and clear discount codes
* Updates backend APIs to handle discount code redemption following the established rules
* Updates the Django Admin to include screens for discount redemptions, and to display discounts applied on each order.

Ensure your setup has a valid product and course combination set up. 
Add a discount code in via djadmin and note the code you create.
Add a product to your cart, and then apply the discount code using the form in Order Summary.
The discount should be applied successfully and the price should reflect the discounted amount.

To test adding multiple codes to an order: follow the same instructions as above but add more than one discount. Then, apply a discount code after applying a discount code. The new code should take precedence over the old one.

To test user discounts: follow the first two steps under the standard test, then also add a User Discount record. Log in as the user you added the discount for and add a product to the cart. You should see that the discount has been applied to the cart automatically. Applying another discount over the user discount should override the user discount, but clearing the discount should re-apply the user discount. Other accounts in the system should not have the discount applied to their carts automatically. 

Rules as follows:
- Only a single discount can be applied to a given cart - any additional discounts added after the first override the last added discount. 
- User discounts are always applied, unless overridden by another discount.
- Discounts that have hit the maximum number of redemptions (either by hitting the value in max_redemptions, or by virtue of being a "one-time" or "one-time-per-user" discount type) will return a "not found" and won't be applied. 
- Discounts are applied to the total amount in the cart (there's only one allowed item in the cart right now, though)

Also of note:
- Error and success states right now pop a standard banner notification. This wasn't in the InVision mockup - seemed reasonable so included it. 
- The "Adding another discount" text disappears if there's not already a discount applied.
- Clear Discount only appears if there's a discount to clear. The front-end has no idea (right now) if a discount is a user discount or not so you can clear a user discount; it just won't apparently do anything as the discount will just get reapplied. 
- Fixed Price discounts probably need some styling help - I added Fixed Price text in there to differentiate them from the dollar off or percent off discounts. Input appreciated for this. 

Desktop:
Screen Shot 2022-02-25 at 2 25 00 PM
Screen Shot 2022-02-25 at 2 25 34 PM
Screen Shot 2022-02-25 at 2 26 07 PM

Mobile:
Screen Shot 2022-02-25 at 2 25 18 PM
Screen Shot 2022-02-25 at 2 25 27 PM
Screen Shot 2022-02-25 at 2 26 11 PM


ecommerce/models.py
```diff
@@ -431,9 +468,37 @@ class BasketDiscount(TimestampedModel):
...
+        Converts basket discounts to order redemptions. Discounts
+        applied to the basket are meant to be ephemeral; this makes them
+        permanent.
+
+        Args:
+            - order (Order): The order to add the discount to.
+        Returns:
+            - DiscountRedemption for the converted redemption.
+        """
+        redemption = DiscountRedemption(
```
rhysyngsun, line 492: Would it be safer to do a `get_or_create` on this to make it idempotent?

ecommerce/discounts_test.py
```diff
@@ -69,3 +110,94 @@ def test_discount_factory_adjustment(discounts, products):
...
+    redemption.save()
+
+
+def test_one_time_discount(user, onetime_discount):
+    """
+    Tests single-use discounts. These should be redeemable once, and then not
+    again.
+    """
+
+    assert onetime_discount.check_validity(user) == True
```
rhysyngsun, line 135: Assertions against singletons like `True` and `False` should be done with `is`:
```suggestion
    assert onetime_discount.check_validity(user) is True
```

Same for similar usages in this file.

ecommerce/discounts_test.py
```diff
@@ -69,3 +110,94 @@ def test_discount_factory_adjustment(discounts, products):
...
+    redemption = DiscountRedemption(
+        redeemed_discount=discount,
+        redemption_date=now_in_utc(),
+        redeemed_order=order,
+        redeemed_by=user,
+    )
+    redemption.save()
+
+
+def test_one_time_discount(user, onetime_discount):
```
rhysyngsun, line 129: These tests for model methods should probably live in a `models_test.py` file instead.

rhysyngsun (changes requested):
Code review

jkachel:
https://github.com/mitodl/mitxonline/pull/444/commits/68b5c4546671ef03019856f0c41bd18ac710feb4 should handle all the comments from Friday (3/4).

static/js/containers/pages/checkout/CartPage.js
```diff
@@ -2,34 +2,140 @@
...
+    }
+
+    addUserNotification({
+      "add-discount-notification": {
+        type:  messageType,
+        props: {
+          text: userMessage
+        }
+      }
+    })
```
rhysyngsun, line 105: I think it's fine to just drop this, it shifts the page layout, which is a bit distracting and there's already an inline error message.

rhysyngsun (changes requested):
I'm getting stuck on this error when checking out:

```
web_1     | [2022-03-07 22:15:10] ERROR 13 [django.request] log.py:224 - [39ff2ccd9958] - Internal Server Error: /checkout/to_payment
web_1     | Traceback (most recent call last):
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
web_1     |     response = get_response(request)
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
web_1     |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
web_1     |   File "/usr/local/lib/python3.9/site-packages/sentry_sdk/integrations/django/views.py", line 67, in sentry_wrapped_callback
web_1     |     return callback(request, *args, **kwargs)
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
web_1     |     return self.dispatch(request, *args, **kwargs)
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/contrib/auth/mixins.py", line 71, in dispatch
web_1     |     return super().dispatch(request, *args, **kwargs)
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
web_1     |     return handler(request, *args, **kwargs)
web_1     |   File "/src/./ecommerce/views/v0/__init__.py", line 474, in get
web_1     |     checkout_payload = api.generate_checkout_payload(request)
web_1     |   File "/src/./ecommerce/api.py", line 19, in generate_checkout_payload
web_1     |     order = PendingOrder.create_from_basket(basket)
web_1     |   File "/usr/local/lib/python3.9/contextlib.py", line 79, in inner
web_1     |     return func(*args, **kwds)
web_1     |   File "/src/./ecommerce/models.py", line 273, in create_from_basket
web_1     |     total += line.discounted_price
web_1     |   File "/usr/local/lib/python3.9/site-packages/django/utils/functional.py", line 48, in __get__
web_1     |     res = instance.__dict__[self.name] = self.func(instance)
web_1     |   File "/src/./ecommerce/models.py", line 432, in discounted_price
web_1     |     DiscountType.get_discounted_price(
web_1     |   File "/src/./ecommerce/discounts.py", line 77, in get_discounted_price
web_1     |     discount_cls = DiscountType.for_discount(discount)
web_1     |   File "/src/./ecommerce/discounts.py", line 61, in for_discount
web_1     |     DiscountTypeCls = cls._CLASSES[discount.discount_type]
web_1     | AttributeError: 'DiscountRedemption' object has no attribute 'discount_type'
```

I think this is because my suggestion [here]() wasn't correct and that it needs to step into `DiscountRedemption.redeemed_discount`.

rhysyngsun (approved):

PR #462, opened 2022-03-08
Account for baskets that end up being zero-value after discounts

### LINKED ISSUE #461: Ecommerce: handle orders that have been discounted to $0
Cybersource fails payments that are $0, so we need to not send the learner over there:

- [ ] Immediately fulfill the order (including enrolling the learner)
- [ ] Record a transaction indicating that there was no payment required (`amount=0, data={"reason": "No payment required"}`)
- [ ] Redirect to the dashboard with a note saying they're now enrolled in the course (ensure this doesn't indicate that a payment occurred)

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

#461

#### What's this PR do?
Fixes #461 

Create a coupon that is either: a) 100% off, b) a fixed-price coupon set to $0, or c) set to the purchase price of a product in the database.
Redeem said coupon. The total after discount should display $0, and you should be able to check out (see notes below). 
Clicking Place Order should not redirect you through to the interstitial page and then on to CyberSource - instead, you should land back on the dashboard with the message "Success! You are now enrolled for the certificate version of <course>" and your enrollments should be updated to be "verified" rather than "audit". 

The above workflow should also work if the discount in question is set up as a user discount. 

Additionally, viewing the order in djadmin should now display the transactions associated with the order. Transactions should be view-only (no create, add, or delete). A zero-value transaction should look like this:
image

Two things to note here:
- There is a new constant for the success text for zero-value carts. The only difference right now is that instead of saying "paid version", it says "certificate version". This probably needs adjustment. (The issue for this states that the text should not indicate that payment has been accepted for the enrollment - I took that to mean saying "paid version" would do so.) 
- This does not handle the case where all of the products in the user's basket are themselves zero-value. That is, if the product itself has a price of $0, you won't be able to check out; the items themselves have to have a set value first. (I can update this to handle that case but it didn't seem like a thing that would actually happen.)

This is just the new success message; there are no other UI changes.
image


annagav (approved):
👍

PR #468, opened 2022-03-10
Adds transaction_type to the Transaction table

### LINKED ISSUE #467: Add `transaction_type` field to `Transaction`
This should be a `ChoiceField`, with supported options `payment` and `refund`. The default should be `payment` so that existing records are treated appropriately.

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

#467 

#### What's this PR do?
Fixes #467 

Run migrations - after this, the transaction_type field should be present in Transaction and all existing records should be set to "payment". 
After completing checkout, the new transaction record should be of type "payment".


ecommerce/constants.py
```diff
@@ -1,5 +1,8 @@
 """Constants for ecommerce."""
 
+from tkinter import ALL
```
annagav, line 3: I don't think this is used anywhere

annagav (changes requested):
Just a minor comment about unused import

annagav (approved):
👍

PR #469, opened 2022-03-10
Adds logic to avoid stepping on an in-progress basket when processing checkout responses

### LINKED ISSUE #453: Ecommerce: when checkout completes, delete only the basket that was checked out
Currently when a user checks out their basket, we defer deletion of the basket under later - this could be a) when the browser navigates back to the site or b) when the payment processor sends us an API request. Right now we just delete whichever basket the user has, but that may be a _new_ basket, particularly if the API call is deferred, and we don't want to inadvertently stomp on that state.

I'd propose that we adjust the callback and api urls to accept a basket id as part of the url path. Then we selectively `Basket.objects.filter(user=order.purchaser, id=basket_id)` and delete that one. Note that we want to filter both on the basket id and user.

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

Fixes #453 

#### What's this PR do?
Fixes #453 

Create a basket, convert to an Order, and then set the is_review flag to True on the resultant order. Then, create a new basket and attempt to process the previous Order. The basket should remain untouched. (Not sure how you'd get CyberSource to put the transaction into a review state but you can do that too if possible.) 

You should also be able to create a basket, go through the checkout process, hit the payment page, then go back (in a separate window), create a new basket again, and then complete the payment process for the first basket. The cart in the separate window should not be affected (you will have to reload the page). Checking out that cart normally should still clear the cart.


ecommerce/models.py
```diff
@@ -194,6 +228,10 @@ def choices(cls):
         decimal_places=5,
         max_digits=20,
     )
+    # Flag to determine if the order is in review status - if it is, then
+    # we need to not step on the basket that may or may not exist when it is
+    # accepted
+    is_review = models.BooleanField(default=False)
```
rhysyngsun, line 234: This would be better as an `@property` derived from `state` rather than its own column so we don't have to worry about keeping those synchronized.

jkachel:
Reworked the flag stuff to use the state engine instead in https://github.com/mitodl/mitxonline/pull/469/commits/09f407010180e20e845c060f6bb579951e11b912

rhysyngsun (approved):
LGTM

PR #472, opened 2022-03-11
Fix cart total display when no discounts are applied

### LINKED ISSUE #471: Cart total does not display properly when there's no discount applied
### Steps to Reproduce

On RC:
- Add something to the cart.
- Observe total - it will be a plain 0 (no currency formatting, and won't reflect the actual price of the cart)

### Expected Behavior

Cart total should be properly displayed with currency formatting.

### Actual Behavior

It doesn't. The display is correct when there's a discount applied. 

### Related Issues

#431

### Screenshot or Screencast

image

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#471 

#### What's this PR do?
Fixes #471 

Add an item to the cart - the total display should be correct. Alternatively, add an item to the cart and hit the cart API (`api/v0/checkout/cart/`) - the discounted_price should match the total_price.

### DIFF
--- a/ecommerce/serializers.py
+++ b/ecommerce/serializers.py
@@ -164,15 +164,14 @@ class BasketWithProductSerializer(serializers.ModelSerializer):
 
     def get_total_price(self, instance):
         return sum(
-            product.price
-            for product in [item.product for item in instance.basket_items.all()]
+            basket_item.base_price for basket_item in instance.basket_items.all()
         )
 
     def get_discounted_price(self, instance):
         discounts = instance.discounts.all()
 
         if discounts.count() == 0:
-            return None
+            return self.get_total_price(instance)
 
         return sum(
             basket_item.discounted_price for basket_item in instance.basket_items.all()
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -97,6 +97,11 @@ class BasketItem(TimestampedModel):
             * self.quantity
         )
 
+    @cached_property
+    def base_price(self):
+        """Returns the total price of the basket item without discounts."""
+        return self.product.price * self.quantity
+
 
 class Discount(TimestampedModel):
     """Discount model"""


ecommerce/serializers.py
```diff
@@ -172,7 +172,9 @@ def get_discounted_price(self, instance):
         discounts = instance.discounts.all()
 
         if discounts.count() == 0:
-            return None
+            return sum(
+                basket_item.product.price for basket_item in instance.basket_items.all()
```
annagav, line 176: The only thing that concerns me here, is that in the discounted amount the the discounted price is being multiplied by the quantity attribute of BasketItem, and here it is just the price independently of quantity of items. Maybe let's make it consistent.

annagav (changes requested):

ecommerce/serializers.py
jkachel: Good call; I added a method to the BasketItems model to handle the base price (before discounts) of the item in https://github.com/mitodl/mitxonline/pull/472/commits/a5d67a2057a9ef96698896b90d553f53dc2170f8

ecommerce/serializers.py
```diff
@@ -172,10 +172,14 @@ def get_discounted_price(self, instance):
...
         if discounts.count() == 0:
-            return None
+            return sum(
+                (basket_item.quantity * basket_item.product.price)
+                for basket_item in instance.basket_items.all()
+            )
 
         return sum(
-            basket_item.discounted_price for basket_item in instance.basket_items.all()
+            (basket_item.quantity * basket_item.discounted_price)
```
annagav, line 181: I think discounted_price is already price*quantity. I am looking here: https://github.com/mitodl/mitxonline/blob/main/ecommerce/models.py#L83

annagav (changes requested):

ecommerce/serializers.py
jkachel: Think I missed something in the last commit.. should be fixed now though.

PR #475, opened 2022-03-15
fix email unsubscription inconsistency after unenrollment

### LINKED ISSUE #445: Open edX - Course email subscription concern require confirmation
Although you mentioned that edX emails won't work if a user is unenrolled. Even if that's true we might end up with data inconsistency in case:
- Enroll User, Both MITxOnline and edX would have email subscribed (This is fine)
- Only unsubscribe from the email through MITxOnline Dashboard
- Unenroll the course and then reenroll the course. In this case, we will have Email Subscription=True in MITxOnline but in Edx the email subscription would still be inactive. Let me know if this would not be the case.

umarmughal824:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

fixes #445 

#### What's this PR do?
fix email unsubscription inconsistency after un-erollment

- Just unsubscribe to emails before unsubscribe
- enroll back in should be subscribed back 
- that should have deleted the popout entry in edX against email subscription

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -37,6 +37,7 @@ from openedx.api import (
     enroll_in_edx_course_runs,
     get_edx_api_course_detail_client,
     unenroll_edx_course_run,
+    subscribe_to_edx_course_emails,
 )
 from openedx.exceptions import (
     EdxApiEnrollErrorException,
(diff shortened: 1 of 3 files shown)


courses/api.py
```diff
@@ -218,7 +219,8 @@ def create_run_enrollments(
             )
             if not created and not enrollment.active:
                 enrollment.edx_enrolled = edx_request_success
-                enrollment.edx_emails_subscription = True
+                subscribed = subscribe_to_edx_course_emails(user, run)
+                enrollment.edx_emails_subscription = True if subscribed else False
```
annagav, line 223: do we want to update similar logic here as well? https://github.com/mitodl/mitxonline/blob/cc12b7556e44e6600eb0b7ba30043f077d2c1d5f/courses/views/v1/__init__.py#L220

annagav (changes requested):

annagav (approved):
👍

PR #477, opened 2022-03-16
Check for blocked countries during checkout

### LINKED ISSUE #447: Ecommerce: perform check on course blocked countries for payments
Currently this check is only done when enrolling in the free course, we also need to make this check when the learner is trying to pay for a course.

- [ ] Refactor [this code](https://github.com/mitodl/mitxonline/blob/cc12b7556e44e6600eb0b7ba30043f077d2c1d5f/courses/views/v1/__init__.py#L118) into a model method.
- [ ] Add verification of the products the learner is trying to pay for [here](https://github.com/mitodl/mitxonline/blob/main/ecommerce/views/v0/__init__.py#L415). Redirect to the cart with an error indicating they cannot complete their purchase (I'll get back to this on some language).

annagav:
Fix #447

#### What's this PR do?
Check for blocked countries during checkout

in the admin add a BlockedCountry instance for your user's country, and link it to the course you are going to checkout with. 
Try to enroll your user in a course.
Screen Shot 2022-03-22 at 3 55 22 PM

### DIFF
--- a/ecommerce/views/v0/__init__.py
+++ b/ecommerce/views/v0/__init__.py
@@ -9,6 +9,7 @@ from main.constants import (
     USER_MSG_TYPE_PAYMENT_ERROR,
     USER_MSG_TYPE_PAYMENT_ERROR_UNKNOWN,
     USER_MSG_TYPE_PAYMENT_REVIEW,
+    USER_MSG_TYPE_ENROLL_BLOCKED,
 )
 from main.utils import redirect_with_user_message
 from rest_framework import mixins, status
@@ -415,7 +416,13 @@ class CheckoutCallbackView(View):
                 )
             )
             if basket:
-                basket.delete()
+                if basket.has_user_blocked_products(order.purchaser):
+                    return redirect_with_user_message(
+                        reverse("user-dashboard"),
+                        {"type": USER_MSG_TYPE_ENROLL_BLOCKED},
+                    )
+                else:
+                    basket.delete()
 
             return redirect_with_user_message(
                 reverse("user-dashboard"), {"type": USER_MSG_TYPE_PAYMENT_REVIEW}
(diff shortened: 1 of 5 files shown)


ecommerce/views/v0/__init__.py
```diff
@@ -413,7 +414,13 @@ def post(self, request, *args, **kwargs):
                 )
             )
             if basket:
-                basket.delete()
+                if basket.has_user_blocked_products(order.purchaser):
```
jkachel, line 417: I don't think this check should only be done here. (I know the issue indicated to put this check here specifically - not sure if this was really the intent, though.) Review is a state that most transactions won't hit, so learners will generally not hit the check. Following this pattern, I think it should also run this check for the Accepted state.

That said, this is pretty late in the payment process - at this point we've taken their money (potentially - if it's in Review, we haven't, but we're like 50% of the way there). Should this check be run at an earlier stage of the transaction? If we want to stick with it revolving around the handoff to CyberSource, it might be better to have it run before we send them off to pay. Or, I can see a good argument for having the check way further up at the point where they're adding the course to the cart - if they're in a blocked country, do they even need to be able to try to pay into a certificate? I get an error when attempting to enroll in the free course using my blocked country user, so that might be the better flow.

jkachel (changes requested):

ecommerce/views/v0/__init__.py
annagav: yes, That make much more sense.

annagav:
@jkachel This PR is ready for review again

ecommerce/views/v0/__init__.py
```diff
@@ -465,9 +472,8 @@ def get(self, request):
         except ObjectDoesNotExist:
             return HttpResponse("No basket")
 
-        if "no_checkout" in checkout_payload:
+        if "no_checkout" or "country_blocked" in checkout_payload:
```
jkachel, line 475: This needs to be two separate "in checkout_payload"s so the if doesn't short circuit early. (If the user can check out normally, you get a 500 error from this line.)

jkachel (changes requested):
One small change - the blocked countries check otherwise looks good to me.

jkachel (approved):
LGTM 👍

Issue #465, opened 2022-03-09
Support an order being refunded in the django-admin interface


rhysyngsun:
A staff user should be able to manage the refund of an order.

- [x] Ideally this admin action is only available at the level of an individual order (i.e. the change page for an order) and not as a bulk operation.
- [x] This action should have an [intermediate page](https://docs.djangoproject.com/en/4.0/ref/contrib/admin/actions/#actions-that-provide-intermediate-pages) with a form with the following inputs:
  - A numeric input for the refund amount.
  - A text field for the `reason` of the refund, this is an opaque value.
  - A checkbox (default unchecked) to optionally unenroll the learner.
- [x] Based on the inputs above, `Order.refund` should be called.
- [x] Update the signature to `Order.refund(amount: float, reason: str, unenroll_learner: bool)` and update the implementation to:
  - Add a `Transaction` record with `transaction_type="refund"`, the `amount` passed, and `data={"reason":reason}`
  - Conditionally unenroll the learner if specified. At the very least, the enrollment record should get updated. If the edx call fails this will be reattempted later and we should present an error message to the staff user explaining this. See also https://github.com/mitodl/mitxonline/issues/456

Issue #456, opened 2022-03-03
Ecommerce: enrollments should still be created in the db if openedx api call fails


rhysyngsun:
### To Reproduce:
- Pay for a course that doesn't have a verified track
- You'll see an error, but no `CourseRunEnrollment` gets created

### Expected:

- A `CourseRunEnrollment` record is created so we can reattempt later

PR #480, opened 2022-03-17
Handle refunded orders

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

#465 
#456 

#### What's this PR do?
Adds ability to refund orders in Django Admin. Also updates the fulfill code to keep local enrollments when enrollments in edX fail. 

In Django Admin:
1. Find a Fulfilled Order
2. There should be a Refund Order button on the submit line
3. Clicking the button should bring up a simple form with options for reason, amount, and unenroll. Amount should be pre-filled with the purchase price of the order.
4. Leaving either reason or amount unfilled should re-render the form with an error message (in the same manner as it would in djadmin)
5. Successfully refunding the order should redirect you to the Refunded Orders list page with a success notification. A Transaction record should be created with the data being a dict of reason and amount, with their values being what was supplied in the form. 
6. If the refund succeeds but the unenrollments in edX fails, you should instead get a warning message indicating that. (For testing, I adjusted the refund method in FulfilledOrder to always return False.) 

The URL can be adjusted to attempt to refund an order that does not exist at all. In this case, you should be redirected to the Fulfilled Order list page in djadmin with a "not found" error.

The URL can be adjusted to attempt to refund an order that's already refunded (or otherwise isn't Fulfilled). In this case, you should be redirected to the Fulfilled ORder list page in djadmin with a "not found, might be fulfilled" error. 

If you attempt to refund an order whose state changes from Fulfilled to something else (and thus won't have a refund method anymore), you should be redirected to the Refunded Orders list page with a "can't be refunded" message. 

#### Notes

- The refund order screen may need more data
- The landing pages after you perform an action may need to change - they work like this now:
	- Refund successful: RefundedOrder change (for selected order)
	- Hit NotImplemented: RefundedOrder changelist
	- Hit ObjectDoesNotExist in POST: FulfilledOrder changelist
	- Hit ObjectDoesNotExist in GET: FulfilledOrder changelist

Button on order screen:
fulfilled-order-with-refund-button

Refund screen:
refund-order-screen

Success message:
unenroll-succeeded

Error messages:
unenroll-failed
refund-failed-not-found
refund-failed-not-FulfilledOrder

### DIFF
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -4,6 +4,8 @@ from django.contrib.admin.decorators import display
 from fsm_admin.mixins import FSMTransitionMixin
 from reversion.admin import VersionAdmin
 from mitol.common.admin import TimestampedModelAdmin
+from django.http import HttpResponse, HttpResponseRedirect
+from django.urls import reverse
 
 from ecommerce.models import (
     Product,
@@ -51,7 +53,7 @@ class BasketItemAdmin(VersionAdmin):
 class DiscountAdmin(admin.ModelAdmin):
     model = Discount
     search_fields = ["discount_type", "redemption_type"]
-    list_display = ["id", "discount_type", "amount", "redemption_type"]
+    list_display = ["id", "discount_code", "discount_type", "amount", "redemption_type"]
 
 
 @admin.register(UserDiscount)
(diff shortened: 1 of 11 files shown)


ecommerce/models.py
```diff
@@ -334,11 +337,55 @@ class FulfilledOrder(Order):
...
+        self.transactions.create(
+            data={"reason": reason}, amount=amount, transaction_type="refund"
+        )
+
+        self.state = Order.STATE.REFUNDED
+
+        unenrolls_successful = True
+
+        if unenroll_learner:
+            from courses.api import deactivate_run_enrollment
```
annagav, line 368: why are we importing this here?

annagav (changes requested):

ecommerce/admin.py
```diff
@@ -158,14 +160,36 @@ def get_queryset(self, request):
...
     model = FulfilledOrder
 
+    @display(description="Purchaser")
+    def get_purchaser(self, obj: Order):
+        return f"{obj.purchaser.name} ({obj.purchaser.email})"
+
     def get_queryset(self, request):
-        """Filter only to fulfilled orders"""
-        return super().get_queryset(request).filter(state=Order.STATE.FULFILLED)
+        """Filter only to pending orders"""
```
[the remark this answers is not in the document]
jkachel: Fulfilled - typo.

ecommerce/models.py
jkachel: It was only used here. I moved it to the top in https://github.com/mitodl/mitxonline/pull/480/commits/69b947a8412a5ccaed5ea275414feaa3041371b6

annagav (approved):
👍

PR #483, opened 2022-03-18
Jkachel/466 display refund tags on receipt

### LINKED ISSUE #466: Support the display of the status of a refund on the receipt page
- [ ] Display an indication of the order state:
    - `REFUNDED` -> Refunded
    - `FULFILLED` -> Paid
- [ ] Display a line item indicating the amount refunded (e.g. "Refunded $200")

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

Fixes #466 

#### What's this PR do?
Fixes #466. Adds a label and readout for refunds applied to an order.

_This would be better tested after #480 is merged. You can manually add Transactions to test, though._

From the Order History page, select an order that has been refunded. The refund(s) should show in the order summary on the right. If the refund has multiple refunds, each will be listed. (The system doesn't currently have UI support for multiple partial refunds but it can be done by either manually adding Transaction records in or by flipping an order to Fulfilled after it has been set to Refunded, and then refundng it again.) 

#### Notes
- Wasn't real sure if the total displayed should be the order total or the actual amount paid (which, after a refund, would be the order total minus refunds), so it doesn't adjust the total now. (In the screenshots below, the effective cost for that order is zero since the entire order was refunded.) 
- The refund date is available in the refund store, but not displayed - do we want to display this? (The reason is explicitly not  transmitted as that seems like a bad idea as there may be order reasons that are not to be viewed by the learner.) 

image
image


static/js/components/OrderSummaryCard.js
```diff
@@ -69,6 +70,48 @@ export class OrderSummaryCard extends React.Component<Props> {
       </div>
     )
   }
+
+  renderIndividualRefund(refund: Refund) {
+    if (refund === null) {
+      return null
+    }
+
+    console.log(refund)
```
annagav, line 79: could you remove this line

annagav (changes requested):

jkachel:
removed the extra console log (oops) and added the Paid tag - it's in under the total price if the order is Fulfilled. It shouldn't show up otherwise. It can be moved elsewhere; that just seemed like a decent enough place to put it.

static/js/components/OrderSummaryCard.js
```diff
@@ -107,6 +152,8 @@ export class OrderSummaryCard extends React.Component<Props> {
           ? this.renderAppliedCoupons()
           : null}
 
+        {this.renderRefunds()}
+
```
annagav, line 156: I think we should add the paid tag here, and make them conditional on each other. Maybe something like this ` {(refunds === null || refunds.length === 0) ? this.renderFulfilledTag() : this.renderRefunds()}`.

Also, this variable doesn't reflect the precise status of the order, it just reflects before payment (cart checkout stage) or after payment( order receipt page)

annagav (changes requested):

static/js/components/OrderSummaryCard.js
annagav: so you will notice now on the refunded page both tags are being displayed

jkachel: Changed this accordingly (see screenshots). 

No refund:
image
Refund:
image

annagav (approved):
👍

PR #498, opened 2022-04-01
Adjusts tests to make them more reliable

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#### What's this PR do?

These two tests were reworked a bit to make them more reliable - they seemed to work OK locally but would trip up the CI runner:
- test_checkout_result - would fail basket check after processing in the Review state. For testing orders in the Review state, the test creates an order, sets it to Review, and then attempts to create a new basket and processes the order to make sure the basket isn't touched. The new basket sometimes gets created with the same product in it as the old basket, though, which causes it to fail. It doesn't do that now.
- test_discounted_price - needed simplification and a small change to the Discount API to make all the types return a consistent type (they all return Decimal now; the fixed price one would return a bare number).


annagav (approved):
👍

PR #503, opened 2022-04-05
Add support for yarn workspaces

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
This PR switches the app over to using yarn workspaces. The goal here is to be able to have multiple projects where `package.json` requirements may differ. The immediate need is to be able to add [Refine](https://refine.dev) to this project and support IDEs being able to recognize that project correctly (our current setup doesn't work because it ends up being a project within a project).

To accomplish this, a few things were done:

- Moved sources under `static/js` and `static/scss` to `frontend/public` (I'm open to a better name). I did this because I didn't want a `package.json` in `static/`, which is served via django and because `static/` is meant for actually static files, which our sources are not. The actual static files are the ones output from webpack to `static/bundles`.
- Updated to a newer version of `node` - this also had the side effect of needed to upgrade a few other dependencies so I handled that as well.

- Remove all the files under `static/bundles` to be sure that the sources build.
- Remove `webpack-stats.json` from the project root to be sure that part of the build still works.
- Run `docker-compose up` and verify there's no error output form the `watch` container and that the app still functions.

- Remove those files again but this time run `NODE_ENV=production ./webpack_if_prod.sh` in the `watch` container instead of `docker-compose up`, you should see those files get generated again but with production settings.

- You should be able to run the yarn scripts `lint`, `scss_lint`, `flow`, `test`, `coverage`, `fmt` via both:

- From the repo root: `yarn workspaces foreach run <script_name>`
- From `frontend/public`: `yarn run <script_name>`


jkachel:
Getting some errors running the watch container after pull and rebuild:

```
mitxonline-watch-1  |  - configuration.module.rules[1].type should be one of these:
mitxonline-watch-1  |    "javascript/auto" | "javascript/dynamic" | "javascript/esm" | "json" | "webassembly/experimental"
mitxonline-watch-1  |    -> Module type to use for the module
mitxonline-watch-1  |  - configuration.optimization has an unknown property 'emitOnErrors'. These properties are valid:
mitxonline-watch-1  |    object { checkWasmTypes?, chunkIds?, concatenateModules?, flagIncludedChunks?, hashedModuleIds?, mangleWasmImports?, mergeDuplicateChunks?, minimize?, minimizer?, moduleIds?, namedChunks?, namedModules?, noEmitOnErrors?, nodeEnv?, occurrenceOrder?, portableRecords?, providedExports?, removeAvailableModules?, removeEmptyChunks?, runtimeChunk?, sideEffects?, splitChunks?, usedExports? }
mitxonline-watch-1  |    -> Enables/Disables integrated optimizations
```

Full log output is in a gist: https://gist.github.com/jkachel/fbf4f60fa07824983ff527a90734b4da

I did follow the steps up to the initial `docker compose up` run and for completeness also rebuilt the watch container with `--no-cache`. For what it's worth, I did `yarn install` manually in a one-off container and then ran the `webpack_dev_server.sh` script as described in the docker-compose file, and that seemed to work.

Issue #436, opened 2022-02-25
Epic: Financial Aid (Flexible Pricing)


rhysyngsun:
- [ ] Learner can input financial data
- [ ] Automatic approval is supported
- [ ] If manual approval is required, prompt learner to upload it
- [ ] Admin interface to approve financial aid

umarmughal824:
https://github.com/mitodl/mitxonline/issues/559

rhysyngsun:
Closing this as I'm going to reorganize the issues around this.

PR #511, opened 2022-04-12
Adds administrative discount APIs

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#436 (to an extent)

#### What's this PR do?
Adds some simple APIs for working with discount codes via REST API. These are meant to be used for upcoming Refine development.

Navigate to /api/v0/discounts - you should be able to work through the API via the standard DRF interface.
The discount detail should also include redemptions (with detail) for the discount. 

* This is a backport of stuff I set up while testing out Refine - it made some intuitive sense to have a Refine console to manage discount codes (since they'll be used in flexible pricing). 
* These are limited to admin-level users but can be bumped down to staff or something else (depending on how we want to limit access to the Refine bits that haven't been done yet). Admin made the most sense now.

### DIFF
--- a/ecommerce/serializers.py
+++ b/ecommerce/serializers.py
@@ -299,3 +299,58 @@ class OrderHistorySerializer(serializers.ModelSerializer):
         ]
         model = models.Order
         depth = 1
+
+
+class DiscountSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.Discount
+        fields = [
+            "id",
+            "amount",
+            "automatic",
+            "discount_type",
+            "redemption_type",
+            "max_redemptions",
+            "discount_code",
+        ]
+        depth = 2
+
+
+class DiscountRedemptionSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.DiscountRedemption
+        fields = [
+            "id",
+            "redemption_date",
+            "redeemed_by",
+            "redeemed_discount",
+            "redeemed_order",
+        ]
+        read_only_fields = [
+            "redemption_date",
+            "redeemed_by",
+            "redeemed_discount",
+            "redeemed_order",
+        ]
+        depth = 1
+
+
+class UserDiscountSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.UserDiscount
+        fields = [
+            "id",
+            "discount",
+            "user",
+        ]
+
+
+class UserDiscountMetaSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.UserDiscount
+        fields = [
+            "id",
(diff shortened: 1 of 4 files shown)


ecommerce/views/v0/__init__.py
```diff
@@ -199,6 +207,36 @@ def get_queryset(self):
...
+    authentication_classes = (SessionAuthentication, TokenAuthentication)
+    permission_classes = (IsAuthenticated, IsAdminUser)
+
+    @action(
+        detail=True,
+        methods=["get"],
+        name="List Discount Redemptions",
+        url_name="redemptions",
+    )
+    def redemptions(self, request, pk):
```
rhysyngsun, line 224: Could you use `rest_framework_extensions.routers.NestedRouterMixin` with a dedicated readonly view instead of a single detail view so that both list and detail for the redemptions?

rhysyngsun (changes requested):

ecommerce/serializers.py
```diff
@@ -299,3 +299,38 @@ class Meta:
...
+        model = models.Discount
+        fields = [
+            "id",
+            "amount",
+            "automatic",
+            "discount_type",
+            "redemption_type",
+            "max_redemptions",
+            "discount_code",
+            "order_redemptions",
```
rhysyngsun, line 315: I'm thinking maybe because `order_redemptions` could be a large set, we should remove this and defer to the nested API which should be paginated.

ecommerce/views/v0/__init__.py
```diff
@@ -199,6 +207,22 @@ def get_queryset(self):
...
+    queryset = Discount.objects.all()
+    authentication_classes = (SessionAuthentication, TokenAuthentication)
+    permission_classes = (IsAuthenticated, IsAdminUser)
+
+
+class DiscountRedemptionViewSet(NestedViewSetMixin, ModelViewSet):
+    serializer_class = DiscountRedemptionSerializer
+    queryset = DiscountRedemption.objects.all()
+    authentication_classes = (SessionAuthentication, TokenAuthentication)
+    permission_classes = (IsAuthenticated, IsAdminUser)
```
rhysyngsun, line 223: Could you add pagination to both of these?

rhysyngsun (changes requested):

jkachel:
Should be fixed now - also worth noting is there's now tests for the redemption stuff (missed that), and added some support in for user discounts (as a nested view within a discount, and as just a general viewset). The user discount stuff is for the Refine discounts code (currently in branch jkachel/refineadmin-discounts )

rhysyngsun (approved):

PR #513, opened 2022-04-12
Ports finaid/flex pricing models over from MicroMasters

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

#436 

#### What's this PR do?
Adds the base models - CountryIncomeThreshold, CurrencyExchangeRate, and FlexiblePrice - and the basic API for each. Adds the app for flexiblepricing. 

This also adds an admin api set for FlexiblePrice that does not limit the result set by the authenticated user.

APIs are testable through the DRF UI in the usual fashion - the APIs are generally at /api/v0/flexible_pricing/*.


flexiblepricing/constants.py
```diff
@@ -0,0 +1,33 @@
+class FlexiblePriceStatus:
+    """Statuses for the FlexiblePrice model"""
+
+    APPROVED = "approved"
+    AUTO_APPROVED = "auto-approved"
+    CREATED = "created"
+    DOCS_SENT = "docs-sent"
+    PENDING_DOCS = "pending-docs"
```
rhysyngsun, line 8: You can remove these, these are specific to the high degree of customization for DEDP that we won't be using in MITx Online.

flexiblepricing/models.py
```diff
@@ -0,0 +1,67 @@
...
+from flexiblepricing.constants import FlexiblePriceStatus
+
+
+class CurrencyExchangeRate(TimestampedModel):
+    """
+    Table of currency exchange rates for converting foreign currencies into USD
+    """
+
+    currency_code = models.CharField(null=False, max_length=3)
+    exchange_rate = models.FloatField(
```
rhysyngsun, line 17: This should be a `DecimalField` to avoid floating point wonkiness.

flexiblepricing/models.py
```diff
@@ -0,0 +1,67 @@
...
+from ecommerce.models import TimestampedModel
+from flexiblepricing.constants import FlexiblePriceStatus
+
+
+class CurrencyExchangeRate(TimestampedModel):
+    """
+    Table of currency exchange rates for converting foreign currencies into USD
+    """
+
+    currency_code = models.CharField(null=False, max_length=3)
```
rhysyngsun, line 16: This should be unique.

flexiblepricing/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
rhysyngsun, line 1: This file can be removed.

rhysyngsun (changes requested):

rhysyngsun (approved):
LGTM

Issue #519, opened 2022-04-14
Integrate Refine for internal dashboard


rhysyngsun:
We'd like a way to build internal dashboards using [Refine](https://refine.dev).

- [x] Add a minimal implementation that defers to the django app for authentication.
- [x] Update the build so that the dashboard gets built for heroku.
- [x] Update any necessary nginx configuration to ensure the dashboard is served.

PR #524, opened 2022-04-20
Cleanup and simplify configuration/localdev

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
- [x] Settings

Part of #519 

#### What's this PR do?
This updates some of the docker setup as well as nginx configuration. Some key things:

- Removes `docker-compose.override.yml` and `docker-compose.notebook.yml` in favor of merging these all down to `docker-compose.yml`, which using [profiles](https://docs.docker.com/compose/profiles/) to make the notebook service optional.
- Removes bespoke `config/nginx.conf` that is only used locally in favor of a single nginx config file that already exists at `config/nginx.conf.erb`
- Creates a `Dockerfile` that extends the `nginx` docker image in order to emulate the compilation of the config ERB template.

Run `docker-compose build` and `docker-compose up`, the app should be reachable and you should see the output of the `20-compile-nginx-conf-erb.sh` script in the nginx container output.


annagav (approved):
Works for me 👍

blarghmatey:
Looks good to me :+1:

rhysyngsun:
Marking this as blocked for now because I'd like to see Heroku deploy this in a PR app before merging.

PR #544, opened 2022-04-27
Adds discounts to Refine admin

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

n/a

#### What's this PR do?
Adds CRUD operations for discounts to the Refine admin UI. Also adds in some infrastructure stuff:
- AuthProvider in the API that doesn't check for the CSRF token (which Refine doesn't know about)
- A DRF-aware DataProvider for Refine; this is mostly the simple data provider with a couple of methods reworked a bit to make them usable with a DRF-based API

Once the code is checked out and the refine app rebuilds, you should have a Discounts menu item on the left. Clicking that should show you the list of discounts in the system. Create button should work, as should edit and view. The view button in particular should additionally show you redemptions and assigned users (UserDiscounts) for that discount. (You can't add new UserDiscounts yet.) 

The things that needed adjustment in the data provider are the pathnames that it constructs and the handling of metadata that comes back from a paginated DRF request. 
- Pathnames: these needed a trailing slash. Django usually takes care of this with a redirect, but it can't do that for POST-like requests. 
- Paginated DRF requests: The simple data provider expects a simple array of results back (and then to do the pagination itself). DRF sends back a bunch of metadata with it (item counts, prev/next, etc.) so rather than try to adjust the API to work with Refine, I opted to make Refine work with it so that it'd be easier to use with other DRF-based APIs. There was also a discrepancy in limit and offset parameters, so this takes care of that too. 

This code also implements the Bearer token for OIDC auth; it doesn't do that by default (that I saw) so it's implemented here using an axios-http interceptor. 

Screen Shot 2022-04-27 at 3 48 08 PM
Screen Shot 2022-04-27 at 3 48 16 PM
Screen Shot 2022-04-27 at 3 48 21 PM

### DIFF
--- a/frontend/staff-dashboard/src/App.tsx
+++ b/frontend/staff-dashboard/src/App.tsx
@@ -2,7 +2,6 @@ import { Refine } from "@pankod/refine-core";
 import { Icons, notificationProvider } from "@pankod/refine-antd";
 import routerProvider from "@pankod/refine-react-router-v6";
 import "@pankod/refine-antd/dist/styles.min.css";
-import dataProvider from "@pankod/refine-simple-rest";
 import { useAuthProvider } from "hooks/useAuthProvider";
 import {
   Title,
@@ -14,20 +13,50 @@ import {
 } from "components/layout";
 import LoginPage from "pages/login";
 import { DashboardPage } from "pages/dashboard";
+import { DiscountList, DiscountEdit, DiscountShow, DiscountCreate } from "pages/discounts";
+import axios from "axios";
+import useDrfDataProvider from "hooks/useDrfDataProvider";
 
-const {UserOutlined} = Icons;
+const {UserOutlined, BarcodeOutlined} = Icons;
+const axiosInterface = axios.create();
+
+axiosInterface.interceptors.request.use((config: any) => {
+  let token = sessionStorage.getItem(`oidc.user:${OIDC_CONFIG.authority}:${OIDC_CONFIG.client_id}`);
+
+  if (token !== null) {
+    token = JSON.parse(token).access_token;
+    config.headers.Authorization = `Bearer ${token}`;
+  }
+
+  return config;
+}, (error: any) => Promise.reject(error));
 
 export default function App() {
   const authProvider = useAuthProvider();
+  const xonlineProvider = useDrfDataProvider("http://mitxonline.odl.local:8013/api/v0");
+
   return (
     <Refine
       routerProvider={routerProvider}
       notificationProvider={notificationProvider}
-      dataProvider={dataProvider("https://api.fake-rest.refine.dev")}
+      dataProvider={xonlineProvider}
       authProvider={authProvider}
       LoginPage={LoginPage}
       DashboardPage={DashboardPage}
-      resources={[{name: "learners", icon: <UserOutlined/>}]}
+      resources={[
+        {
+          name: "learners", 
+          icon: <UserOutlined/>
+        },
+        {
+          name: "discounts",
+          icon: <BarcodeOutlined/>,
+          show: DiscountShow,
+          list: DiscountList,
+          edit: DiscountEdit,
+          create: DiscountCreate,
+        },
+      ]}
       Title={Title}
       Header={Header}
       Sider={Sider}
--- a/frontend/staff-dashboard/src/pages/discounts/edit.tsx
+++ b/frontend/staff-dashboard/src/pages/discounts/edit.tsx
@@ -1,0 +1,45 @@
+import { 
+    useForm,
+    Form,
+    Input,
+    InputNumber,
+    Select,
+    Edit,
+ } from "@pankod/refine-antd";
+
+import { IDiscount } from "interfaces";
+
+
+export const DiscountEdit = () => {
+    const { formProps, saveButtonProps, queryResult } = useForm<IDiscount>();
+    const discount_type = queryResult?.data?.data.discount_type
+
+    return (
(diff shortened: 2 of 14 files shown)


frontend/staff-dashboard/src/App.tsx
```diff
@@ -14,20 +14,59 @@ import {
 } from "components/layout";
 import LoginPage from "pages/login";
 import { DashboardPage } from "pages/dashboard";
+import { DiscountList, DiscountEdit, DiscountShow, DiscountCreate } from "pages/discounts";
+import { UserDiscountCreate } from "pages/discounts/user";
```
rhysyngsun, line 18: This import is failing for me:

```
refine_1  | ERROR in src/App.tsx:18:36
refine_1  | TS2307: Cannot find module 'pages/discounts/user' or its corresponding type declarations.
refine_1  |     16 | import { DashboardPage } from "pages/dashboard";
refine_1  |     17 | import { DiscountList, DiscountEdit, DiscountShow, DiscountCreate } from "pages/discounts";
refine_1  |   > 18 | import { UserDiscountCreate } from "pages/discounts/user";
refine_1  |        |                                    ^^^^^^^^^^^^^^^^^^^^^^
refine_1  |     19 | import axios from "axios";
refine_1  |     20 | import DrfDataProvider from "hooks/drfDataProvider";
refine_1  |     21 |
```

frontend/staff-dashboard/src/hooks/drfDataProvider.ts
```diff
@@ -0,0 +1,172 @@
...
+            const { _sort, _order } = generatedSort;
+            query._sort = _sort.join(",");
+            query._order = _order.join(",");
+        }
+
+        console.log(query);
+
+        const uri = `${url}?${stringify(query)}&${stringify(queryFilters)}`;
+
+        console.log(`uri is ${uri}`);
```
rhysyngsun, line 121: You can condense all of this down to a few lines:

```typescript
const query = {
    ...generateFilter(filters),
    ...generatePagination(pagination),  // note this should be implemented as a refactor
    ...generateSort(sort), // this should now return {} instead of undefined if no sort
}
const uri = `${url}?${stringify(query)}`;
```

rhysyngsun (changes requested):
I'm having difficulties with django crashing, I'm not sure if it's your changes or some bad state on my end though. Here's a review of the code while I sort that out.

ecommerce/views/v0/__init__.py
```diff
@@ -84,6 +84,11 @@
 log = logging.getLogger(__name__)
 
 
+class CsrfExemptSessionAuthentication(SessionAuthentication):
```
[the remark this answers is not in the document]
jkachel: New commits fix this by exposing CSRF_ALLOW_CREDENTIALS (so axios can attach cookies to the response) and adjusting the defaults in axios to provide credentials and to set the CSRF cookie and header to the right values. (This does mean there's a change in the main/settings.py - I've defaulted this to True for now.)

jkachel:
Last commit addresses all comments - there's some other changes that relate to CSRF protection that I noted in the comment for that.

frontend/staff-dashboard/src/App.tsx
```diff
@@ -14,20 +13,50 @@ import {
...
+    token = JSON.parse(token).access_token;
+    config.headers.Authorization = `Bearer ${token}`;
+  }
+
+  return config;
+}, (error: any) => Promise.reject(error));
 
 export default function App() {
   const authProvider = useAuthProvider();
+  const xonlineProvider = useDrfDataProvider("http://mitxonline.odl.local:8013/api/v0");
```
rhysyngsun, line 36: It looks like this is hardcoded to the local hostname, but we can figure out separately how to deal with this since we need it to work both locally with the live server and in deployed environments with the static build (once I get that working on heroku), could you open an issue for that?

frontend/staff-dashboard/src/pages/discounts/edit.tsx
```diff
@@ -0,0 +1,42 @@
...
+                    </Form.Item>
+                    <Form.Item label="Discount Type" name="discount_type">
+                        <Select options={[
+                            { label: 'Percent Off', value: 'percent-off' },
+                            { label: 'Dollars Off', value: 'dollars-off' },
+                            { label: 'Fixed Price', value: 'fixed-price' }, 
+                        ]}></Select>
+                    </Form.Item>
+                    <Form.Item label="Amount" name="amount">
+                        <Input />
```
rhysyngsun, line 36: Is there a way to adjust the precision on this to only show 2 digits after the decimal? Even better if there's a currency input you can utilize for both this and the create form.

rhysyngsun (approved):
One small nit but otherwise looking good.

frontend/staff-dashboard/src/App.tsx
jkachel: Added issue https://github.com/mitodl/mitxonline/issues/550 for this one.

frontend/staff-dashboard/src/pages/discounts/edit.tsx
jkachel: There was a InputNumber component for this that allows for a set precision - I swapped these over (on create and edit) to use that. (It has a bit of a styling bug, as the field isn't full width like the rest, but since there'll need to be a general styling overhaul here later I'm ignoring it for now.)

Issue #532, opened 2022-04-22
Receipt: custom email 


pdpinch:
As a user, I'd like to get a complete and customized receipt via email.

#### Designs and Mockups

As close as possible to what we do for xPRO:

<img width="940" alt="image" src="https://user-images.githubusercontent.com/430126/164752128-3a07dbc4-ab54-43d8-bb0a-fa8dd2344148.png">

#### Acceptance Criteria:

- [ ] same add'l information as #531 

#### Related issues

- #531

pdpinch:
On Hold until @rhysyngsun provides some guidance on how we can put this into an ol-django library.

jkachel:
Question (probably for @rhysyngsun) - do we need the more complete user record for this (like xPro has, which includes address fields and such) or can I pull that data from the CyberSource payload? My thinking is that, since the receipt is a record of a particular transaction, the address should be the one that was supplied in the checkout process (which may be different from the learner's actual address), but xPro does this differently.

rhysyngsun:
@jkachel I agree that it should match the actual payment information from cybersource.

PR #558, opened 2022-05-11
Adds email receipts

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#532 

#### What's this PR do?
This is mostly a cross-port of the xPro receipt email code into MITxOnline. Upon successful payment, this will send the learner an emailed copy of their receipt. 

1. Ensure you have Mailgun credentials in your local environment, or you won't be able to send email. The credentials from RC can be used for testing.
2. Complete the payment process. 
3. You should receive an email from the system with the receipt for your purchase. It should look mostly like the email that is sent by xPro (see #532 for a screenshot of that).

Additionally, there's a pytest test for the serializer itself. You can manually send emails for other (completed) orders by loading the order and calling mail_api.send_ecommerce_order_receipt. 

Depending on the CyberSource settings, you still may get a CyberSource receipt email. That functionality is set in the CS account so CyberSource receipts should be turned off when this is ready for production.

(My local dev doesn't have the logo - that is part of the base email template so should work elsewhere; I just don't have it locally. Color changes below are from Gmail colorization, not the email template.) 

Screen Shot 2022-05-12 at 2 20 51 PM

### DIFF
--- a/ecommerce/mail_api.py
+++ b/ecommerce/mail_api.py
@@ -1,0 +1,64 @@
+"""Ecommerce mail functions"""
+import pycountry
+import logging
+
+from ecommerce.messages import OrderReceiptMessage
+from mitol.mail.api import get_message_sender
+
+log = logging.getLogger()
+
+
+def send_ecommerce_order_receipt(order):
+    """
+    Send emails receipt summarizing the user purchase detail.
+
+    Args:
+        order: An order.
+    """
+    from ecommerce.serializers import OrderReceiptSerializer
+
+    data = OrderReceiptSerializer(instance=order).data
+    purchaser = data.get("purchaser")
+    coupon = data.get("coupon")
+    lines = data.get("lines")
+    order = data.get("order")
+    receipt = data.get("receipt")
+    country = pycountry.countries.get(alpha_2=purchaser.get("country"))
+    recipient = purchaser.get("email")
+
+    try:
+        with get_message_sender(OrderReceiptMessage) as sender:
+            sender.build_and_send_message(
+                recipient,
+                {
+                    "coupon": coupon,
+                    "content_title": lines[0].get("content_title") if lines else None,
+                    "lines": lines,
+                    "order_total": format(
+                        sum(float(line["total_paid"]) for line in lines),
+                        ".2f",
+                    ),
+                    "order": order,
+                    "receipt": receipt,
+                    "purchaser": {
+                        "name": " ".join(
+                            [
+                                purchaser.get("first_name"),
+                                purchaser.get("last_name"),
+                            ]
+                        ),
+                        "email": purchaser.get("email"),
+                        "street_address": purchaser.get("street_address"),
+                        "state_code": purchaser.get("state_or_territory").split("-")[
+                            -1
+                        ],
+                        "postal_code": purchaser.get("postal_code"),
+                        "city": purchaser.get("city"),
+                        "country": country.name if country else None,
+                        "company": purchaser.get("company"),
+                    },
+                },
+            )
+
+    except:  # pylint: disable=bare-except
+        log.exception("Error sending order receipt email.")
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -26,7 +26,8 @@ from ecommerce.constants import (
 )
 from users.models import User
 from decimal import Decimal
-from courses.api import deactivate_run_enrollment
+from courses.api import deactivate_run_enrollment, create_run_enrollments
+from ecommerce.tasks import send_ecommerce_order_receipt
 
 from mitol.common.utils.datetime import now_in_utc
 
@@ -357,7 +358,6 @@ class PendingOrder(Order):
     @transition(field="state", source=Order.STATE.PENDING, target=Order.STATE.FULFILLED)
     def fulfill(self, payment_data):
         """Fulfill this order"""
-        from courses.api import create_run_enrollments
 
         # record the transaction
         self.transactions.create(
@@ -372,6 +372,9 @@ class PendingOrder(Order):
             mode=EDX_ENROLLMENT_VERIFIED_MODE,
             keep_failed_enrollments=True,
         )
+
+        # send the receipt emails
+        send_ecommerce_order_receipt.delay(self.id)
 
     @transition(field="state", source=Order.STATE.PENDING, target=Order.STATE.CANCELED)
     def cancel(self):
--- a/ecommerce/serializers.py
+++ b/ecommerce/serializers.py
@@ -1,14 +1,15 @@
 """
 MITxOnline ecommerce serializers
 """
-from this import d
+import json
 from rest_framework import serializers
+from decimal import Decimal
 
 from ecommerce import models
 from courses.models import CourseRun, ProgramRun, Course
-from ecommerce.models import Basket, Product, BasketItem
+from ecommerce.models import Basket, Product, BasketItem, Order
 from ecommerce.discounts import DiscountType
-from ecommerce.constants import TRANSACTION_TYPE_REFUND
+from ecommerce.constants import TRANSACTION_TYPE_REFUND, CYBERSOURCE_CARD_TYPES
 
 from cms.serializers import CoursePageSerializer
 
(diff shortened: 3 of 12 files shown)


mail/templates/product_order_receipt/body.html
```diff
@@ -0,0 +1,88 @@
+{% extends "email_base.html" %}
+
+{% block content %}
+<tr class="receipt-new-wrapper" style="font: 14px/22px 'Arial', sans-serif; margin: 0 auto; padding: 0 20px">
+    <table style="margin: 0 20px !important;">
+    <td class="rec-head" style="color: rgba(0,0,0,0.85); font: 14px/20px 'Arial', sans-serif; max-width: 696px; ">
+        <p style="font-weight: normal; margin: 0 0 20px;">Dear {{ purchaser.name }},</p>
+        <p style="font-weight: normal; margin: 0 0 20px;">You have been enrolled {% if content_title %} in {{ content_title }}{% endif %}.
```
pdpinch, line 8: Can we include the course number here as well?

e.g. "14.740x Foundations of Development Policy"

mail/templates/product_order_receipt/body.html
```diff
@@ -0,0 +1,88 @@
...
+            <h3 style="color: #000000; font-size: 16px; font-weight: 700; line-height: 18px; margin: 0 0 20px;">Order Information</h3>
+            <p>
+                {% if order.reference_number  %}<strong style="font-weight: 700;">Order Number:</strong> {{ order.reference_number }}<br> {% endif %}
+                <strong style="font-weight: 700;">Order Date:</strong><span> {{ order.created_on|date:"F j, Y"}} </span><br>
+                <strong style="font-weight: 700;">Order Total:</strong> ${{ order_total }}<br>
+            </p>
+            {% for line in lines %}
+            <p>
+                <strong style="font-weight: 700;">Order Item:</strong> {{ line.content_title }}<br>
+                <strong style="font-weight: 700;">Product Number:</strong> {{ line.readable_id }}<br>
```
pdpinch, line 26: `readable_id` sounds like the right field to me, but in your screenshot this renders as `2T2022` when I think it should render as something like `course-v1:MITxT+14.740x+2T2022`

mail/templates/product_order_receipt/body.html
```diff
@@ -0,0 +1,88 @@
...
+            <h2 style="color: #000000; font-size: 22px; font-weight: 700; line-height: 24px; margin: 0 0 22px;">Receipt</h2>
+            <h3 style="color: #000000; font-size: 16px; font-weight: 700; line-height: 18px; margin: 0 0 20px;">Order Information</h3>
+            <p>
+                {% if order.reference_number  %}<strong style="font-weight: 700;">Order Number:</strong> {{ order.reference_number }}<br> {% endif %}
+                <strong style="font-weight: 700;">Order Date:</strong><span> {{ order.created_on|date:"F j, Y"}} </span><br>
+                <strong style="font-weight: 700;">Order Total:</strong> ${{ order_total }}<br>
+            </p>
+            {% for line in lines %}
+            <p>
+                <strong style="font-weight: 700;">Order Item:</strong> {{ line.content_title }}<br>
```
pdpinch, line 25: Probably should include course number here as well. The names aren't very distinct.

mail/templates/product_order_receipt/body.html
```diff
@@ -0,0 +1,88 @@
...
+            </p>
+            {% if receipt or coupon %}
+                <h3 style="color: #000000; font-size: 16px; font-weight: 700; line-height: 18px; margin: 0 0 20px;">Payment Information</h3>
+                <p>
+                    {%  if receipt %}
+                        <strong style="font-weight: 700;">Name:</strong> {{ receipt.name }}<br>
+                        {% if receipt.payment_method == 'card' %}
+                            <strong style="font-weight: 700;">Payment Method:</strong> {{ receipt.card_type }} | {{ receipt.card_number }}<br>
+                        {% elif receipt.payment_method == 'paypal' %}
+                            <strong style="font-weight: 700;">Email:</strong> {{ receipt.email }} (Paypal account email)<br>
```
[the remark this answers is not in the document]
jkachel: Double-checked this against the CS test account and it does capture the email if they're regular credit card users. Prob makes sense to have the template differ here - I think that PayPal transactions pull the email from their PayPal account, rather than entering it like they do for credit cards.

On a related note: this template is 90% verbatim from xPRO so might be worth considering if this change needs to happen there too?

ecommerce/api.py
```diff
@@ -17,9 +18,14 @@
...
 from mitol.common.utils.datetime import now_in_utc
 from ipware import get_client_ip
+import pycountry
+import logging
 
+from mail import api
 from ecommerce.models import Basket, PendingOrder, UserDiscount, BasketDiscount
 
+log = logging.getLogger()
+
```
rhysyngsun, line 28: These changes don't appear to be used?

ecommerce/factories.py
```diff
@@ -2,6 +2,7 @@
 from factory.django import DjangoModelFactory
 import faker
 import random
+import reversion
```
rhysyngsun, line 5: Doesn't appear to be used.

ecommerce/mail_api.py
```diff
@@ -0,0 +1,80 @@
...
+import pycountry
+import logging
+
+from mail import api
+from mail.constants import EMAIL_PRODUCT_ORDER_RECEIPT
+
+log = logging.getLogger()
+
+
+def send_ecommerce_order_receipt(order, cyber_source_provided_email=None):
```
rhysyngsun, line 11: Could you update this to used `mitol-django-mail` instead? We should do away with the `mail` app so there's not the temptation to use it in new code.

ecommerce/models.py
```diff
@@ -373,6 +373,9 @@ def fulfill(self, payment_data):
             keep_failed_enrollments=True,
         )
 
+        # send the receipt emails
+        send_ecommerce_order_receipt(self)
```
rhysyngsun, line 377: It would be a good measure to put this in a celery task so we have a guarantee this email will eventually send even if mailchimp fails.

ecommerce/serializers.py
```diff
@@ -354,3 +355,188 @@ class Meta:
...
+    Serializer for extracting receipt info from an Order object
+    This hews pretty closely to the data format in xPro but modified a bit
+    for MITxOnline's data model.
+    """
+
+    lines = serializers.SerializerMethodField()
+    purchaser = serializers.SerializerMethodField()
+    coupon = serializers.SerializerMethodField()
+    order = serializers.SerializerMethodField()
+    receipt = serializers.SerializerMethodField()
```
rhysyngsun, line 371: I think these are all (except `order`, but let's do it anyway for consistency) complex enough to warrant refactoring into their own serializers, which would also allow you to write smaller tests for each of those and then in the test for `OrderReceiptSerializer` just use something like `CouponSerializer(order.coupon).data` in your assertions since you'd have already tested that serializer independently.

An example of this is here in xPro: https://github.com/mitodl/mitxpro/blob/master/courses/serializers_test.py#L89-L94

rhysyngsun (changes requested):

ecommerce/serializers.py
jkachel: https://github.com/mitodl/mitxonline/pull/558/commits/fc631001025f20db1e8e711ddfa13b7a810e8c6a refactors these, except for the coupon one where I just re-used the existing DiscountRedemptionSerializer and changed the template to work with it instead.

ecommerce/models.py
```diff
@@ -373,6 +373,9 @@ def fulfill(self, payment_data):
             keep_failed_enrollments=True,
         )
 
+        # send the receipt emails
+        queue_ecommerce_order_receipt(self)
```
rhysyngsun, line 377: This is calling the task synchronously - the celery `@task` adds a `.delay()` method  on the task function to make an asynchronous call.

ecommerce/mail_api.py
```diff
@@ -0,0 +1,64 @@
...
+import pycountry
+import logging
+
+from ecommerce.messages import OrderReceiptMessage
+from mitol.mail.api import get_message_sender
+
+log = logging.getLogger()
+
+
+def send_ecommerce_order_receipt(order):
```
rhysyngsun, line 11: Could you add a test for this? I don't think there's an example of comparable complexity on how to do this, but the gist of it would be mocking out `send_message` [like this](https://github.com/mitodl/ol-django/blob/main/tests/mitol/mail/test_api.py#L230), which should be passed an instance of your message class, created via its [`create()` method](https://github.com/mitodl/ol-django/blob/main/src/mitol/mail/messages.py#L43-L69). You'd just want to assert on the title and verify the contents of the body contain the textual content you're expecting.

jkachel:
https://github.com/mitodl/mitxonline/pull/558/commits/1042ab32e02ed15a9567f281f4ccaef52cd99c74 should address remaining comments.

rhysyngsun (approved):
Functionality is looking good, a few more comments but feel free to merge once you've addressed those.

jkachel (commented):
LGTM 👍

Issue #531, opened 2022-04-22
Receipt: additional information and formatting 


pdpinch:
As a user, I'd like to have enough information on the receipt so that I can share it with an employer for reimbursement. 

#### Designs and Mockups

as close as possible to this (from xPRO):

![image](https://user-images.githubusercontent.com/430126/164749837-0950ab39-1fca-4844-9387-32e550ead4c9.png)

#### Acceptance Criteria:

- [ ] include purchase date
- [ ] include learner name and purchaser name (these can be different)
- [ ] include address
        - I don't think we capture learner's address, but we should get the purchaser's address from Cybersource 
- [ ] include MITx Online logo
- [ ] include MITx Online address

pdpinch:
This is on hold until @rhysyngsun can provide some guidance on how we can make this a library in ol-django

rhysyngsun:
@pdpinch this is done in the react part of the application so I don't think it's within the scope of `ol-django`

asadiqbal08:
@pdpinch I know, you added the screenshot from MitxPro as reference here. I looked around the current implementation of receipt page in MitxOnline and found these two screens that show receipt details. 


<img width="1676" alt="Screenshot 2022-05-16 at 5 08 13 PM" src="https://user-images.githubusercontent.com/7334669/168589585-e381988c-c159-495e-b00d-6e5714e5b638.png">
<img width="1665" alt="Screenshot 2022-05-16 at 5 08 05 PM" src="https://user-images.githubusercontent.com/7334669/168589611-9b7118cd-3a7c-437c-a97a-afa648e9b62d.png">

1- Let me know, where do you want to show some more details from the above screens. 
2- We have `purchaser` details e.g. name, email etc referring this [field](https://github.com/mitodl/mitxonline/blob/main/ecommerce/models.py#L246).
3- But, we do not have any receipt model in MitxOnline as we are dealing in MitxPro in order to get `card_type`, `card_number` or `purchaser name` etc from [data field](https://github.com/mitodl/mitxpro/blob/master/ecommerce/models.py#L650-L657) there.

Can you put more thoughts over it ?

annagav:
Hi @asadiqbal08 I am sorry I forgot to assign this issue to me, because there was a change of intention with this issue https://github.com/mitodl/ol-npm-libraries/issues/18, and I was out few days last week. Did you already start working on this?

annagav:
[...]

PR #578, opened 2022-05-19
asadiqbal08/Updated the Receipt Page with additional Details

asadiqbal08:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Testing
  - [ ] Code is tested

fixes: #531   
#### What's this PR do?
As a user, I'd like to have enough information on the receipt so that I can share it with an employer for reimbursement.

- Visit the receipt page after an enrolment or by visiting `/orders/receipt/{RECEIPT_ID}/`
- Verify the required data is present as requested over the issue.

screencapture-localhost-8013-orders-receipt-2-2022-05-20-13_52_11

screencapture-localhost-8013-orders-receipt-2-2022-05-20-15_41_51


pdpinch:
It seems like we're working on this in two places, and we should be sure the information (and maybe the styling?) is the same in both places. 

See #558

pdpinch:
A couple comments on the screenshot: 

- can we display the discount code?
- If the address is actually coming from the Cybersource receipt, and not the user model, it should be displayed under "payment information" instead of "customer information" 
- The font seems off to me, but I know we don't have a final design for this page. Please just make sure we're using the same basic body styles as the rest of the site.

asadiqbal08:
I guess, Discount is there in [code](https://github.com/mitodl/mitxonline/pull/578/files#diff-17a03998399852a17aa3600e93122905838e5d4536942be8bf5c75349589c78eR163-R165) but hidden behind a condition

asadiqbal08:
I updated the screenshots, You can see the: 
- `Discount code` and `Discount text` there. 
- I removed the conditional way for Discount. 
- I manually added values in model so one can ignore the actual factual values.

frontend/public/scss/receipt.scss
```diff
@@ -0,0 +1,257 @@
+// sass-lint:disable mixins-before-declarations
+$label-color: #8a8b8c;
+
+.receipt-wrapper {
+
+  padding: 20px 15px;
+  margin: 0 auto;
+  max-width: 1018px;
+  font: 500 18px/23px 'Rajdhani', Times New Roman, serif;
```
jkachel, line 9: Can this `font` rule (and where it appears elsewhere in this file - there's 11 total occurrences of Rajdhani in the file) be removed? This should use the standard fonts from the base layout.scss.

frontend/public/scss/receipt.scss
```diff
@@ -0,0 +1,257 @@
...
+  }
+}
+
+@media print {
+  .print-btn,
+  .print-row,
+  .footer,
+  .header-holder {
+    display: none !important;
+  }
```
jkachel, line 251: Can you add "header" (the tag) to this list? It'd be good to not have the site header on the printed version of the receipt.

jkachel (changes requested):
Few small layout changes - otherwise the functionality looks good to me.

frontend/public/scss/receipt.scss
asadiqbal08: Removed. Done. 

Screenshot 2022-05-26 at 4 33 19 PM

frontend/public/src/components/ReceiptPageDetailCard.js
```diff
@@ -0,0 +1,225 @@
...
+                  line.product.purchasable_object.start_date
+                )
+                const endDate = parseDateString(
+                  line.product.purchasable_object.end_date
+                )
+                return (
+                  <tr key={line.product.purchasable_object.readable_id}>
+                    <td>
+                      <div>
+                        {line.product.purchasable_object.title} <br />
```
jkachel, line 198: Can you update this to get the course's readable_id rather than the one for the course run? Also, can you add the `course_number` to the beginning of the title (if it's a course that's been purchased)? That will make it match the email receipt code. (Take a look at the `TransactionLineSerializer` to see how we're doing it there.)

jkachel (changes requested):
One more change after reviewing against the email receipt.

frontend/public/src/components/ReceiptPageDetailCard.js
asadiqbal08: @jkachel If I am not wrong, It should not be CourseRun instead of a course here ?

asadiqbal08: @jkachel If I am not wrong, It should not be CourseRun instead of a course here ?

asadiqbal08: You mean that `line.product.purchasable_object.course.readable_id` but it can be a program.

asadiqbal08: @jkachel Do you mean the change set as in [commit here](https://github.com/mitodl/mitxonline/pull/578/commits/4e02dd6f4633213382ae113a4edc1bfde1611fbe) ?

asadiqbal08: Screenshot 2022-05-27 at 3 58 44 PM


Showing like this one.

jkachel: Yup - tested with an order locally and looks good to me.

jkachel (approved):
Flow types need fixing but this is good to go once that's done. LGTM 👍

PR #580, opened 2022-05-23
Flexible Pricing: Automatically approve if the Learner is elligible when they request it

### LINKED ISSUE #542: Flexible Pricing: Automatically approve if the Learner is elligible when they request it
We want to be able to automatically approve learners for flexible pricing based on their income and location. This should occur immediately when the learner submits their income information.

There is a preexisting implementation of this in MicroMasters that we should pull in: https://github.com/mitodl/micromasters/blob/master/financialaid/api.py#L45

This diagram roughly illustrates the way we want our models setup:

```mermaid
flowchart LR
    Course1 --> Program
    Course2 --> Program
    Course3 --> Program

    Product1 --> Course1
    Product2 --> Course2
    Product3 --> Course3

    FlexiblePriceTier1[Tier: Full Price]
    FlexiblePriceTier2[Tier: 50% Off] --> Discount2[Discount: 50% Off] --> Product1 & Product2 & Product3
    FlexiblePriceTier3[Tier: 75% Off] -->  Discount3[Discount: 75% Off] --> Product1 & Product2 & Product3

    FlexiblePriceTier1 & FlexiblePriceTier2 & FlexiblePriceTier3 --> Program

    Learner1[Sally Ride] --> FlexiblePrice1["Flexible Price Request (High Income)"] --> FlexiblePriceTier1 & Program
    Learner2[Ada Lovelace] --> FlexiblePrice2["Flexible Price Request (Medium Income)"] -->FlexiblePriceTier2 & Program
    Learner3[Marie Curie] --> FlexiblePrice3["Flexible Price Request (Low Income)"] -->FlexiblePriceTier3 & Program

    linkStyle 6,7,8,9 stroke:#3ff
    linkStyle 10,11,12,13 stroke:#f3f
    
    linkStyle 14,15,16,17,18,19,20,21,22,23,24,25 stroke:#999

```

- [x] https://github.com/mitodl/mitxonline/issues/596
- [x] https://github.com/mitodl/mitxonline/issues/597
- [x] https://github.com/mitodl/mitxonline/issues/598

umarmughal824:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested

fixes #542, #596, #597, #598

#### What's this PR do?
Flexible Pricing: Automatically approve if the Learner is elligible when they request it

- Add Flexible Price Tiers against the course (you could also follow the attached screenshots)
- Apply for flexible pricing request

**USD Currency**

Screenshot 2022-06-03 at 4 21 04 PM

**Other Country**

Screenshot 2022-06-03 at 4 30 16 PM


flexiblepricing/models.py
```diff
@@ -46,6 +51,28 @@ def __str__(self):
         )
 
 
+class FlexiblePriceTier(TimestampedModel):
```
jkachel, line 54: Can you add a `__str__` method here so that it's easier to distinguish between individual objects?

cms/models.py
```diff
@@ -676,36 +683,41 @@ def serve(self, request, *args, **kwargs):
...
+            converted_income = determine_income_usd(
+                float(form.cleaned_data["your_income"]),
+                form.cleaned_data["income_currency"],
+            )
+        except NotSupportedException:
+            raise ValidationError("Currency not supported")
+
+        courseware = self.get_parent_product()
+        income_usd = round(converted_income, 2)
+        tier = determine_tier_courseware(courseware, income_usd)
```
jkachel, line 697: This ends up delivering an error message to the learner if there's no tier, which is undesirable. We'd still want to capture the request so that it can be processed manually. Can you wrap this in a try/except block and leave it in `PENDING_MANUAL_APPROVAL` if there's no tier? It looks like the tier FK in FlexiblePrice is already nullable.

flexiblepricing/models.py
```diff
@@ -46,6 +51,28 @@ def __str__(self):
...
+
+    discount = models.ForeignKey(
+        Discount,
+        null=False,
+        related_name="flexible_price_tiers",
+        on_delete=models.PROTECT,
+    )
+    courseware_content_type = models.ForeignKey(
+        ContentType, on_delete=models.CASCADE
+    )  # add a filter to limit this to program/course contenttypes
```
jkachel, line 67: A filter here would be very handy (since now it gives you a list of all models). We do this in the ecommerce Product model - can you cross port that over?

jkachel (changes requested):
A handful of changes. I did have a question about auto-approval. In testing, I set up three tiers for the same course - one zero-income one, one at $45,000, and one at $75,000 with different discounts for each - then submitted flex pricing requests that would slot in to each tier. The first tier worked fine: it auto-approved the request. The third tier also worked fine: it set it to manual approval, which is also fine. The middle tier, though, was also set to manual approval. Do we want these to be manually approved if they hit an in-between tier or should they be autoapproved with the applicable discount? (This may be a better question for @rhysyngsun or @pdpinch .)

cms/models.py
umarmughal824: In my opinion, there is no need to apply for the flexible pricing if we don't have any tier. @pdpinch / @rhysyngsun what's your opinion on that?

flexiblepricing/models.py
umarmughal824: Sure, thanks

cms/models.py
pdpinch: This is fine as is.

jkachel (approved):
LGTM 👍

PR #585, opened 2022-05-24
Adds Refine Admin documentation

jkachel:
Updates the documentation to add configuration information for Refine.

### DIFF
--- a/docs/source/configuration/refine_admin.rst
+++ b/docs/source/configuration/refine_admin.rst
@@ -1,0 +1,52 @@
+Refine Admin
+============
+
+Some administrative functionality is built out using `Refine <https://refine.dev>`_. The Refine admin is a separate application that uses OAuth2/OIDC for authentication and communicates with MITxOnline via the standard REST API. It compliments the Django Admin interface, providing an interface for operations that would otherwise be hard to implement in Django Admin, and for users that don't necessarily need the level of access that Django Admin provides.
+
+Accessing
+---------
+
+The Docker compose environment builds a ``refine`` container that runs a dev server available at port 8016. 
+
+Set Up
+------
+
+You will need to make some adjustments to the MITxOnline configuration to allow the Refine admin to work.  
+
+  Currently, the application expects to be accessible at ``mitxonline.odl.local``. 
+
+**Step 1: CORS/CSRF Settings**
+
+Your ``.env`` file needs these two variables to be set::
+
+  CORS_ALLOWED_ORIGINS=http://mitxonline.odl.local:8016
+  CSRF_TRUSTED_ORIGINS=http://mitxonline.odl.local:8016
+
+**Step 2: Generating the Key** 
+
+The following command will generate the key:
+
+``openssl genrsa 4096 2>/dev/null | sed -e 'H;${x;s/\n/\\n/g;s/^\\n//;p;};d'``
+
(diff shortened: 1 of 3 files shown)


rhysyngsun (approved):

docs/source/configuration/refine_admin.rst
```diff
@@ -0,0 +1,35 @@
...
+---------
+
+The Docker compose environment builds a ``refine`` container that runs a dev server available at port 8016. 
+
+Set Up
+------
+
+The Refine app uses OAuth2/OIDC for authentication with the core MITxOnline app. You will need to generate a shared private key for this authentication flow to work. 
+
+*Generating the Key:* The following command will generate the key:
```
asadiqbal08, line 16: A note about the location of file where it will be placed after running the command. Know it will be current path but good to add for more understanding.

docs/source/configuration/refine_admin.rst
```diff
@@ -0,0 +1,35 @@
...
+
+``openssl genrsa -out oidc.key 4096``
+
+The key then needs to be put into the ``.env`` file into a setting called ``OIDC_RSA_PRIVATE_KEY``. Remove linebreaks and replace with the string ``\n``. Both the Refine and Django apps will pick the key up from this setting.
```
asadiqbal08, line 20: Note to add: You need to restart the server after add the key in .env file

asadiqbal08: I noted, you added the note below but It should be here regarding restarting container as when I was adding the below confituration in `Application` model then showing error to have `OIDC_RSA_PRIVATE_KEY`.

docs/source/configuration/refine_admin.rst
```diff
@@ -0,0 +1,35 @@
...
+
+(The key generation process is from the `Django OAuth Toolkit <https://django-oauth-toolkit.readthedocs.io/en/latest/oidc.html#creating-rsa-private-key>`_ docs.)
+
+*Configuring MITxOnline:* You will need to add a new Application in the Django Oauth Toolkit section in Django Admin (``/admin/oauth2_provider/application/``). Navigate there and create a new Application. The settings you will need are:
+
+* **Client Id**: ``refine-local-client-id``
```
asadiqbal08, line 26: Do we need to define client_id and client_secret somewhere in .env file ?

asadiqbal08: Screenshot 2022-05-25 at 2 51 42 PM


getting this while clicking on SignIn button.

asadiqbal08 (changes requested):
Readme needs to update to fill error gaps to setup as discussed 1-1

docs/source/configuration/refine_admin.rst
jkachel: No, Client Secret is ignored (OIDC uses the key) and Client Id doesn't have an exposed configuration setting in the .env file right now so it's just that value.

docs/source/configuration/refine_admin.rst
jkachel: I updated this a bit - there's a one-liner in there now that will output the key in the right format directly to the terminal, so you won't have to deal with key files at all (which removes the possibility that the key gets committed to the repo accidentally too).

jkachel:
Tested the last round of documentation changes from https://github.com/mitodl/mitxonline/pull/585/commits/3c4c0dfd5488dff49d7088f9497084f8d5117091 in a fresh copy of the app so should have gotten everything this time - please let me know if I've missed anything.

asadiqbal08:
@jkachel I am still getting the error while logged in, I have these configs in `.env` file. 

- `.env`
Screenshot 2022-05-26 at 3 38 37 PM

- Oauth Application

Screenshot 2022-05-26 at 3 38 55 PM


Please correct me, If I am missing something else that required to login there ?

asadiqbal08 (approved):
LGTM 👍 now

PR #595, opened 2022-05-27
Adds a check for exchange rate description when constructing the currency list

### LINKED ISSUE #594: Object of type 'NoneType' has no len() error if exchange rate doesn't have a description
### Steps to Reproduce

1. Add a Currency Exchange Rate through the Django Admin that does not include a description.
2. Attempt to view a Flexible Pricing Request form. 
3. You should receive an error message.

### Expected Behavior

The form should render, and the currency field should show the available currencies regardless of whether they have a description.

### Actual Behavior

Object of type 'NoneType' has no len()

### Related Issues

https://github.com/mitodl/mitxonline/issues/542#issuecomment-1139528001

### Screenshot or Screencast

image
(this is from #542)

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #594 

#### What's this PR do?

Adds a check for None in the currency field options generation code. (It checked for length > 0 but not `None` specifically.) 

Steps to reproduce are in #594 - you should see the form render successfully, and exchange rates without descriptions should render as just the currency code.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -80,7 +80,7 @@ class FlexiblePricingFormBuilder(FormBuilder):
 
         for record in CurrencyExchangeRate.objects.all():
             desc = record.currency_code
-            if len(record.description) > 0:
+            if record.description is not None and len(record.description) > 0:
                 desc = "{code} - {code_description}".format(
                     code=record.currency_code, code_description=record.description
                 )


umarmughal824 (approved):
👍

Issue #571, opened 2022-05-17
Refunds: When a refund is issued call CyberSource and issue a refund of the payment


rhysyngsun:
This will build on the work from #572, which ports the Google Sheets app.

The API for this appears to be [here](https://developer.cybersource.com/api-reference-assets/index.html#payments_refund_refund-a-payment).

### Acceptance Criteria

- [ ] When a refund is issued via Google Sheets, within a transaction:
- [ ] Mark the order as refunded 
- [ ] Make an API call to CyberSource to refund the payment
  - [ ] Handle failure states of the API request:
    - [ ]  If CyberSource indicates that this is a duplicate request, this means we already issued the refund and that we can treat this as a success
  - [ ] If the call to CyberSource fails:
    - [ ] No changes in state should be made (i.e. the learner should not be unenrolled and the spreadsheet should not be updated).
    - [ ] An error should be logged

rhysyngsun:
@arslanashraf7 If you're looking for the next bit of work to do you could probably implement the CyberSource API call wrappers in a smaller PR before #572 is done.

arslanashraf7:
@rhysyngsun I was looking through the existing usage of CyberSource in MItxOnline and xPRO and there are a few things I wanted to confirm on this:
- Is it fine to use RC's CyberSource keys for testing this? I also created a test account of my own but I'm confused about the key generation steps (Specifically the error I see says that there is no Security profile which I couldn't find a way to create).
- MITxOnline has been migrated to use the [PaymentGateway from ol-django](https://github.com/mitodl/ol-django/tree/main/src/mitol/payment_gateway) which is not the case for xPRO
    - As of now, the CyberSource gateways have the make payment functionality, Do you think a refund is something that should also go into the `ol-django`'s payment_gateway?
- We also have refunds in xPRO are we porting them over? I suppose currently they don't call CyberSouce refund API?

rhysyngsun:
> @rhysyngsun I was looking through the existing usage of CyberSource in MItxOnline and xPRO and there are a few things I wanted to confirm on this:
> 
>     * Is it fine to use RC's CyberSource keys for testing this? I also created a test account of my own but I'm confused about the key generation steps (Specifically the error I see says that there is no Security profile which I couldn't find a way to create).
> 
>     * MITxOnline has been migrated to use the [PaymentGateway from ol-django](https://github.com/mitodl/ol-django/tree/main/src/mitol/payment_gateway) which is not the case for xPRO

Yes that's fine.

[...]

PR #599, opened 2022-05-31
feat: refund orders CyberSource - Integrate [mitol-django-payment-gateway]

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
- [x] Settings
  - [x] Opened issue for DevOps regarding necessary configuration changes to deployed environments

#571 

#### What's this PR do?
- Adds the initial integration of order refunds with the `mitol-django-payment-gateway` from https://github.com/mitodl/ol-django/pull/64.
- Removes all the Cybersource related settings from MITxOnline Marketing site (They have been migrated to the common app ttps://github.com/mitodl/ol-django/pull/64)
- Implements the refund flow using API and adds integration for refunds done through Django Admin
- Also, updates other ol-django packages to for compatibility reasons

- Setup the all the environment-related config for CyberSource, (You can see them all [here](https://github.com/mitodl/ol-django/pull/64/files#diff-8925307c256c81211d35d1a950207fb606d513e45d72cc6ab6ccaa4c44f5d431))
- Setup a product with a price and place an order for it (Can be done while enrolling in a course or by `Get Certificate` links on the course details and dashboard pages) - You should complete all Cybersource steps successfully for placing an order
- You can test it through Django Shell or Admin. The full flow should be testable as:
    - Place an order
    - Open the Fulfilled order from above step in Django Admin and click the `Refund Order` button
    - The Order should be refunded, and the user should be unenrolled of the checkbox was selected while refunding in the above step
    - Notice that a refund Transaction with Cyber Source API response details gets created and the order status is changed to `Refunded`

Setup, product, place order, complete CyberSource order steps successfully before a refund

#### NOTE:
- If you don't have the configuration keys for CyberSource you can ask me or a fellow developer

Below is Done:
- ~This PR is associated with its counterpart work done in ol-django in https://github.com/mitodl/ol-django/pull/64.~
- ~This PR should ideally be reviewed after https://github.com/mitodl/ol-django/pull/64~


arslanashraf7:
@rhysyngsun This would be the integration of https://github.com/mitodl/ol-django/pull/64. Most of the things w.r.t the integration are done. Let me know if would like to take an initial pass at this.

A few things that will be done once the package from ol-django is out:
- Failing tests would pass
- I'll remove the cybersource-related settings.

The end to end would actually depend upon sheets work. but I think we can merge this first if we'd like and then use it in sheets.

main/settings.py
```diff
@@ -981,6 +981,24 @@
     description="CyberSource access key",
 )
 
+MITOL_PAYMENT_GATEWAY_CYBERSOURCE_MERCHANT_ID = get_string(
```
rhysyngsun, line 984: i think we should actually upstream these settings into `mitol-django-payment-gateway` as a boilerplate settings file. You can take a look at https://github.com/mitodl/ol-django/blob/main/src/mitol/google_sheets/settings/google_sheets.py as an example. Can you move the other `MITOL_PAYMENT_GATEWAY_` ones in addition to the ones you're adding?

ecommerce/cybersource_api.py
```diff
@@ -0,0 +1,91 @@
...
+        # We treat duplicate refund requests as a success
+        update_order_state(
+            order,
+            refund_exception.amount,
+            refund_exception.reason_code,
+            refund_exception.body,
+        )
+    except Exception as ex:
+        log.exception(str(ex))
+        raise ex
```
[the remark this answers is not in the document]
arslanashraf7: Initially, the idea was that this function independently logs the exception details and then passes the exception to the caller to handle that in anyway it wants. But since we'd now make it a bool-based return we can assume handling the exception ourselves and in a way just log it and pass False for the caller so it knows the operation wasn't successful. I'll update this while addressing feedback.

ecommerce/cybersource_api.py
```diff
@@ -0,0 +1,91 @@
...
+                "",  # There is no reason/code in case of successful refund
+                response.response_data,
+            )
+
+        else:
+            print(f"THERE WAS AN ERROR WITH THE REQUEST {response.message}")
+    except RefundDuplicateException as refund_exception:
+        # Refund got an error during the API call (Check if the error was a DUPLICATE_REQUEST)
+        # We treat duplicate refund requests as a success
+        update_order_state(
```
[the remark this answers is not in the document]
arslanashraf7: Since we were considering it a success, So what I had in mind regarding this was to at least create a transaction object and in this case, the API data for the transaction would not be like the success response. But your suggestion about the negative balance and the frontend inconsistency makes sense. Good call!

ecommerce/cybersource_api.py
```diff
@@ -0,0 +1,91 @@
...
+log = logging.getLogger(__name__)
+
+REFUND_SUCCESS_STATES = [
+    ProcessorResponse.STATE_ACCEPTED,
+    ProcessorResponse.STATE_DUPLICATE,
+    ProcessorResponse.STATE_PENDING,
+]
+
+
+def update_order_state(order_to_update, refunded_amount, reason=None, data=None):
```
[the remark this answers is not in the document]
arslanashraf7: I removed this function altogether, it was created as a common thing to call from refund success and duplicate requests. Now that we won't be adding any update to the order/transaction itself in a duplicate request as per your above comment. So, this happens only in success response now and we can call the associated functionality right from there(a one timer).

main/settings.py
arslanashraf7: As per [my comment here](https://github.com/mitodl/mitxonline/pull/599#issuecomment-1155166560), These settings have already been moved to ol-django in [this PR](https://github.com/mitodl/ol-django/pull/64/files#diff-8925307c256c81211d35d1a950207fb606d513e45d72cc6ab6ccaa4c44f5d431R1). Since we haven't yet released that I'm kept these here for testing because the ol-django packages can't directly be installed right from git because of pants build and without these there are settings errors.

These should be removed once that ol-django PR is merged and released.

The priority from my side would be to get the ol-django PR reviewed/merged/deployed first.

.gitignore
```diff
@@ -122,3 +122,4 @@ celerybeat-schedule
 
 # Local Docker customizations
 docker-compose.*.yml
+/sources
```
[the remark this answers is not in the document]
arslanashraf7: This was a dir where I was keeping the package from ol-django to install locally to MITxOnline. This should be removed once the PR is ready for review.

arslanashraf7:
@rhysyngsun This is ready for review, IMO, it should ideally be reviewed after https://github.com/mitodl/ol-django/pull/64.

ecommerce/cybersource_api.py
```diff
@@ -0,0 +1,99 @@
...
+                ECOMMERCE_DEFAULT_PAYMENT_GATEWAY,
+                refund_gateway_request,
+            )
+            if response.state in REFUND_SUCCESS_STATES:
+                # There is no default reason for response of a Refund, so we append the response json with the reason
+                # provided by the caller
+                if refund_reason:
+                    response.response_data["reason"] = refund_reason
+
+                unenrolled = order.refund(
```
arslanashraf7, line 81: We can have two types of implementations here:

1- We keep it less coupled and perform only API calls and return the results and let the caller update the other things e.g. making appropriate changes in database for marking the order as `Refunded`. But in this case the callers from different places will have to perform the database level refunds at multiple places/.

2- We keep it the way it is right now, More coupled but it will make almost all the necessary changes upon the API call e.g. Marking order refunded, unenrolling based on the passed flag.

@rhysyngsun Let me know what you think.

ecommerce/cybersource_api.py
```diff
@@ -0,0 +1,99 @@
...
+        # Extract transaction_id from order CyberSource json
+        order_recent_transaction = Transaction.objects.filter(order=order_id).first()
+
+        if not order_recent_transaction:
+            log.error(
+                f"There is no associated transaction against order_id {order_id}."
+            )
+            return False, False
+
+        transaction_dict = json.loads(json.dumps(order_recent_transaction.data))
```
rhysyngsun, line 56: Could you explain why you're cycling this through json?

ecommerce/cybersource_api.py
```diff
@@ -0,0 +1,99 @@
...
+from mitol.payment_gateway.api import (
+    Refund as GatewayRefund,
+    PaymentGateway,
+)
+from mitol.payment_gateway.exceptions import RefundDuplicateException
+
+log = logging.getLogger(__name__)
+
+
+def refund_order(*, order_id: int, **kwargs):
```
rhysyngsun, line 20: I don't think this should be part of `cybersource_api.py` since it should ideally be generic.

ecommerce/cybersource_api.py
```diff
@@ -0,0 +1,99 @@
...
+        transaction_id = transaction_dict["transaction_id"]
+
+        order_amount = refund_amount or transaction_dict.get("req_amount")
+        order_currency = transaction_dict.get("req_currency")
+
+        refund_gateway_request = GatewayRefund(
+            transaction_id=transaction_id,
+            refund_amount=order_amount,
+            refund_currency=order_currency,
+        )
```
rhysyngsun, line 68: So I'm thinking this bit is actually leaking the specifics of Cybersource into this function which should be generic. What do you think of adding a new method to `PaymentGateway` like this:

```python
refund_gateway_request = PaymentGateway.create_refund_request(
                ECOMMERCE_DEFAULT_PAYMENT_GATEWAY,
               transaction_dict
)
```
And we'd have the specifics of this logic in the library instead?

ecommerce/cybersource_api.py
```diff
@@ -0,0 +1,99 @@
...
+        try:
+            response = PaymentGateway.start_refund(
+                ECOMMERCE_DEFAULT_PAYMENT_GATEWAY,
+                refund_gateway_request,
+            )
+            if response.state in REFUND_SUCCESS_STATES:
+                # There is no default reason for response of a Refund, so we append the response json with the reason
+                # provided by the caller
+                if refund_reason:
+                    response.response_data["reason"] = refund_reason
```
rhysyngsun, line 79: Instead of putting this in the response data (we shouldn't be modifying it), can you add a new column to `Transaction` for `reason`? It can default to blank because we won't have a reason for purchases. `Order.refund()` will need to be updated to remove this kind of logic because it's doing a similar thing if `api_response_data` is `None`.

rhysyngsun (changes requested):

ecommerce/cybersource_api.py
rhysyngsun: Hmm, yeah, this is a bit tricky here. I think we need to decouple this because if things fail a certain way we could end up having made the api call to refund the learner via cybersource but then the `Order` save fails and rolls back so we end up in a bad state. And `Order.refund()` is doing way too much - it shouldn't be doing the unenrollments. I'd move that to a separate function and then reorder the logic in `refund_order` to this:

- Start a DB transaction, within that transaction:
  - Get the order and call `order.refund()`. This will ensure that the DB has the correct state before we make the call to Cybersource. This way if the DB write fails, we don't attempt the API call and if the API call fails the DB update rolls back
  - Make the call to the Cybersource refund API
- Exit the transaction so it commits
- Trigger the unenrollments, this might be best as a celery task because that could also fail temporarily, but at this point the most important work of issuing the refund has been completed so it's not absolutely necessary that this happen immediately.

I hope that answers your question?

ecommerce/cybersource_api.py
arslanashraf7: It shouldn't be, I initially had some bad data in the transaction somehow, maybe because I created some orders through the factory and that wasn't loading correctly as JSON as is, so I had to cycle it but I'm sure it can be removed and we can use `transaction.data` as is. Will be done as part of the feedback.

ecommerce/cybersource_api.py
arslanashraf7: This does, It's a great suggestion. I had a thought on how the end-user behavior would work in the case of Refund through Admin. Running the unenrollments in a Celery task will have to change that behavior a bit. Currently, when a user clicks the refund what happens is Order refunds in DB, CyberSource API called, and lastly unenrollment, and when all that is done the User sees that the order got refunded.

After using the celery task, should we care for the unenrollment status or just return the user with success before the while the unenrollment celery task runs?

arslanashraf7: @rhysyngsun an update on this part:

>Get the order and call order.refund(). This will ensure that the DB has the correct state before we make the call to Cybersource. This way if the DB write fails, we don't attempt the API call and if the API call fails the DB update rolls back

Ideally, when refunding the order we also set the acuired response json from CyberSoure refund API in transaction data which isn't available before the API call, So, i tweaked it in a way that it will refund the order, create the transaction and when we have api response it will update that transaction with response data.

>Trigger the unenrollments, this might be best as a celery task because that could also fail temporarily, but at this point the most important work of issuing the refund has been completed so it's not absolutely necessary that this happen immediately.

Updates, so in this case when an order is refunded from Django Admin, it refunds and starts the celery task. I updated the success message on admin to something like `Order <id> refunded and unenrollment is in progress.`.

Let me know if above seems feasible to you?

rhysyngsun (changes requested):

Issue #619, opened 2022-06-13
Receipt page is empty when there is no discount code


pdpinch:
### Steps to Reproduce

1. Purchase a course on RC, using the instructions at https://docs.google.com/document/d/1JUMIABviilgnctn9mJcrFdlw8xzd0HUyn9kXNg4Xvts/edit#
2. You should get an email with a link to your receipt
3. View the receipt

### Expected Behavior

- receipt should display, with all the relevant info 
 
### Actual Behavior

- page is empty 

### Screenshot or Screencast

![image](https://user-images.githubusercontent.com/430126/173445744-aecbaa5f-b458-4b7b-9189-5a052443bb50.png)

jkachel:
In testing, this appears to be an issue with the code that displays the discounts codes on the page. 
* Testing with a user that has used (method doesn't matter) a discount code: receipt displays fine
* Testing with a user that has not used a discount code: blank page

Relevant test transactions on RC: orders 263 and 264

pdpinch:
I'm reopening this because it's still happening. I'm not sure if it was a regression or if the fix was incomplete. 

I just tested this on RC, and my email receipt comes with a link to just https://rc.mitxonline.mit.edu/orders/receipt/ which seems incomplete to me. I think the link should have been https://rc.mitxonline.mit.edu/orders/receipt/302/

jkachel:
Yup, closing this

PR #621, opened 2022-06-14
Fixes: Receipt page is empty when there is no discount code

asadiqbal08:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

fixes: #619 

#### What's this PR do?
Fixes the Receipt page issue when there is no discount code then the page renders empty.

Visit the receipt page with and without discount code.

### DIFF
--- a/frontend/public/src/components/ReceiptPageDetailCard.js
+++ b/frontend/public/src/components/ReceiptPageDetailCard.js
@@ -16,32 +16,32 @@ type Props = {
 export class ReceiptPageDetailCard extends React.Component<Props> {
   render() {
     const { orderReceipt, discounts } = this.props
+    let discountAmount = null
+    let discountAmountText = null
+    const stateCode = null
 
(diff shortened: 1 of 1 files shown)


frontend/public/src/components/ReceiptPageDetailCard.js
```diff
@@ -205,9 +207,11 @@ export class ReceiptPageDetailCard extends React.Component<Props> {
                     <td>
                       <div>${line.price}</div>
                     </td>
-                    <td>
-                      <div>${discountAmountText}</div>
-                    </td>
+                    {discountAmountText !== null ? (
+                      <td>
+                        <div>${discountAmountText}</div>
```
jkachel, line 212: Small thing: remove the `$` here and move it to the appropriate spots in the switch statement starting on line 28. (it displays "$35% off" for a 35% discount.)

jkachel (changes requested):
One very small change.

Issue #608, opened 2022-06-06
Flexible Pricing Staff UI: Capture a justification for Flexible Pricing request status changes


jkachel:
As a staff member, when I change the status of a Flexible Pricing request, I'd like to provide a justification of why I am changing the status of the request. The justification should be one from a set of pre-defined justifications available in the app. 

### Acceptance Criteria
- [ ] The justification field should appear in the confirmation modal that's displayed when the status buttons are activated.
- [ ] The options in the justification field should mirror the ones in MicroMasters (see [this blob](https://github.com/mitodl/micromasters/blob/7546f4e962bd65b5a341351bf47f411b3d9d9084/financialaid/constants.py#L9-L19))
- [ ] If the user confirms the change, the justification should be sent as part of the payload and should be captured in the appropriate column in the database.
- [ ] Justifications should be shown (but not editable) ~~when the request is in the Approved state~~ in the table.

This should work much like it does in MicroMasters. I think issue #607 should be done before this.

jkachel:
Question for @pdpinch / @rhysyngsun regarding placement of the justification field: In MicroMasters, the justification field is in the table. However, since we're using a modal for confirmation in the Refine admin, we can potentially have that field be there. (MM uses an `alert` for this.) Having it in the confirmation modal keeps it out of the table (and makes it less cramped looking, especially with the search pane on the left and/or the nav expanded) and would give us the potential to filter the justification list by status. Thoughts on this?

pdpinch:
@jkachel I like the idea of having it in the confirmation modal, if it's not too much work.

asadiqbal08:
@jkachel regarding this
>> Justifications should be shown (but not editable) when the request is in the Approved state.

I guess, we should show the justification in the `Admin JS <Table/>` there whatever the state regardless of showing just in case of "Approved" status ? thoughts ?

PR #622, opened 2022-06-14
Show justification once status changed

asadiqbal08:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

fixes: #608 

#### What's this PR do?
As a staff member, when I change the status of a Flexible Pricing request, I'd like to provide a justification of why I am changing the status of the request. The justification should be one from a set of pre-defined justifications available in the app.

Visit the admin portal in Refine UI to test it.

screencapture-mitxonline-odl-local-8016-flexible-pricing-2022-06-14-16_40_29

screencapture-mitxonline-odl-local-8016-flexible-pricing-2022-06-14-16_39_59
screencapture-mitxonline-odl-local-8016-flexible-pricing-2022-06-14-16_39_40

Screenshot 2022-06-14 at 4 40 19 PM

screencapture-mitxonline-odl-local-8013-admin-flexiblepricing-flexibleprice-1-change-2022-06-14-16_40_57

### DIFF
--- a/frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
+++ b/frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
@@ -155,7 +155,7 @@ export const FlexiblePricingList: React.FC = () => {
     };
 
     const handleChange = (e: any) => {
-        setJustification(e.target.options[e.target.selectedIndex].text)
+        setJustification(e.target.options[e.target.selectedIndex].value)
       }
 
 
@@ -268,7 +268,7 @@ export const FlexiblePricingList: React.FC = () => {
                                                         </span>
                                                         <select onChange={(e) => handleChange(e)} style={{ marginLeft: "20px" }}>
                                                             {All_Justifications.map((option) => (
-                                                            <option value={option.value} selected={justification == option.value}>{option.label}</option>
+                                                            <option value={option.value} selected={justification == option.value}>{option.value}</option>
                                                             ))}
                                                         </select>
                                                     </p>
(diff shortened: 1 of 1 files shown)


frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
```diff
@@ -231,6 +262,16 @@ export const FlexiblePricingList: React.FC = () => {
                                                         <strong>Country of Income:</strong>
                                                         <div>{modaldata.country_of_income}</div>
                                                     </p>
+                                                    <p>
+                                                        <span>
+                                                            <strong>Justification:</strong>
+                                                        </span>
+                                                        <select onChange={(e) => handleChange(e)} style={{ marginLeft: "20px" }}>
+                                                            {All_Justifications.map((option) => (
+                                                            <option value={option.value} selected={justification == option.value}>{option.label}</option>
```
jkachel, line 271: Should display `option.value` here.

jkachel (changes requested):
One more change other than the dropdown in the modal: either the Justification field needs to display the value (it displays the label now, so I get "INSUFFICIENT" rather than "Insufficient docs") or the code should store the value of selection and not the label. 

(In MicroMasters, it does the latter - looking at financial aid record 226 on RC, for instance, shows "Documents in order" for the justification, so doing it that way would be consistent with MicroMasters. But, I think it'd be better to use the lookup table because then we can more easily change or internationalize the text in the future.)

frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
asadiqbal08: Screenshot 2022-06-15 at 4 04 43 PM

Issue #610, opened 2022-06-07
`/cart/add/` source alternatively accept a `course_id` parameter


rhysyngsun:
We've previously added a plugin to openedx to perform a redirect to mitx online (#567 and https://github.com/mitodl/open-edx-plugins/pull/31) for the upgrade button over there. Now we need to add some logic on our end that converts that incoming course key into a product that's added to the cart.

We have an existing view that does this for product ids: https://github.com/mitodl/mitxonline/blob/main/ecommerce/views/v0/__init__.py#L496

- Update the view linked above to also handle `course_id` being passed
- Update the `ol-infrastructure` configs to set `MARKETING_SITE_CHECKOUT_URL` to this url

rhysyngsun:
This should be done, need to verify

arslanashraf7:
Tested this on RC and it's working. Closing it via https://github.com/mitodl/ol-infrastructure/pull/903, https://github.com/mitodl/mitxonline/pull/626, https://github.com/mitodl/open-edx-plugins/pull/63

FYI, @briangrossman

PR #626, opened 2022-06-15
feat: update cart to handle products from external checkout

arslanashraf7:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [x] Testing
  - [x] Code is tested

#610 

#### What's this PR do?
- Adds logic to handle course links from external checkout e.g. `?course_id=<>`
- Converts the course_id into product and adds that to the cart
- Adds the associated course_id product to the cart and redirects the user to checkout

- With edX:
    - Setup edx
    - Install the latest external-checkout plugin from https://github.com/mitodl/open-edx-plugins/tree/main/src/ol_openedx_checkout_external (Follow the setup process mentioned in README)
    - Setup Course modes in edx (Verified & Audit) & enroll
    - Make sure your MITxOnline course run's `courseware_id` matches the edx's course Id and a product against the course run is present
    - In edX, You'll see `Upgrade button` click that and you should be redirected to `/cart/add?course_id=?` which should redirect you to checkout page with the product added to the cart

- Without edX:
    - If you don't want to use or can't use it with edX you can manually hit the URL `http://mitxonline.odl.local:8013/cart/add/?course_id=course-v1%3AMIT%2BMIT_RR_1%2BR1?`. Note that course_id is quoted as done [here](https://github.com/mitodl/open-edx-plugins/pull/63/files#diff-34466ba0b1f7e3c516578a76aeec05141923f313b441ae0cc795a163a0e1a30bR66). This is quoted because edX course ID's are generally of format e.g. `course-v1:MIT+MIT_RR_1+R1` and when this is sent as plain text in URL the param would be converted to `course-v1:MIT MIT_RR_1 R1` removing the special character `+`. So to fix that I've updated the plugin itself([see this PR](https://github.com/mitodl/open-edx-plugins/pull/63/files)).

### DIFF
--- a/ecommerce/views/v0/__init__.py
+++ b/ecommerce/views/v0/__init__.py
@@ -506,9 +506,20 @@ class CheckoutProductView(LoginRequiredMixin, RedirectView):
             )
             basket.basket_items.all().delete()
 
-            for product in Product.objects.filter(
-                id__in=self.request.GET.getlist("product_id")
-            ):
+            # Incoming product ids from internal checkout
+            all_product_ids = self.request.GET.getlist("product_id")
+
+            # If the request is from an external source we would have course_id as query param
+            course_ids = self.request.GET.getlist("course_id")
+
+            all_product_ids.extend(
+                list(
+                    CourseRun.objects.filter(courseware_id__in=course_ids).values_list(
+                        "products__id", flat=True
+                    )
+                )
+            )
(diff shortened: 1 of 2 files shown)


jkachel (approved):
Tested w/out edX - raw URL worked OK (as long as I remembered to grab a course that had a valid product). LGTM 👍

PR #627, opened 2022-06-15
Adds username normalization

### LINKED ISSUE #259: Normalize accented characters when validating username uniqueness

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

#259 

#### What's this PR do?

Fixes #259.

This adds a `normalized_username` field to the users model and an API function to perform the normalization. It also updates the openedx `create_edx_user` call to pass the `normalized_username` field instead of the regular `username` field. 

The migrations to add the `normalized_username` field will automatically normalize the usernames. Because of that, it's pretty important that your user database does not contain accounts that will conflict. (Per #259, the `normalized_username` field is uniqued.) There is a management command to check for and fix usernames that would conflict that's in PR https://github.com/mitodl/mitxonline/pull/629.

After performing the migrations, create a user with a username using accented characters (áüô, etc.). The username should be created on the Open edX side with those characters converted to non-accented characters (auo, etc.). 

Some characters will _not_ be normalized and will instead be dropped. This includes ø so testing with that character will result in a normalized username without that character (for example, Søren will normalize to Sren).


users/models_test.py
```diff
@@ -123,3 +132,16 @@ def test_user_is_editor(is_staff, is_superuser, has_editor_group, exp_is_editor)
         user.groups.add(Group.objects.get(name=CMS_EDITORS_GROUP_NAME))
         user.save()
     assert user.is_editor is exp_is_editor
+
+
+@pytest.mark.django_db
+def test_username_normalization():
```
annagav, line 138: Is this a duplicate of the test in the api.py? Is this test incomplete?

annagav (changes requested):
Just a minor comment

annagav (approved):
👍

pdpinch:
@jkachel can you remind me why you put this on hold?

Would this have avoided the situation in https://odl.zendesk.com/agent/tickets/139242 and https://sentry.io/organizations/mit-office-of-digital-learning/issues/2616447470/events/9ab540c0d80f4983933e71bb7b9cd728/?project=5864687

jkachel:
> @jkachel can you remind me why you put this on hold?
> 
> Would this have avoided the situation in https://odl.zendesk.com/agent/tickets/139242 and https://sentry.io/organizations/mit-office-of-digital-learning/issues/2616447470/events/9ab540c0d80f4983933e71bb7b9cd728/?project=5864687

This was put on hold because it needed some coordination to do - there was a part to this that was normalizing the existing usernames in the database. (See https://github.com/mitodl/mitxonline/issues/259#issuecomment-946706104) This was done about the time Brian left, though; I was going to get him to help with that. Otherwise, the code worked and it would have caught this, I think. It'd need some revising to work in the current version of the app before we could deploy it now.

pdpinch:
@jkachel doesn't look like we're going to get back to this anytime soon. Can we close this PR? 

I'm not even sure it would still be relevant once we move to keycloak SSO

jkachel:
> @jkachel doesn't look like we're going to get back to this anytime soon. Can we close this PR?

Oh wow, I forgot this was out there. This will definitely get superseded by the Keycloak integration, so closing it.

PR #628, opened 2022-06-16
Nl/refine builds

rhysyngsun:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

(Required)

#### What's this PR do?
(Required)

(Required)

(Optional)

(Optional)

(Optional)


frontend/public/babel.config.js
```diff
@@ -0,0 +1,24 @@
...
+            "@babel/env",
+            "@babel/preset-react"
+        ] : [
+          ["@babel/preset-env", { modules: false }],
+          "@babel/preset-react",
+          "@babel/preset-flow"
+        ],
+        plugins: [
+          "@babel/plugin-transform-flow-strip-types",
+          "react-hot-loader/babel",
```
ChristopherChudzicki, line 15: This is getting used in all environments right now. Any harm in that, or should it be non-prod only?

rhysyngsun:
@jkachel this should be good to try again, I added a script that waits for both refine and webpack dev server to run before running the django scripts. I also made the django settings resilient against those directories not existing.

scripts/run-django-dev.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# This script runs the django app, waiting on the react applications to build
+
+python3 manage.py collectstatic --noinput --clear
```
ChristopherChudzicki, line 5: I thought `collectstatic` needed to run after the `wait-for`, once the build directories have completed?

How will collectstatic / webpack build work in production? I saw this note in the django-webpack-loader docs (https://github.com/django-webpack/django-webpack-loader#usage-in-production):

>⚠️ Heroku is one platform that automatically runs collectstatic for you, so you need to set DISABLE_COLLECTSTATIC=1 environment var. Instead, you must manually run collectstatic after running webpack. In Heroku, this is achieved with a post_compile hook. You can see an example on how to implement this flow on [django-react-boilerplate](https://github.com/vintasoftware/django-react-boilerplate/tree/master/bin).

jkachel:
The main site appears to work OK. The Refine admin, though, seems like it's not aware that it's in a subdirectory. The front page loads (at `http://mitxonline.odl.local:8013/staff-dashboard/` - it doesn't load on port 8012) but the URLs it redirects to don't contain `/staff-dashboard`. (For example, it changes the URL to `http://mitxonline.odl.local:8013/login?to=%2Fstaff-dashboard%2F` after the app loads.)

scripts/http-ok.sh
```diff
@@ -0,0 +1,22 @@
...
+        }));
+        return;
+    }
+    res.writeHead(500);
+    res.end(JSON.stringify({
+        success: false,
+        server: "healthcheck-ok",
+        message: "This service is only meant to be queried at /health"
+    }));
+}).listen(process.env.PORT);
```
jkachel, line 22: `process.env.PORT` is undefined in the watch container - just needs to be added to the environment in the docker-compose but this does cause it to not build when things are set to production (I think - that worked for me, at least to get it to start up).

jkachel (changes requested):
Apart from staff-dashboard issues (noted earlier) and the port bit - when building for production, it seems to want to use URLs from the watch container (rather than relative URLs to `/static/`). (It's using the compiled sources - the URLs have hashes in them - but still pointing at `http://mitxonline.odl.local:8012/`. )

rhysyngsun:
@jkachel I'm pretty sure this is fixed now

jkachel (approved):
LGTM 👍 - both production and dev builds worked for me locally.

PR #629, opened 2022-06-16
Adds management command to find possible username conflicts

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#259 

#### What's this PR do?

Adds a management command to identify usernames that would conflict after normalization. This is step one of the process for resolving #259. 

The management command is `find_username_conflicts` and running it will load all user IDs and usernames from the database and run through them to see if a normalized username would be generated more than once. If the normalized username would be generated more than once, it outputs that normalized username with the IDs of the users that would cause a conflict. These can then be manually edited (in Django Admin or in the database directly). 

The command does not change usernames at all and the standard caveats apply to modifying usernames manually - they will need to be modified on the Open edX side as well. 

This command will need to be run and the affected accounts adjusted before https://github.com/mitodl/mitxonline/pull/627 can be merged, as it adds a uniqued `normalized_username` field and automatically normalizes existing usernames. 

Create two users: their usernames should share spelling but one should contain accented characters where the other doesn't. (For example, `taylor` and `tâylör`.) Then, run the `find_username_conflicts` command. It should identify those two accounts specifically as accounts that may conflict after normalization.

### DIFF
--- a/users/management/commands/find_username_conflicts.py
+++ b/users/management/commands/find_username_conflicts.py
@@ -1,0 +1,52 @@
+import sys
+import unicodedata
+from argparse import RawTextHelpFormatter
+
+from django.core.management import BaseCommand
+
+from users.models import User
+
+
+class Command(BaseCommand):
+    """Finds usernames that would cause conflicts after normalization"""
+
+    help = """
+  Finds usernames that would cause conflicts after normalization. 
+  """
+
+    def make_normalized_username(self, username):
+        """Strips non-ASCII characters from the username."""
+        normalized_username = unicodedata.normalize("NFD", username)
+        normalized_username = normalized_username.encode("ascii", "ignore").decode(
+            "utf-8"
+        )
+        return str(normalized_username)
+
+    def create_parser(self, prog_name, subcommand):  # pylint: disable=arguments-differ
+        """
+        create parser to add new line in help text.
+        """
+        parser = super().create_parser(prog_name, subcommand)
+        parser.formatter_class = RawTextHelpFormatter
+        return parser
+
+    def handle(self, *args, **kwargs):
+        all_users = User.objects.only("id", "username").all()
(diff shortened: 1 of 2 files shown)


briangrossman:
@jkachel 

Is it correct that the process will be:
1. While `find_username_conflicts` returns conflicts:
   - Manually update MITx Online usernames
2. Merge https://github.com/mitodl/mitxonline/pull/627
?

Note that the steps to update an MITx Online username are:
- Update the username for the user record in the MITx Online admin
- Update the username for the user record in the MITx Online Open edX admin
- If the learner has a Micromasters account:
   - Update the username (UID) for the user's social auth in the Micromasters admin (the mitxonline one, not the edX one)

jkachel:
@briangrossman yes, that'll be exactly the process I had in mind and should prevent https://github.com/mitodl/mitxonline/pull/627 from failing to deploy successfully. (One of the migrations in there will automatically normalize the incoming usernames and it'll fail if it hits a conflict.)

arslanashraf7:
@jkachel Can this PR be reviewed independently of is should be reviewed after https://github.com/mitodl/mitxonline/pull/627?

jkachel:
@arslanashraf7 this actually needs to be reviewed/merged before #627

arslanashraf7 (approved):
I tested it with the following cases:

- Create normal users (No results)
- Create a user with a normal username and then a user with some accented character in the same username (Outputs usernames & ids)
- Create a user with accented characters based username & then a new user with a normal username without accented chars (Prints ids of the conflicts)
- Above step with multiple users with the same username but different accented chars (Prints multiple IDs of those users)

For more readability, we could print email ids too for the conflicted users because even the usernames would be the same considering accents the email ids might be different. But I'm okay with this since this looks more like a one-timer.


Looks good 👍

PR #630, opened 2022-06-17
If a coupon is entered it should replace the financial aid discount only if it's a higher discount.

### LINKED ISSUE #612: If a coupon is entered it should replace the financial aid discount only if it's a higher discount.
If a coupon is entered it should replace the financial aid discount only if it's a higher discount.

umarmughal824:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

fixes #612 

#### What's this PR do?
If a coupon is entered it should replace the financial aid discount only if it's a higher discount.

- Make sure you have flexible price details for the user and it's applied on the cart
- apply the coupon on the cart.

### DIFF
--- a/ecommerce/views/v0/__init__.py
+++ b/ecommerce/views/v0/__init__.py
@@ -54,6 +54,7 @@ from courses.models import (
 )
 
 from ecommerce import api
+from ecommerce.discounts import DiscountType
 from ecommerce.serializers import (
     OrderHistorySerializer,
     ProductSerializer,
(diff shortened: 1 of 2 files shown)


jkachel (approved):
Tested with 5 different coupon codes:
- With just flex pricing discount applied: applying lower value coupon codes was not successful (flex pricing applied) while applying higher value codes did as expected
- With flex pricing discount and a higher value learner-provided discount applied: applying lower-value coupon reset to the flex pricing discount and a higher value coupon was applied successfully. (I think this fits the flow we've established for this operation.) 

This looks good to me. 👍

Issue #631, opened 2022-06-17
Dropdown `Justification` is not maintaining the state after refresh or after fetching data


asadiqbal08:
### Steps to Reproduce
After merging the change set for [issue#608](https://github.com/mitodl/mitxonline/issues/608), Found an issue that justification dropdown did not maintain the state after fetching data from server. 

Steps to produce:
- In order to change the status of flexible price request in Refine Admin portal, Click on any button e.g. `Deny`.
- Choose the justification from the list for denial and save the change set.
- Change save in local react state but refreshed to initial state once refresh the page.

### Expected Behavior

Justification dropdown should maintain the state that choosed last time for the flexible price request even after refreshing the page.

### Actual Behavior

Justification dropdown was not maintaining the state after fetching data from server again.

PR #632, opened 2022-06-17
asadiqbal08/Dropdown Justification is not maintaining the state after refresh

asadiqbal08:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

fixes: #631 

#### What's this PR do?
Fixes the issue for Dropdown `Justification` was not maintaining the state after refresh.


jkachel (changes requested):
The justification dropdown doesn't reflect state. (I think the problem is on line 276: `selected={justification == option.value}` doesn't look in `modaldata`.)

asadiqbal08:
@jkachel I could not get your point. Can you elaborate a bit more ?

jkachel (approved):
LGTM 👍

Issue #635, opened 2022-06-21
Confusing message after submitting forgot password


briangrossman:
### Steps to Reproduce

- Go to: https://mitxonline.mit.edu/
- Click `Sign In`
- Enter your email and click `Continue`
- Click `Forgot Password`
- Enter your email and click `Continue`

### Expected Behavior

- The instructions for the case where the learner has input the wrong email address should ask them to try entering their password again
- Please change the text to: 
```
Is your email correct? If you made a typo, no problem, just try to reset your password again.
```
This is consistent with our other products

### Actual Behavior

- The instructions for  the case where the learner has input the wrong email address asks them to `create an account again`

### Screenshot or Screencast

MITx Online
<img width="639" alt="image" src="https://user-images.githubusercontent.com/21342568/174838390-b7edbdfa-3dff-4ef9-b872-c0fb23091afe.png">

xPRO - for reference
<img width="592" alt="image" src="https://user-images.githubusercontent.com/21342568/174838566-29411b17-cf4e-461f-9fd0-d1afe0cb6f58.png">

Bootcamps - for reference
<img width="706" alt="image" src="https://user-images.githubusercontent.com/21342568/174838498-ca2ed194-d508-4ff3-b2db-6f87b00c2903.png">

briangrossman:
@arslanashraf7 discovered that the email verification screen when creating a new account has the wrong text as well. It looks like it may have been swapped with the Forgot Password instructions. See https://github.com/mitodl/mitxonline/pull/636

PR #636, opened 2022-06-21
Updating copy on the Forgot Password confirmation screen

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

#635 

#### What's this PR do?

Updates the text on the Forgot Password confirmation screen.

Attempt to reset your password. The text under the "Is your email correct?" heading should reference resetting your password again rather than creating a new account. 

(Just desktop - this was strictly a copy change.)

image


arslanashraf7 (approved):
Tested it both with forgot password & create account flows and it works well. 👍

arslanashraf7:
@jkachel Since this ticket is only related to `Forgot Password` flow. But I just wanted to mention that in `Create Account` flow the message doesn't show `just try to`. Adding a screenshot of that below just in case.

It would be great if you could confirm that with @briangrossman before merging.

Screenshot 2022-06-22 at 3 36 31 PM

pdpinch:
So this isn't part of ol-django?

jkachel:
Hmm. That message has more problems - it also uses "password reset" when you're signing up. (For comparison, xPRO:
image

@briangrossman should this be a separate issue or can I take care of that here (or both)?

briangrossman:
@arslanashraf7  - Nice catch!
This is making sense now... it looks like the instructions may have been reversed for Forgot Password and Create Account. Since it's mostly a swap, we can probably take care of this in one PR - I took the liberty of updating the PR name. 

@jkachel 
Peter's question:
> So this isn't part of ol-django?

is a good one. Is this something that is / could be shared from ol-django?

arslanashraf7 (approved):
Thanks for making the changes in the register form. Looks good 👍

PR #639, opened 2022-06-23
Flexible Pricing: email notifications should be sent when statuses change

### LINKED ISSUE #616: Flexible Pricing: email notifications should be sent when statuses change
The Flexible Pricing system doesn't proactively provide the learner feedback about the status of their request. We should add email notifications to the system, similar to the email notifications in MicroMasters, to alert the learner of status changes. (This just means email when the status changes administratively.) 

### Acceptance Criteria
- [ ] The learner should receive an email when their flexible pricing request's status has changed in the admin interface. 
- [ ] The email should generally follow the text of the 
- [ ] The email received should include the discounted amount of the course in question, in the case where the notification is for an approval.

This functionality was mentioned in https://github.com/mitodl/mitxonline/pull/611 .

umarmughal824:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

fixes #616 

#### What's this PR do?
Flexible Pricing: email notifications should be sent when statuses change

Try to change the status of flexible price request on refine admin

#### Other

From @jkachel - the statuses that send email are approved, pending manual approval, and reset. This should match what was in MicroMasters. In addition, the email text (including formatting) is directly from MicroMasters other than changes to the product name and a minor change to use the standard email template. (MicroMasters just used constants for this; here we're using the constants plus the standard email template.)


umarmughal824:
@jkachel I followed the suggested approach for email but it's not really sending out any email and don't show any message either but I could see the email content being generated here

Screenshot 2022-06-24 at 11 47 03 AM

umarmughal824:
Celery tasks did not show any error either but still, no email is being received.

`[2022-06-24 12:49:04,741: INFO/ForkPoolWorker-2] Task flexiblepricing.tasks.notify_flexible_price_status_change_email[242f7847-d7fd-4219-b946-28fcdaf21388] succeeded in 1.0752656819968252s: None` (edited)

mail/api.py
```diff
@@ -20,18 +20,24 @@
 """
 import logging
 import re
+import requests
```
arslanashraf7, line 23: There are imports in this file but there is no direct usage. I think we should remove them,

flexiblepricing/mail_api.py
```diff
@@ -0,0 +1,89 @@
...
+    if flexible_price.status == FlexiblePriceStatus.APPROVED:
+        courserun = CourseRun.objects.filter(
+            course=flexible_price.courseware_object
+        ).first()
+        product = Product.objects.get(object_id=courserun.id)
+        price = DiscountType.get_discounted_price(
+            [flexible_price.tier.discount], product
+        )
+        message = FLEXIBLE_PRICE_APPROVAL_MESSAGE.format(
+            program_name=flexible_price.courseware_object.title, price=price
```
arslanashraf7, line 50: Should we rename this to something more generic to reflect both courses and programs?

flexiblepricing/mail_api.py
```diff
@@ -0,0 +1,89 @@
...
+
+from courses.models import CourseRun
+from flexiblepricing.constants import (
+    FLEXIBLE_PRICE_APPROVAL_MESSAGE,
+    FLEXIBLE_PRICE_APPROVAL_SUBJECT,
+    FLEXIBLE_PRICE_DOCUMENTS_RECEIVED_MESSAGE,
+    FLEXIBLE_PRICE_DOCUMENTS_RESET_MESSAGE,
+    FLEXIBLE_PRICE_RESET_SUBJECT,
+    FLEXIBLE_PRICE_DOCUMENTS_RECEIVED_SUBJECT,
+    FLEXIBLE_PRICE_EMAIL_BODY,
```
arslanashraf7, line 17: nit: We can adjust the imports in a more readable way e.g. importing `FLEXIBLE_PRICE_DOCUMENTS_*` in order and the same for the others.

arslanashraf7 (commented):
The code overall looks good to me, but I've suggested some minor changes.

I wasn't able to functionally test the PR yet since the refine configuration isn't letting me log on so I'm working on that.

Q: I also tried to test it through Django Admin by changing flexible price status but the celery task doesn't trigger for me, is it supposed to be working through Django Admin too ? or just refine dashboard?

arslanashraf7 (approved):
+ Continuing to my last comment, I was able to set up refine and functionality test it through `/staff-dashboard` and get the email notifications for the Flexible Price status change.

So, Just some minor comments that I added above, and then it looks good to me. 👍

flexiblepricing/mail_api.py
jkachel: FYI I renamed and regrouped these so that they're a bit cleaner but ultimately I'm leaving this mostly alone - I'd prefer we get an actual email template (like we have for receipts/etc.) and use that instead, which would eliminate these constants eventually.

flexiblepricing/mail_api.py
```diff
@@ -0,0 +1,89 @@
...
+    try:
+        with get_message_sender(FlexiblePriceStatusChangeMessage) as sender:
+            # import pdb
+            # pdb.set_trace()
+            # print(sender.can_email_user(flexible_price.user))
+            # built = sender.build_message(
+            #     flexible_price.user.email,
+            #     {"subject": subject, "body": body},
+            # )
+            # sender.send_message(built)
```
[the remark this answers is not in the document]
arslanashraf7: @jkachel Thanks for making the changes, Just a quick reminder on this just in case, you might already be making this change.

PR #651, opened 2022-06-28
Capture learner's country when saving flexible pricing request

### LINKED ISSUE #650: Flexible Pricing: Auto-approval does not find the learner's country
### Steps to Reproduce

Note: the default threshold is $75,000, so you will need to either choose a country with a lower threshold or modify the threshold for that country. 

1. Log into Django Admin and find the auto-approval threshold in `Country income thresholds` you want to test with.
2. Log back in as a user in the country that you chose in step 1. 
3. Submit a flexible pricing request (which specific course/program doesn't matter) for an amount below the threshold.
4. Observe the status of the request - it should be pending manual approval.
5. Submit a new flexible pricing request (which specific course/program doesn't matter) for an amount above the threshold.
6. Observe the status of the request - it should be auto-approved.

### Expected Behavior

In step 5, the status should be `auto-approved` when the request is above the threshold.

### Actual Behavior

An error message is logged: `Country code None does not exist in CountryIncomeThreshold for flexible price id X` (where X is the flexible pricing request object ID). This error message also means that the default threshold of $75,000 is used, so if the income level reported is under that amount (but above the threshold for the country the learner is in), it will be pending approval.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#650

#### What's this PR do?

Fixes #650 - pulls the learner's country from their profile when submitting their flexible pricing request so the auto-approve code has something to consider (outside of the default)

Submit a flexible pricing request for a course as a user in a country that has a threshold of < $75,000. The request should be for an income level above the threshold. It should auto-approve the request.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -706,6 +706,7 @@ class FlexiblePricingRequestForm(AbstractForm):
             user=form.user,
             original_income=form.cleaned_data["your_income"],
             original_currency=form.cleaned_data["income_currency"],
+            country_of_income=form.user.legal_address.country,
             income_usd=income_usd,
             date_exchange_rate=datetime.datetime.now(),
             cms_submission=form_submission,
--- a/flexiblepricing/api.py
+++ b/flexiblepricing/api.py
@@ -121,26 +121,26 @@ def determine_tier_courseware(courseware, income):
     return tier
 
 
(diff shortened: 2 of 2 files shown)


arslanashraf7 (approved):
Looks good 👍

Issue #643, opened 2022-06-23
Flaky test in `ecommerce/views_test.py`


rhysyngsun:
See the log output from this:

https://github.com/mitodl/mitxonline/runs/7026103146?check_suite_focus=true#step:14:321

This is occasionally failing on github actions.

rhysyngsun:
As a temporary workaround I added the `flaky` package and annotated this test so that if it fails once it doesn't require a rerun of the entire test suite: https://github.com/mitodl/mitxonline/blob/main/ecommerce/views_test.py#L184

jkachel:
Think this is OK now with the rework done in #653.

PR #653, opened 2022-06-30
Reworks flexible pricing discount test to reduce flakiness

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#643 

#### What's this PR do?

Reworks `test_redeem_discount_with_higher_discount` so that it doesn't randomly fail.

Run pytest tests. The affected test should be successful.


arslanashraf7 (approved):
👍 Ran the test multiple times and it passes now and the change seems seasonable.

PR #655, opened 2022-07-05
Revert "Flexible Pricing: email notifications should be sent when statuses change"

jkachel:
Reverts mitodl/mitxonline#639 - didn't rebase from main and it reverted some other merged changes


rhysyngsun (approved):
👍 when tests pass

Issue #410, opened 2022-02-15
Warning: Change *coursepage.price* to use StreamFieldPanel (Wagtail)


arslanashraf7:
### Description

A warning is logged about the usage of `FieldPanel` in the `price` field of our `CoursePage`. Although this is just a warning, for now, it might break in the future so we should refactor the model to use StreamFieldPanel instead of the `FieldPanel`.

### Steps to Reproduce
- Simply fire up your containers(`docker-compose up`) and you should see this warning being logged into the terminal. 

### Expected Behavior
- No, warning about the StreamField usage

### Actual Behavior
- A warning is logged about the usage of the StreamField instead of the S

### Stacktrace
```
web_1     | WARNINGS:
web_1     | cms.CoursePage: (wagtailadmin.W003) CoursePage.price is a StreamField, but uses FieldPanel
web_1     | 	HINT: Ensure that it uses a StreamFieldPanel, or change the field type
```
### Related Issues
This might be related to some recent wagtail version updates but I'm not sure. It might be related to https://github.com/mitodl/mitxonline/pull/383

### Screenshot or Screencast
<img width="994" alt="Screenshot 2022-02-15 at 5 34 34 PM" src="https://user-images.githubusercontent.com/34372316/154064005-fd08fc97-4e7f-424d-8267-ae45737a3c2f.png">

pdpinch:
Can this be closed?

arslanashraf7:
Yes, Sorry there have been a couple of issues that might not be closed but I'll reiterate them and close which ones need to be closed. Closed via #662 .

PR #662, opened 2022-07-06
fix(warning): use StreamFieldPanel instead of FieldPanel

arslanashraf7:
#410 

#### What's this PR do?
Fixes wagtail warning regarding usage of `FieldPanel` instead of `StreamFieldPanel` for price field(Which is a `StremField`)

- In a course make sure you have added a price field, No move to this branch rebuild and check if the Price field on the course page still works fine.
- In a new course you can also try creating a Price field in the course through CMS and check that users can see that on the course details page.
- Check the logs and there should be no warning mentioned in the associated ticket
- Tests pass

#### Screenshots
Screenshot 2022-07-06 at 3 24 13 PM
Screenshot 2022-07-06 at 3 24 26 PM

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -389,7 +389,7 @@ class ProductPage(Page):
         FieldPanel("description"),
         FieldPanel("length"),
         FieldPanel("effort"),
-        FieldPanel("price"),
+        StreamFieldPanel("price"),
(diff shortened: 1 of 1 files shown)


jkachel (approved):
LGTM 👍

PR #668, opened 2022-07-07
Fixes error on justification field when changing a flex pricing request status 

### LINKED ISSUE #665: Flexible Pricing: Error on staff dashboard when changing request status without setting a justification
### Steps to Reproduce

1. Log into the staff dashboard.
2. Attempt to change the status of a request. Do not change the Justification.
3. An error occurs:
image

### Expected Behavior

The request should succeed.

### Actual Behavior

The request fails. The noted error message is displayed and the error from Django is 
```
{"errors":{"justification":["This field may not be blank."]}}
```

The payload in these cases has a justification field that is, in fact, blank, and not the currently selected option. Since there's no change event if you don't actually change the Justification dropdown, it should grab the justification that was there if Justification is blank. (This is especially noticeable when approving a request, since the first justification is "Documents in order" so you wouldn't normally have a reason to change it.)

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#665

#### What's this PR do?

Fixes #665 - refactors the approval modal out into its own component so that the state of the approval doesn't become weird. This also refactors the dialog to use the Ant Design `<Select>` component rather than a bare HTML `select` and adds a no-choice option for requests that haven't had a justification set yet.

1. Log into the staff dashboard and navigate to the Flexible Pricing screen.
2. Attempt to change the status of a request that has a justification set. Do not change the justification. The change should succeed. 
3. Attempt to change the status of a request that has not had a justification set. Leave the justification field blank. You should receive a toast informing you to choose a justification. 
4. Change the justification of a request and save the change. The justification should change. 
5. Open the status dialog on several requests. The Justification field should report the current justification that's set for each of them.

### DIFF
--- a/frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
+++ b/frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
@@ -156,13 +156,16 @@ export const FlexiblePricingList: React.FC = () => {
                 description: "Error",
                 key: "bad-justification-error",
                 type: "error",
+                undoableTimeout: 3000,
             });
 
             return;
         }
-        console.log(justification);
-        setmodaldata({...modaldata, justification: justification})
-        console.log(modaldata);
+
+        let settable_modaldata = modaldata;
+        settable_modaldata.justification = justification;
+        setmodaldata(settable_modaldata)
+        
         handleUpdate(modaldata, modaldata.action)
         setIsModalVisible(false);
     };
--- a/frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
+++ b/frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
@@ -1,4 +1,4 @@
-import { useUpdate, useNotification, CrudFilters, HttpError } from "@pankod/refine-core";
+import { CrudFilters, HttpError } from "@pankod/refine-core";
 import React from "react"
 const { useState } = React;
 import {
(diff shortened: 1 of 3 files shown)


umarmughal824:
@jkachel I am still getting the same error on the setup guide, seems like the README is not updated. Could you please ask someone else to review it who has the function refine dashboard on his local machine?

frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
```diff
@@ -98,6 +73,8 @@ const FlexiblePricingFilterForm: React.FC<{ formProps: FormProps }> = ({ formPro
 }
 
 export const FlexiblePricingList: React.FC = () => {
+    const { open: displayToast } = useNotification();
```
collinpreston, line 76: Is this still being used?

collinpreston:
@jkachel, I am having trouble creating a flexible price record with an empty justification field in order to test this.  I'm likely over looking how a step or two.  I wasn't able to generate an empty justification with and without your changes.

jkachel:
The justification is left blank when the learner submits the request initially, so you'll need to put in a new request to get one without a justification set. You shouldn't be able to save a blank justification through the staff dashboard interface.

frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
jkachel: Fixed in https://github.com/mitodl/mitxonline/pull/668/commits/eaee7b46d3a1e30a049498afbefa1d8201a8319d

collinpreston (approved):
Look good. Added a comment.

Issue #673, opened 2022-07-07
Order History: unauthenticated user can view (empty) order history page


pdpinch:
### Steps to Reproduce

1. Log out. 
2. visit /orders/history/ 

### Expected Behavior

1. User should be redirected to login page, i.e. /signin
2. After successful authentication, user should be returned to /orders/history

### Actual Behavior

- user can view an order history page, but with 0 orders

<img width="1115" alt="Screen Shot 2022-07-06 at 3 21 12 PM" src="https://github.mit.edu/storage/user/5/files/5572da00-fd3f-11ec-82b1-26b080b336bc">

### Related Issues

- In the future, we would like to be able to support customer support team members being able to view learner's order history

collinpreston:
@pdpinch can this issue be closed since https://github.com/mitodl/mitxonline/pull/681 was merged?

jkachel:
Closing this; the cart routes are PrivateRoutes as of #681 so this'll redirect to the login screen as expected.

PR #680, opened 2022-07-08
Order History redirects if user isn't logged in

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#673 

#### What's this PR do?

Adds a http status code check when loading the order history. It will now redirect to the sign in page if the code that's returned is an error code. 

Navigate to `/orders/history/` as a user that isn't logged in. You should be redirected to the sign in page. 

Navigate to the order history page after logging in. You should see the order history as per normal.


jkachel:
Closing - this was taken care of more effectively in https://github.com/mitodl/mitxonline/pull/681

PR #681, opened 2022-07-08
Order history and order receipt accessible when logged out

collinpreston:
#### Pre-Flight checklist
- [ ] Testing
  - [ ] Code is tested

#673 
#### What's this PR do?
Allows only users who are logged in the ability to view the `orders/history/` and `orders/receipt/` pages.  If those pages are accessed by a user who is not logged in, then the user will be redirected to the sign in page.

Navigate to /orders/history/ as a user that isn't logged in. You should be redirected to the sign in page.

Navigate to the order history page after logging in. You should see the order history as per normal.


jkachel (approved):
LGTM 👍

PR #696, opened 2022-07-13
Fixes link to order receipt in the email

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#619

#### What's this PR do?

The URL for the order receipt was bad; this tested OK locally but that was due to also having some local config set wrong. This also fixes a typo in the text ("Below you will find a copy of your receipt" vs. "Below you will find a copy of you receipt"). 

Generate a receipt email. you can do this either by creating a new order, or through the shell. (In the shell, I just import the Order model and the `send_ecommerce_order_receipt` function from mail_api and then run `send_ecommerce_order_receipt` on the last Order object to test.) The receipt link should be valid and the text immediately following it should be correct.


arslanashraf7 (approved):
👍

PR #699, opened 2022-07-15
Use master branch and don't clone mitodl edx

collinpreston:
Removal of out-of-date instructions.

#### What's this PR do?
Removes instructions in order to ensure that the reader will use the master branch for `devstack` instead of a specific version.  Removes instructions for the reader to clone the mitodl edx repo which is no longer required.


annagav (approved):
👍

Issue #687, opened 2022-07-12
Ecommerce: ability to limit discounts to only apply to specific products


rhysyngsun:
Summary
---
This seems to have been missed somewhere, but as we have it in our other products we will very likely need it for MITx Online as well. Right now discounts work globally on all products. This is a good option to have but it's more likely that discounts will be applicable for a handful of courses.

#### Acceptance Criteria

- [ ] Ability to specify zero or more products as applicable for the discount
- [ ] If at least one product is specified, limit usage of that discount based on that (in APIs, UI, backend code, etc)

pdpinch:
@jkachel @rhysyngsun can this be closed, now that #708 is merged?

PR #708, opened 2022-07-15
Adds support for tying a discount to a specific product

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

#687 

#### What's this PR do?

Adds support for tying a discount to a product. If the discount is tied to a product, it will only be redeemable if the learner's basket contains a product that the discount is associated with. 

Automated tests should pass.

1. Select a discount code and product to test with.
2. In Django Admin, add a new Discount Product record for the discount and product. 
3. Attempt to add that product to the cart. 
4. Then, apply the discount code. It should apply successfully.
5. Clear your cart and add a different item to the cart.
6. Apply the discount code again. You should get a message saying the code is invalid. 

Similarly, you should be able to add another Discount Product for the selected discount and a different product and the discounts should apply (or not) in the same manner.

You should also be able to load the discount in the staff dashboard and see any products that are assigned to the discount. 

#### Other

You can't assign products to discounts in the staff dashboard at this point. There are other things you can't do (like assign users) and other things it needs so I'd rather wrap all that up in a separate issue/PR.


ecommerce/api.py
```diff
@@ -107,6 +109,37 @@ def generate_checkout_payload(request):
...
+    Args:
+        - basket (Basket): the current basket
+        - discount (Discount|string: the discount to apply (if a string, loads the discount code specified)
+    Returns:
+        boolean
+    """
+    if not isinstance(discount, Discount):
+        discount = Discount.objects.filter(discount_code=discount).first()
+
+    basket_item_products = [item.product for item in basket.basket_items.all()]
```
annagav, line 129: I think this could be put into a model method like `get_all_products`or ` get_basket_products`, since somewhere else in the code you are also getting all the products from a basket, like here https://github.com/mitodl/mitxonline/blob/main/ecommerce/models.py#L339. So

annagav (changes requested):

ecommerce/api.py
jkachel: Got this done in https://github.com/mitodl/mitxonline/pull/708/commits/0da926720553395f9c4c9a6506dcecc3b9596699

ecommerce/api.py
```diff
@@ -107,6 +109,37 @@ def generate_checkout_payload(request):
...
+    Args:
+        - basket (Basket): the current basket
+        - discount (Discount|string: the discount to apply (if a string, loads the discount code specified)
+    Returns:
+        boolean
+    """
+    if not isinstance(discount, Discount):
+        discount = Discount.objects.filter(discount_code=discount).first()
+
+    basket_item_products = basket.get_products()
```
annagav, line 129: I think it would make more sense to name this `basket_products`, meaning products in a basket.

ecommerce/api.py
```diff
@@ -107,6 +109,37 @@ def generate_checkout_payload(request):
...
+
+    if (
+        not discount.products.count() == 0
+        and DiscountProduct.objects.filter(product__in=basket_item_products)
+        .filter(discount=discount)
+        .count()
```
annagav, line 135: I think calling .exits() would be a little cleaner here

annagav: Also you could rewrite this if statement and returns as:
```
return discount.products.count() == 0 or not DiscountProduct.objects.filter(product__in=basket_item_products)
        .filter(discount=discount).exists()
```

ecommerce/api.py
```diff
@@ -135,6 +168,10 @@ def apply_user_discounts(request):
             discount = user_discount.discount
 
     if discount:
+        # check for product specificity in the discount
```
annagav, line 171: Could you add a Returns section in the docstring for this function

annagav (changes requested):

ecommerce/api.py
jkachel: Done in https://github.com/mitodl/mitxonline/pull/708/commits/fe65eb05b4c6f846d4b157a46804f0f800656674

ecommerce/api.py
jkachel: Done in https://github.com/mitodl/mitxonline/pull/708/commits/fe65eb05b4c6f846d4b157a46804f0f800656674

ecommerce/api.py
jkachel: Done in https://github.com/mitodl/mitxonline/pull/708/commits/fe65eb05b4c6f846d4b157a46804f0f800656674

ecommerce/api.py
```diff
@@ -115,6 +144,9 @@ def apply_user_discounts(request):
 
     Args:
         - user (User): The currently authenticated user.
+
+    Returns:
+        True
```
annagav, line 149: does it always return True? :D

annagav (approved):
Looks good. Just a minor comment.

ecommerce/api.py
jkachel: it really doesn't return anything.. not sure why I set it to True but fixed that in https://github.com/mitodl/mitxonline/pull/708/commits/7ceb1391089ac4e01660d9204101bdddb1127f42

PR #711, opened 2022-07-18
Update steps for accessing and configuring devstack

collinpreston:
#### What's this PR do?
Add a reference to the default django admin login credentials for `devstack`.  Add steps to define `backend` and `enable` oauth configuration in devstack.  Add instructions on where to update the lms.yml file and inform the reader to first search for existing settings before updating/adding to the file.

### DIFF
--- a/docs/source/configuration/open_edx.rst
+++ b/docs/source/configuration/open_edx.rst
@@ -200,6 +200,7 @@ Configure Open edX to support OAuth2 authentication from MITx Online
 
   * Set ``OPENEDX_API_CLIENT_ID`` to the client id
   * Set ``OPENEDX_API_CLIENT_SECRET`` to the client secret
+  * Set ``OPENEDX_API_BASE_URL`` to ``http://host.docker.internal:18000`` when running ``devstack`` and ``mitxonline`` locally through Docker
 
 Configure Logout
 ################
--- a/docs/source/configuration/open_edx.rst
+++ b/docs/source/configuration/open_edx.rst
@@ -134,7 +134,7 @@ In MITx Online:
 
 In Open edX (derived from instructions `here <https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/configuration/tpa/tpa_integrate_open/tpa_oauth.html#additional-oauth2-providers-advanced>`_):
 
-* ``make lms-shell`` into the LMS container and ensure the following settings:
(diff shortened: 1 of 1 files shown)


docs/source/configuration/open_edx.rst
```diff
@@ -134,7 +134,7 @@ In MITx Online:
 
 In Open edX (derived from instructions `here <https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/configuration/tpa/tpa_integrate_open/tpa_oauth.html#additional-oauth2-providers-advanced>`_):
 
-* ``make lms-shell`` into the LMS container and ensure the following settings:
+* ``make lms-shell`` into the LMS container and ensure the following settings are configured in ``/edx/etc/lms.yml``.  Search for the setting names before appending anything new to the file.:
```
arslanashraf7, line 137: A very similar thing is done in https://github.com/mitodl/mitxonline/pull/683/files#diff-a90f6fb11bd555c7c3cf903cafc0798bf2abc9e796c6dfce046f9ffba5fbbd0dR137. That might conflict with this. If you agree with that you can remove this change from here.

docs/source/configuration/open_edx.rst
```diff
@@ -176,6 +177,9 @@ In Open edX (derived from instructions `here <https://edx.readthedocs.io/project
 
     * Linux users: The gateway IP of the docker-compose networking setup for mitxonline as found via ``docker network inspect mitx-online_default``
     * OSX users: Use ``host.docker.internal``
+    
+  * Save the configuration.  Then, while on the ``/admin/third_party_auth/oauth2providerconfig/`` page, select 'update' for the configuration you just created.
+    * Check the 'enabled' box at the top, and save the configuration.
```
arslanashraf7, line 182: I think this could be a redundant step, We have the `Enabled` checkbox while creating the configuration for the first time and the configuration doesn't really have to be created and then updated to check that box.

 Instead, I think we can add a bullet [here](https://github.com/mitodl/mitxonline/pull/711/files#diff-a90f6fb11bd555c7c3cf903cafc0798bf2abc9e796c6dfce046f9ffba5fbbd0dR160) for the `Enabled` checkbox where we mention about other checkboxes.

arslanashraf7 (changes requested):
I requested just a couple of small changes.

docs/source/configuration/open_edx.rst
collinpreston: Ah yes, I just saw that this morning.  Thanks for the catch.

PR #716, opened 2022-07-19
MicroMasters data migration command and model

### LINKED ISSUE #721: Implement a management command to run MM data migrations
Related to #438 

We need a management command that can execute a set of SQL queries in sequence that migrate data from MicroMasters. As part of this we'll also need a way to map object IDs from the MicroMasters database to the MITx Online one.

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing

Fixes #721

#### What's this PR do?

- Adds data models for mapping course and program ids
- Adds a management command that runs SQL files in alphanumeric order by filename.

- You should be able to go to django-admin and add new records for `CourseId` and `ProgramId`
- You should be able to run the `./manage.py import_micromasters_data` command


micromasters_import/management/commands/import_micromasters_data.py
```diff
@@ -0,0 +1,53 @@
...
+        if proceed.lower() != "y":
+            self.stdout.write(self.style.ERROR("Aborting operation"))
+            exit(1)
+
+        with connection.cursor() as cursor:
+            self.stdout.write(self.style.SUCCESS("Connected to database"))
+
+            for sql in sqls:
+                self.stdout.write(f"Executing SQL query: {sql.file_name}")
+                cursor.execute(sql.raw_sql)
```
rachellougee, line 51: If we could print out what's being executed in the sql file, that'd be helpful. It also would be nice to know how many rows are affected if it's insert using cursor.rowcount? I know that might get messed here. But we would want to get some insight of what's happening in these queries. 
Other than that, this import works with migration queries I had

rhysyngsun:
@rachellougee Addressed your feedback, I'm not sure exactly how the postgres driver determines the row count, but that property is what the documentation indicates should do the job.

rhysyngsun:
@rachellougee addressed your feedback, also noticed that `ProgramId` had an FK to `Course`, so I fixed that and redid the migration. You should run `./manage.py migrate micromasters_import zero` locally before pulling my latest changes otherwise you'll end up in a weird db state.

rachellougee (approved):
@rhysyngsun Re-tested. It looks good to me.

Issue #670, opened 2022-07-07
Flexible Pricing: display personalized price 


pdpinch:
As a user, I'd like to see my personalized price, including financial assistance, in every place where I see a course price. 

In particular, I'd like to see it on the course about page.  

#### Designs and Mockups

<img width="1532" alt="image" src="https://user-images.githubusercontent.com/430126/177823073-df78f574-a167-46d7-a0a8-46994820e033.png">

#### Acceptance Criteria:

- [ ] For logged in users whose financial assistance has been approved, display their personalized price instead of the undiscounted price. 

#### Out of Scope

- we don't want to change the link labelled "personalized pricing" in the screenshot at this time. The link text and its destination are managed in the CMS.

collinpreston:
@pdpinch - on the page shown in the issue description, what should the `Personal pricing` hyperlink direct to?

pdpinch:
That link is specified in the CMS. 

In production ([for example](https://mitxonline.mit.edu/courses/course-v1:MITxT+JPAL102x/)), it links to a FAQ about how personal pricing works. The text and link is set in the CMS -- in this example, at https://mitxonline.mit.edu/cms/pages/11/edit/

On RC, you can see that we've changed the destination of that link to be the form for requesting financial aid.

PR #720, opened 2022-07-20
Flexible pricing display personalized price

collinpreston:
#### Pre-Flight checklist
- [ ] Testing
  - [ ] Code is tested

https://github.com/mitodl/mitxonline/issues/670

#### What's this PR do?
Displays the course price, with the deducted flex price discount, for a logged in user who has been approved for flexible pricing.

1. Ensure you have the following already created: standard user, course, course run, flexible pricing tier, flexible price that is approved for the course and user.
3. Login to mitxonline
4. From the dashboard, click 'browse our courses'
5. Click on the course card
6. Click 'Enroll now'
7. A modal will display.  You should see the flexible pricing adjusted certificate price in the modal.
8. Click 'No thanks, I'll take the course....'
9. From the dashboard, click 'Course details'
10. You should see the flexible pricing adjusted certificate price in the card.

Screen Shot 2022-07-20 at 1 25 03 PM
Screen Shot 2022-07-20 at 1 24 26 PM

### DIFF
--- a/frontend/public/src/containers/ProductDetailEnrollApp.js
+++ b/frontend/public/src/containers/ProductDetailEnrollApp.js
@@ -50,6 +50,8 @@ function calculateCoursePriceWithFlex(coursePrice, flexDiscountAmount, flexDisco
     return coursePrice - ((flexDiscountAmount / 100) * coursePrice)
   case DISCOUNT_TYPE_FIXED_PRICE:
     return flexDiscountAmount
+  default:
+    return coursePrice
   }
 }
 
--- a/frontend/public/src/containers/ProductDetailEnrollApp.js
+++ b/frontend/public/src/containers/ProductDetailEnrollApp.js
@@ -55,7 +55,6 @@ export class ProductDetailEnrollApp extends React.Component<
     const { courseRuns } = this.props
     const { upgradeEnrollmentDialogVisibility } = this.state
     const product = run.products ? run.products[0] : null
-    const flexPriceDiscountAmount = product.product_flexible_price ?? 0
     return product ? (
       <Modal
         id={`upgrade-enrollment-dialog`}
@@ -71,7 +70,7 @@ export class ProductDetailEnrollApp extends React.Component<
             <div className="flex-grow-1 align-self-end">
               Learn online and get a certificate
             </div>
-            <div className="text-right align-self-end">${product.price - flexPriceDiscountAmount}</div>
+            <div className="text-right align-self-end">${product.price - product.product_flexible_price.amount}</div>
           </div>
           <div className="row">
             <div className="col-12">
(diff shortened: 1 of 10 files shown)


frontend/public/src/containers/ProductDetailEnrollApp.js
```diff
@@ -70,7 +71,7 @@ export class ProductDetailEnrollApp extends React.Component<
             <div className="flex-grow-1 align-self-end">
               Learn online and get a certificate
             </div>
-            <div className="text-right align-self-end">${product.price}</div>
+            <div className="text-right align-self-end">${product.price - flexPriceDiscountAmount}</div>
```
jkachel, line 74: The amount that's coming through here isn't always a dollar amount - they can be a set amount off, or they can be a percentage of the original amount or a fixed price. In the latter two cases, this calculation fails. For example, this was a $999 course that I ran through locally and set up with a 75% discount for the tier I was in: 
Screen Shot 2022-07-21 at 12 35 41 PM

(This should be $249.75.)

jkachel (changes requested):
FP calculation needs some work - this does seem to be OK otherwise.

frontend/public/src/containers/ProductDetailEnrollApp.js
```diff
@@ -55,7 +72,9 @@ export class ProductDetailEnrollApp extends React.Component<
     const { courseRuns } = this.props
     const { upgradeEnrollmentDialogVisibility } = this.state
     const product = run.products ? run.products[0] : null
-    const flexPriceDiscountAmount = product && product.product_flexible_price ? product.product_flexible_price.amount : 0
+    const flexDiscountAmount = product && product.product_flexible_price ? product.product_flexible_price.amount : 0
+    const flexDiscountType = product && product.product_flexible_price ? product.product_flexible_price.discount_type : null
+    const flexAdjustedCoursePrice = Number(calculateCoursePriceWithFlex(product.price, flexDiscountAmount, flexDiscountType)).toFixed(2)
```
jkachel: Nit: this gets repeated a couple times - can you refactor this into a utility function? Those are kept in `src/lib/utils.js`. Also - instead of `Number().toFixed()`, there's a utility function called `formatLocalePrice` that should be used here.

jkachel (changes requested):
Functionality looks good. Just a couple minor changes and should be ready to merge.

frontend/public/src/lib/util.js
```diff
@@ -231,10 +231,9 @@ export const formatLocalePrice = (amount: number | null) => {
   return amount.toLocaleString("en-US", { style: "currency", currency: "USD" })
 }
 
-export const getFlexiblePriceForProduct = product => {
+export const getFlexiblePriceForProduct = (product:any) => {
```
jkachel, line 234: The type for this is in `flow/cartTypes` (it'll need to be updated slightly to accommodate the new field) but once that is changed this should be good to go.

jkachel (approved):
One minor change regarding typing - otherwise LGTM 👍

frontend/public/src/lib/util.js
collinpreston: Let me know if my most recent change looks good

Issue #717, opened 2022-07-19
Course Certificates: Poll open edX for updated grades


pdpinch:
As a user, I'd like to have a certificate generated within 24-48 hours of my having passed the course. 

- This is a simpler alternative to https://github.com/mitodl/mitxonline/issues/690

#### Acceptance Criteria:

- [ ] Update learners' grades (in open courses) on mitxonline, from open edX, on a regular basis
- [ ] If the learner has passed a course, generate a certificate for them
- [ ] some courses won't want certificates generated until proctored exam results have been reviewed

#### out of scope

- alert learners by email when a certificate has been generated for them

pdpinch:
Virtual proctored exam grades will be included in the overall course grade, so one query per course should be sufficient.

pdpinch:
I can’t think of any reason to change the enrollment status when the certificate is generated. Nothing else at this time. 

> Once we know the learner has passed the course and we generate the certificate for them, is there anything else specific w.r.t to MITxOnline that we might want to do? e.g. Any changes in enrollment etc?

arslanashraf7:
Update:

So, I've ported most of the things from MITxPRO and wanted to take input before the PR is actually finalized. So the current workflow is like this:

1. We will have a cron job/scheduled celery task for certificates generation that by default runs daily at midnight just like xPRO (We have settings through we can customize it if we'd like [reference](https://github.com/mitodl/mitxonline/pull/722/files#diff-96677aa35debfefd031d9d34d9c70369754ee3acb2d9a9d4090e98612efee6f5R86-R93) ).

2. While generating certificates, the task will consider all the course runs whose end date has passed and their certificates haven't been generated yet. (We also have a delay setting in xPRO which will allow us to generate the certificates with a delay when the end date has passed [reference](https://github.com/mitodl/mitxonline/pull/722/files#diff-96677aa35debfefd031d9d34d9c70369754ee3acb2d9a9d4090e98612efee6f5R62-R65))

3. Certificates will be generated for all the users that have passed the course

**Things that are still in progress:**
>some courses won't want certificates generated until proctored exam results have been reviewed

Currently, I'm trying to set the proctored exams in edX to check if the grades API respects the status of the Proctored Exam e.g:
1. If the overall grades are above passing percentage, does that grades API respect if the `Proctored Eams` were reviewed or not? This means if a user has passing grades in a course but his proctored exam hasn't been. reviewed, does the send `passed=False` (If Yes, we might not need to change anything )
[...]

PR #722, opened 2022-07-21
feat: poll grades and generate certificates

arslanashraf7:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [x] Settings

#717 

#### What's this PR do?
- Ports parts of certificates and grades polling-based Certificates from xPRO ([ref](https://github.com/mitodl/mitxpro/blob/master/courses/tasks.py#L26))
- Integrate with edX grades API and fetch grades and save them in MITxOnline
- Generate certificates based on available grades passed & paid status of user
- A cron job to be scheduled to run every 24 hours which would fetch grades of all the eligible course runs and save the grades and create certificates if the criteria meet
- Some refactoring

Below might be good to read before you start testing it and get familiarized with certain criteria.

**1. Eligible Course runs for cron Job:**
- Run end date is in the future or empty
- Since this job will run every 24 hrs and it is possible a course was ended after the last time the job ran, to align with it we have a `CERTIFICATE_CREATION_DELAY_IN_HOURS (Default=24hrs)` - This is the time we use for picking course runs that ended recently e.g. in the last 24 hours.

**2. Eligible Certificates*
1. If a course run has`self_paced=True` we generate the certificates right away when the job runs (For eligible students)
2.  If a course has `self_paced=False` we generate the certificates only if the `end_date` has passed for the run
3. For a User, The certificate generation would be fulfillment of `Point#1/Point#2`, his grade from edX should be `Passed`, and his enrollment should be Paid(`Verified`) (In xPRO there were no free enrollments)

**3. Now to actually test these scenarios:**
1. You need to configure your local edX devstack with MITxOnline.
2. Create a course in edX and link it to MITxOnline.
3. In edX, create grading for the course by visiting (advanced settings in Studio) with passing grade and grade division for Lab/Home work, Mid, Final. (Try to set max grading portion to lab/Homework for easy testing).
4. Go through the course and get your user to pass the course in edX.
5. You can now run `generate_course_certificates` from Django Admin shell and verify the above scenarios for course run/and certificate eligibility
6. Check `CourseRunGrade` and `CourseRunCertificate` models in Admin for the entries

- Configure edX with MITxOnline locally
- Create a course in edX through Studio, set grading policy, and link the course to MITxOnline


arslanashraf7:
Marking this blocked by https://github.com/mitodl/mitxonline/pull/740. I believe that PR should merge first but I'll reiterate on that to see if we can still get this PR reviewed and/or merged.

courses/api.py
```diff
@@ -579,3 +581,34 @@ def is_program_text_id(item_text_id):
...
+        (courses.models.CourseRunCertificate, bool, bool) that depicts the CourseRunCertificate, created, deleted values
+    """
+    user = course_run_grade.user
+    course_run = course_run_grade.course_run
+
+    # A grade of 0.0 indicates that the certificate should be deleted
+    should_delete = not bool(course_run_grade.grade)
+    should_create = course_run_grade.is_certificate_eligible
+
+    if should_delete:
```
rhysyngsun, line 603: I'm not sure that this operation should be deleting certificates and I don't see anything about it in the issue Peter wrote, is there a reason you included this behavior?

courses/models.py
```diff
@@ -392,6 +392,7 @@ class CourseRun(TimestampedModel):
     )
 
     live = models.BooleanField(default=False)
+    self_paced_certificates = models.BooleanField(default=False)
```
rhysyngsun, line 395: Nit: I think this should probably be renamed a bit - a) boolean fields should be prefixed with either `is_` or `has_` b) the certificates being self paced sounds a bit odd, is it rather that the course run is self paced?

courses/tasks.py
```diff
@@ -39,3 +47,71 @@ def subscribe_edx_course_emails(enrollment_id):
...
         enrollment.edx_emails_subscription = subscribed
         enrollment.save()
+
+
+@app.task
+def generate_course_certificates():
+    """
+    Task to generate certificates for courses.
+    """
+    now = now_in_utc()
```
rhysyngsun, line 57: The body of this task is complex enough that it should be moved into an `api.py` method (or two).

rhysyngsun (changes requested):
Ran short on time today, but here's a partial code review based on a quick skim of the code.

courses/api.py
arslanashraf7: Most of the things are migrated from xPRO and this has been part of the certificate process there, I also wondered why. And on thing i could think of was that we have an option for resetting grades for a user in edX and in that case if a user gets his grades reset by Course Admins he is going to have `0` grade in which case I suppose we delete the certificate if it has been created already. But we can change this, let me know what you think?

arslanashraf7:
@rhysyngsun Thanks for the review, I've been addressing the partial feedback and it should be done shortly. Just wanted to sync in case you have continued the review, if not, maybe it might be good idea if you could take the look once i address the current feedback.

courses/models.py
arslanashraf7: Atleast for now the self paces would only make sense for certificates but you're right if we make it generic it would make more sense for other conditions that we might place based on self paced of a course run. I've changed it to `is_self_paced`

main/settings.py
```diff
@@ -815,6 +831,16 @@
         "task": "sheets.tasks.process_refund_requests",
         "schedule": crontab(minute=CRON_PROCESS_REFUND_REQUESTS_MINUTES),
     },
+    "generate-course-certificate": {
+        "task": "courses.tasks.generate_course_certificates",
+        "schedule": crontab(
```
rhysyngsun, line 836: I'm wondering what your thoughts are on this, I think this would be a good opportunity to make crontab configurations a lot more flexible, what do you think of this PR?

https://github.com/mitodl/ol-django/pull/92

arslanashraf7: I think that will surely make it cleaner and easier for defining the crontab jobs. I can review that PR. 

Although i wonder if we should start migrating to that right from this PR or would it be a good idea to create a PR that migrates all our jobs simultaneously to using the the parser from the ol-django.

rhysyngsun: That's a good point, probably worth doing as a separate PR also because it'll probably require a PR in `saltops` so it'd be best to just have one PR on each repo instead of a few here and there.

arslanashraf7: I agree, In that case, I think you might be able to continue the review on this PR.

pdpinch:
Thinking ahead a little bit, I have a question:

We have some courses in mitxonline production  that have been up for some time now, as audit only. When we turn on payments, and a user pays for a course, will we create a certificate for the user the next time we sync grades? Or is some other action necessary?

arslanashraf7:
> Thinking ahead a little bit, I have a question:
> 
> We have some courses in mitxonline production that have been up for some time now, as audit only. When we turn on payments, and a user pays for a course, will we create a certificate for the user the next time we sync grades? Or is some other action necessary?

@pdpinch So the current implementation is like mentioned below:

**Criteria for picking course runs for this task:**
1. The course runs that have `end_date=None`
2. The runs which have end_date in future
3. The runs that have ended in the last 24 hrs (This hour value is configurable and the default is **24hrs**).

**Criteria to update grade**
1. Since we want to update the grade on regular bases so for all the course runs picked above we update/create the grades for all the enrolled users without any additional conditions.

**Criteria for generating certificates**
1. First thing we now check is if the course run has `is_self_paced=True` or course `end_date` has passed (For non self-paced we want to generate at the end of the course run).
2. If the criteria in above point#1 are met, we check if the user(s) has a `passed` grade and his/her enrollment is paid, then we generate a certificate.

The answer to your question depends upon the above criteria, in addition, we are also creating a command which we should be able to run to generate grades& certificates and the implementation of the command is like "We won't check for the above criteria mentiond for picking course runs and picking the users based on course end_date or self paced, we just pick the run sync the grades and generate the certificates if the user has passed & paid enrollment".

courses/api.py
rhysyngsun: That makes sense.

rhysyngsun (commented):
Code looks good, I'll give this a functional test in the morning.

courses/api.py
```diff
@@ -579,3 +584,106 @@ def is_program_text_id(item_text_id):
...
+    )
+    return course_runs
+
+
+def generate_course_run_certificates():
+    """
+    Hits the edX grades API for eligible course runs and generates the certificates and grades for users for course runs
+    """
+    now = now_in_utc()
+    course_runs = get_certificate_grade_eligible_runs()
```
rhysyngsun, line 645: I think you should be passing `now` to this so that the logic across this function uses the same value (this is realistically probably not an issue, but consistency is probably good).

Other than that, LGTM so feel free to merge after changing that.

rhysyngsun (approved):

PR #723, opened 2022-07-21
Adding Google Sheets Refunds functionality to mitxonline

### LINKED ISSUE #669: Refunds: Integrate google sheets library with MITx Online
Summary
---
We need to integrate the work done in the previous work on refunds together so that a refund can be issued via google sheets.

#### Blocked By

- https://github.com/mitodl/mitxonline/issues/571
- https://github.com/mitodl/ol-django/issues/67 

#### Acceptance Criteria

This issue will be considered complete when as a staff member I can:

- [ ] Enter refund information in the configured google sheet
- [ ] MITx Online updates the order status to reflect the refund with a refund transaction
- [ ] MITx Online calls the CyberSource API to issue the refund to the learner
- [ ] MITx Online updates the google sheet indicating that the refund has been processed successfully or with an indication of failure

annagav:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

Fix #669

#### What's this PR do?
Installs mitol-google-sheets, and mitol-google-sheets-refunds libraries
Adds implementation of the refund process 

When google sheet is configured for processing refunds, the following process takes place:
1. Staff enters refund information in the configured google sheet
2. An scheduled task in MITx Online picks up the new change and  updates the order status to reflect the refund with a refund transaction
3. MITx Online calls the CyberSource API to issue the refund to the learner
4. MITx Online updates the google sheet indicating that the refund has been processed successfully or with an indication of failure

Follow the instructions [here](https://github.com/mitodl/ol-django/blob/85bea3ec5da01180ef943deb89b14d1463eb7c21/src/mitol/google_sheets/README.md) for developer setup.
When you Google Drive is set up correctly, proceed to testing:

1. Fill out and submit the spreadsheet request form. You need to link the refund form responses form to output to the spreadsheet. You can do that by opening the form 
and Responses->Settings->Select response destination-> Select existing spreadsheet.
When you fill out the form it will create a new worksheet, called something like "Form Response 1". You can
rename this tab to "Refund Form Response".
In the "Refund Form Response" sheet make sure that the "Timestamp" column format is set to "Date" and not to "Date Time".
The main google worksheet gets updated by the "Refund Response" worksheet by the following query:

```markdown
={QUERY({'Refund Form Response'!A2:G, ARRAYFORMULA(if(isblank('Refund Form Response'!A2:A),"",ROW('Refund Form Response'!A2:G)))},"SELECT Col8, Col1, Col3, Col4, Col2, Col5, Col6, Col7",0)}
```
Add this query to the first data row and column (4:A).

2. Run the management command to process the sheet:
`./manage.py process_refund_requests -i "<spreadsheet id>"`. This should update the "Order Completed Date" column for the row you added.


sheets/management/commands/process_refund_requests.py
```diff
@@ -0,0 +1,33 @@
+"""
+Parses refund request row(s), reverses orders/enrollments, and updates the spreadsheet
+to reflect the processed request(s).
+"""
+from django.core.management import BaseCommand
+
+from mitol.google_sheets_refunds.api import RefundRequestHandler
+
+
+class Command(BaseCommand):
```
rhysyngsun, line 10: This command should probably be in `ol-django`, but for the sake of getting this merged today let's go ahead with this as-is and refactor in a follow up.

sheets/plugins.py
```diff
@@ -0,0 +1,93 @@
...
+    deactivate_program_enrollment,
+    deactivate_run_enrollment,
+)
+from courses.constants import ENROLL_CHANGE_STATUS_REFUNDED
+from courses.models import ProgramEnrollment, CourseRunEnrollment
+from ecommerce.models import Order
+from users.models import User
+
+
+class RefundPlugin:
```
rhysyngsun, line 17: Could you write some tests around this?

rhysyngsun (changes requested):
Code review

sheets/plugins.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        if isinstance(enrollment, ProgramEnrollment):
+            deactivated_enrollment, _ = deactivate_program_enrollment(
+                enrollment, change_status=ENROLL_CHANGE_STATUS_REFUNDED
+            )
+        else:
+            deactivated_enrollment = deactivate_run_enrollment(
+                enrollment, change_status=ENROLL_CHANGE_STATUS_REFUNDED
+            )
+        # When #1838 is completed, this logic can be removed
```
arslanashraf7, line 60: This comment might be invalid.

sheets/plugins.py
```diff
@@ -0,0 +1,93 @@
...
+            enrollment (ProgramEnrollment or CourseRunEnrollment):
+        """
+        if isinstance(enrollment, ProgramEnrollment):
+            deactivated_enrollment, _ = deactivate_program_enrollment(
+                enrollment, change_status=ENROLL_CHANGE_STATUS_REFUNDED
+            )
+        else:
+            deactivated_enrollment = deactivate_run_enrollment(
+                enrollment, change_status=ENROLL_CHANGE_STATUS_REFUNDED
+            )
```
arslanashraf7, line 59: @rhysyngsun I don't have much context on sheets, And since you are reviewing, shouldn't we call [refund_order](https://github.com/mitodl/mitxonline/blob/main/ecommerce/api.py#L180) from ecommerce api, I believe that will call CyberSource and handle the refunding state of the Order in the database and return a bool value for success/failed refuned.

rhysyngsun: @arslanashraf7 thanks, great catch, @annagav I think this means at least for now that this entire method would be replace by that call.

rhysyngsun (changes requested):

sheets/plugins.py
```diff
@@ -0,0 +1,20 @@
...
+        self, refund_request_row: RefundRequestRow
+    ) -> RefundResult:
+
+        refund_api_success = refund_order(
+            order_id=refund_request_row.order_id, unenroll=True
+        )
+        if refund_api_success:
+            return RefundResult(ResultType.PROCESSED.value)
+
+        return RefundResult(ResultType.FAILED.value)
```
rhysyngsun, line 20: I think you need to remove `.value` here since the code calling this expects the enum property instead of the `str` value.

rhysyngsun (changes requested):

rhysyngsun (approved):
LGTM :+1:

PR #733, opened 2022-07-22
Update refine docs to conform to poorly formatted URL

collinpreston:
#### Pre-Flight checklist
- [x] Testing

#### What's this PR do?
Updates the refine admin configuration document to use the poorly formatted `//` in the redirect URL which acts as a workaround for a bug.  The bug should be fixed at some point, and this PR should be reverted after the fix is released.

Negative test - Set the redirect URL without the `//`
Positive test - Set the redirect URL with the `//` shown in this PR.


rachellougee (approved):
Thanks for the fix. Looks good to me

jkachel (approved):
LGTM 👍

Issue #678, opened 2022-07-07
Flexible Pricing: flexible pricing requests should be applied to programs instead of courses


rhysyngsun:
Summary
---

Right now flexible pricing is using the course that the learner has linked from, we'd rather this be program-based so that learners don't have to apply to every course because this would be burdensome and confusing.

We also need a way to apply program-level discounts to the courses. This issue is blocked until we determine how to do that.

#### Acceptance Criteria

- [ ] The flexible pricing page should probably be hosted under the program instead
- [ ] The flexible pricing link on the course page should still link to this page
- [ ] When the learner submits their request the program should be associated with that program
- [ ] When the request is approved the learner should get a discount on the program

pdpinch:
Is there a particular issue blocking this?

jkachel:
I don't think there's an issue that's blocking this (at this point in time) but there's some thought that needs to go into how this will be implemented. This is my thinking on the subject:

The link between a flexible pricing request (a `FlexiblePricingFormSubmission`) and the courseware object that it is associated to is weak. There's not an FK in there for this; it just has a helper function to figure out what the parent page is and grabs the courseware object it is associated with. (This is why the flexible pricing form itself needs to be a child page of the course - it's how the `FlexiblePrice` model figures out what courseware object it needs to be associated with.) That's only courses at the moment as we don't have a way to add a page for a program into the CMS. Expanding that to work with programs can be done in one of two ways:
1. We can add program pages into the CMS (I'd think in the same way the course pages are done up now). Then, the Flexible Pricing request forms can be updated slightly to allow having a program page as a parent. Any Flexible Pricing request form that's created underneath a program's page, then, would be able to grab the program as the courseware object for submitted requests. 
2. We can add a dropdown to set the courseware object for the Flexible Pricing request forms, similar to how the courses are set for course pages. Then, the form itself would be explicitly linked to the courseware object whether or not it's a program or a course. 
Doing both here is also a viable option - I'm thinking we need program pages in general anyway so the question is really whether or not we also do the second one. 

Flexible pricing tiers and products can be established now for programs (or program runs, in the case of products - we only sell runs). So, they're already set up right. We would need to amend the code that applies flexible pricing discounts to consider a couple new scenarios:
1. If the product being purchased is a program run, then it should look for tiers that are associated with the underlying program. (Right now it's hard-coded to look for courses.) 
[...]

PR #735, opened 2022-07-22
Adds program pages to the CMS

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

#678

#### What's this PR do?

Adds program pages to the CMS. These are essentially course pages, just with programs instead. (This was initially done to provide something for Flexible Price requests to attach and because we'd probably need marketing pages for programs anyway. Things have changed here but we'll probably need marketing pages for programs at some point so this code might as well be in there.) This should work - you should be able to create a program page - but it's not really exposed in the frontend.   

You should be able to create a program page in the CMS. You will need to create a program index page first (there's generally already a course index in place). The program page form should be the same as a course page form, other than the courseware object dropdown (the first field) should just list programs instead.

There's some routing issues so you can't view the pages on the front end in this PR. (That was about where I stopped.) This would just use the existing course templates but there'll likely be separate UI for program marketing pages, so not being able to load the program on the site yet is really sort of a feature rather than a bug.


jkachel:
Closing; redid this entirely due to cruft issues in https://github.com/mitodl/mitxonline/pull/797

Issue #676, opened 2022-07-07
Ecommerce: don't allow user to pay for the same course run twice 


pdpinch:
### Steps to Reproduce

I haven't been able to reproduce this with my own user, but you can see it on RC if you masquerade as the user `ddoyle`. 

They also three receipts for the same course run

Receipt 1: https://rc.mitxonline.mit.edu/orders/receipt/172/
Receipt 2: https://rc.mitxonline.mit.edu/orders/receipt/285/
Receipt 3: https://rc.mitxonline.mit.edu/orders/receipt/292/

### Expected Behavior

- user should not be able to pay for the same course run twice. Perhaps there is an edge case where a user gets a refund and then wants to pay again, but that would require customer service intervention. 
- Normally we shouldn't even be displaying buttons to allow the user to pay for the same course twice. But we should also prevent this from having in the checkout page, in case the user somehow circumvents the UI (for example, by using a direct URL). 

### Actual Behavior

- Somehow this user paid for the same course run 3 times.

rhysyngsun:
@pdpinch what are you thinking in terms of a refund and a customer service intervention? To fix this we'll need to filter out orders in certain states (e.g. pending, canceled, failures). The safest way to go about this is to filter to the statuses we want to include when determining if a course run is paid for.

pdpinch:
This user is in a bad state: they have paid and don't have anything to show for it. They need to be either:
1. refunded and unenrolled in mitxonline
2. manually enrolled as verified in open edX. 

> To fix this we'll need to filter out orders in certain states (e.g. pending, canceled, failures). The safest way to go about this is to filter to the statuses we want to include when determining if a course run is paid for.

I don't understand what you mean by this☝️. All 3 orders are "fulfilled"

rhysyngsun:
@pdpinch sorry, i think there's a couple of things here that I might've been muddling together:

##### The specific user issue on RC
This is currently a data issue I think, so repairing that isn't necessarily a thing we can do with a code change. If you want to fix the data I'd suggest:
- Fix the state to have the learner verified:
  - The enrollment dates need to be changed to a future date (if we can't do this, the other option is required)
  - Mark all but one order as refunded
  - Update the learner's `CourseRunEnrollment` record setting `edx_enrolled=False` and `enrollment_mode="verifed" (this will instruct MITx Online to set the enrollment to the correct mode in openedx)
- Update the learner's state to reflect their current `audit` 
[...]

PR #737, opened 2022-07-25
add validation to block payment for the paid courserun

rachellougee:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/676

#### What's this PR do?
Adding validation on checkout to payment view to see if user has already purchased for that course run, displaying a error message if that happens

- enroll a course on the frontend, but leave it in the cart
- go to /admin/ecommerce/order/ and find this order, manually change it to 'Fullfilled'
-  go back to frontend to enroll this course again, when you click on place your order, you should see
image

(Optional)

(Optional)

(Optional)

### DIFF
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -80,6 +80,25 @@ class Basket(TimestampedModel):
             ]
         )
 
+    def has_user_purchased_same_courserun(self, user):
+        """
+        Return true if any of the courses in the basket has already purchased
+        """
+        basket_items = self.basket_items.prefetch_related("product")
+        for item in basket_items:
+            purchased_object = item.product.purchasable_object
+            if isinstance(purchased_object, CourseRun):
+                # PaidCourseRun should only contain fulfilled or review orders
+                # but in order to avoid false positive passing in order__state__in here
+                if PaidCourseRun.objects.filter(
+                    user=user,
+                    course_run=purchased_object,
+                    order__state__in=[Order.STATE.FULFILLED, Order.STATE.REVIEW],
+                ).exists():
+                    return True
+
+        return False
+
     def compare_to_order(self, order):
         """
         Compares this basket with the specified order. An order is considered
@@ -363,7 +382,6 @@ class FulfillableOrder:
         target=Order.STATE.FULFILLED,
     )
     def fulfill(self, payment_data):
-
         # record the transaction
         self.create_transaction(payment_data)
 
--- a/frontend/public/src/lib/notificationsApi.js
+++ b/frontend/public/src/lib/notificationsApi.js
@@ -7,6 +7,7 @@ import {
   USER_MSG_COOKIE_NAME,
   USER_MSG_TYPE_COMPLETED_AUTH,
   USER_MSG_TYPE_ENROLL_BLOCKED,
+  USER_MSG_TYPE_ENROLL_DUPLICATED,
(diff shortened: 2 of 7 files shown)


ecommerce/models.py
```diff
@@ -81,6 +81,28 @@ def has_user_blocked_products(self, user):
             ]
         )
 
+    def has_user_purchased_same_courserun(self, user):
+        """
+            Return true if any of the courses in the basket has already purchased
+            meaning user has orders for that run in a FULFILLED or REVIEW state
+         """
+        purchased_orders = Order.objects.filter(purchaser=user).filter(state__in=[Order.STATE.FULFILLED, Order.STATE.REVIEW])
```
jkachel, line 89: This works but I think there's a better way to do this. If the user's successfully paid for and enrolled in the course, there should be a CourseRunEnrollment that is in the `verified` state. Can this check for that instead? This would get a few cases where checking for a completed order might be a bit problematic. (Thinking specifically of the case where a learner may have been refunded for an order for a course run but we're not changing their enrollment status - in that case, they probably shouldn't be able to pay for the course run again.)

jkachel (changes requested):
Overall structure looks good but had a question about the algorithm.

ecommerce/models.py
rachellougee: @jkachel Thanks for the feedback. Do you mean that it should check for CourseRunEnrollment.enrollment_mode = 'verified' instead of ecommerce_order.status in ('fulfilled', 'reviewed')? That works for my test enrollment on RC (I can see mine has enrollment_mode = 'verified'), but when I checked the 3 orders peter linked in the ticket, they all have 'audit' state, not verified. Is this just data issues for these 3 orders?
image
I can update the logic, just wanted to confirm.

rhysyngsun: Like Rachel mentioned, we have some situations that have come up where someone has paid but isn't enrolled as verified. That's what we're trying to avoid here so that if this happens a learner doesn't pay repeatedly and will instead contact support.

rachellougee:
Per discussion, I will pause this PR until https://github.com/mitodl/mitxonline/issues/738 is done to simplify the queries from Order to CourserRunEnrollment

rachellougee:
@jkachel I've updated PR to use the new table PaidCourseRun and added a test. Please take a look when you get a chance. Thank you

rachellougee:
@pdpinch Not sure if you would like to take a look before I merge this PR. To manually test as user ddoyle, it would needs the backfill migration to be ran on RC https://github.com/mitodl/mitxonline/pull/760/

frontend/public/src/lib/notificationsApi.js
```diff
@@ -55,6 +56,10 @@ export function parseStoredUserMessage(
     msgText =
         "We're sorry, your country is currently blocked from enrolling in this course."
     break
+  case USER_MSG_TYPE_ENROLL_DUPLICATED:
+    alertType = ALERT_TYPE_DANGER
+    msgText = "You have already enrolled and paid for this course."
```
pdpinch, line 61: Please include an email address in the message. 

"You have already enrolled and paid for this course. If this is unexpected, please contact customer support at {support_email}"

rachellougee: @pdpinch I've updated the message

Issue #738, opened 2022-07-25
Orders: Add a tracking table for course run purchases


rhysyngsun:
We need to be able to have a simpler way of determining which runs a learner has purchased. Currently you have to go through `User` -> `Order` -> `Line` -> `Product` -> `CourseRun`, which makes querying inefficient and messy.

I've proposed a join table that would track information on what course runs have been paid for, something like this:

```python
class PaidCourseRun(TimestampedModel):
    user = models.ForeignKey(
        User, on_delete=models.CASCADE, related_name="paid_course_runs"
    )
    course_run = models.ForeignKey(
        CourseRun, on_delete=models.CASCADE, related_name="paid_course_runs"
    )
    order = models.ForeignKey(
        Order, on_delete=models.CASCADE, related_name="paid_course_runs"
    )
```

This would simplify the joins down to `User` -> `PaidCourseRun` -> `CourseRun`.

We would only create these for orders in the `FULFILLED` or `REVIEW` state and keep the records in the event of a refund.

#### Tasks

- [x] Add a model similar to the above psuedocode
- [x] Where we currently mark an order as fulfilled or in review state, add code to create and/or update a record
- [ ] In a PR that is merged and deployed after the above logic change is deployed add a migration that backpopulates records for existing orders

dseaton:
@rhysyngsun @rachellougee I have confirmed that this works for me. Thank you!

PR #745, opened 2022-07-27
add a tracking table for course run purchases

rachellougee:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/738)

#### What's this PR do?
 Add a new table to track when order is fulfilled or in review state. see https://github.com/mitodl/mitxonline/issues/738 for detail
 The insert logic is implemented in the following places
- When order is transition from pending or review to fulfilled
- When order is transition from pending to review

Note that 
- If order is manually updated to fulfilled or review from admin directly, It would not work.
- No delete is implemented

enroll a course from frontend, it should display this if order is paid successful
image
Go to admin/courses/paidcourserun/, you should see the fulfilled order shows up there
image

(Optional)

(Optional)

(Optional)


ecommerce/api.py
```diff
@@ -164,6 +167,21 @@ def fulfill_completed_order(
...
+    """
+    order = Order.objects.get(id=order.id)
+    if order.STATE.FULFILLED or order.STATE.REVIEW:
+        for run in order.purchased_runs:
+            if not PaidCourseRun.objects.filter(
+                order=order, course_run=run, user=order.purchaser
+            ).exists():
+                PaidCourseRun.objects.create(
+                    order=order, course_run=run, user=order.purchaser
+                )
```
jkachel, line 182: Small change: there's a `update_or_create()` [function](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#update-or-create) that could be used here. (Or, for that matter, `get_or_create()` if you don't want it to update the timestamps.)

jkachel (changes requested):
One other change: can you add in a test for this? There probably doesn't need to be an entirely new test - in `ecommerce/views_test.py` there's an existing `test_checkout_result` that should also check for a valid `PaidCourseRun` record. Looks good otherwise.

rachellougee:
@jkachel I've addressed your feedback and added the test. I added some extra commits when tried to resolve merge conflict, but the file changed should reflect the code change in this PR. Please let me know if any additional feedback. Thank you

jkachel (approved):
LGTM 👍

ecommerce/api.py
```diff
@@ -188,6 +189,8 @@ def fulfill_completed_order(
     order.fulfill(payment_data)
     order.save()
 
+    add_fulfilled_order_to_paid_courserun(order)
```
rhysyngsun, line 192: I'd suggest this should be added to `Order.fulfill()` so that we can be sure if that's called from anywhere else that this logic is always executed.

ecommerce/views/v0/__init__.py
```diff
@@ -508,6 +508,9 @@ def post(self, request, *args, **kwargs):
             order.review(payment_data)
             order.save()
 
+            # order is in review state, record it as paid courserun
+            api.add_fulfilled_order_to_paid_courserun(order)
```
rhysyngsun, line 512: Similar to above, I'd suggest this should be added to `Order.review()` .

rhysyngsun (changes requested):

ecommerce/api.py
```diff
@@ -200,6 +203,18 @@ def fulfill_completed_order(
...
+def add_fulfilled_order_to_paid_courserun(order):
+    """
+    When order is fulfilled or reviewed, record it to paid courserun table if not exist
+    """
+    order = Order.objects.get(id=order.id)
+    if order.STATE.FULFILLED or order.STATE.REVIEW:
+        for run in order.purchased_runs:
+            PaidCourseRun.objects.get_or_create(
+                order=order, course_run=run, user=order.purchaser
+            )
```
rhysyngsun, line 215: This could probably be a method on `Order` so it can just use `self`. I'd also say we already have a property `is_review` so we should use that and also add `is_fulfilled`.

Probably worth adding ones for the other states too and updating any usages, but that's probably best as a separate improvement issue.

rhysyngsun (changes requested):

rachellougee:
@rhysyngsun I've moved this logic to the following places, It's also where we create enrollments, It makes sense to keep it there.

- PendingOrder.fulfill()
- PendingOrder.review()
- ReviewOrder.fulfill()

I had trouble adding the logic to Order model directly due to state issues as it's need to wait until order is saved per discussed.
I ran the pytest after the code changes. Please let me know if any other feedback. Thanks

ecommerce/models.py
```diff
@@ -495,6 +510,12 @@ def fulfill(self, payment_data):
             mode=EDX_ENROLLMENT_VERIFIED_MODE,
         )
 
+        # record all the courserun in the order
+        for run in self.purchased_runs:
```
rhysyngsun, line 514: A few things we could improve overall:

 - Move this code above the `create_run_enrollments` call so that if that API call to openedx fails for some reason we don't get in an inconsistent state.
 - Could you create an abstract base class that subclasses `Order` and that this proxy class would the subclass?
   - Move the `self.transaction.create` and `PaidCourseRun.objects.get_or_create` into their own dedicated functions there
   - Move the `create_run_enrollments()` call into its own dedicated function
   - Move the implementation of `fulfill()` there as well so we don't have duplicate code

ecommerce/api.py
```diff
@@ -29,6 +29,7 @@
 
 from mitol.common.utils.datetime import now_in_utc
 from ipware import get_client_ip
+from courses.models import PaidCourseRun
```
rhysyngsun, line 32: Unused import?

rhysyngsun (changes requested):

rachellougee:
@jkachel Do you mind taking another look based on @rhysyngsun 's new feedback? The changes now have new class to reduce code duplication for existing code. Thanks

jkachel (approved):
Needs a two character change - then good to merge. LGTM 👍

Issue #732, opened 2022-07-21
don't allow products to be deleted


pdpinch:
On RC, it seems that a couple of product records were deleted, causing a data inconsistency problem with orders that were still related to the product. 

We shouldn't allow products to be deleted, at least not through the admin. If a product needs to be removed, it should be deactivated somehow.

dseaton:
Related issue: https://github.com/mitodl/ol-data-platform/issues/284

dseaton:
@pdpinch Can we close this issue?

dseaton:
I would like to add that we need to fix/remove the duplication in the ecommerce_product table for MITx Online RC. 
@pdpinch

pdpinch:
@arslanashraf7 can this be closed?

PR #753, opened 2022-07-28
fix: protect Product model from deletion

arslanashraf7:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

#732

#### What's this PR do?
- Disables the delete option for the products model through Admin
- Introduce Active Manager for products
- Introduce (Disable delete through Django Shell too)
- Adds db level protection for the Product model
- The deletion through Django would result in marking the product as in-active.

- You should not be able to delete product objects through Django Admin
- You should not be able to delete Products through Django Shell via `Product(instance).delete()` or `Product.objects.all().delete()` - try with all/filter
- Upon deleting a product it should be marked as inactive instead
- Go to db shell `./manage.py dbshell` and try to delete a product with SQL e.g. (`delete from ecommerce_product where id=1;`, the record should not be deleted)

(Optional)

(Optional)


rhysyngsun:
@arslanashraf7 could you also add postgres-level delete protections? Here's where we did it in xPro: https://github.com/mitodl/mitxpro/blob/master/ecommerce/migrations/0015_db_protect.py#L33-L35

rhysyngsun (approved):

Issue #734, opened 2022-07-22
Registration: validate username against openedx


rhysyngsun:
We've seen instances of username conflicts due to an undetermined character set conversion, so we should put some code in place to validate the inputs:

Openedx has an API call to validate the input:

https://courses.edx.org/api-docs/#/user/user_v1_validation_registration_create

When the user submits the form after confirming their email we should make this validation call. We're only concerned about `username` at this time. I'm not sure if other fields are required, but we should ignore error output from them for now.

collinpreston:
@rhysyngsun, what should occur if there if the validation fails (there is a matching username in openedx)?

pdpinch:
Do we have an existing method to display an error message and ask the user to choose something else?

collinpreston:
@pdpinch, yes
![Screen Shot 2022-07-28 at 12 30 09 PM](https://user-images.githubusercontent.com/8311573/181590189-98df2232-863a-4359-8074-31e34d9f4799.png)

collinpreston:
@rhysyngsun @pdpinch , what should occur if `mitxonline` cannot connect to the `openedx` API endpoint?

pdpinch:
The screenshot looks great. If you can make that happen, with that message, that will be good.

collinpreston:
@pdpinch I've set it up so that if `mitxonline` cannot connect to `openedx`, the user account will be created on `mitxonline` and the user will not receive any sort of message.  This is not the same as if `openedx` responds with a non-200 HTTP code, then the user will receive an error message.

I believe this follows suit with our current workflow for which if `mitxonline` is unable to connect to `openedx` during the user registration process, the user is still created in `mitxonline` and allowed to proceed.

pdpinch:
> @pdpinch <https://github.com/pdpinch> I've set it up so that if mitxonline cannot connect to openedx, the user account will be created on mitxonline and the user will not receive any sort of message. This is not the same as if openedx responds with a non-200 HTTP code, then the user will receive an error message.

This makes sense to me. It gives us some time to detect and fix an failure in openedx. It runs the risk of username collisions during the downtime, but we can fix that up manually if/when it comes to it.

PR #757, opened 2022-07-28
734 - registration validate username against openedx

collinpreston:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/734

#### What's this PR do?
Restricts a new `mitxonline` user from registering with a username that is currently taken by an OpenEdx user.  If a new `mitxonline` user attempts to use a username which is already taken by an OpenEdx user, they will be unable to submit the user registration form and presented with an error message.

1. Create an OpenEdx user with a known username.
2. Register a new user through `mitxonline` and attempt to use the same username as in Step 1.
3. Verify that you are unable to create a new user with the username from Step 1, and that you receive an error message.

If you are running `mitxonline` and `openedx` locally, you can also test out the workflow when `mitxonline` cannot connect to `openedx` by only running `mitxonline` locally.  You should perform the same steps listed above, but you will not receive an error message and you should be allowed to successfully submit the registration form.

- There is a new method added to the `openedx` API service which handles making the request to the `openedx` registration verification service.
- I have added a call to the `openedx` API service method to the User Serializer username validation.
- There is a custom exception that is thrown if the `openedx` API call is unsuccessful.

Screen Shot 2022-07-29 at 12 10 07 PM

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -42,6 +42,7 @@ from openedx.exceptions import (
     UserNameUpdateFailedException,
     EdxApiEmailSettingsErrorException,
     UnknownEdxApiEmailSettingsException,
+    EdxApiRegistrationValidationException,
 )
 from openedx.models import OpenEdxApiAuth, OpenEdxUser
 from openedx.utils import edx_url, SyncResult
@@ -50,6 +51,7 @@ log = logging.getLogger(__name__)
 User = get_user_model()
 
 OPENEDX_REGISTER_USER_PATH = "/user_api/v1/account/registration/"
+OPENEDX_VALIDATION_REGISTRATION_PATH = "/api/user/v1/validation/registration"
 OPENEDX_REQUEST_DEFAULTS = dict(country="US", honor_code=True)
 
 OPENEDX_SOCIAL_LOGIN_XPRO_PATH = "/auth/login/mitxpro-oauth2/?auth_entry=login"
@@ -712,3 +714,32 @@ def unsubscribe_from_edx_course_emails(user, course_run):
     except Exception as exc:
         raise UnknownEdxApiEmailSettingsException(user, course_run, exc) from exc
     return result
+
+def username_exists_in_openedx(username):
+    """
+    Returns true if the username exists within Open edX.
+    Returns false if the username does not exists in Open edX.
+
+    Args:
+        username (str): the username
+        
+    Raises:
+        EdxApiRegistrationValidationException: Raised if the edX API HTTP request fails
+    """
+    
+    req_session = requests.Session()
+    req_session.headers.update(
+                {
+                    ACCESS_TOKEN_HEADER_NAME: settings.MITX_ONLINE_REGISTRATION_ACCESS_TOKEN
+                }
+            )
+    resp = req_session.post(
+        edx_url(OPENEDX_VALIDATION_REGISTRATION_PATH),
+        data=dict(
+            username=username,
+        ),
+    )
+    if resp.status_code != status.HTTP_200_OK:
+            raise EdxApiRegistrationValidationException(username, resp)
+    result = resp.json()
+    return result['validation_decisions']['username'] == "It looks like {} belongs to an existing account. Try again with a different username.".format(username) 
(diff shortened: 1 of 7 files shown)


openedx/api.py
```diff
@@ -712,3 +714,35 @@ def unsubscribe_from_edx_course_emails(user, course_run):
...
+        data=dict(
+            username=username,
+        ),
+    )
+    if resp.status_code != status.HTTP_200_OK:
+        raise EdxApiRegistrationValidationException(username, resp)
+    result = resp.json()
+    return result["validation_decisions"][
+        "username"
+    ] == "It looks like {} belongs to an existing account. Try again with a different username.".format(
```
arslanashraf7, line 746: In my opinion, we should keep this message the same for the username validation error. There is [an existing message](https://github.com/mitodl/mitxonline/blob/main/users/serializers.py#L98) when the username exists. 

Personally, I like this string better, Maybe we can move this to a constant and use it in both places.

openedx/api_test.py
```diff
@@ -93,6 +96,12 @@ def test_create_edx_user(user, settings, application, access_token_count):
         json=dict(success=True),
         status=status.HTTP_200_OK,
     )
+    responses.add(
+        responses.POST,
+        settings.OPENEDX_API_BASE_URL + OPENEDX_VALIDATION_REGISTRATION_PATH,
+        json={"validation_decisions": {"username": ""}},
+        status=status.HTTP_200_OK,
+    )
```
arslanashraf7, line 104: We are adding the same thing in responses, right here and [below](https://github.com/mitodl/mitxonline/pull/757/files#diff-353bcf6e9efd054b7874f0b78245634f9b837e421e1cd56e457016f4c2763e95R149-R154). Is this duplicate or intentional, Plus if we need the same responses additions it might be better to go for a helper mock function that can be reused.

I think there is the same case [here](https://github.com/mitodl/mitxonline/pull/757/files#diff-353bcf6e9efd054b7874f0b78245634f9b837e421e1cd56e457016f4c2763e95R165-R174) too.

users/serializers.py
```diff
@@ -31,6 +36,9 @@
 USERNAME_RE_PARTIAL = r"[\w ._+-]+"
 USERNAME_RE = re.compile(rf"(?P<username>{USERNAME_RE_PARTIAL})")
 USERNAME_ERROR_MSG = "Username can only contain letters, numbers, spaces, and the following characters: _+-"
+USERNAME_ALREADY_EXISTS_MSG = (
```
arslanashraf7, line 39: Oh, you've already created one, I have a comment above where I mentioned creating a constant, maybe this is something we can reuse.

arslanashraf7 (changes requested):

openedx/api.py
collinpreston: This is the string which comes in the response from Edx when there is an existing username found.  So this message here doesn't actually end up going anywhere, it's just used for checking Edx response.

users/serializers.py
```diff
@@ -110,6 +118,20 @@ def validate_email(self, value):
...
+            HTTPError,
+            RequestsConnectionError,
+            EdxApiRegistrationValidationException,
+        ):
+            log.exception(
+                "edX username verification failure for username: %s",
+                trimmed_value,
+            )
+        if openedx_username_taken:
+            raise serializers.ValidationError(USERNAME_ALREADY_EXISTS_MSG)
```
arslanashraf7, line 134: I realized this only when I was testing the PR, I think this block of code can/should be moved after we've validated the username against the regex that is [below](https://github.com/mitodl/mitxonline/pull/757/files#diff-4349bb2be4e3bfa15b93d2cf390f4df197e2a0b11e6edfe8abd3f09beec21273R135-R136).

So, Currently, We are bearing a cost of an extra edX username validation API call even if the predefined username fails the regex validation. When we register with a username e.g. `arslan@` (Invalid as per regex) we send a call to edX and regardless of edX API result we then validate the regex and if it fails, we send an error message to frontend. So, Changing the order will make the logic better and save an API call.

arslanashraf7 (changes requested):
Just a small change, It's basically a change of order in how we are validating the username.

arslanashraf7 (approved):
@collinpreston This looks great 👍, Thanks for making all the changes.

I was able to test both the mentioned cases too.

Issue #701, opened 2022-07-15
Story: legacy MM users course enrollment on dashboard


pdpinch:
As a returning DEDP MicroMasters learner, I can see my past edx.org course enrollment on my dashboard. 

For example, on RC my user has enrollments in 14.100x+2T2017. I expect to see that enrollment on my dashboard on mitxonline as well. 

#### Designs and Mockups

no change

#### requires the following MM data:

- [ ] Course runs
- [ ] Course run enrollments

| MicroMasters (Model/Table) | MITx Online (Model/Table) |
| --- | --- |
| `CourseRun` / `courses_courserun` | `CourseRun` / `courses_courserun` 
| `CachedEnrollment` / `dashboard_cachedenrollment` | `CourseRunEnrollment` / `courses_courserunenrollment`

These should be separate SQl queries and will need to be run in the sequence above. 

We'll join on those social auth `uid as username` for determining the user.

- `CourseRun`: the `edx_course_key` / `courseware_id`
- `CachedEnrollment`: the combination of `user_id` and `run_id`

Note: the `Course` data model is going to be manually generated on our systems (they should already be in place). These will get mapped.

#### Out of Scope

- grades and certificates are not required for this issue

rachellougee:
@rhysyngsun @pdpinch I am not sure the best place to put up the queries for review, but I've saved it on BI https://bi.odl.mit.edu/queries/1161/source#2017 
I tested the courseRun/Enrollment migration from MicroMaster RC to my local MITxOnline postgres DB using postgres_fdw https://www.postgresql.org/docs/current/postgres-fdw.html per suggested.
I verified Peter's enrollments and can see 14.100x+2T2017 on my test
```----Peter's enrollment On MITxOnline after migration from MicroMaster
SELECT run.courseware_id
FROM public.courses_courserunenrollment as enroll
JOIN public.users_user AS users on enroll.user_id = users.id
JOIN public.courses_courserun as run ON enroll.run_id = run.id 
where users.username ='pdpinch';

        courseware_id          
--------------------------------
 course-v1:MITx+14.100x+2T2017
 course-v1:MITx+14.310x+3T2021
 MITx/14.73x_1/1T2015
 course-v1:MITx+JPAL102x+3T2021
 course-v1:MITx+14.740x+2T2022
 course-v1:MITx+14.100x+1T2022
(6 rows)
```
[...]

PR #758, opened 2022-07-28
add migration queries for courserun and enrollment

rachellougee:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/701

#### What's this PR do?
This PR contains the queries need to run against MITxOnline to bring over DEDP courserun and enrollment data from MicroMaster

- Go to /admin/micromasters_import/courseid/ to map McroMaster course id to MITxOnline course for DEDP 
- run ```docker-compose run --rm web ./manage.py import_micromasters_data ```

18 courses migrated from MicroMaster RC
image
3 enrollments migrated
image

### DIFF
--- a/micromasters_import/management/commands/queries/002_import_courserun.sql
+++ b/micromasters_import/management/commands/queries/002_import_courserun.sql
@@ -1,0 +1,38 @@
+--courserun migration from MicroMaster to MITxOnline
+INSERT INTO public.courses_courserun(
+    title,
+    courseware_id,
+    run_tag,
+    courseware_url_path,
+    start_date,
+    end_date,
+    enrollment_start,
+    enrollment_end,
+    expiration_date,
+    live,
+    created_on,
+    updated_on,
+    course_id)
+SELECT
+    mm_courserun.title,
+    mm_courserun.edx_course_key,
+    (regexp_match(mm_courserun.edx_course_key, '[^(+|/)]+$'))[1],
+    mm_courserun.enrollment_url,
+    mm_courserun.start_date,
+    mm_courserun.end_date,
+    mm_courserun.enrollment_start,
+    mm_courserun.enrollment_end,
+    mm_courserun.end_date + INTERVAL '1 day', -- expiration date must be later than end_date per validation
+    true, --live
+    coalesce(mm_courserun.enrollment_start, now()),
+    coalesce(mm_courserun.enrollment_end, now()),
+    pk_map.course_id
+FROM micromasters.courses_courserun AS mm_courserun
+JOIN public.micromasters_import_courseid AS pk_map
+  ON mm_courserun.course_id = pk_map.micromasters_id
+LEFT JOIN public.courses_courserun AS mo_courserun
+  ON mo_courserun.courseware_id = mm_courserun.edx_course_key
+WHERE mm_courserun.is_discontinued = false
+  AND mm_courserun.courseware_backend = 'edxorg'
+  AND mo_courserun.courseware_id IS NULL ---course runs that don't exist in MITxOnline
+ON CONFLICT DO NOTHING;
--- a/micromasters_import/management/commands/queries/003_import_enrollment.sql
+++ b/micromasters_import/management/commands/queries/003_import_enrollment.sql
@@ -1,0 +1,32 @@
+--courserun Enrollment migration from MicroMaster to MITxOnline
+INSERT INTO public.courses_courserunenrollment(
+    run_id,
+    user_id,
+    change_status,
+    active,
+    edx_enrolled,
+    edx_emails_subscription,
+    enrollment_mode,
+    created_on,
+    updated_on)
+SELECT
+    mo_courserun.id,
+    mo_user.id,
+    '' AS change_status,
+    (mm_enrollment.data->>'is_active')::boolean,
+    true,
+    false,
+    mm_enrollment.data->>'mode',
+    coalesce(((mm_enrollment.data->'course_details'->>'enrollment_start')::timestamp), now()),
+    coalesce(((mm_enrollment.data->'course_details'->>'enrollment_end')::timestamp), now())
+FROM micromasters.dashboard_cachedenrollment AS mm_enrollment
+JOIN micromasters.courses_courserun AS mm_courserun
+  ON mm_enrollment.course_run_id = mm_courserun.id
+JOIN micromasters.social_auth_usersocialauth AS mm_social
+  ON mm_enrollment.user_id = mm_social.user_id
+JOIN public.users_user AS mo_user
+  ON mm_social.uid = mo_user.username
+JOIN public.courses_courserun AS mo_courserun
+  ON mm_courserun.edx_course_key = mo_courserun.courseware_id
+WHERE mm_social.provider = 'mitxonline'
+ON CONFLICT DO NOTHING;


micromasters_import/management/commands/queries/002_courserun_enrollment.sql
```diff
@@ -0,0 +1,83 @@
...
+
+--- create a foreigner server to connect to external MicroMaster DB
+CREATE SERVER IF NOT EXISTS micromaster_bridge
+	FOREIGN DATA WRAPPER postgres_fdw
+    OPTIONS (host '--------', dbname '--------');
+
+-- setup a user mapping to authenticate:
+CREATE USER MAPPING IF NOT EXISTS FOR postgres
+	SERVER micromaster_bridge
+    OPTIONS (user '-------', password '------');
```
rhysyngsun, line 12: I think you can remove these queries - the db user won't have permission to do these kinds of operations and it's going to be something devops sets up ahead of time anyway. Locally it'll need to be run manually or as part of an init script.

rachellougee: oh ok. Good to know. I will remove that

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
```diff
@@ -0,0 +1,65 @@
...
+    mm_courserun.edx_course_key,
+    LEFT(CASE WHEN position('+' in mm_courserun.edx_course_key) > 0
+    THEN (regexp_match(mm_courserun.edx_course_key, '[^+]+$'))[1]
+    ELSE (regexp_match(mm_courserun.edx_course_key, '[^/]+$'))[1] END, 10) AS run_tag,
+    mm_courserun.enrollment_url,
+    mm_courserun.start_date,
+    mm_courserun.end_date,
+    mm_courserun.enrollment_start,
+    mm_courserun.enrollment_end,
+    mm_courserun.enrollment_end + INTERVAL '1 day',
```
pdpinch, line 27: If expiration dates are used to restrict access to a closed course, we should leave them blank instead of generating an expiration date. Unlike xPRO courses, where access is restricted after some time, access to mitxonline courses is perpetual.

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
```diff
@@ -0,0 +1,65 @@
...
+    NOT mm_courserun.is_discontinued,
+    coalesce(mm_courserun.enrollment_start, now()), --- need to confirm the default if enrollment_start is null
+    coalesce(mm_courserun.enrollment_end, now()),
```
pdpinch, line 30: I don't understand the logic here. Why would we use `enrollment_start` for the record creation date?

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
```diff
@@ -0,0 +1,65 @@
...
+    pk_map.course_id
+FROM micromasters.courses_courserun AS mm_courserun
+JOIN public.micromasters_import_courseid AS pk_map ON mm_courserun.course_id = pk_map.micromasters_id
+LEFT JOIN public.courses_courserun AS mo_courserun ON mo_courserun.courseware_id = mm_courserun.edx_course_key
+WHERE mo_courserun.courseware_id IS NULL ---bring in course runs that don't exist in MITxOnline
+ON CONFLICT DO NOTHING; -- Do we want to update?
```
pdpinch, line 36: If a courserun already exists in mitxonline, we don't want to update it with data from micromasters

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
```diff
@@ -0,0 +1,65 @@
...
+SELECT
+    mo_courserun.id,
+    mo_user.id,
+    '' as change_status,
+    (mm_enrollment.data->>'is_active')::boolean,
+    TRUE,
+    FALSE,
+    mm_enrollment.data->>'mode',
+    ((mm_enrollment.data->>'created')::timestamp), -- need to verify
+    ((mm_enrollment.data->>'created')::timestamp)
```
pdpinch, line 58: created, or updated?

pdpinch (changes requested):
I left some specific comments, but I have a couple of general questions:

1. At this time, we only _need_ to import courseruns and enrollment data for DEDP (program 2). I don't oppose bringing in additional data from other programs, but I want to make sure it doesn't have any side effects.

2. How will the imported courseruns be associated with existing courses? All the DEDP courseruns should have correspoding course records on mitxonline. But the other programs' courseruns will _not_ have courses in mitxonline. 

cc @rhysyngsun

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
rachellougee: @pdpinch I don't recall specifically why I chose enrollment_start, but created is required field so we needed a similar field to populate it

rachellougee:
@pdpinch Thanks for your feedback. 
This query would only bring in DEDP courserun and enrollment per discussion with @rhysyngsun. 
There is a new mapping table that map MicroMaster courseid to MITxOnline course id. The join to the new table limits to the existing courses
```JOIN micromaster_migration.courses_courserun AS mm_courserun ON mm_enrollment.course_run_id = mm_courserun.id```
 If we want to bring in other program, we would need to add courses in MITxOnline first if they don't exist, then add mapping, which should allow us to bring in courserun and enrollment, we will have to adjust the logic accordingly. I don't think there will be any side effects, but I can double check @rhysyngsun your thought?

File: micromasters_import/management/commands/queries/002_courserun_enrollment.sql (hunk 3 above)
Comment (rachellougee) [reply micromasters_import/management/commands/queries/002_courserun_enrollment.sql:30]: @pdpinch 
Here are available fields from MicroMaster courseruns
image
How should we populate created_on and updated_on? As you can see, there are blank for enrollment_start, start_date, enrollment_end and end_date in MicroMaster, so we need default value in case of blank for import

File: micromasters_import/management/commands/queries/002_courserun_enrollment.sql (hunk 5 above)
Comment (rachellougee) [reply micromasters_import/management/commands/queries/002_courserun_enrollment.sql:58]: @pdpinch @rhysyngsun Looks like there isn't updated field in McroMaster enrollment table
There are also some enrollments on RC that have blank 'created'. For those cases, how should we populate created_on and updated_on ?
image
Also, just to confirm that we are not filtering out any enrollments as long as mm_social.provider = 'mitxonline'?

File: micromasters_import/management/commands/queries/002_courserun_enrollment.sql [hunk 6]
```diff
@@ -0,0 +1,65 @@
...
+    mm_courserun.enrollment_end,
+    mm_courserun.enrollment_end + INTERVAL '1 day',
+    NOT mm_courserun.is_discontinued,
+    coalesce(mm_courserun.enrollment_start, now()), --- need to confirm the default if enrollment_start is null
+    coalesce(mm_courserun.enrollment_end, now()),
+    pk_map.course_id
+FROM micromasters.courses_courserun AS mm_courserun
+JOIN public.micromasters_import_courseid AS pk_map ON mm_courserun.course_id = pk_map.micromasters_id
+LEFT JOIN public.courses_courserun AS mo_courserun ON mo_courserun.courseware_id = mm_courserun.edx_course_key
+WHERE mo_courserun.courseware_id IS NULL ---bring in course runs that don't exist in MITxOnline
```
rachellougee, line 35: Do we need to add a where clause to only bring in active course runs? (mm_courserun.is_discontinued = 1). 
https://micromasters-rc.odl.mit.edu/admin/courses/courserun/?q=course%3A Peter marked these courseruns inactive, but these would still be imported if we don't filter out

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
```diff
@@ -0,0 +1,65 @@
...
+    LEFT(CASE WHEN position('+' in mm_courserun.edx_course_key) > 0
+    THEN (regexp_match(mm_courserun.edx_course_key, '[^+]+$'))[1]
+    ELSE (regexp_match(mm_courserun.edx_course_key, '[^/]+$'))[1] END, 10) AS run_tag,
+    mm_courserun.enrollment_url,
+    mm_courserun.start_date,
+    mm_courserun.end_date,
+    mm_courserun.enrollment_start,
+    mm_courserun.enrollment_end,
+    mm_courserun.enrollment_end + INTERVAL '1 day',
+    NOT mm_courserun.is_discontinued,
```
rhysyngsun, line 28: So we actually want this to be `true`, because `live` indicates whether or not the run shows up in the interface. It won't be enrollable because of the enrollment dates.

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
rhysyngsun: I believe `expiration_date` is used to determine when we stop syncing course run data from mitxonline. None of the production runs have this set AFAIK, but we don't want the app trying to query openedx for runs that only exist on edx.org.

Given that I think the current expression makes sense.

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
rhysyngsun: I think this was a mapping we decided on in a call @rachellougee, and as you said, they're more or less arbitrary, I don't believe we use them in any user-facing logic or UI. The only place these values would show up is `django-admin`.

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
rhysyngsun: You want to only bring in the discontinued runs, because ones that aren't discontinued are the mitxonline ones, which are already over there with correct data. So you should filter to `mm_courserun.is_discontinued  = true`

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
```diff
@@ -0,0 +1,65 @@
...
+    active,
+    edx_enrolled,
+    edx_emails_subscription,
+    enrollment_mode,
+    created_on,
+    updated_on)
+SELECT
+    mo_courserun.id,
+    mo_user.id,
+    '' as change_status,
```
rhysyngsun, line 52: This value I'm unsure of, @pdpinch (how) do we handle deferrals and transfers in MicroMasters?

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
rhysyngsun: I think we should filter the enrollments just down to DEDP ones (we'll also need to do this to orders). My thinking is those other programs only use edx.org and I think there could be a lot of learner confusion seeing enrollments over in mitxonline that may not be current and not integrated with edx.org. @pdpinch do you agree?

As far as the two fields, we might just have to use `enrollment_start` and `enrollment_end` as best guesses, similar to my comment about course runs.

rhysyngsun (changes requested):
Could you also split the two operations here into separate files to delineate that they're doing two different things?

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
rachellougee: @rhysyngsun My thinking is that the new table 'micromasters_import_courseid' will only be DEDP courses mapping if we just want to bring in DEDP courserun and enrollment for now, so we don't have to add any hardcoded filter here. Also the JOIN public.courses_courserun here should be enough to limit the enrollments down to DEDP. If we want to be sure, I would add a filter to courserun query

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
pdpinch: I see. Ok to keep as is.

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
pdpinch: Ok

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
pdpinch: I agree with the goal of only importing DEDP enrollments. I won't comment on the implementation.

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
pdpinch: Nathan and I discussed this about an hour ago, and I think we agreed that we should only pull in courses where the `courseware_backend = 'edxorg' AND is_discontinued = false` 

The `is_discontinued` flag was used to omit mistakenly-created duplicate courses in micromasters that exist in mitxonline. 

Obviously we don't need to import courseruns that have a `courseware_backend = mitxonline`.

rachellougee: @pdpinch Just to confirm if we add ```courseware_backend = 'edxorg' AND is_discontinued = false```, It would only bring in 'MITx/14.73x_1/1T2015' enrollment for your account as all other 6 would all be filtered out. 
image

I asked because you mentioned "For example, on RC my user has enrollments in 14.100x+2T2017. I expect to see that enrollment on my dashboard on mitxonline as well." in the original task.

rachellougee:
@rhysyngsun @pdpinch Thank you both for the detail feedback. 
I made the changes and re-tested it. However, with the updated filters there were only 9 runs and 3 enrollment imported from MicroMaster RC.  For user - pdpinch,  it only imported MITx/14.73x_1/1T2015 enrollment as all other 6 got filtered out. Let me know if I misunderstand anything.

rhysyngsun (changes requested):

pdpinch:
> For user - pdpinch, it only imported MITx/14.73x_1/1T2015 enrollment as all other 6 got filtered out. Let me know if I misunderstand anything.

Surprising, but that looks correct. 

Would you like me to generate another enrollment to include?

rachellougee:
@rhysyngsun @pdpinch After fixing the run_tag issues https://github.com/mitodl/mitxonline/pull/788, I was able to import 18 DEDP courses from MicroMaster RC, still 3 enrollments though.

micromasters_import/management/commands/queries/002_courserun_enrollment.sql
rachellougee: I had to change expiration_date to use end_date+1 as I run into a validation issue 'Expiration date must be later than end date.' on admin after the import

rhysyngsun (approved):
LGTM

PR #760, opened 2022-07-29
add backfill migration for new table paid courserun

rachellougee:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

(https://github.com/mitodl/mitxonline/issues/738)

#### What's this PR do?
To backfill the new table paid_course_runs for existing orders

-  Run ```docker-compose run --rm  web ./manage.py migrate``` locally
-  Go to /admin/courses/paidcourserun/, all the paid courserun orders should be populated there.
-  Verify the new records with admin/ecommerce/order/


courses/migrations/0013_backfill_paid_courserun.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def backfill_paidcourserun(apps, schema_editor):
+    """Backfill PaidCourseRun for existing orders"""
+    PaidCourseRun = apps.get_model("courses", "PaidCourseRun")
+    CourseRun = apps.get_model("courses", "CourseRun")
+    Order = apps.get_model("ecommerce", "Order")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+
+    for order in Order.objects.filter(state__in=["fulfilled", "review"]):
```
asadali145, line 13:
```suggestion
    for order in Order.objects.filter(state__in=[Order.STATE.FULFILLED, Order.STATE.REVIEW]):
```

rachellougee: I tried it with Order.STATE.FULFILLED, It says "AttributeError: type object 'Order' has no attribute 'STATE'". By debugging it, it seems like these constant are not available in app defined model - Order = apps.get_model("ecommerce", "Order").
I could import Order model directly, then Order.STATE.FULFILLED is available, but from what I read, we should always use app model.  Please let me know if I am missing anything

courses/migrations/0013_backfill_paid_courserun.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("courses", "0012_paidcourserun"),
+    ]
+
+    operations = [
+        migrations.RunPython(backfill_paidcourserun),
```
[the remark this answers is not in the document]
rachellougee: I tried it with or without migrations.RunPython.noop, It didn't seem to make much difference when I revert this migration. Could you explain more about the benefit of passing migrations.RunPython.noop here? sorry but I just wanted to understand the difference in case I miss anything

rachellougee:
@asadali145 I've addressed your feedback. I had to import Order model to get these state constants as these are not available from get_model call. Please take a look again. Thanks

courses/migrations/0013_backfill_paid_courserun.py
asadali145: Yeah, We should always use `apps.get_model`.  You are right. If this is the case then I am okay with this one. You can ignore my comment. Thanks

courses/migrations/0013_backfill_paid_courserun.py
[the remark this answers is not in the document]
asadali145: By reverting a migration, I mean to go back to the previous migration. I am not sure why it isn't raising an exception in your case.
Here is a bit of my understanding:

1. There can be cases where you want to revert a migration like in this case we have a migration 0013. Let's say we have created and applied this migration without `noop`. Now, If we want to go back to a previous migration like 0012 then it will raise an exception that this operation is not reversible (As in Screenshot).

2. In the case of Data Migrations, we usually provide `noop` which means that we don't want to perform any action on the database while reverting a migration. We can also provide a callable in place of `noop` if we want to reverse our changes.
3. the last para in this guide explains it a bit. https://docs.djangoproject.com/en/3.2/topics/migrations/#data-migrations

Screenshot 2022-07-30 at 11 32 21 AM

courses/migrations/0013_backfill_paid_courserun.py
asadali145: @jkachel What's your say on this? I think we can go without model constants.

courses/migrations/0013_backfill_paid_courserun.py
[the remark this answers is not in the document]
rachellougee: @asadali145 Thank you for the detail. Yeah I understand that migrations.RunPython.noop is just not doing anything if we want to revert. Without it, it would raise exception. In this case, once the migration is applied we shouldn't do anything as this is data migration as you mentioned, not something we can revert. In short, I agree that we should pass noop for good practice

courses/migrations/0013_backfill_paid_courserun.py
jkachel: After some more thought, we actually shouldn't be using the constants here. Future revisions may change the values of the constants and, if that happens, those changes would invalidate this migration. So, these should just be hard-coded in order to make the migration idempotent.

rachellougee:
@asadali145 @jkachel I've updated the PR. The only change since the original PR is to add migrations.RunPython.noop to migrations.RunPython per Asad's feedback. If any of you could take a look, that'd be great. Thank you

Issue #762, opened 2022-07-29
Ecommerce: upgrading an enrollment does not change the enrollment mode


jkachel:
### Steps to Reproduce

User doesn't appear to be a factor here, so testing this with a newly created account is recommended.

1. Enroll into a course that has a certificate without upgrading to the certificate level track. The enrollment should succeed.
2. Load your dashboard. The course should appear on the dashboard, and the Get Certificate button should be displayed.
3. Click the Get Certificate button. You should be redirect to the cart. 
4. Complete the purchase. You should receive a receipt via email and should be redirected to the MITx Online dashboard with a success message. Specifically, the message should note that you have enrolled in the paid version of the course. 
5. Observe the dashboard. The entry for the course should say "Enrolled in certificate course" and the Get Certificate button should not be visible.

### Expected Behavior

In step 5, the course should be tagged as enrolled in a certificate course and the Get Certificate button should not be visible.

### Actual Behavior

The upgrade does not happen, so the tag used is the "Enrolled in free course" tag and the Get Certificate button is still present.

### Notes

The transaction details should be recorded successfully. There will be a Fulfilled Order record, and a Transaction record will exist as well. Purchasing the certificate level course for a course with no enrollment does work.

pdpinch:
#763 is merged and tested on RC. Do you think we can close this?

arslanashraf7:
Yes, Closing this.

PR #763, opened 2022-08-01
fix: enrollment upgrade from free to paid version

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#762 

#### What's this PR do?
- Refactors the logic in enrollment to add a case when the free enrollment exists and user upgrades to paid

There are a couple of paths you should test it with (The enrollment should properly reflect the state as per the below scenarios):

1. A new user new enrolment _Unpaid_
2. A new user new enrolment _Paid_
3. User enrolled in _Unpaid_, now Pays and enrolles in _Paid Mode_
4. User enrolled unpaid, then _Unenrolls_, now _pays and enrolls again_

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -33,11 +33,11 @@ from courses.models import (
     ProgramEnrollment,
     Course,
 )
+from courses.tasks import subscribe_edx_course_emails
 from openedx.api import (
     enroll_in_edx_course_runs,
     get_edx_api_course_detail_client,
     unenroll_edx_course_run,
-    subscribe_to_edx_course_emails,
 )
 from openedx.exceptions import (
     EdxApiEnrollErrorException,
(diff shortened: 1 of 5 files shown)


courses/api.py
```diff
@@ -258,11 +258,21 @@ def create_run_enrollments(
                 run=run,
                 defaults=dict(edx_enrolled=edx_request_success, enrollment_mode=mode),
             )
-            if not created and not enrollment.active:
-                enrollment.edx_enrolled = edx_request_success
-                subscribed = subscribe_to_edx_course_emails(user, run)
-                enrollment.edx_emails_subscription = True if subscribed else False
```
arslanashraf7, line 264: I've moved this to a celery task, We aren't doing anything extra if the result of this API is true/false that needs to be here. So we can let celery task do this and continue with enrollment.

jkachel (approved):
I think this looks fine (and it successfully update the enrollment in all the test cases I ran through), so LGTM 👍

Issue #770, opened 2022-08-02
flexible pricing: too many decimal places


pdpinch:
### Steps to Reproduce

1. As a learner, submit a request for flexible pricing with a low income on https://rc.mitxonline.mit.edu/courses/course-v1:MITx+14.100x/microeconomics-request-flexible-pricing/
2. As an admin, approve the request

### Expected Behavior

- The price on the enroll pop-up should be "$250"

### Actual Behavior

- The price on the enroll pop-up is "$250.00000"

### Screenshot or Screencast

<img width="565" alt="image" src="https://user-images.githubusercontent.com/430126/182451955-8efef2c3-8a97-4bad-8fd5-ebdcaafa30c9.png">

pdpinch:
This also shows up in the email:

<img width="729" alt="image" src="https://user-images.githubusercontent.com/430126/182519503-670909ac-69f8-49b0-992f-3750b445554b.png">

PR #772, opened 2022-08-02
770 - flexible pricing: too many decimal places

collinpreston:
Just like James recommended in my prior PR but I never did...

#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

https://github.com/mitodl/mitxonline/issues/770

#### What's this PR do?
Displays the course price in a format such as `$xxx.xx`.  

1. Login as a learner and navigate to a course that is not free
2. Click the "Enroll" button
3. A modal should appear which will display the certificate price in the top right.  Verify the price is in the correct format.

James suggested that I use `formatLocalePrice` in a prior PR and I completely forgot to ensure that I did.

#### Screenshots
Desktop:
Screen Shot 2022-08-02 at 3 35 50 PM

Mobile:
Screen Shot 2022-08-02 at 3 40 41 PM

### DIFF
--- a/ecommerce/serializers_test.py
+++ b/ecommerce/serializers_test.py
@@ -201,7 +201,7 @@ def test_basket_with_product_serializer():
     serialized_basket = BasketWithProductSerializer(basket_item.basket).data
 
     logic = DiscountType.for_discount(discount)
-    discount_price = logic.get_product_price(basket_item.product)
+    discount_price = logic.get_discounted_price([discount], basket_item.product)
 
     assert serialized_basket["total_price"] == basket_item.product.price
     assert serialized_basket["discounted_price"] == discount_price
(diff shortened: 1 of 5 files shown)


asadali145 (changes requested):
It seems like there are a few issues in displaying the prices. For me, It is not showing the Formatted price and `$` is missing as well.

After a bit of research here are the things that I found:

- API returns the `price` of a product as `String`.
- `getFlexiblePriceForProduct`CAN return `String` in case it returns the default `product.price`.
- `formatLocalePrice` does not work on the `String` type, although it works on the `Number` type as we are expecting here.
- In your case, It might be working because `getFlexiblePriceForProduct` is applying some kind of a discount and returning a `Number` type.

Here are a few things that we can do:
- There is a setting for DRF, which currently defaults to False. If we change it to True, our API will return a `Decimal Object` which will eventually be converted to a `Number` in the frontend app. The issue is that it will change this in all the APIs. https://www.django-rest-framework.org/api-guide/settings/#coerce_decimal_to_string
- Convert `amount` to `Number` in `formatLocalePrice` before applying currency styling.

Let me know if you have any questions or you're facing problems reproducing this issue.

Screenshot 2022-08-03 at 1 44 50 PM

asadali145 (approved):
LGTM! Great Work!

PR #774, opened 2022-08-02
Restrict single active product per course ID

collinpreston:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested

(Required)

#### What's this PR do?
There was an error occurring when a user was applying for Flexible Pricing on a Course which was associated with two Product that were `active`.  https://sentry.io/organizations/mit-office-of-digital-learning/issues/3426908655/?environment=rc&project=5864687&query=is%3Aunresolved+&statsPeriod=14d

This PR adds a check when retrieving the Product related to the Course to ensure that it has `is_active` == `true`.

This PR also adds a migration to the Product model to ensure that only a single Product can be active if multiple Products are linked to the same Course ID.

- Login to the Django admin
- Create a Course
- Create a Product record linked to the Course that is active.
- Create a second Product record linked to the same course that is also active.
- You should receive an error message.  Right now this is a 500 error.  I would like to make it nicer for the user, but I'm not sure of how to do that.
- Create a second Product record linked to the same course that is not active.  This should succeed.
Screen Shot 2022-08-03 at 9 52 53 AM


ecommerce/migrations/0020_product_unique_object_id_validated.py
```diff
@@ -0,0 +1,21 @@
...
+
+    operations = [
+        migrations.AddConstraint(
+            model_name="product",
+            constraint=models.UniqueConstraint(
+                condition=models.Q(("is_active", True)),
+                fields=("object_id", "is_active"),
+                name="unique_object_id_validated",
+            ),
+        ),
```
rhysyngsun, line 20: So if there's currently duplicate products (I know RC does/did have one at one point). We can deal with this either by deleting the conflicts prior to the migration or deal with it in the migration.

collinpreston: Yeah, I figured it would be best to deal with them pre-migration.  What is likely to happen if they are left and I perform the migration?

rhysyngsun: the migration will fail because existing data violates the constraint

collinpreston: So would you suggest one method of resolving this over another?

flexiblepricing/mail_api.py
```diff
@@ -41,7 +41,7 @@ def generate_flexible_price_email(flexible_price):
         courserun = CourseRun.objects.filter(
             course=flexible_price.courseware_object
         ).first()
-        product = Product.objects.get(object_id=courserun.id)
+        product = Product.objects.get(object_id=courserun.id, is_active=True)
```
arslanashraf7, line 44: I was just thinking, Instead of getting the products with `is_active` filter in different places, Would it be good that we change the manager in a way that whenever `Product.objects` manager is called it returns the active products only? I created [that manager](https://github.com/mitodl/mitxonline/blob/main/ecommerce/models.py#L45-L51) recently with similar intent:
so if we replace
`
       return ProductsQuerySet(self.model, using=self._db)
`

with

`
        return ProductsQuerySet(self.model, using=self._db).filter(is_active=True)
`

Calling the `Product.objects` will always return Active product everywhere in the existing code, and if we want all products we could use [Product.all_objects](https://github.com/mitodl/mitxonline/blob/main/ecommerce/models.py#L79).

rhysyngsun: I think we'd want to be very careful here with that kind of change - it'd be very easy to introduce a regression - I also think it warrants a separate PR.

arslanashraf7: @rhysyngsun you're correct.

ecommerce/migrations/0020_product_unique_object_id_validated.py
rhysyngsun: Production has no products, so if you're fine cleaning up RC before this gets deployed you can just go ahead with that.

rhysyngsun (approved):
LGTM

PR #785, opened 2022-08-04
Updated ecommerce docs to include max product price and unique CVN

collinpreston:
#### What's this PR do?
Updates the ecommerce documentation with some specifics which are needed or else errors can occur.

### DIFF
--- a/docs/source/configuration/ecommerce.rst
+++ b/docs/source/configuration/ecommerce.rst
@@ -54,7 +54,7 @@ Setting Up a Product
 ####################
 
 1. Log into MITxOnline Django Admin.
-2. Under Ecommerce, open Products and create a new Product. Set *Content type* to Course Run and *Object Id* to the ID of the course run you created earlier (it's probably 1 if you're working from a new install). Price should ideally be set to a non-zero value, that is less than $999. Description needs to be filled in but can be anything - for clarity, it's recommended to use the course name. Make sure Is active is checked.
+2. Under Ecommerce, open Products and create a new Product. Set *Content type* to Course Run and *Object Id* to the ID of the course run you created earlier (it's probably 1 if you're working from a new install). Price should ideally be set to a non-zero value, that is less than $999, in RC/Sandbox environments. Description needs to be filled in but can be anything - for clarity, it's recommended to use the course name. Make sure Is active is checked.
 
 You should now be able to enroll in the upgraded course. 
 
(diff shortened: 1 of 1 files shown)


docs/source/configuration/ecommerce.rst
```diff
@@ -54,7 +54,7 @@ Setting Up a Product
 ####################
 
 1. Log into MITxOnline Django Admin.
-2. Under Ecommerce, open Products and create a new Product. Set *Content type* to Course Run and *Object Id* to the ID of the course run you created earlier (it's probably 1 if you're working from a new install). Price should ideally be set to a non-zero value. Description needs to be filled in but can be anything - for clarity, it's recommended to use the course name. Make sure Is active is checked.
+2. Under Ecommerce, open Products and create a new Product. Set *Content type* to Course Run and *Object Id* to the ID of the course run you created earlier (it's probably 1 if you're working from a new install). Price should ideally be set to a non-zero value, that is less than $999. Description needs to be filled in but can be anything - for clarity, it's recommended to use the course name. Make sure Is active is checked.
```
pdpinch, line 57: Why less than $999?

collinpreston: @pdpinch, I ran into trouble yesterday when trying to purchase a course which was $1000.  The I found this forum: https://groups.google.com/g/edx-code/c/cNY90EYg2PU.  So I lowered the price of the course to $125 and I no longer had issues.  James mentioned that it needs to be less than $999.  I'm not sure why.

arslanashraf7: I remember hearing (Can't find any specific written link on this though) that for test cards the single transaction is `$999` and I suppose this would not be the case on production environments. So, Maybe while mentioning `$999` we might mention that this would only be the case for RC/Sandbox environments with test cards.

collinpreston: @arslanashraf7 That's a good distinction to add.

jkachel (approved):
LGTM 👍

docs/source/configuration/ecommerce.rst
pdpinch: [This Cybersource knowledgebase article](https://support.cybersource.com/knowledgebase/knowledgearticle/?code=000002274) says that transactions between $1000 and $5000 will always fail, to provide a reliable test failure test case.

I'm trying to find more complete documentation on this, but haven't been successful yet.

pdpinch (approved):
I left a comment about the $1000 limit with a link to Cybersource support.

Issue #724, opened 2022-07-21
Flexible Pricing: Allow courseware object to be assigned to Flexible Pricing Request Forms


jkachel:
As an administrator, I'd like to be able to explicitly set what courseware object a Flexible Pricing form applies to. 

Learners request Flexible Pricing through a custom form page that can be configured through the Wagtail CMS system. Each Flexible Pricing request needs to be tied to the courseware object that it's for (either a course or a program), and this is done currently through the page hierarchy: the form itself is created as a child page for a course, and the code determines what the parent page is and grabs its course when a learner submits a request. 

We'd like to replace that functionality with a dropdown in the form itself that allows the form to be explicitly linked to a courseware object. The parent page code works well for requesting flexible pricing for standalone courses but (at least for now) programs won't have a marketing page so there won't be a way in the CMS to do the same for programs. So, the easiest thing to do is to just have a dropdown in the editor in the CMS that allows the admin or whoever to select the course or program that requests should be for. 

There is some prior art for this in the `CoursePage` model in the `cms` app - course pages have a dropdown for the course they belong to. We should do something similar here but it should also incorporate available programs. (My suggestion would be to have this be a single dropdown with the courses and programs separated into groups but there's a few ways this can be done.) Additionally, `cms.models.FlexiblePricingRequestForm` doesn't have a FK to a courseware object at all - there's a class method in there to work it out based on ownership - so one will need to be added; this should ideally be a generic relationship, which we do in a number of other spots (like in `flexiblepricing.models.FlexiblePrice`). 

#### Acceptance Criteria:

- [ ] The Flexible Pricing Request Form form in Wagtail should have a dropdown to choose a specific program or course that requests should be tied to.
- [ ] Flexible Pricing Request Forms should only be tied to either a course or a program. (A relation will need to be added for this.) 
- [ ] Existing Flexible Price requests should be ignored. (Flexible Pricing isn't live yet so data that does exist on RC or in local instances can just be cleared for this.)

pdpinch:
Which form is this? Are you saying we need a program version of, for example, https://rc.mitxonline.mit.edu/cms/pages/22/edit/ ?

jkachel:
Yes and no - yes, it's that form, but no not a totally separate version of it. We should be able to just add a dropdown to select what courseware object it belongs to rather than have that be determined by where it is in the CMS's page hierarchy.

asadali145:
I just saw in standup notes that @jkachel is about to work on this one, so, unassigned myself.

PR #787, opened 2022-08-04
Adds explicit binding of flexible price request forms to courseware objects

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

#724 

#### What's this PR do?

Adds a couple of fields and relations to the FlexiblePriceForm to link them directly to a Course or Program object. Also updates the submit logic to retrieve the correct courseware object based on some rules. 

The courseware object can be retrieved through the `get_parent_courseware` function (was `get_parent_product`), which returns one of (in order):
- The selected program, if there is one
- The selected course, if there is one
- The parent page's course, if there is one

Prior to this, the model would only look at the form's place in the page hierarchy to determine the courseware object. This allows us to set up a Flexible Pricing request form for a program without having to have program pages in the system. When the system games program pages, the logic here should be updated to pull the parent page's program as well. 

All automated tests should pass.

1. Create or edit a flexible pricing request form.
2. You should see dropdowns for Course and Program. Set one or the other. (If you set both, the Program takes precedence.) 
3. Save the form, and submit a request through the form.
4. The resulting submission should be linked with the appropriate courseware object.


arslanashraf7 (approved):
Looks good 👍

PR #788, opened 2022-08-04
Alter courserun.run_tag to a text field with max_length 100

rachellougee:
https://github.com/mitodl/mitxonline/issues/701

#### What's this PR do?
This is discovered while testing https://github.com/mitodl/mitxonline/pull/758 as some of the courses have long run_tag like "2013_Spring" ("MITx/14.73x/2013_Spring") from MicroMaster, it doesn't fit char(10) that is currently defined.
Per discussion with Nathan, we need to make it a text field with max_length 100. It might not need that long, just something we shouldn't worry about later

Run the migration, then go to course run edit page, you should be able edit the run_tag field to be more than 10 characters
image

(Optional)

(Optional)

(Optional)


rhysyngsun (approved):
LGTM

PR #797, opened 2022-08-05
Adds product pages to the CMS

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

#678

#### What's this PR do?

Adds program pages to the CMS. These follow the definition of a course page, but allow you to choose a program instead. In addition to that, this allows Flexible Pricing Request Forms to be created under a ProgramPage, and should update the logic to return the underlying Program if the form is created under a ProgramPage. Also updated are the `configure_wagtail` management command and routing to support ProgramPages (but not the HomePage model).

Automated tests should pass.

To run manual tests, you will need courses set up within a program, and flexible pricing tiers/discounts set up appropriately for the _program_. 

1. Run the `configure_wagtail` command. This will create the base Program index page in the CMS.
2. Add a Program page under the "Programs" in the CMS. 
3. Fill out the form as per usual and publish. The page should be viewable. 
4. Create a Flexible Pricing Request Form under the Program page, in the same manner that you'd create one for a course. 
5. Submit a Flexible Pricing Request via the new form. Courses that are in the specified program should see the proper discount. 

The homepage code hasn't been updated to allow you to choose a Program Page for the Featured Content section. This is by design; we don't necessarily want marketing pages for programs to be seen yet. 

The cards and things that are displayed on a Program Page are cut down - it doesn't look for start dates, eligible course runs, or provide a way to enroll directly in the program. This is also on purpose.


pdpinch:
Please add some screenshots, of the wagtail form and the rendered template. I know the template itself isn't critical path, but I'd like to know what the baseline is if/when we need to actually create a program page.

jkachel:
> Please add some screenshots, of the wagtail form and the rendered template. I know the template itself isn't critical path, but I'd like to know what the baseline is if/when we need to actually create a program page.

Added these in.

cms/migrations/0022_merge_20220805_1914.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("cms", "0020_add_productpage_and_base_tables"),
+        ("cms", "0021_flexiblepricingrequestform_selected_program"),
+    ]
+
+    operations = []
```
annagav, line 13: It looks like this migration is empty?

cms/migrations/0020_add_productpage_and_base_tables.py
```diff
@@ -0,0 +1,207 @@
+# Generated by Django 3.2.14 on 2022-08-05 16:53
```
annagav, line 1: I see you are adding migrations 0020 and 0022, but where is 0021?

annagav (changes requested):
The cms migrations look odd, I left couple of comments.

cms/migrations/0022_merge_20220805_1914.py
jkachel: It's a merge migration; it doesn't have any database operations in it. (There was a migration 20 and 21 in `cms` in a different branch, so this provides a spot for future migrations to base their dependencies on.)

annagav: I see, usually I would undo my migrations, rebase my branch and then create new migrations. I think it works both ways, but I am wondering if we have a preferred way of doing this.

annagav (commented):
I was able to set the program page and create the flexible pricing form. But I was not able to see the discount apply to the individual course in the program. Not sure what I am missing

jkachel:
@annagav not sure about the pricing... I tested it locally with a flexible price form associated with a program, and was able to get the discount to apply successfully on a course in the program. I also tried with a flexible price form under a program in the CMS hierarchy but not explicitly set (to anything) and that also worked. Can we double-check your settings when you get a chance?

annagav (changes requested):
The flexible pricing works for me, and it applied to the courses in the program.
I am not sure if it is the scope of this PR, but there are couple of things missing:
1. there is no way to add a program to the featured products section in the cms
2. the program doesn't have courses attribute under it.

jkachel:
> 1. there is no way to add a program to the featured products section in the cms

It doesn't do that (as noted in the description). This is just to add the program page type to the CMS. 

> 2. the program doesn't have courses attribute under it.

It doesn't do this either (by design, or, rather, lack thereof). We'd need some guidance on how the course list should be presented that we don't have currently, so I've just omitted it for now (especially since people aren't necessarily supposed to be able to get to these yet).

annagav (approved):
👍

PR #800, opened 2022-08-08
Updates status filtering to send an undefined if there's nothing selected

### LINKED ISSUE #731: Flexible pricing: allow searching for users with any status
As a customer support rep, I'd like to be able to find a user's financial assistance request without having to know their current status. 

If I do a search without choosing a status at all, this works. 

But once I choose a status, I can't remove it. From then on I'm limited to searches filtered by status.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #731

#### What's this PR do?

Updates the filter processing code in the Staff Dashboard's Flexible Pricing list screen to clear the Status filter when the blank option is selected. 

You will need to have some Flexible Price requests in the system to test. For best results, they should be a mix of statuses. 

1. Navigate to Flexible Pricing in the Staff Dashboard.
2. Filter by any status. Only the requests matching the status you selected should be shown.
3. Clear the filter by selecting the blank option at the top. Upon selecting the Filter button, all requests should be shown (i.e., it shouldn't be filtering by Status). 

Implementation detail: clearing the filter requires passing `undefined` as the option. So, the change here was just to check for something non-true and explicitly set the filter argument as `undefined`. Simply omitting the filter does not work. 

(Optional)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,3 +124,5 @@ celerybeat-schedule
 
 # Local Docker customizations
 docker-compose.*.yml
+
+.python-version
--- a/frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
+++ b/frontend/staff-dashboard/src/pages/flexible_pricing/list.tsx
@@ -23,10 +23,6 @@ import { FlexiblePricingStatusModal } from "components/flexiblepricing/statusmod
 
 const FlexiblePricingStatuses = [
     {
-        label: null,
-        value: null,
-    },
-    {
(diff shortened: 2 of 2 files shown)


rhysyngsun (changes requested):
A better option might be to just remove the blank option entirely and instead set `allowClear={true}` on the `<Select/>` which will enabled an X icon to clear the input.

rhysyngsun (approved):
LGTM

Issue #794, opened 2022-08-04
Flexible pricing: program-level discount isn't reflected in enrollment pop-up


pdpinch:
### Steps to Reproduce

1. Set up a flexible pricing tier that is associated with a program, [like this one on RC](https://rc.mitxonline.mit.edu/admin/flexiblepricing/flexiblepricetier/41)
2. Make and approve a flexible pricing request with a course in that program, like [14.100x on RC](https://rc.mitxonline.mit.edu/courses/course-v1:MITx+14.100x/)
3. Click the Enroll button

or, if you are already enrolled, 

Look at the upsell box in the lower right of the about page. 

### Expected Behavior

- The price in the pop-up should reflect the program-level discount
- Clicking the "Get Certificate" button takes you to the checkout page, with the program-level discount
- The discounts match. 

### Actual Behavior

- The price in the pop-up does not reflect the program-level discount
- Clicking the "Get Certificate" button takes you to the checkout page, with the program-level discount
- The discounts don't match.

collinpreston:
Hi @pdpinch, I am having trouble replicating this issue using the Flexible Price and Course you shared in the description.  I noticed that the discount for the Flexible Price is set to 0% off. 

I tested this by submitting a Flexible Price request for the course you referenced with my yearly income being less than $1, which would then trigger the use of this `course | program` Flexible Price Tier: https://rc.mitxonline.mit.edu/admin/flexiblepricing/flexiblepricetier/38/change/.  

- I approved my Flexible Price request.
- When I clicked 'Enroll', the modal shown displayed the adjusted price (now $250).
- If I clicked the modal option to get the certificate, the checkout page also displayed $250.
- If I clicked enroll without paying in the modal, and then navigated to the course details page (https://rc.mitxonline.mit.edu/courses/course-v1:MITx+14.100x/), I was also shown $250 in the upsell card shown in the bottom right of the page.

Sorry if I did this incorrectly.

PR #801, opened 2022-08-08
Add financial aid migration query and tier program mapping table

rachellougee:
#### Pre-Flight checklist

- [x] Migrations
  - [x] Migration is backwards-compatible with current production code
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested
#### What are the relevant tickets?
https://github.com/mitodl/mitxonline/issues/702

#### What's this PR do?
- Adding a Tier Program mapping table to map MicroMaster TierProgram to MITxOnline FlexiblePriceTier
- Adding a migration query to bring over the MicroMaster DEDP financial aid records to MITxOnline

#### How should this be manually tested?

- Run the migrate command to create a new mapping table locally

> docker-compose run --rm web ./manage.py migrate

- Go to http://mitxonline.odl.local:8013/admin/micromasters_import/programtierid/, and manually add the following tier program mapping:

Tier Program | MicroMaster Tier Program ID | MITxOnline Flexible Price  (This is your local flexible price instance)
-- | -- | --
Tier 1 | [51](https://micromasters-rc.odl.mit.edu/admin/financialaid/tierprogram/51/change/ )| [38](http://mitxonline.odl.local:8013//admin/flexiblepricing/flexiblepricetier/38/change/)
Tier 2 | [52](https://micromasters-rc.odl.mit.edu/admin/financialaid/tierprogram/52/change/) | [39](https://mitxonline.odl.local:8013/admin/flexiblepricing/flexiblepricetier/39/change/)
Tier 3 | [53](https://micromasters-rc.odl.mit.edu/admin/financialaid/tierprogram/53/change/ )| [40](https://mitxonline.odl.local:8013/admin/flexiblepricing/flexiblepricetier/40/change/)
Tier 4 | [54](https://micromasters-rc.odl.mit.edu/admin/financialaid/tierprogram/54/change/) | [41](https://mitxonline.odl.local:8013/admin/flexiblepricing/flexiblepricetier/41/change/)

- Run the following queries on your local DB to connect to our MicroMaster RC

    (Note that this step is only needed to test locally, our MITxOnline RC already has '**micromasters**' foreign data wrapper configured by DevOps)
    (Replace host/user/password with **micromasters-rc**  credential from Heroku config)

```-- load postgres_fdw extension on MITxOnline DB
CREATE EXTENSION IF NOT EXISTS postgres_fdw;

--- create a foreigner server to connect to external MicroMaster DB
CREATE SERVER IF NOT EXISTS micromaster_bridge
[...]

PR #807, opened 2022-08-09
Updates flexible pricing to add a unique constraint on submissions

### LINKED ISSUE #803: Flexible Pricing: learners can submit multiple requests for same courseware object if separate forms are available for it
As a learner, my ability to request flexible pricing should depend on the courseware object that I used to apply for flexible pricing, and not the actual form that was used. In addition, any flexible pricing requests that have been migrated from MicroMasters for a given courseware object should additionally disallow me from submitting a new request (unless the migrated request itself is in "reset" status). 

### Steps to Reproduce

(For this to be testable, you need a system that allows you to specify what courseware object a given Flexible Pricing Request Form belongs to; at the time of writing, the RC instance isn't one of them.) 

1. Set up two Flexible Pricing Request Forms for the same object. 
2. Submit a request through one.
3. Submit a request through the other.

### Expected Behavior

Step 3 should fail - you should instead see the status of the request you entered in step 2 when you load the form.

### Actual Behavior

You are allowed to submit a new request.

### LINKED ISSUE #804: Flexible Pricing: add unique constraint to FlexiblePrice objects
FlexiblePrice objects are unique based on the submitting user and the courseware object they apply to; the database should reflect this. 

#### Acceptance Criteria:

- [ ] Unique constraint is added on the user and courseware object fields. 
- [ ] Flexible Price submissions continue to work as expected.

See also #803.

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

Fixes #803 
Fixes #804 

#### What's this PR do?

Adds a unique constraint on the `user` and `courseware_object` fields (or, rather, the GenericRelation fields it uses) to the `FlexiblePrice` object. Additionally updates the logic for the `FlexiblePriceRequestForm` to check for an existing request:
- If there's an existing request not in the Review state, submitting will fail. (The user should never get to this point but they will see a 500 if they do.) 
- The logic to determine the last submission from the user and the related courseware object now takes into account the set courseware object for the form. If the object is a standalone course or a program, it will use that. If it's a course in a program, it will return the program it belongs to. (You cannot submit a flexible price request, nor create a form for, an individual course in a program.) 

Automated tests should pass. Manual testing will require a set of courses created, ideally one independent of a program and two that are in the same program.

1. Create Flexible Price Request Forms under each course. The ones for the program courses can be associated with either the course _or_ the program. (One of each is a good test.)
2. Attempt to submit a flexible price request using one of the program course forms. The request should be successful.
3. Attempt to submit a flexible price request using the other program course form. You should not see the form and should instead see the status page (as configured in the form) for the request you entered. 
4. Attempt to submit a flexible price request using the independent course form. The request should be successful. 

You can also use a different user and run the same test as above, except with steps 3 and 4 reversed, if you've assigned one form to a course and one form to a program. In either case, the generated `FlexiblePrice` record should be for the program.


collinpreston:
@jkachel, I think you might have a hat-trick here as this resolves https://github.com/mitodl/mitxonline/issues/794 for me when testing locally.

courses/models.py
```diff
@@ -90,7 +90,6 @@ class Program(TimestampedModel, ValidateOnSaveMixin):
         max_length=255, unique=True, validators=[validate_url_path_field]
     )
     live = models.BooleanField(default=False)
-    # products = GenericRelation(Product, related_query_name="programs")
```
collinpreston, line 93: Should this line be removed?

collinpreston (approved):
Tested and looks good to me

courses/models.py
jkachel: That's been commented out for a while now so I pulled it. (I don't think that GenericRelation can work since pulling in the Products model will create a circular import.)

rachellougee:
@jkachel This looks good to me. I've tested MicroMaster financial aid import https://github.com/mitodl/mitxonline/pull/801 with this PR. It successfully blocks the duplicated import.

rachellougee (approved):

courses/models.py
rhysyngsun: GenericRelation should accept a string argument for the model name I think.

Issue #806, opened 2022-08-09
Ecommerce: Implement a cybersource notification api endpoint


rhysyngsun:
To integrate with CyberSource webhooks (called "Notifications" in their app), we need a plain restful API that they can call. Currently we just have the browser-friendly callback view, but CyberSource probably wants us to respond with a `200 OK` rather than a redirect (different status code).

An example of how we've implemented this in the past can be found here in MicroMasters:

https://github.com/mitodl/micromasters/blob/master/ecommerce/views.py#L139

- This endpoint should respond with a `200 OK` response if it was able to successfully process the request, regardless of the outcome of processing it (e.g. if the order was cancelled or declined).

PR #817, opened 2022-08-10
806: ecommerce implement a cybersource notification api endpoint

collinpreston:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/806

#### What's this PR do?
Adds an additional API endpoint that Cybersource will send a POST request to in response to a user's payment submission.  This PR does a lot of refactoring within the project to separate much of the validation and processing of the Cybersource request so that it can be used by the new API endpoint `checkout_result_api` and `checkout-result-callback`.  

The `checkout-result-callback` is where the user is redirected to after completing the payment submission on Cybersource, and so this view contains extra functionality to show the user a message based on the payment status.

The `checkout_result_api` will always respond with a 200, unless the Order status is `PENDING` and the Cybersource signature on the request could not be validated.

I spent a lot of time building the unit tests to ensure that they were able to run through as much functionality as possible.  I also made sure that I could hit the endpoint locally using Postman.  I'm not sure what's the best way to test this locally though.


jkachel (approved):
LGTM - tested the API with a cURL command retrieved from the web inspector and that seemed to process things OK (once I also changed the order back to Pending status). 👍

ecommerce/views/v0/__init__.py
```diff
@@ -425,110 +423,76 @@ def post(self, request, *args, **kwargs):
...
+        """
+        This endpoint is called by Cybersource as a server-to-server call
+        in order to respond with the payment details.
+        """
+        order = api.get_order_from_cybersource_payment_response(request)
+
+        # We only want to process responses related to orders which are PENDING
+        # otherwise we can conclude that we already received a resonse through
+        # the user's browser.
+        if order is not None and order.state == Order.STATE.PENDING:
```
rhysyngsun, line 491: If we can't find an order in this case, we should probably log an error regarding it.

collinpreston: @rhysyngsun should this still respond with a `200` if there is no order found and an exception is raised, or a `404`?

ecommerce/views/v0/__init__.py
```diff
@@ -425,110 +423,83 @@ def post(self, request, *args, **kwargs):
...
+        # otherwise we can conclude that we already received a response through
+        # the user's browser.
+        result_status = status.HTTP_200_OK
+        try:
+            if order is not None and order.state == Order.STATE.PENDING:
+                api.process_cybersource_payment_response(request, order)
+            elif order is None:
+                raise ObjectDoesNotExist()
+        except ObjectDoesNotExist:
+            result_status = status.HTTP_404_NOT_FOUND
```
rhysyngsun, line 498: I think you can simplify this down to avoid the `try...except` - it adds unnecessary complexity.

ecommerce/api_test.py
```diff
@@ -201,3 +252,37 @@ def test_unenrollment_unenrolls_learner(mocker, user):
...
+        "message": "payment processor message",
+    }
+
+    order = Order.objects.get(state=Order.STATE.PENDING, purchaser=user)
+
+    assert order.reference_number == payload["req_reference_number"]
+
+    request = HttpRequest()
+    request.method = "POST"
+    request.POST = payload
```
rhysyngsun, line 281: Django has a [`RequestFactory`](https://docs.djangoproject.com/en/3.2/topics/testing/advanced/#django.test.RequestFactory) helper and `pytest` [wraps it in a convenient fixture](https://pytest-django.readthedocs.io/en/latest/helpers.html#rf-requestfactory), so you can simplify this down to `rf.post("/url", payload)`

rhysyngsun (changes requested):
A few small things

ecommerce/views/v0/__init__.py
collinpreston: @rhysyngsun Should I just return the `404` and not raise the exception, or is there a way in which both can be done?

ecommerce/views/v0/__init__.py
```diff
@@ -425,110 +423,80 @@ def post(self, request, *args, **kwargs):
...
-            ECOMMERCE_DEFAULT_PAYMENT_GATEWAY
-        ).convert_to_order(payment_data)
-
-        order_id = Order.decode_reference_number(converted_order.reference)
-
-        try:
-            order = Order.objects.get(pk=order_id)
-        except ObjectDoesNotExist:
+        order = api.get_order_from_cybersource_payment_response(request)
+        if order == None:
```
rhysyngsun, line 428: When comparing to `None` you need to use `is` as it's a singleton value.
```suggestion
        if order is None:
```

rhysyngsun (changes requested):

rhysyngsun (approved):
LGTM

Issue #792, opened 2022-08-04
Courseruns need upgrade deadlines


pdpinch:
We can't allow users to pay for a course after the upgrade deadline set in open edX, because the enrollment will fail. 

As a course team member, I want to set an upgrade deadline for instructor-paced courses, so that learners don't pay for courses that they cannot pass. 

As a marketing team member, I want to know the upgrade deadline so I can motivate learners to pay for a course, "before time runs out"  

#### Acceptance Criteria:

- [ ] Add an upgrade deadline to courseruns
- [ ] don't allow payments after the upgrade deadline has passed (because they are going to fail) 
- [ ] ~synchronize the upgrade deadline from open edX to the mitxonline application~

#### Out of scope from this ticket (Based on comments)
- [ ] synchronize the upgrade deadline from open edX to the mitxonline application (See https://github.com/mitodl/mitxonline/issues/842)

arslanashraf7:
A few questions/updates that I wanted to share/ask while working on this:

>don't allow payments after the upgrade deadline has passed (because they are going to fail)

Is there any behavior specified for this on the frontend side, e.g. if a course run upgrade deadline has passed?

1. (Not enrolled) Will we show enroll button? Should we allow free enrollments?
2. (Enrolled) We should hide `Get Certificate` button from everywhere. (Course Details page, Dashboard page)
3. If a user has multiple products and for some of them the `Upgrade Deadline` passed, should we not show those products on the cart or just remove them from the cart?

>synchronize the upgrade deadline from open edX to the mitxonline application

I tried to look around MIcromasters and MITxPRO for the usage of `Upgrade Deadline`, [MicroMasters has it](https://github.com/mitodl/micromasters/blob/master/courses/models.py#L227), but I'm not sure if we are syncing upgrade deadline with edX there?

In xPRO we use course detail API to sync some course attributes with edX e.g, Name, Enrollment start/end, etc. https://github.com/mitodl/mitxpro/blob/master/courses/utils.py#L271-L275.

I suppose when we say `Upgrade Deadline` we are talking about the field we set while creating `Verified` Course Mode in edX ?

[...]

PR #820, opened 2022-08-11
feat: add course run upgrade deadline

arslanashraf7:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested

#792 

#### What's this PR do?
- Adds upgrade deadline to a course run
- Manages the enroll visibility based on upgrade deadline
- Manages cart checkout w.r.t upgrade deadline
- Manages Dashboard enrollments w.r.t upgrade deadline
- Manages course detail visibility on the basis of upgrade deadline
- Removed unused imports from some frontend js file

There are a couple of cases this should be tested with But also take a look at the communication on the ticket:
- Make sure to have courses sets with & without products
- Setup on this branch and without doing anything this should work the same as above (In this case the upgrade_deadline will be empty so it should behave the same for courses with/without product)
- Set an upgrade deadline in the future for the course run, Open the course detail page, and enroll. (The enroll button should work fine as it used to e.g. A modal in case of a product and directly enroll in case of no product)
- Now set the upgrade_deadline for the course run in past, The enroll button should not show the Upgrade modal.
- Enroll in the course (free mode), Visit Dashboard, and see you don't see the `Get Certificate` button there as well
- Visit the course detail page(of the above-enrolled course) and check that you see `Get Certificate` card.
- Now set a future upgrade_deadline and add the course to the cart, now change the upgrade deadline in past and try to checkout, it should show an error that a course in the cart can not be upgraded
- If you think of more scenarios to check this with (try them)

- Would be good to give a quick look at removed imports in frontend/js code to make sure they were really unused

**COURSE UPGRADE DEADLINE IN PAST**

**When a user tries to checkout with a course with upgrade_deadline in the past**
Screenshot 2022-08-15 at 5 10 37 PM

**When a user is enrolled in a course with valid product but the upgrade_deadline is in past**
Screenshot 2022-08-15 at 5 10 12 PM

**When a user is enrolled in a course free mode with a product but upgrade_deadlin is in past (He doesn't see `Get Certificate` button the dashboard)**
Screenshot 2022-08-15 at 5 10 21 PM

**COURSE UPGRADE DEADLINE IS EMPTY OR IN THE FUTURE**

**Course Detail Page**
Screenshot 2022-08-15 at 5 17 37 PM

**Dashboard Page**
Screenshot 2022-08-15 at 5 17 45 PM

**Enroll button click**
Screenshot 2022-08-15 at 5 18 05 PM


frontend/public/src/lib/courseApi.js
```diff
@@ -52,3 +52,9 @@ export const generateStartDateText = (run: CourseRunDetail) => {
 export const isFinancialAssistanceAvailable = (run: CourseRunDetail) => {
   return run.page ? !!run.page.financial_assistance_form_url : false
 }
+
+export const isUpgradable = (run: CourseRunDetail) => {
```
rhysyngsun, line 56: Since you already have this ;logic on the backend (`CourseRun.is_upgradable`), we should just serialize that in addition to the deadline itself and use that on the frontend. That way if the logic ever changes there's only one place to do so.

frontend/public/src/lib/notificationsApi.js
```diff
@@ -56,6 +57,11 @@ export function parseStoredUserMessage(
     msgText =
         "We're sorry, your country is currently blocked from enrolling in this course."
     break
+  case USER_MSG_TYPE_COURSE_NON_UPGRADABLE:
+    alertType = ALERT_TYPE_DANGER
+    msgText =
+        "The upgrade deadline for a course in your cart has passed. Please try later."
```
rhysyngsun, line 63: This message could be interpreted as meaning the learner could try tomorrow or next week, which won't be the case.

Maybe having it be something like:

> The upgrade deadline for a course in your cart has passed. You can retake and pay for this course the next time it runs.

CC @pdpinch

rhysyngsun (changes requested):
Code review

frontend/public/src/lib/courseApi.js
arslanashraf7: This makes sense. I've applied this while addressing feedback.

frontend/public/src/lib/notificationsApi.js
arslanashraf7: This probably makes more sense, I've updated this text.

rhysyngsun (approved):
LGTM

Issue #818, opened 2022-08-10
Bug: `unique_object_id_validated` contraint on `Product` makes it impossible to add products across different content types


rhysyngsun:
Because the index is across only `(object_id, is_active)`, having a Product for a course run prevents the addition of a product for a program run with the same id. The index needs to include `content_type`.

Issue #777, opened 2022-08-03
Ecommerce: Only a single Product can be active for a Course Run


collinpreston:
### Steps to Reproduce

1. Create a new Course Run through the Django admin (http://mitxonline.odl.local:8013/admin/courses/courserun/)
2. Create a new Product that is associated with the Course Run in step 1, and is active through the Django admin (http://mitxonline.odl.local:8013/admin/ecommerce/product/)
3. Create a second Product that is associated with the Course Run in Step 1, and is active through the Django admin.

### Expected Behavior

You should not be allowed to create a second Product which is active and associated with a Course Run for which an already existing active Product is associated with.

### Actual Behavior

You are able to create as many active Products that can all be associated with the same Course Run.

### Stacktrace

https://sentry.io/organizations/mit-office-of-digital-learning/issues/3426908655/?environment=rc&project=5864687&query=is%3Aunresolved+&statsPeriod=14d

PR #823, opened 2022-08-11
818 bug unique object id validated contraint on product makes it impossible to add products across different content types

rachellougee:
Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/818

#### What's this PR do?
This is to modify the unique key unique_object_id_validated previously added in https://github.com/mitodl/mitxonline/issues/777
to allow the same object id for the course and program

No data issue needs to be fixed on RC as the existing data is still validated with the old and new unique key

The automated test should still pass because it uses purchasable_object, which refers GenericForeignKey("content_type", "object_id"))

To manually test, go to http://mitxonline.odl.local:8013/admin/ecommerce/product/
Add a course and program with the same object id, they both should be allowed now

image

image

(Optional)

(Optional)

(Optional)


rhysyngsun (approved):
L

PR #802, opened 2022-08-08
Switch some settings to use urljoin

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?

N/A - this fixes a small avoidable issue w/ configuration

#### What's this PR do?
This fixes the refine dashboard settings to have correct urls regardless of `SITE_BASE_URL` having a trailing slash or not. I wasted a bit of time trying to debug oauth issues related to this.

#### How should this be manually tested?
You should be able to login to the refine dashboard having a base url with and without a trailing slash.

PR #835, opened 2022-08-12
Update refine callback URL to remove double slash

collinpreston:
#### What's this PR do?
Previously you need to use a double slash for the oauth callback URL used when configuring the refine connection.  With https://github.com/mitodl/mitxonline/pull/802, the double slash hack is no longer needed.

- Open the refine oauth configuration at http://mitxonline.odl.local:8013/admin/oauth2_provider/application/ and set the redirect URI to `http://mitxonline.odl.local:8013//staff-dashboard/oauth2/login/`.
- Confirm that the refine dashboard does not allow you to login successfully as an admin user.
- Update the redirect URI to `http://mitxonline.odl.local:8013/staff-dashboard/oauth2/login/`
- Confirm that the refine dashboard does allow you to login successfully as an admin user.


arslanashraf7 (approved):
👍 Thanks for creating this one.

PR #796, opened 2022-08-05
Online-664 Add additional data to Financial Assistance Request Table and Dialog

asadali145:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Fixes https://github.com/mitodl/mitxonline/issues/664

#### What's this PR do?
Adds additional data in the staff dashboard for Financial Assistance requests. This PR adds courseware information and discount information for a financial assistance request. While approving a request admin can also change the discount from the dropdown.

#### How should this be manually tested?
(Required)

- Setup financial assistance for a course/program using https://github.com/mitodl/hq/discussions/133.
- Submit a financial assistance request.
- Visit flexible pricing in Refine dashboard.
- In the flexible pricing list, you can see the `courseware readable id` and `Discount` information.
- When you click on approve button for a request, courseware information is displayed. Also, a dropdown for discounts is displayed. So, you can change the discount for a request. You can change it from the refine dashboard and verify it from Django-admin.
- While denying or resetting a request, the discount is a static field.

#### Screenshots (if appropriate)
(Optional)
Screenshot 2022-08-12 at 11 17 54 AM
Screenshot 2022-08-12 at 11 18 00 AM
Screenshot 2022-08-12 at 11 18 05 AM
Screenshot 2022-08-12 at 11 18 11 AM
Screenshot 2022-08-12 at 11 18 52 AM

PR #781, opened 2022-08-04
online-779 Display `Documents in order` as default

asadali145:
#### Pre-Flight checklist

- [x] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
Fixes https://github.com/mitodl/mitxonline/issues/779

#### What's this PR do?
Displays `Documents in order` as a default choice in Staff Dashboard while approving a pending financial assistance request.

#### How should this be manually tested?
1. Setup financial assistance for a course using https://github.com/mitodl/hq/discussions/133. 
2. Visit the financial assistance form for a course and request financial assistance.
3. Visit the staff dashboard and go to the flexible prices tab.
4. Click on `approve` for the request that you just submitted.
5. Popup will show `Documents in order` as a default choice.

#### Screenshots (if appropriate)
(Optional)
Screenshot 2022-08-04 at 4 51 01 PM
Screenshot 2022-08-04 at 4 51 16 PM

PR #837, opened 2022-08-12
Release 0.41.1

odlbot:
## James Kachel
  - [x] Fixes some conditionals to return good values if there's no CMS page for a courseware object ([0f9d4872](../commit/0f9d4872d557cf0ef12ae133e2084e7170428b10))
  - [x] Adding an extra retry and extending startup grace period to 45s; should help with starting up on Apple Silicon ([585367c4](../commit/585367c40e5aaafc0ce7bc5da0b263bcb180fe81))

## Asad Ali
  - [x] Online-664 Show courseware and discount info for a financial assistance request (#796) ([45f4f306](../commit/45f4f306b8f5f8d986fbd0f43047e922f4501566))
  - [x] online-779 Display `Documents in order` as default (#781) ([a502a295](../commit/a502a295a1d602db4ec21f8581e9899cbd65dcc9))


asadali145:
FYI @jkachel @pdpinch 
There is a minor issue in the refine dashboard due to https://github.com/mitodl/mitxonline/pull/796.

- In the last release, we added support for financial assistance for programs. 
- There are a few old financial assistance requests which belong to a course and we have disabled the Flexible Price Tiers for that course as it belongs to the DEDP program and enabled tiers for the program. So, when we try to approve such a request the discount dropdown is a bit out of order. SS is attached (48 is the Discount Id that we calculated initially but now we don't have such a discount option available in the dropdown).
    Screenshot 2022-08-15 at 10 59 54 AM
- Everything will be fine for future requests.

Possible Solutions:

- Leave this as it is as the admin cannot do anything wrong with the old requests (Admin will see discount id in the dropdown).
- Display static discount string as we do in case of `RESET` & `DENY`.
   Screenshot 2022-08-15 at 11 22 31 AM


Let me know if we want to fix this for Old data in RC.

asadali145:
> FYI @jkachel @pdpinch There is a minor issue in the refine dashboard due to #796.
> 
> * In the last release, we added support for financial assistance for programs.
> * There are a few old financial assistance requests which belong to a course and we have disabled the Flexible Price Tiers for that course as it belongs to the DEDP program and enabled tiers for the program. So, when we try to approve such a request the discount dropdown is a bit out of order. SS is attached (48 is the Discount Id that we calculated initially but now we don't have such a discount option available in the dropdown).
>   Screenshot 2022-08-15 at 10 59 54 AM
> * Everything will be fine for future requests.
> 
> Possible Solutions:
> 
> * Leave this as it is as the admin cannot do anything wrong with the old requests (Admin will see discount id in the dropdown).
> * Display static discount string as we do in case of `RESET` & `DENY`.
>   Screenshot 2022-08-15 at 11 22 31 AM
> 
> Let me know if we want to fix this for Old data in RC.

I think we can go with the current implementation for now. Maybe we can improve this in a follow-up ticket. Checking the boxes for now.

PR #838, opened 2022-08-12
Purges the Fastly cache for a page once the page has been modified

### LINKED ISSUE #749: CMS: after updating a page, the CDN cache is not cleared
### Steps to Reproduce

1. On RC or Prod, view a course about page like https://rc.mitxonline.mit.edu/courses/course-v1:MITx+14.100x/. Viewing it will cause it be cached in your local CDN node. 
2. Examine the HTTP headers to confirm that you got a cache hit, and the version stamp
2. Edit the page in the CMS and publish it
3. Reload the published page in your browser
4. Examine the HTTP headers to confirm that you are getting the same cached version as before. 

### Expected Behavior

- The change should be reflected immediately

### Actual Behavior

- No change
- Same version coming from the cache 

### Related Issues

We dealt with a similar issue on OCW by making explicit API calls to the CDN to clear the cache on publishing. 

### Screenshot or Screencast

https://user-images.githubusercontent.com/430126/181348812-a6c0d343-458d-4a6e-a2cd-2d25ed7f7274.mov

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested
- [X] Settings
  - [X] New settings are documented and present in `app.json`

Fixes #749 

#### What's this PR do?

Adds some receivers for some Wagtail signals that are sent when a page is (un)published or moved to invalidate the Fastly cache for the affected pages. This should ensure that new edits made in the CMS override the cache when they happen.

Fully testing this manually will attempt to invalidate pages in the Fastly cache. This will fail unless your system is configured to look like a system it knows about (e.g. you need to set up your testing machine to respond to `rc.mitxonline.mit.edu` locally for it to really work). However, you should be able to see log entries where the app queues and then invalidates the cache regardless of your local instance's settings.

Do any number of these operations:
1. Publish a page.
2. Move a page such that the URL for it will change.
3. Unpublish a page. 

Each of these should cause Wagtail to signal, which should then be picked up by the code in this PR and will queue a Fastly purge request via Celery. 

#### Other

It doesn't appear that the purge API is set up to require authentication at the moment - this (can be changed)[https://docs.fastly.com/en/guides/authenticating-api-purge-requests] and if/when it does, the `MITX_ONLINE_FASTLY_AUTH_TOKEN` setting will need to be set.

### DIFF
--- a/app.json
+++ b/app.json
@@ -99,6 +99,14 @@
       "description": "Timeout (in seconds) for requests made via the edX API client",
       "required": false
     },
+    "FASTLY_AUTH_TOKEN": {
+      "description": "Optional token for the Fastly purge API.",
+      "required": false
+    },
+    "FASTLY_URL": {
+      "description": "The URL to the Fastly API.",
+      "required": false
+    },
     "GA_TRACKING_ID": {
       "description": "Google analytics tracking ID",
       "required": false
--- a/cms/api.py
+++ b/cms/api.py
@@ -12,6 +12,8 @@ from wagtail.core.models import Page, Site
 
 from cms import models as cms_models
 from cms.exceptions import WagtailSpecificPageError
+from cms.models import Page
+
 
 log = logging.getLogger(__name__)
 DEFAULT_HOMEPAGE_PROPS = dict(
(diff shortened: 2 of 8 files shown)


rachellougee (approved):

rachellougee:
LGTM. For testing, I can see that editing a page causes a signal queuing the request via celery log

Issue #816, opened 2022-08-10
Refunds: Update the order lookup for refunds to use the new `reference_number` field


rhysyngsun:
Blocked by #808 

We've previously added ( https://github.com/mitodl/mitxonline/pull/808 ) a `reference_number` to `Order` to be able to disambiguate MicroMasters orders from MITx Online ones given that the primary keys will overlap and also that the MicroMasters orders will actually get different primary keys when they're imported anyway.

When learners contact support for a refund, the number they'll end up providing for MicroMasters orders won't be currently matched using the current strategy of looking up the numeric primary key.

To accommodate this, we want to update the current lookup to use the full reference number (e.g. MM-prod-5830, mitxonline-production-85474, etc)

pdpinch:
I don't think this is blocked anymore.

PR #847, opened 2022-08-15
Refund order based on id or reference number

annagav:
- [ ] Testing
  - [x] Code is tested

Fix https://github.com/mitodl/mitxonline/issues/816

#### What's this PR do?
Refund order based on id or reference number

Try to refund an order through django admin, and through google sheets


ecommerce/api.py
```diff
@@ -224,12 +224,13 @@ def establish_basket(request):
     return basket
 
 
-def refund_order(*, order_id: int, **kwargs):
+def refund_order(*, order_id: int = None, reference_number=None, **kwargs):
```
rhysyngsun, line 227: Could you type `reference_number` as a `str` in the signature?

ecommerce/api.py
```diff
@@ -242,18 +243,26 @@ def refund_order(*, order_id: int, **kwargs):
...
-
+        if reference_number is not None:
+            order = FulfilledOrder.objects.select_for_update().get(
+                reference_number=reference_number
+            )
+        elif order_id is not None:
+            order = FulfilledOrder.objects.select_for_update().get(pk=order_id)
+        else:
+            log.debug(
+                f"Either order_id or reference_number is required to fetch the Order. "
```
rhysyngsun, line 254: Doesn't need to be an f-string and has an extra trailing space. This should also probably be an error.

ecommerce/api.py
```diff
@@ -242,18 +243,26 @@ def refund_order(*, order_id: int, **kwargs):
...
+            )
+            return False
         if order.state != Order.STATE.FULFILLED:
-            log.debug(f"Order with order_id {order_id} is not in fulfilled state.")
+            log.debug(f"Order with order_id {order.id} is not in fulfilled state.")
             return False
 
-        order_recent_transaction = Transaction.objects.filter(order=order_id).first()
+        order_recent_transaction = Transaction.objects.filter(order=order.id).first()
```
rhysyngsun, line 261: Nit: could be `order.transactions.first()`

ecommerce/api.py
```diff
@@ -242,18 +243,26 @@ def refund_order(*, order_id: int, **kwargs):
...
+        elif order_id is not None:
+            order = FulfilledOrder.objects.select_for_update().get(pk=order_id)
+        else:
+            log.debug(
+                f"Either order_id or reference_number is required to fetch the Order. "
+            )
+            return False
         if order.state != Order.STATE.FULFILLED:
-            log.debug(f"Order with order_id {order_id} is not in fulfilled state.")
+            log.debug(f"Order with order_id {order.id} is not in fulfilled state.")
```
rhysyngsun, line 258: For this and the other log strings, it should use `%s` formatting so they get deduplicated in sentry.

ecommerce/api_test.py
```diff
@@ -74,6 +74,15 @@ def test_cybersource_refund_no_fulfilled_order(order_state):
     assert refund_response is False
 
 
+def test_cybersource_refund_no_order_id():
+    """Test that refund_order returns logs properly and False when there is no Fulfilled order against
+    the given order_id"""
+
+    refund_response = refund_order()
+    assert f"Either order_id or reference_number is required to fetch the Order."
```
rhysyngsun, line 82: Doesn't need to be an f-string.

rhysyngsun (changes requested):
Code review, will do a functional test tomorrow morning.

rhysyngsun (changes requested):
I'm getting this error trying to test, which I think needs to be fixed in the refunds library which is trying to parse the order id column as an int still:

```
[2022-08-17 22:41:01] ERROR 7 [mitol.google_sheets_refunds.api] api.py:61 - [d28c18aed9c6] - Parsing failure
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/mitol/google_sheets_refunds/utils.py", line 79, in parse_raw_data
    order_id=int(raw_row_data[6]),
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/mitol/google_sheets_refunds/api.py", line 59, in process_row
    refund_req_row = RefundRequestRow.parse_raw_data(row_index, row_data)
  File "/usr/local/lib/python3.9/site-packages/mitol/google_sheets_refunds/utils.py", line 95, in parse_raw_data
    raise SheetRowParsingException(str(exc)) from exc
mitol.google_sheets.exceptions.SheetRowParsingException: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

annagav:
closed by accident

rhysyngsun:
@annagav I've cut a new release of mitol-django-common that should fix the test failure: https://pypi.org/project/mitol-django-common/2.5.2/

rhysyngsun:
@annagav I've been trying to test this but I'm finding that it's failing because my google sheet isn't copying over reference numbers - it's still copying over historical numeric order ids but not the ones with the prefix. Any ideas on why or what I need to fix?

sheets/plugins.py
```diff
@@ -12,7 +12,7 @@ def refunds_process_request(
     ) -> RefundResult:
 
         refund_api_success = refund_order(
-            order_id=refund_request_row.order_id, unenroll=True
+            reference_number=refund_request_row.order_id, unenroll=True
```
rhysyngsun, line 15: Needs to be fixed to use the new property:

```suggestion
            reference_number=refund_request_row.order_ref_num, unenroll=True
```

rhysyngsun (changes requested):
I don't think you'd meant to change the other two files in your last commit.

rhysyngsun (approved):
:+1: when tests pass

PR #848, opened 2022-08-15
Ecommerce: adds activation and expiration dates to discount codes

### LINKED ISSUE #832: activation and expiration dates for discounts
As a marketer, I'd like to be able set the dates for when a discount is available, so I can coordinate the discount with ad campaigns. 

As a AR team member, I'd like discounts (particularly, enrollment codes) to have expiration dates so we don't have an open-ended obligation to deliver a course. 

As an admin, I'd like to have a way to turn off discounts that are no longer needed. 

I think we can address all these stories with an activation date, and an expiration date on discounts, similar to coupons in xPRO.

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

Fixes #832 

#### What's this PR do?

Adds an optional activation and expiration date to discount codes. If set, the discount code cannot be applied if the current date/time is before the activation date or after the expiration date. This applies to user discounts and flexible pricing discounts as well. 

All automated tests should pass.

1. In Django Admin, create or modify a discount code such that it has an activation date or an expiration date (or both).
2. Attempt to redeem the discount through the cart. If today's date is before the activation date, or if it's past the expiration date, it should fail.


ecommerce/api.py
```diff
@@ -2,12 +2,14 @@
 
 import json
 import logging
+import datetime
+import pytz
```
collinpreston, line 6: Can these be removed since they are not being used in this file?

ecommerce/api.py
```diff
@@ -2,12 +2,14 @@
...
 
 from functools import total_ordering
 from django.urls import reverse
 from django.db import transaction
 from django.core.exceptions import ObjectDoesNotExist
-from main.settings import ECOMMERCE_DEFAULT_PAYMENT_GATEWAY
+from main.settings import ECOMMERCE_DEFAULT_PAYMENT_GATEWAY, TIME_ZONE
```
collinpreston, line 12: Can `TIME_ZONE` be removed?

ecommerce/api.py
```diff
@@ -182,6 +184,9 @@ def apply_user_discounts(request):
         if not check_discount_for_products(discount, basket):
             return
 
+        if not discount.check_validity(user):
+            return
+
```
collinpreston, line 189: nit picking, but can the two `if` statements between line 184 and 189 be combined with an OR and just `return`?

ecommerce/views_test.py
```diff
@@ -332,6 +335,56 @@ def test_redeem_discount_with_higher_discount(
...
+    )
+
+    resp_json = resp.json()
+
+    if expects:
+        assert "message" in resp_json
+        assert resp_json["message"] == "Discount applied"
+    else:
+        assert "not found" in resp_json
+
```
collinpreston, line 386: Nice

collinpreston (approved):
Easy to follow and looks good.  I did notice that if the discount became invalid while still applied to my current Cart, it would still be valid.  I didn't try performing a payment with it like this, so maybe it would eventually show up as invalid, or maybe this isn't much of a concern.

jkachel:
https://github.com/mitodl/mitxonline/pull/848/commits/19eab06ec4e000b82bb1662b0cf38f602e74b589 adds in clean up and updates the checkout code to make sure the discounts are valid before proceeding. At the moment, if this check fails, you get a notification banner and it clears your discounts (and reapplies any user/flex pricing discounts you may have).

collinpreston (approved):
Looks good and works well.  Thanks for finding out how other applications were handling the expiration of discounts while they're applied to the Cart/Basket!

Not to throw another wrench into this, but when testing today I noticed that the ProductDetailEnrollApp (https://github.com/mitodl/mitxonline/blob/cd945449f31800544366e4c5e7de733f32662a04/frontend/public/src/containers/ProductDetailEnrollApp.js#L99) is displaying the discounted price from the Flexible Price applied to the course even when the discount is expired.  I imagine the same thing occurs for the UpSellCard.

jkachel:
https://github.com/mitodl/mitxonline/pull/848/commits/ec4b16b61b5ae5486e4345f07de46893c393b0d2 updates `determine_courseware_flexible_price_discount` to factor in the tier discount dates (if they exist), which is ultimately where the `ProductDetailEnrollApp` amongst others gets that data.

collinpreston (approved):
Nice!

Issue #811, opened 2022-08-10
"Need financial assistance?" link on the checkout page


pdpinch:
When we did issue #677 / PR #764 it looks like we overlooked the last item on the list, possibly because I forgot to link to InVision. 

4. "Need financial assistance?" link on the checkout page

#### Designs and Mockups

https://projects.invisionapp.com/d/main?origin=v7#/console/21535972/462885073/preview 

note the link in the top right

#### Acceptance Criteria:

- [ ] The link should only be available when the course in question offers flexible pricing / financial assistance
- [ ] The link should not display if the learner has already been approved for financial assistance
- [ ] The link should go to the appropriate flexible pricing form for the current course

collinpreston:
@pdpinch can you please verify if these screenshots meet the requirements?
![Screen Shot 2022-08-15 at 12 20 36](https://user-images.githubusercontent.com/8311573/184674463-8152aedd-9d9b-4876-976b-e21298070b14.png)
![Screen Shot 2022-08-15 at 12 20 30](https://user-images.githubusercontent.com/8311573/184674465-81e1096e-625d-41cd-bb6b-bf539a08ce27.png)
![Screen Shot 2022-08-15 at 12 20 21](https://user-images.githubusercontent.com/8311573/184674467-86a307d3-a4c5-4004-96ba-6fc4f17526e9.png)

pdpinch:
It's not a blocker, but in the mobile view:

<img width="50%" src="https://user-images.githubusercontent.com/8311573/184674465-81e1096e-625d-41cd-bb6b-bf539a08ce27.png">

Why does the "need financial assistance link?" move after the "Help and FAQ" ? 

In any case, can it be right aligned?

PR #855, opened 2022-08-16
811: need financial assistance link on the checkout page

collinpreston:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing

https://github.com/mitodl/mitxonline/issues/811

#### What's this PR do?
If a user adds a Product that offers Flexible Pricing, but the user does not currently have a Flexible Price, to their Cart then the user should be presented with a link that takes them to the Flexible Pricing Form for the Product in their Cart. 

Things to note: The  'Need financial assistance?' link is only hidden when the user has an approved flexible pricing request.

1. Login as a user and add a Course which the user does not have any approved Flexible Pricing to your Cart.
2. View the Cart and ensure that you can see the 'Need financial assistance?' link.
3. Click the link and ensure that you are navigated to the Flexible Pricing form associated with the Product in your Cart.
4. Submit a Flexible Pricing request and approve the request.
5. Visit the Cart again and ensure that you no longer see the 'Need financial assistance?' link.

Screen Shot 2022-08-16 at 9 06 12
Screen Shot 2022-08-16 at 9 06 03
Screen Shot 2022-08-16 at 9 05 43


jkachel (commented):
This seems to work OK but there's secondary problem, I think, that came up when I was testing this. 

A prior PR added in the `financial_assistance_form_url` to the data here and that part only accounts for items that have financial assistance pages that are directly situated underneath the given page. We've decoupled the flexible pricing request forms from the CMS hierarchy so this really needs to be changed to look for a form specifically for the program or the course (in that order) first before defaulting back to the CMS hierarchy. 

(For example, my local instance has 5 courses set up - 2 each across 2 programs and one standalone - and flexible pricing forms for each program and the standalone course. The flexible pricing forms for the programs are under the program page and the standalone course has its directly underneath its page. In this case, I'm using an account that has no flexible pricing submissions or enrollments at all. When I add the standalone course to the cart, it shows the Need Flexible Pricing? link as expected. This doesn't happen with the other 4 courses, though, since they don't have any child pages of their own - the flexible pricing form is instead under the program. I should still be able to see that link, though.) 

So, the serializer needs to look for a flexible price form that's explicitly assigned to the program or course first before it looks at the page hierarchy, and then return that data. This should all be in the `CoursePageSerializer` in `cms.serializers`. 

This may be outside the scope of this PR - what do you think? I can throw this into a separate issue if that makes more sense.

cms/serializers.py
```diff
@@ -30,11 +32,42 @@ def get_financial_assistance_form_url(self, instance):
         """
         Returns URL of the Financial Assistance Form.
         """
-        financial_assistance_page = (
-            instance.get_children().type(FlexiblePricingRequestForm).live().first()
-        )
+        financial_assistance_page = None
+        base_url = None
+        if instance.product.program:
+            print("CP", instance.product.program)
```
jkachel, line 38: stray debug statement

cms/serializers.py
```diff
@@ -30,11 +32,42 @@ def get_financial_assistance_form_url(self, instance):
...
+                selected_course=instance.product
+            ).first()
+
+        if financial_assistance_page is None:
+            base_url = instance.get_url()
+            financial_assistance_page = (
+                instance.get_children().type(FlexiblePricingRequestForm).live().first()
+            )
+
+        print("CP", " 64 ", financial_assistance_page)
```
jkachel, line 67: stray debug statement

cms/serializers.py
```diff
@@ -30,11 +32,42 @@ def get_financial_assistance_form_url(self, instance):
...
+
         return (
-            f"{instance.get_url()}{financial_assistance_page.slug}/"
+            f"{base_url}{financial_assistance_page.slug}/"
```
jkachel, line 70: The Wagtail `get_url` method should probably be used here on the request form page. It should return the URL with the special formatting we use for course/program pages (which use the readable ID rather than the page slug).

cms/serializers_test.py
```diff
@@ -28,3 +35,84 @@ def test_serialize_course_page(mocker, mock_context):
...
+    course = CourseFactory(program=program)
+    course_page = CoursePageFactory(course=course)
+
+    data = CoursePageSerializer(instance=course_page).data
+    assert_drf_json_equal(
+        data,
+        {
+            "feature_image_src": fake_image_src,
+            "page_url": None,
+            "financial_assistance_form_url": f"{program_page.get_url()}{financial_assistance_form.slug}/",
```
jkachel, line 62: See earlier comment about using get_url here (and below, on lines 89 and 116). Caveat: sometimes that doesn't work in a test suite - if it doesn't, this is fine as is.

frontend/public/src/containers/pages/checkout/CartPage.js
```diff
@@ -154,6 +155,20 @@ export class CartPage extends React.Component<Props, CartState> {
     )
   }
 
+  renderFinancialAssistanceOffer() {
+    const { cartItems, discounts } = this.props
+    const userFlexiblePriceExists = discounts && discounts.length > 0 ? discounts.find(discount => discount.for_flexible_pricing === true) : false
```
jkachel, line 160: This seems to return false whenever a discount is applied, whether or not it's flagged for flexible pricing.

jkachel (changes requested):
Some minor changes. Otherwise, seems to work - I get the right flex pricing form when I add a program course to my cart and when I add a standalone course in.

jkachel (approved):
LGTM 👍

Issue #657, opened 2022-07-05
MM Data Migration: Migrate Learner orders from MicroMasters


rhysyngsun:
Summary
---

We need historical purchase data to be brought over from MicroMasters. There are some aspects of this that are a bit complicated, we should address them individually though:

- Historically DEDP learners were permitted to take 2 exams for every purchases
- Recently, we shifted this to 1 exam per purchase
- Some learners have purchased exams, but not taken them yet. MicroMasters limits learners to 1 unused purchase per course.

##### Order

This is pretty straightforward since most of the fields are identical.

##### Line

Line is a little trickier, but maps 1 to 1:

- `quantity=1`
- `purchased_content_type` - you'll need to join on the contenttypes table and get the id that represents course runs
- `purchased_object_id` - this will be the course run id, which we'll need to make a best guess on. Is it possible to select the run that meets the criteria of having been made after `enrollment_start` but before `upgrade_deadline`? I'm not sure if that will work without conflicts or not.
- `product_version` - TBD???
- 

##### Transaction

Transaction will get derived from `Receipt`:

- `data` is the same
- `amount` should come from the order's `total_price_paid`
- `transaction_type="payment"`
- `reason=""`

rachellougee:
@rhysyngsun I plan to work on https://github.com/mitodl/mitxonline/issues/702 first as this one is marked as blocked. Could you give some details like what tables are involved and the business logic with this order migration?

rhysyngsun:
@rachellougee I updated the description with what I think the mappings will be

@pdpinch:
 - Could you sanity check those?
 - MM has a `partially_refunded` state, any thoughts on what to do with those as we never implemented them in xPro or MITx Online? I see 3 orders in this state in production.
- In MITx Online we record refund `Transaction` records, should we create equivalents here? The cybersource `data` would have to be empty or better have some kind of marker json to indicate that it's generated. Not sure how this would be affected by partial refunds as I'm guessing we probably don't have data on it, although if we do we could mark those transactions somehow and then manually repair them since they're only 3 orders.
- I have a question about order numbers, but I'm having trouble describing it in text so we should chat tomorrow

pdpinch:
Regarding migrating lines:

> `purchased_object_id` - this will be the course run id, which we'll need to make a best guess on. Is it possible to select the run that meets the criteria of having been made after enrollment_start but before upgrade_deadline? I'm not sure if that will work without conflicts or not.

[...]

PR #859, opened 2022-08-16
Add queries to migrate order, line and transaction from MicroMaster

rachellougee:
#### Pre-Flight checklist
- [x] Migrations
- [x] Testing
  - [x] Code is tested
https://github.com/mitodl/mitxonline/issues/657

#### What's this PR do?

- Adding a new state 'partially_refunded' to bring over partially refunded orders from MM
- Making 'reference_number' as readly-only field in django admin (This field can't be manually modified as we rely on it for migration)
- Adding migration queries to bring over orders/lines/transactions from MicroMaster

- Run the following query to check if there are corresponding products for the course runs in the local DB
   (If any course run doesn't have a product, then add them from Django admin)
  
```
     SELECT * FROM courses_courserun AS run 
     LEFT JOIN public.ecommerce_product AS product 
       ON run.id = product.object_id AND product.content_type_id = 60
    WHERE product.id is NULL
```
 - Create a local foreign data connection named 'micromasters'  (Only need this step when testing locally)
 ```
CREATE EXTENSION IF NOT EXISTS postgres_fdw;

CREATE SERVER IF NOT EXISTS micromaster_bridge
	FOREIGN DATA WRAPPER postgres_fdw
    OPTIONS (host 'micromasters-rc-db.cbnm7ajau6mi.us-east-1.rds.amazonaws.com', dbname 'micromasters', port '15432');

CREATE USER MAPPING IF NOT EXISTS FOR postgres
	SERVER micromaster_bridge
    OPTIONS (user 'micromasters', password '------');

DROP SCHEMA IF EXISTS micromasters CASCADE;
CREATE SCHEMA micromasters;
IMPORT FOREIGN SCHEMA public 
FROM SERVER micromaster_bridge INTO micromasters;
```

- Run the migration script
```
docker-compose run --rm web ./manage.py import_micromasters_data
```

- Verify orders/line/transactions are imported correctly (compare with MicroMaster RC using reference_number)

Orders
   
image

Line:

image

Transaction
image

Receipt on frontend
image

### CODE REVIEW

**Comment by rachellougee** [comment]
@rhysyngsun This is ready for review as the blocker is merged. Thank you

File: micromasters_import/management/commands/queries/007_import_transation.sql [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+    updated_on
+)
+SELECT
+    mm_receipt.data->>'transaction_id',
+    mm_order.total_price_paid,
+    mm_receipt.data,
+    mo_order.id,
+    CASE
+         WHEN mm_order.status IN ('refunded', 'partially_refunded') THEN 'refund'
+         ELSE 'payment'
```
rhysyngsun, line 18: You added the partially refunded state, so should this be mapped to that?

rhysyngsun (changes requested):

micromasters_import/management/commands/queries/007_import_transation.sql
rachellougee: This is different from the order state. This is the transaction state that only has two:
TRANSACTION_TYPE_REFUND = "refund"
TRANSACTION_TYPE_PAYMENT = "payment"
I thought the transaction type is just a way to identify if it's a refund or payment.  Should we add a third transaction type?

rhysyngsun: Oh, no you're right. I need more coffee. Carry on.

rhysyngsun (approved):
LGTM

Issue #809, opened 2022-08-09
flexible pricing: remove up/down arrows from income input field


pdpinch:
As a learner applying for financial assistance, I don't want to adjust my income in $1 increments. 

#### Designs and Mockups

![image](https://user-images.githubusercontent.com/430126/183742481-f1f0b90a-4197-435f-9c16-2973a6172e8f.png)

#### Acceptance Criteria:

- [ ] remove the up and down arrows

rachellougee:
@pdpinch Simply removing the up/down arrow is just a visual thing. 

<img width="481" alt="image" src="https://user-images.githubusercontent.com/3138890/185208420-cdfb3c44-c9f4-406d-810b-34b58cf30ac1.png">

I am still able to press my keyboard up and down to increment it by $1. Is that ok? 
Currently, it's set step="any" which I believe defaults to step = 1. What incremental should we set here besides removing the arrow? or do you think removing the arrow is enough?

PR #867, opened 2022-08-17
Hide up/down arrow for income number field

rachellougee:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/809

#### What's this PR do?
- Removes the up/down arrow for the income field. It is a number field and the step is set to any, which defaults to 1. The CSS is just to hide the arrow, so adding Javascript to disable the up/down from the keyboard (If its overkill, I can remove the JS)

image


asadali145 (approved):
LGTM!

Issue #873, opened 2022-08-18
Bug: inactive products are added to the cart


rhysyngsun:
It looks like even though we now enforce only one active product per courseware when the `/cart/add?course_id=X` url is used it will add _all_ products to the cart regardless of their active status.

The relevant code is here: https://github.com/mitodl/mitxonline/blob/main/ecommerce/views/v0/__init__.py#L520-L530

We should probably solve this by updating `ProductsQuerySet` to filter to active record only. That should address the query on `Product` in the above code and also prevent an issue like this in the future. Some areas of code may need to be updated to use the `Product.all_objects` manager because there may be some usages of `Product.objects` that depend on that return all products. Django-admin is almost definitely one of those but as part of fixing this other usages should be investigated too.

arslanashraf7:
Just mentioning for anyone reading through this,

As a result of [the fix PR for this](https://github.com/mitodl/mitxonline/pull/874) the other visible issues that were fixed except for the mentioned `/cart/add?course_id=X` issue:

1. The same behavior was applicable for product Ids too e.g. if we hit `/cart/add?product_id=X` the inactive product was getting added to the cart.
2. On the Course detail Page and Dashboard Page we'd see `Get Certificate` button in an enrolled free course, the associated PR also fixes this issue.

arslanashraf7:
Closed via #874

PR #874, opened 2022-08-19
fix: active products to cart only

arslanashraf7:
#873

#### What's this PR do?
Some of the refactors were quite straightforward, e.g. The places we were filtering products based on `is_active` should use `objects`. For the places where were won't care for `is-active` were likely to use `all_objects`.

- Makes the default Model manager for Product to return active products only
- For all products uses `all_objects`
- Refactors the usage of Product query set on multiple places to either use `all_objects` or `objects` based on its usage

- Majorly this relies on code coverage but testing is needed
- In general checking, everything works just like before in every place the product is used
    - I can think of `Cart`, `Dashboard`, `Course Details`, `Order History`, `Payment`, `Discounts`, `Flexible Pricing`

Try to reproduce the issue by putting the URL add URL manually with the course id:
- Uncheck `is_active` for a Product
- Now for this product run, something like `http://mitxonline.odl.local:8013/cart/add/?course_id=<courseware_id>` (make sure your courseware_id is quoted before being sent to URL An example of that is `course-v1:Arbisoft+CT_AR_001+1` it should be sent as `course-v1%3AArbisoft%2BCT_AR_001%2B1` it should add the product without this PR, With this PR, it shouldn't.

(Optional)

(Optional)

### DIFF
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -38,6 +38,8 @@ def valid_purchasable_objects_list():
 
 
 class ProductsQuerySet(models.QuerySet):
+    """Queryset to block delete and instead mark the items in_active"""
+
     def delete(self):
         self.update(is_active=False)
 
@@ -48,7 +50,7 @@ class ActiveUndeleteManager(models.Manager):
     # This can be used generally, for the models that have `is_active` field
     def get_queryset(self):
         """Getting the active queryset for manager"""
-        return ProductsQuerySet(self.model, using=self._db)
+        return ProductsQuerySet(self.model, using=self._db).filter(is_active=True)
 
 
 @reversion.register(exclude=("created_on", "updated_on"))
@@ -706,7 +708,7 @@ class Line(TimestampedModel):
         return (
             DiscountType.get_discounted_price(
                 discounts,
-                Product.objects.get(pk=self.product_version.object_id),
+                Product.all_objects.get(pk=self.product_version.object_id),
                 product_version=self.product_version,
             ).quantize(Decimal("0.01"))
             * self.quantity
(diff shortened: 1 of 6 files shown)


collinpreston:
@arslanashraf7, I'm having trouble reproducing the issue locally and on RC.  The error I'm seeing locally is `ecommerce.models.BasketItem.DoesNotExist: BasketItem matching query does not exist.`

arslanashraf7:
> @arslanashraf7, I'm having trouble reproducing the issue locally and on RC. The error I'm seeing locally is `ecommerce.models.BasketItem.DoesNotExist: BasketItem matching query does not exist.`

@collinpreston 

Did you try these steps (When on the main branch):
1. Make sure you have a fully functional purchasable product with course run (e.g. when you can see `Get Certificate` button on Course details/Dashboard)
2. Now, Deactivate this product - You should not see `Get Certificate` button anywhere so the user can't add it to the cart, But there is another path to add a product to the cart, which is by hitting the `cart/add?` URL directly in browser like `http://mitxonline.odl.local:8013/cart/add/?course_id=<courseware_id>`.
3. On main branch, you should notice that the product actually gets added to the cart, but it shouldn't have.

**Now move to this PR:**
1. Hit the `http://mitxonline.odl.local:8013/cart/add/?course_id=<courseware_id>` again and this time the product shouldn't be added to the cart instead it should show an empty cart.
2. `ecommerce.models.BasketItem.DoesNotExist: BasketItem matching query does not exist.` error occurs when we hit the `/cart` URL when there is no product in the cart and it tries to calculate the discounts - To fix this I added the fix [here in the PR](https://github.com/mitodl/mitxonline/pull/874/files#diff-120717d67e48941e59530fa72e6949e40f620956ae8c9845afa17570d78ed955R396-R400) to not calculate discounts when there is a basket with no products/line items in it.

collinpreston (approved):
Works for me and looks like a good change.  I would suggest adding a test similar to this (https://github.com/mitodl/mitxonline/blob/c05d008d374b41acc62acd82cdf837f6b4bdd90c/ecommerce/views_test.py#L92-L99) to enforce your changes around filter on `is_active=True`

asadali145 (approved):

Issue #872, opened 2022-08-18
Checkout: remove "clear discount" and a few other tweaks


pdpinch:
The behavior of "clear discount" is confusing and inconsistent, especially with financial assistance / flexible pricing. It should be removed. 

As a user, I can apply another coupon at any time. The most valuable one will always be chosen. 

#### Acceptance Criteria:

- [ ] Remove the "clear discount" link 
- [ ] consider remove the API as well

#### related issues 

- might want to combine a PR for this issue with issue #871

Issue #871, opened 2022-08-18
Checkout: standardize course display 


pdpinch:
The course card on the checkout page isn't very user friendly. It would be good if it could be more similar to the dashboard. 

#### Designs and Mockups

current Checkout page: 
- no course number; instead, a full course id 
- no label on the date
- no link to the course info

<img width="751" alt="image" src="https://user-images.githubusercontent.com/430126/185461262-061297d5-0c70-46b7-ab0d-d4b3a595e3e7.png">

current dashboard course card:
- course number
- "Starts" plus start date 
- link to "course details" 

<img width="727" alt="image" src="https://user-images.githubusercontent.com/430126/185461126-e735b39e-c4ea-4d85-90c6-f15c65cfe980.png">

#### Acceptance Criteria:

- [ ] display course number
- [ ] label the course date. Could be "starts" or "active"
- [ ] link to the course about page 

#### related issue 

- might want to combine this with another issue about the checkout page: #872

collinpreston:
@pdpinch clicking the course title (bold red text) and clicking "Course details" from the Dashboard appears to load the same page.  There is currently a PR to fix the course title link for the course in a cart, but should I also add the "Course details" link if it navigates to the same location?

collinpreston:
@pdpinch, this PR here (https://github.com/mitodl/mitxonline/pull/853) adds functionality for when a user clicks "Test External Course" from the Cart, they will be navigated to a new tab showing the course details (`http://mitxonline.odl.local:8013/courses/<course_URL>/`).  

Should that PR (853) actually redirect the user to `courses-qa.mitxonline.mit.edu`?  This is what clicking "Test External Course" from the Dashboard is doing.  
Or, is that other PR (853) correctly navigating the user to the course details page (`http://mitxonline.odl.local:8013/courses/<course_URL>/`)?

PR #877, opened 2022-08-19
872: checkout remove clear discount and a few other tweaks

collinpreston:
#### Pre-Flight checklist

  - [x] Desktop screenshots

https://github.com/mitodl/mitxonline/issues/872
https://github.com/mitodl/mitxonline/issues/871

#### What's this PR do?

- Removes the "Clear Discount" option from the Cart page.
- Removes the `clear_discount` API endpoint.
- Updates the Cart Item card to display additional fields (see https://github.com/mitodl/mitxonline/issues/871)

- Login as a user
- Add a course run to your Cart
- View your Cart and ensure that you no longer see the option to "Clear Discount"
- Add a discount, then ensure that you still no longer see the option to "Clear Discount"
- Attempt to apply another discount which is of lesser value than the current discount.  Ensure that the lesser value discount does not apply.
- Ensure that you see the UI additions specified in https://github.com/mitodl/mitxonline/issues/871

Screen Shot 2022-08-19 at 8 36 11


annagav:
@collinpreston there are some conflicts

frontend/public/src/components/CartItemCard.js
```diff
@@ -21,18 +27,19 @@ export class CartItemCard extends React.Component<Props> {
...
         </a>
       )
 
+    const courseDetail =
+      this.courseAboutLink("Course details", course)
+
     const readableId =
       course !== undefined
-        ? purchasableObject.readable_id
+        ? purchasableObject.readable_id.split('+')[1]
```
annagav, line 42: I think here we can use `course.readable_id` instead

annagav (changes requested):
Code looks good. Just a minor comment.

frontend/public/src/components/CartItemCard.js
collinpreston: @annagav, from my understanding of the issue ticket the intention was to only display the course number and not the full id, which I assume the `readable_id` was showing the full id since that's what was in place.  Maybe I overlooked something somewhere.

collinpreston: @annagav where does the formatting f `readable_id` take place?  When I look at RC, the course readable_id's that I see are all in the format of `XXXX:XXX+XXXX`.

annagav: it's just a field on the model of `Course` and `Program`

collinpreston: @annagav I'm confused though.  Does the `readable_id` provide any additional formatting so it only shows the course number like shown in https://github.com/mitodl/mitxonline/issues/871?

annagav:
The readable_id is not showing yet. I think you need to add 'readable_id' field to `CourseProductPurchasableObjectSerializer` in ecommerce/serializers.py

frontend/public/src/components/CartItemCard.js
rachellougee: If I understand it correctly, we want to display readable_id from the course, which is the format like  [course-v1:MITx+14.740x](https://rc.mitxonline.mit.edu/admin/courses/course/6/change/).

The purchasableObject.readable_id here refers to the courserun id (course-v1:MITx+14.740x+2T2022) https://rc.mitxonline.mit.edu/admin/courses/courserun/7/change/ if product is courserun

purchasableObject.readable_id.split('+')[1] will output '14.740x' if courserun id is like course-v1:MITx+14.740x+2T2022, which doesn't seem to be correct? I could be wrong though

collinpreston: @rachellougee that could be, I guess I was looking at the screenshots included in issue #871 where currently `course-v1:Arbisoft+ARB_TEST_CKT_001+1` is shown, and we want to only show `ARB_TEST_CKT_001`.  But looking into RC I don't see anywhere that only `ARB_TEST_CKT_001` is stored or provided without some sort of pre-processing.

rachellougee: @collinpreston Yeah, I noticed that. I think the test data on RC is causing confusion. The course corresponds to course-v1:Arbisoft+ARB_TEST_CKT_001+1 is https://rc.mitxonline.mit.edu/admin/courses/course/14/change/, which isn't ARB_TEST_CKT_001.  Not sure what the course number means, Is it course id? or does it has to be from the second part of courserun id?

rachellougee: Just for the record, course_number is extracted from courserun.courseware_id  https://github.com/mitodl/ol-django/blob/main/src/mitol/openedx/utils/courses.py. so I was wrong

rachellougee (approved):

frontend/public/src/components/CartItemCard.js
collinpreston: All good.  Thanks for thinking through it with me @rachellougee

PR #878, opened 2022-08-19
Check for program page before checking for child pages

collinpreston:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

#### What's this PR do?
When trying to find the financial assistance form which may be a child of a program page, before checking a program page's child pages we need to make sure that the program page we are checking actually exists!


jkachel (approved):
Seems to work for me - I added a new program and course and a flex pricing form that was under the course (rather than the program so it matches RC) and the dashboard loaded successfully, and the financial assistance links appear as expected.

PR #880, opened 2022-08-19
Fixes some issues with the Fastly API code

### LINKED ISSUE #749: CMS: after updating a page, the CDN cache is not cleared

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #749 

#### What's this PR do?

Fixes some bugs in the original PR https://github.com/mitodl/mitxonline/pull/838 namely:
- Doesn't use the Fastly API library anymore (the stable version is outdated and the updated version is pre-release and doesn't allow for single pages to be purged)
- Adds some better logging if it does fail (which it definitely will locally without proper setup)

This also adds in a management command to purge the entire cache. Purging everything results in a hard purge, where a single page request is handled with a soft purge. See [the documentation](https://developer.fastly.com/reference/api/purging/) for info but the gist of it is a hard purge removes everything right now where a soft one doesn't. 

To properly test, you will need to have your test instance set up to respond as `rc.mitxonline.mit.edu` on port 80. This involves:
- Adding a `HOSTS` file entry for that hostname
- Setting `MITX_ONLINE_BASE_URL` to `http://rc.mitxonline.mit.edu` in your `.env` file
- Either modifying the `docker-compose.yml` file to redirect local port 80 to 8013 in the `nginx` container, making a `docker-compose-override.yml` that does the same, or doing some other sort of proxy thing so it responds on port 80. (This may not be strictly necessary but skipping this step means that generated URLs may be wrong.) 

The cache purge will be requested whenever a page in the CMS is published or moved, so, to test, just do that - it should queue a task and Celery should go do it. If you're using the RC URL, it should be successful; otherwise, you should get a `421 Misdirected Request` error. 

The hard purge can be triggered by running `manage.py purge_fastly_cache`. The log output from this should be pretty similar to the single page call.

### DIFF
--- a/cms/management/commands/purge_fastly_cache.py
+++ b/cms/management/commands/purge_fastly_cache.py
@@ -1,0 +1,13 @@
+"""Purges all pages from the Fastly cache."""
+from django.core.management.base import BaseCommand
+
+from cms.tasks import queue_fastly_full_purge
+
+
+class Command(BaseCommand):
+    """Purges all pages from the Fastly cache."""
+
+    help = __doc__
+
+    def handle(self, *args, **options):
+        queue_fastly_full_purge.delay()
--- a/cms/tasks.py
+++ b/cms/tasks.py
@@ -6,8 +6,44 @@ from main.settings import (
 )
 from cms.models import Page
 from urllib.parse import urljoin, urlparse
-import fastly
 import logging
+import requests
+
+
+def call_fastly_purge_api(relative_url):
+    """
+    Calls the Fastly purge API. (We aren't using the official Fastly SDK here
+    because it doesn't work for this - the version of it that works with the
+    current API only allows you to purge *everything*, not individual pages.)
+
+    Args:
+        - relative_url  The relative URL to purge.
+    Returns:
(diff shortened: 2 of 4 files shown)


rachellougee (approved):
LGTM. Manual test works as well

Issue #869, opened 2022-08-18
Add unique key and new columns to ecommerce_line and ecommerce_transaction


rachellougee:
Currently, there aren't any unique constraints to ecommerce_line and ecommerce_transaction table. It raised issues for migrating the orders from MicroMaster. See https://github.com/mitodl/mitxonline/issues/657

For ecommerce_line, 
- [ ] Add a unique key on (order_id,purchased_content_type_id,purchased_object_id)
- [ ] Dedup any existing data

For ecommerce_transaction,
- [ ] Add two new columns `transaction_id` and `transaction_type` (It will be payment or refund)
- [ ] Populate it with the id from the data column 

> - [ ] For refund, it might be data->'id' 
> - [ ] For other transactions, it's data->'transaction_id'

- [ ] Dedup any existing data
- [ ] Make `transaction_id` unique

PR #882, opened 2022-08-19
869 add unique key to ecommerce line and ecommerce transaction

rachellougee:
- [x] Migrations
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/869

#### What's this PR do?

- Adding uniqueness to `ecommerce_line` and its related code changes to prevent adding duplicates
- Adding a new column `transaction_id` to `ecommerce_transaction`(transaction_type exists), populating it with existing data, and finally making it unique, so 3 migrations for this column change
- Updating the existing automated tests due to unique key changes

- Before running the migration, dedup any existing data that violates the uniqueness:

(I've already cleaned up the dups on the RC database, but I will double-check before merging to RC)
``` 
SELECT COUNT(*), order_id, purchased_content_type_id, purchased_object_id FROM ecommerce_line GROUP BY order_id,purchased_content_type_id, purchased_object_id HAVING COUNT(*) >1;

SELECT COUNT(*) FROM ecommerce_transaction WHERE data->>'transaction_id' is null AND data->>'id' is null;

```

- After fixing the existing data, then run the migrate
- transaction_id should be populated for all the _existing_ transaction http://mitxonline.odl.local:8013/admin/ecommerce/transaction/

- Go to frontend, and enroll in a course. It should go through.

image

- Check transaction admin (transaction Id should be populated for _new_ record)

image

- Go to Fulfilled orders from admin, and request a refund, you should see transaction_id is populated for _new_ refund as well

image

(Optional)

(Optional)

(Optional)


ecommerce/models.py
```diff
@@ -415,10 +414,14 @@ class FulfillableOrder:
     """class to handle common logics like fulfill, enrollment etc"""
 
     def create_transaction(self, payment_data):
-        self.transactions.create(
-            data=payment_data,
-            amount=self.total_price_paid,
-        )
+        # Since transaction_id is unique identifier, only create transaction if transaction_id exists
+        transaction_id = payment_data.get("transaction_id")
+        if transaction_id:
```
collinpreston, line 419: If `transaction_id` do we still want to continue with the enrollment of the user?  Are there any instances where `payment_data` does not contain the `transaction_id`?

collinpreston: We may want to create some log output or something traceable when transaction_id is `None`.

rachellougee: I am not sure if we need to block enrollment because of transaction_id from the API response.

There are only two types of transactions: payment and refund. This code is for payment, the refund is handled separately. For payment, my understanding is that it has to have transaction_id, otherwise it can't be recorded. This is also to block inserting some meaningless or manually added transaction data like I saw RC.

Yeah, I can add some log here if transaction_id is None 

 @rhysyngsun should we block the subsequent enrollment if no transaction_id? I haven't seen any payments that don't have transaction_id in them.

rhysyngsun: Talked to @rachellougee about this and as long as the column is `NOT NULL` then that should protect us from null values since it'll correctly raise an exception. I think she was planning to wrap that call with some exception handling to make the error messaging a bit more specific though.

jkachel (approved):
Code looks ok to me 👍

Issue #693, opened 2022-07-13
Story: As a learner I want to be able to share my certificate with others


rhysyngsun:
As a learner, I need to be able to view my certificate, print it, and share it with others.

As a user who is not logged in, I'd like to be able to view a learner's certificate and confirm that they passed.

#### Acceptance Criteria:

- [ ] logged-in user can share their own certificate with Facebook, Twitter, and LinkedIn
- [ ] logged in user can print their certificate
- [ ] anonymous user can view a certificate
- [ ] permalink

Issue #692, opened 2022-07-13
Story: As a learner I want a page where I can see the certificate I earned


rhysyngsun:
As a learner, I need to be able to view my certificate at a unique URL that is not guessable, and is not indexed by Google. 

#### Acceptance Criteria:

- [ ] permalink using a uuid, similar to xPRO

#### related issues

- #691 
- #693

PR #892, opened 2022-08-24
Learner and anonymous certificate view- issue #692 #693

aliraza-abbasi:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [ ] Migrations
- [ ] Testing
  - [x] Code is tested
- [ ] Settings

Issue #692 #693 

- Enroll in course and create certificate (can be manually created by DB entry)
- Open Learner Dashboard
- Open course look if View Certificate button is there
- Click the View Certificate button, it should open a new certificate window with share options

#### Screenshots
Screenshot 2022-08-24 at 4 58 36 PM
Screenshot 2022-08-24 at 4 59 56 PM
Screenshot 2022-08-24 at 5 02 03 PM
Screenshot 2022-08-24 at 5 02 14 PM

### DIFF
--- a/cms/api.py
+++ b/cms/api.py
@@ -13,6 +13,7 @@ from wagtail.core.models import Page, Site
 from cms import models as cms_models
 from cms.exceptions import WagtailSpecificPageError
 from cms.models import Page
+from cms.constants import CERTIFICATE_INDEX_SLUG
 
 
 log = logging.getLogger(__name__)
@@ -166,6 +167,52 @@ def ensure_program_product_index() -> cms_models.ProgramIndexPage:
         page = Page.objects.get(id=page_id)
         page.move(program_index, "last-child")
     return program_index
+
+
+def ensure_signatory_index() -> cms_models.ProgramIndexPage:
+    """
+    Same as ensure_product_index, but operates on programs instead.
+    """
+    home_page = get_home_page()
+    signatory_index = Page.objects.filter(
+        content_type=ContentType.objects.get_for_model(cms_models.SignatoryIndexPage)
+    ).first()
+    if not signatory_index:
+        signatory_index = cms_models.SignatoryIndexPage(title="Signatories")
+        home_page.add_child(instance=signatory_index)
+        signatory_index.save_revision().publish()
+
+    if signatory_index.get_children_count() != cms_models.SignatoryPage.objects.count():
+        for signatory_page in cms_models.SignatoryPage.objects.all():
+            signatory_page.move(signatory_index, "last-child")
+        log.info("Moved signatory pages under signatory index page")
+    return signatory_index
+
+
+def ensure_certificate_index() -> cms_models.ProgramIndexPage:
+    """
+    Same as ensure_product_index, but operates on programs instead.
+    """
+    home_page = get_home_page()
+    certificate_index = Page.objects.filter(
+        content_type=ContentType.objects.get_for_model(cms_models.CertificateIndexPage)
+    ).first()
+
+    if certificate_index and certificate_index.slug != CERTIFICATE_INDEX_SLUG:
+        certificate_index.slug = CERTIFICATE_INDEX_SLUG
+        certificate_index.save()
+
+    if not certificate_index:
+        cert_index_content_type, _ = ContentType.objects.get_or_create(
+            app_label="cms", model="certificateindexpage"
+        )
+        certificate_index = cms_models.CertificateIndexPage(
+            title="Certificate Index Page",
+            content_type_id=cert_index_content_type.id,
+            slug=CERTIFICATE_INDEX_SLUG,
+        )
+        home_page.add_child(instance=certificate_index)
+    return certificate_index
 
 
 def get_wagtail_img_src(image_obj) -> str:
--- a/cms/api.py
+++ b/cms/api.py
@@ -169,9 +169,9 @@ def ensure_program_product_index() -> cms_models.ProgramIndexPage:
     return program_index
 
 
-def ensure_signatory_index() -> cms_models.ProgramIndexPage:
+def ensure_signatory_index() -> cms_models.SignatoryIndexPage:
     """
-    Same as ensure_product_index, but operates on programs instead.
+    Ensures that an index page has been created for signatories.
     """
     home_page = get_home_page()
     signatory_index = Page.objects.filter(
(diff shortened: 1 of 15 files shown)


cms/api.py
```diff
@@ -168,6 +169,52 @@ def ensure_program_product_index() -> cms_models.ProgramIndexPage:
...
+
+
+def ensure_certificate_index() -> cms_models.ProgramIndexPage:
+    """
+    Same as ensure_product_index, but operates on programs instead.
+    """
+    home_page = get_home_page()
+    certificate_index = Page.objects.filter(
+        content_type=ContentType.objects.get_for_model(cms_models.CertificateIndexPage)
+    ).first()
```
arslanashraf7, line 199: Can this be replaced with:

`    certificate_index = cms_models.CertificateIndexPage.objects.first()
`

cms/api.py
```diff
@@ -168,6 +169,52 @@ def ensure_program_product_index() -> cms_models.ProgramIndexPage:
...
+    if not certificate_index:
+        cert_index_content_type, _ = ContentType.objects.get_or_create(
+            app_label="cms", model="certificateindexpage"
+        )
+        certificate_index = cms_models.CertificateIndexPage(
+            title="Certificate Index Page",
+            content_type_id=cert_index_content_type.id,
+            slug=CERTIFICATE_INDEX_SLUG,
+        )
+        home_page.add_child(instance=certificate_index)
```
arslanashraf7, line 214: Shouldn't we move the existing certificate pages under this index page once created? e.g. something like (from xPRO):

```
    if (
        certificate_index.get_children_count()
        != cms_models.CertificatePage.objects.count()
    ):
        for cert_page in cms_models.CertificatePage.objects.all():
            cert_page.move(certificate_index, "last-child")
        log.info("Moved certificate pages under certificate index page")

```

cms/models.py
```diff
@@ -809,6 +809,16 @@ class CoursePage(ProductPage):
...
 
+    def _get_child_page_of_type(self, cls):
+        """Gets the first child page of the given type if it exists"""
+        child = self.get_children().type(cls).live().first()
+        return child.specific if child else None
+
+    @property
+    def certificate_page(self):
+        """Gets the certificate child page"""
+        return self._get_child_page_of_type(CertificatePage)
```
arslanashraf7, line 820: We should define it into the ProductPage instead, will be reusable for programs too.

cms/serializers_test.py
```diff
@@ -147,7 +147,7 @@ def test_serialize_course_page_with_flex_price_form_as_program_child(
     program = ProgramFactory()
     program_page = ProgramPageFactory(program=program)
     FlexiblePricingFormFactory(parent=program_page)
-    course = CourseFactory(program=program)
+    course = CourseFactory(program=program, page=None)
```
arslanashraf7, line 150: Any specific reason we want the page to be sent as `None`?

courses/serializers.py
```diff
@@ -329,9 +338,45 @@ def create(self, validated_data):
...
 
+    def get_certificate(self, enrollment):
+        """
+        Resolve a certificate for this enrollment if it exists
+        """
+        # When create method is called it returns list object of enrollments
+        if isinstance(enrollment, list):
+            for en in enrollment:
+                self.get_certificate(en)
+        else:
```
arslanashraf7, line 349: Let's take a look at what's happening in `create` see line#336. You will notice that we only have one run for this but the enrollment method take a list and returns a list so we convert that single run into list like `[run]`. We get a list in return of create but it is always going to be a list of one element.

Insread, you can pick that enrollment `enrollments[0]` in your case.

frontend/public/scss/dashboard.scss
```diff
@@ -108,7 +108,7 @@ $program-badge-bg: #DBF2F3;
     a {
       text-decoration: underline;
       color: $light-gray;
-      margin: auto 0;
+      margin: auto 2px;
```
arslanashraf7, line 111: Can we increase it to 5px? how does that look?

arslanashraf7 (changes requested):

cms/api.py
aliraza-abbasi: This will not work in xOnline. The certificates are child of courses if we move them in child of certificate index page they got removed from course and then certificate learner view and anonymous view does not work.

cms/serializers_test.py
aliraza-abbasi: The `page` is added to CourseFactory as discussed earlier for testing the Enrollment Serializer. But it caused the failure of other tests because previously the `page` attribute (which is referred to CoursePage) was None by default and created after CoursePageFactory.create(). Now after adding `page` attribute in CourseFactory the `page` attribute is created as a CoursePage as default. This caused the error `Course page with this Course already exists.` when we call CoursePageFactory. So passing the page None allow to call CoursePageFactory

cms/api.py
arslanashraf7: What is the parent of certifiicates in xPRO? My perception was that they are same for xPRO and MITxOnline.

aliraza-abbasi: I've tested this in xPro. It is also removing the certificate learner view and anonymous view from xPro as well. This peace of code should be removed from xPro as well

aliraza-abbasi: The certificate moved from Course child to Certificate Index Page child. And then you cannot edit this page and cannot view the certificate. by moving the certificate again manually from Certificate Index Page to Course Child page everything started working again.

arslanashraf7: Well, this makes sense now Since the ensure_index_pages are only used in `configure_wagtail` that's run with the fresh installation of the app locally. This is the reason we might not have seen any issues though because in a fresh installation there are no course associates certificate pages.

aliraza-abbasi: But this can be a problematic if the command started by any need in future it'll move all the certificates from courses to Certificate index page.

courses/serializers_test.py
```diff
@@ -271,8 +274,19 @@ def test_serialize_program_enrollments(settings, receipts_enabled):
...
+                CourseRunFactory.create(
+                    course=CourseFactory.create(program=program, position_in_program=2)
+                ),
+                CourseRunFactory.create(
+                    course=CourseFactory.create(program=program, position_in_program=1)
+                ),
+                CourseRunFactory.create(
+                    course=CourseFactory.create(program=None, position_in_program=None)
+                ),
+            ]
```
arslanashraf7, line 288: Is there any specific reason for this?

aliraza-abbasi: The CourseRunFactory generates the CourseFactory with `page` as None (required for other tests) as discussed above. So here we need course with `page` not as None. To do so the CourseFactory is called so the `page` element can be created to pass EnrollmentSerializer test.

arslanashraf7 (changes requested):
I was able to test it able to view my certificate but I can point out a few problems with that.

1- The certificate page shows the Top App bar in both anonymous and authenticated cases which i think it shouldn't.

Sharing the screenshots below:

Screenshot 2022-08-25 at 5 00 49 PM
Screenshot 2022-08-25 at 5 01 00 PM

main/templates/base.html
```diff
@@ -31,10 +32,12 @@
     <a href="#main" class="sr-only sr-only-focusable">Skip to main content</a>
     {% include "partials/gtm_body.html" %}
     {% hijack_notification %}
+    {% if not request.path|startswith:'/ecommerce/bulk/,/certificate/' %}
```
arslanashraf7, line 35: I don't think we have bulk functionality in MITxOnline. @aliraza-abbasi

arslanashraf7 (approved):
Looks great 👍 

Though, We should also add entries for django-robots for certificates.

Issue #658, opened 2022-07-05
MM Data Migration: legacy final grades from MicroMasters


rhysyngsun:
We need to bring over the legacy grades for DEDP, these are spread out over a couple different models on MicroMasters:

- `FinalGrade`: the grade for a course run
- `ProctoredExamGrade`: the grade for an exam a learner sat for

Whenever a learner gets a new record of for either model, a `CombinedFinalGrade` is created or updated (https://github.com/mitodl/micromasters/blob/40deb7afba5cc93b2514fc2dfa050fdb96b1df54/grades/api.py#L299-L326). The combined grade is only at the `Course` level and it's continuously updated so it's not useful in populating historical grades. It's mentioned here for context and because we'll use a similar algorithm.

So the main problem we face here is that historically when a learner takes an exam and when they take a course run are not guaranteed to intersect however in MITx Online they _always_ will, so we need to figure out a reasonable way to map these over.

I think what is probably the most accurate would be to walk the learner's `FinalGrade` and `ProctoredExamGrade` records and compute what their `CombinedFinalGrade` would've been calculated as at that point in time and give them a grade for the most recent run combining the best of both in MITx Online.

#### Use Cases

- Learners that have taken at least one course run and one exam
  - We'll use an adapted version of the algorithm for `CombinedFinalGrade` to import these
- Learners that have taken at least one course run but no exams
  - We need to decide how to treat these grades as there's no exam grade to account for. 

#### Approach

We need to decide how to create `CourseRunGrade` records in MITx Online. Since these are rooted in runs we need to make sure that these records match up to enrollments. `FinalGrade` in MicroMasters follows this convention as well, so that's what we'll end up basing our approach off of.

rachellougee:
@pdpinch @Ferdi I've run the number for DEDP program grades on production (related to the second use case described in the above issue)

- There are 40955 learners who have taken at least one course run but never taken the exam
- There are 2264 learners who have _paid_ for at least one course run (enrollment mode is either _verified_ or _honor_), but never taken the exam

Per discussed with @rhysyngsun last week, we are currently only bringing in DEDP grades in the first use case in the ticket

> Learners that have taken at least one course run and one exam

Ferdi:
>  There are 40955 learners who have taken at least one course run but never taken the exam

Can we get a distribution by year ?  

> There are 2264 learners who have paid for at least one course run (enrollment mode is either verified or honor), but never taken the exam

[...]

PR #893, opened 2022-08-24
add query to migrate grades from MicroMaster

rachellougee:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/658

#### What's this PR do?

- Adding a migration script to bring over grades from MicroMaster
    - Limit to learners who have taken the exam and passed

- Modify the enrollment query to those who have taken the exam and passed

- Similar setup foreign data connection as https://github.com/mitodl/mitxonline/pull/859
- It needs the mapped course id http://mitxonline.odl.local:8013/admin/micromasters_import/courseid/
- Run docker-compose run --rm web ./manage.py import_micromasters_data
- Check the imported course run grade http://mitxonline.odl.local:8013/admin/courses/courserungrade/

**Test result**

- Migrated 1 enrollment from MicroMaster RC

image

- Migrated 2 grades from MicroMaster RC 
  (Note that there isn't a migrated enrollment for MITx/14.73x/2013_Spring because there is no user enrollment for this course on MicroMaster)

image

image

Sources of the grade for "MITx/14.73x/2013_Spring"

-  [best proctored exam grade](https://micromasters-rc.odl.mit.edu/admin/grades/proctoredexamgrade/9/change/)

 - [Best Final grade ](https://micromasters-rc.odl.mit.edu/admin/grades/finalgrade/215/change/)
  
Formula: (0.68 * 100 * **0.4** + 46 * **0.6**) = 54.8
Passed: True (because grade and exam both passed on MicroMaster

The migrated grade is a bit weird but I assume that's just test data?

### DIFF
--- a/micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
+++ b/micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
@@ -1,0 +1,103 @@
+----This script is to migrate the grades for course run from MicroMaster:
+----  It computes score from FinalGrade and ProctoredExamGrade using formula - course run grade * 0.4 + exam grade * 0.6
+----  grades need to exist in both FinalGrade (course run) and ProctoredExamGrade (exam)
+----  grades need to be 'passed'
+
+WITH courses_have_exam AS (
+    SELECT
+        DISTINCT mm_examrun.course_id
+    FROM micromasters.exams_examrun AS mm_examrun
+    JOIN micromasters.courses_course AS mm_course
+        ON mm_examrun.course_id = mm_course.id
+),
+ --- this is to compute the best grade learner gets for the course runs they have enrolled
+best_courserun_grades AS (
+    SELECT
+        mm_grade.user_id,
+        mm_grade.course_run_id,
+        MAX(mm_grade.grade) AS max_grade
+    FROM micromasters.grades_finalgrade AS mm_grade
+    JOIN micromasters.courses_courserun AS mm_courserun
+        ON mm_grade.course_run_id = mm_courserun.id
+    JOIN courses_have_exam
+        ON mm_courserun.course_id = courses_have_exam.course_id
+    WHERE mm_grade.status = 'complete'
+      AND mm_grade.passed = True
+    GROUP BY mm_grade.user_id, mm_grade.course_run_id
+),
+--- this is to compute the best exam grades learner get for every course run in the exam they signed up for
+best_exam_grades AS (
+    SELECT
+        mm_exam_grade.user_id,
+        mm_exam_grade.course_id,
+        mm_courserun.id AS course_run_id,
+        MAX(mm_exam_grade.score) AS max_score
+    FROM micromasters.grades_proctoredexamgrade AS mm_exam_grade
+    JOIN micromasters.courses_courserun AS mm_courserun
+       ON mm_exam_grade.course_id = mm_courserun.course_id
+    JOIN courses_have_exam
+       ON mm_exam_grade.course_id = courses_have_exam.course_id
+    WHERE mm_exam_grade.passed = true
+    GROUP BY mm_exam_grade.user_id, mm_courserun.id, mm_exam_grade.course_id
+),
+combined_final_grades AS (
+    SELECT
+        best_grade.user_id,
+        best_grade.course_run_id,
+        ROUND(CAST(best_grade.max_grade * 100 * 0.4 + best_exam.max_score * 0.6 AS numeric), 1) AS calculated_grade,
+        mm_courserun.course_id,
+        mm_courserun.edx_course_key,
+        mm_grade.created_on,
+        mm_grade.updated_on,
+        True AS passed  -- because both course run and exam are passed
+    FROM best_courserun_grades AS best_grade
+    JOIN best_exam_grades AS best_exam
+       ON (best_grade.course_run_id = best_exam.course_run_id
+           AND best_grade.user_id = best_exam.user_id)
+    JOIN micromasters.courses_courserun AS mm_courserun
+       ON best_grade.course_run_id = mm_courserun.id
+    --- The following joins are to get the corresponding passed and metadate fields matched with previous aggregated CTEs
+    JOIN micromasters.grades_finalgrade AS mm_grade
+       ON (best_grade.course_run_id = mm_grade.course_run_id
+           AND best_grade.user_id = mm_grade.user_id
+           AND best_grade.max_grade = mm_grade.grade)
+    JOIN micromasters.grades_proctoredexamgrade AS mm_exam
+       ON (best_exam.course_id = mm_exam.course_id
+           AND best_exam.user_id = mm_exam.user_id
+           AND best_exam.max_score = mm_exam.score)
+)
+
+INSERT INTO public.courses_courserungrade(
+   created_on,
+   updated_on,
+   passed,
+   grade,
+   letter_grade,
+   set_by_admin,
+   course_run_id,
+   user_id
+)
+SELECT
+  combined_grades.created_on,
+  combined_grades.updated_on,
+  combined_grades.passed,
+  combined_grades.calculated_grade,
+  CASE
+      WHEN combined_grades.calculated_grade >= 82.5 THEN 'A'
+      WHEN combined_grades.calculated_grade >= 65 THEN 'B'
+      WHEN combined_grades.calculated_grade >= 55 THEN 'C'
+      WHEN combined_grades.calculated_grade >= 50 THEN 'D'
+      ELSE 'F'
+  END AS letter_grade,
+  False AS set_by_admin,
+  mo_courserun.id,
+  mo_user.id
+FROM combined_final_grades AS combined_grades
+JOIN public.courses_courserun AS mo_courserun
+    ON combined_grades.edx_course_key = mo_courserun.courseware_id
+JOIN micromasters.social_auth_usersocialauth AS mm_social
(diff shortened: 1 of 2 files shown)


annagav (approved):
Works for me 👍

rachellougee:
@rhysyngsun @pdpinch Do you want to take a look before I merge?

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
```diff
@@ -0,0 +1,106 @@
...
+       ON mm_exam_grade.course_id = mm_courserun.course_id
+    JOIN courses_have_exam
+       ON mm_exam_grade.course_id = courses_have_exam.course_id
+    GROUP BY mm_exam_grade.user_id, mm_courserun.id, mm_exam_grade.course_id
+),
+combined_final_grades AS (
+    SELECT
+        best_grade.user_id,
+        best_grade.course_run_id,
+        ROUND(CAST(best_grade.max_grade * 100 * 0.4 + best_exam.max_score * 0.6 AS numeric), 1) AS calculated_grade,
```
pdpinch, line 45: Is it true that we don't have a materialized combined_final_grade in the micromasters database? I feel like someone told me that, but it always surprises me.

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
```diff
@@ -0,0 +1,106 @@
...
+    JOIN micromasters.courses_courserun AS mm_courserun
+       ON mm_exam_grade.course_id = mm_courserun.course_id
+    JOIN courses_have_exam
+       ON mm_exam_grade.course_id = courses_have_exam.course_id
+    GROUP BY mm_exam_grade.user_id, mm_courserun.id, mm_exam_grade.course_id
+),
+combined_final_grades AS (
+    SELECT
+        best_grade.user_id,
+        best_grade.course_run_id,
```
pdpinch, line 44: Are we capturing two different course_run_ids? I see `best_grade.course_run_id` here and then `mm_courserun.course_id` below. Sorry my SQL is so weak, I'm not clear about what the source of these two are, nor how they might be used.

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
```diff
@@ -0,0 +1,106 @@
...
+  combined_grades.updated_on,
+  combined_grades.passed,
+  combined_grades.calculated_grade,
+  CASE
+      WHEN combined_grades.calculated_grade >= 82.5 THEN 'A'
+      WHEN combined_grades.calculated_grade >= 65 THEN 'B'
+      WHEN combined_grades.calculated_grade >= 55 THEN 'C'
+      WHEN combined_grades.calculated_grade >= 50 THEN 'D'
+      ELSE 'F'
+  END AS letter_grade,
```
pdpinch, line 92: Again, is this not materialized already?

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
```diff
@@ -0,0 +1,106 @@
...
+)
+
+INSERT INTO public.courses_courserungrade(
+   created_on,
+   updated_on,
+   passed,
+   grade,
+   letter_grade,
+   set_by_admin,
+   course_run_id,
```
pdpinch, line 78: Maybe the same question again -- which course run id will this be? The online course, the exam run, or something else?

pdpinch (changes requested):
Thanks for the nudge. I have some questions. 

I think it would be good to have Nathan's eyes on this as well.

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
rachellougee: `best_grade.course_run_id` comes from [grades_finalgrade](https://micromasters-rc.odl.mit.edu/admin/grades/finalgrade/1/change/) . In order to match MicroMaster course_run_id with MITxOnline, we need the corresponding _edx_course_key_ so it can be inserted into MITxOnline. it's like this:
`MM grades_finalgrade.course_run_id => MM courses_courserun.edx_course_key => MO courses_courserun.courseware_id => MO.courses_courserun.id`

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
rachellougee: There is a grade field in CombinedFinalGrade in micromasters, but it's at the course level, not the course run level.

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
rachellougee: Right. No letter grade in micromaster

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
rachellougee: The course_run_id here is the primary key of the course run from MITxOnline.
For example 2 in the URL https://rc.mitxonline.mit.edu/admin/courses/courserun/2/change/.

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
pdpinch: Did you consider migrating that calculated grade, instead of recalculating? I am pretty sure you have the correct formula here, but I feel like there's always some chance that the data in MM is different from this calculation.

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
pdpinch: Ok. That makes sense.

pdpinch:
What happens when the user took a course run that was delivered on mitxonline? Their `final_grade` currently stored on MM should match the `final_grade` currently stored on mitxonline. But then we're going to calculate the combined_final_grade, and then override the grade stored on mitxonline?

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
rhysyngsun: @pdpinch I hope maybe this helps clarify what we're trying to do here so you can confirm this is a valid approach. 

To start, there's a difference in how MicroMasters and MITx Online represent grade data in the database because MicroMasters was historically tailored to how DEDP operated these courses and MITx Online departs from that in a way that doesn't map one-to-one:

#### MicroMasters

- Learners can have multiple grades for a course, which are stored in `grades_finalgrade`. These are mapped to a course run.
- Learners can take one or more exams, separately from when they take the course itself. These are stored in `grades_proctoredexamgrade`.
- There is only **one** combined grade tracked in the database, at the course level. This grade is recalculated every time a record in `grades_finalgrade` or `grades_proctoredexamgrade` is created or updated. This is calculated as the weighted combination of the best course run and exam grades.

#### MITx Online

- Courses and exams have been combined on the openedx side and as a result the grades coming back from openedx are the combined and weighted grade.

Additionally, it was my understanding that MITx Online would be handling the grades in the following manner:
- There will be no computation done on the MITx Online side in terms of the "best grade of runs and exams" logic that MicroMasters had since the grades are opaque to us.
- If the course team wants to support some kind of "forwarding of the best grade to future runs" to support something similar to the MicroMasters behavior, that would be their responsibility to manage through openedx functionality and not a feature of MITx Online itself.
- As a result, when we eventually generate program certificates, we'll simply take the best course run grade.

Assuming all the above holds true, we'd been assuming we wanted to made the legacy grades into MITx Online while keeping to MicroMasters' grade calculation logic:

- For each course run `FinalGrade` in Micromasters, create a corresponding `CourseRunGrade` in MITx Online, weighted using the best exam grade.
- This preserves MicroMasters' legacy logic while making it compatible with MITx Online's data structure and also for any future exam attempts a learner might take because MicroMasters' `max(final_grades) * 0.4 + max(exam_grades) * 0.6` is functionally the same as just taking the `max(course_run_grades)` of legacy run grades that have been migrated as `course_run_grade * 0.4 + max(exam_grades) * 0.6`.
- This keeps the query simple because it doesn't need to do a subquery to also determine the best course run grade. Since we'll just end up using the best course run grade anyway on MITx Online, the lower grades get shadowed anyway so there's no harm in bringing them over.

rhysyngsun:
> What happens when the user took a course run that was delivered on mitxonline? Their `final_grade` currently stored on MM should match the `final_grade` currently stored on mitxonline. But then we're going to calculate the combined_final_grade, and then override the grade stored on mitxonline?

Won't be a problem as there are no grades stored on mitx online production at this time. It is worth noting that we might want to hide grades on MicroMasters before the next round of courses end because then MicroMasters will compute an incorrect incorrect grades for learners who have taken an exam both on MITx Online this semester and have an existing legacy exam grade.

micromasters_import/management/commands/queries/008_import_combined_final_grade.sql
rachellougee: Yes, sorry I wasn't precise. As Nathan said, it's implemented in python, not available in any table

rachellougee:
@rhysyngsun and @pdpinch I've made changes to both grade and enrollment to limit to those who took the exam and passed. I ran the script and put the test result in the description. Please take a look and let me know if it makes sense to you. Thanks

rhysyngsun (approved):
LGTM

pdpinch (approved):
lgtm, thank you

Issue #881, opened 2022-08-19
Certificates: management command for creating and revoking course certificates


pdpinch:
From time to time, course teams would like to grant, or revoke a certificate manually. 

This will also be useful for manual testing scenarios where we need a certificate. 

#### acceptance criteria

- Inputs:
  - courserun readable_id
  - email (optional; default is to attempt certificates for all _enrolled_ users)
  - grade (optional; overrides the final grade for a single user)
- use the grades method of [edx-api-client](https://github.com/mitodl/edx-api-client/)
- if possible, fetch all the grades and the passed status in one request, instead of iterating through each user
- if the user passed, create a course certificate for them

- also need a way to revoke an existing certificate

arslanashraf7:
Posting an update about the working of the command that's created for this:

**Criteria & Conditions (For operations in the command)**
1. A course run (Course Run's `courseware_id`) is needed for all the operations in the command
2. _Grade override and revoke/un-revoke_ operations are supported for a single user only, so for these operations, the command will specifically need a user
3. The _Create_ operation is supported for single or all users
    1. If a user is provided while running `--create` the operation will run only for that specific user
    2. If a user is not provided while running `--create` the operation will for all the enrolled users

**Overall behavior**
1. The command will not check the eligibility of the course run like the regular 24 hours task for certificates, it will just perform the operation specifically for the provided run
2. In the case of `--grade=` if the grade is greater than `0.0` the command will override the grade and mark the user passed and generate the certificate
3. In case the grade is not provided, the command will check the eligibility criteria of certificates e.g. (Passed in edX + Paid enrollment - `Verified`)

**Samples:**

**Revoke, Un-Revoke (This needs a user, it only works for a single user)**

- `./mange.py manage_certificates —revoke —user=<username or email> —run=<course_run_courseware_id> `(Revokes the user certificate for a specific run)

**Create certificates (This supports single user or multiple users if no user is provided, If a user is provided it runs for that single user only):**

- `./manage.py manage_certificates —create —user=<username or email> —run=<course_run_courseware_id>` (This should create/update the grades for the user and create the certificate for the provided user if the user has passed grade from edX and enrolled in Verified/paid mode)

**Override Grade (This operation needs a user and works only for a single user):**
[...]

PR #897, opened 2022-08-25
feat: Add command for certificate management

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#881 

#### What's this PR do?
Adds a command to manage below operations:

- Certificate revoke/unrevoke
- Grade override/Certificate creation
- Certifiicate creation

**Criteria & Conditions (For operations in the command)**
1. A course run (Course Run's `courseware_id`) is needed for all the operations in the command
2. _Grade override and revoke/un-revoke_ operations are supported for a single user only, so for these operations, the command will specifically need a user
3. The _Create_ operation is supported for single or all users
    1. If a user is provided while running `--create` the operation will run only for that specific user
    2. If a user is not provided while running `--create` the operation will for all the enrolled users

**Overall behavior**
1. The command will not check the eligibility of the course run like the regular 24 hours task for certificates, it will just perform the operation specifically for the provided run
2. In the case of `--grade=` if the grade is greater than `0.0` the command will override the grade and mark the user passed and generate the certificate
3. In case the grade is not provided, the command will check the eligibility criteria of certificates e.g. (Passed in edX + Paid enrollment - `Verified`)

To test the command you can follow three paths defined below:

**Revoke, Un-Revoke (This needs a user, it only works for a single user)**

1. `./mange.py manage_certificates --revoke --user=<username or email> --run=<course_run_courseware_id> `(Revokes the user certificate for a specific run)
2. `./mange.py manage_certificates --unrevoke --user=<username or email> --run=<course_run_courseware_id>` (Un-Revokes the user certificate for a specific run) 

**Create certificates (This supports single users or multiple users if no user is provided, If a user is provided it runs for that single user only):**

1. `./manage.py manage_certificates --create --user=<username or email> --run=<course_run_courseware_id>` (This should create/update the grades for the user and create the certificate for the provided user if the user has passed grade from edX and enrolled in Verified/paid mode)
2. `./manage.py manage_certificates --create --run=<course_run_courseware_id>` (This should create/update the grades for the users and create the certificate for all the users (Passed in edX + Enrolled in Verified/Paid) enrolled in the provided course run if the user has passed grade from edX and enrolled in Verified/paid mode)

**Override Grade & create a certificate(This operation needs a user and works only for a single user):**
1. `./manage.py manage_certificates --create --user=<username or email> --run=<course_run_courseware_id> --grade=<a float value between 0.0-1.0>` (This should override the grades for the user with the provided grade in parameters and create the certificate for the provided user regardless of the passing grade and paid enrollment) - This will not create the certificate for the user in the case when `grade=0.0` otherwise it will override the grade+create certificate

### DIFF
--- a/courses/management/commands/manage_certificates.py
+++ b/courses/management/commands/manage_certificates.py
@@ -1,0 +1,189 @@
+"""
+Management command to revoke and un revoke a certificate for a course run or program for the given user.
+"""
+from django.core.management.base import BaseCommand, CommandError
+from users.api import fetch_user
+from courses.api import (
+    manage_course_run_certificate_access,
+    ensure_course_run_grade,
+    process_course_run_grade_certificate,
+    override_user_grade,
+)
+from courses.models import CourseRun
+from openedx.api import get_edx_grades_with_users
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+
+
+class Command(BaseCommand):
+    """
+    Command to revoke/un-revoke a course run or program certificate for a specified user.
+    """
+
+    help = "Revoke and un revoke a certificate for a specified user against a program or course run."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--user",
+            type=str,
+            help="The id, email or username of the enrolled User",
+            required=False,
+        )
+        group = parser.add_mutually_exclusive_group()
+        group.add_argument(
+            "--run", type=str, help="The 'courseware_id' value for a CourseRun"
+        )
+        parser.add_argument(
+            "--grade",
+            type=float,
+            help="Override a grade. Setting grade to 0.0 blocks certificate creation. Setting a passing grade \
+                (>0.0) allows certificate creation. Range: 0.0 - 1.0",
+            required=False,
+        )
+        parser.add_argument(
+            "--revoke", dest="revoke", action="store_true", required=False
+        )
+        parser.add_argument(
+            "--unrevoke", dest="unrevoke", action="store_true", required=False
+        )
+        parser.add_argument(
+            "--create", dest="create", action="store_true", required=False
+        )
+
+        super().add_arguments(parser)
+
+    def handle(self, *args, **options):  # pylint: disable=too-many-locals
+        """Handle command execution"""
+
+        revoke = options.get("revoke")
+        unrevoke = options.get("unrevoke")
+        create = options.get("create")
+        run = options.get("run")
+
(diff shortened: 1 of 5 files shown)


courses/management/commands/manage_certificates.py
```diff
@@ -0,0 +1,189 @@
+"""
+Management command to revoke and un revoke a certificate for a course run or program for the given user.
```
annagav, line 2: Could you add here also the certificate created and update of a grade for a user to the description?

courses/management/commands/manage_certificates.py
```diff
@@ -0,0 +1,189 @@
...
+                    created_grade,
+                    updated_grade,
+                ) = ensure_course_run_grade(
+                    user=user,
+                    course_run=course_run,
+                    edx_grade=edx_grade,
+                    should_update=True,
+                )
+
+                if is_grade_override:
```
annagav, line 136: Could we move this out of the `-create` option into it's own option? something like `./manage.py manage_certificates -update_grade ...`

annagav (changes requested):

courses/management/commands/manage_certificates.py
arslanashraf7: Good suggestion but let me add some more context. So, most of this command is inspired by [this](https://github.com/mitodl/mitxpro/blob/master/courses/management/commands/sync_grades_and_certificates.py) and [this](https://github.com/mitodl/mitxpro/blob/master/courses/management/commands/revoke_certificate.py) in xPRO.

A use case: Let's say we want to create a certificate for a user specifically and also want to override the grade for that along the way without respecting what grades the user might have in edX, we will need to pass the `-grade` value because we won't know what grading policy has been set in edX. Since this is still a create operation but with grade we can use it with `-create`.

Since you mentioned, I though,t or something, How about when we just want to override/update the grade and not create the certificate we have no option for that in command. So, I create `-update_grade` operation which will just update the grade of that user in the database and not touch the certificate. What would you say about that? @annagav

arslanashraf7: @annagav I've addressed the feedback as per my above message, Of course, let me know if you think otherwise and I'll update the PR accordingly.

courses/management/commands/manage_certificates.py
arslanashraf7: Updated the documentation and added usages too.

annagav (approved):
LGTM 👍

PR #898, opened 2022-08-25
Adds setup command to bootstrap financial aid for DEDP

### LINKED ISSUE #883: Management command to set up minimal requirements for DEDP financial aid
As an admin, I'd like to reliably set up the production environment for flexible pricing. There are [a lot of steps](https://github.com/mitodl/hq/discussions/133), and I don't want to make any mistakes. 

As an engineer, I'd like to be able to easily replicate the production environment when I need to work on flexible pricing code. 

#### Acceptance Criteria:

- [ ] Based on a known good configuration in RC
- [ ] can create all the settings from scratch
- [ ] some reasonable, documented behavior if the application is already configured

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #883

#### What's this PR do?

Adds a management command that bootstraps DEDP financial aid. This includes:
- making sure a DEDP program exists
- creating discounts for the tiers (or identifying and reusing discounts that already exist, if their parameters are correct)
- creating tiers for DEDP financial assistance requests

The command will also deactivate (by setting the expiration date to a date in the past) discounts that follow the naming scheme (`DEDP-fa-tierX-YYYY` where YYYY is the year), and will deactivate tiers that aren't associated with a current set of DEDP discounts. 

Automated tests should pass.

See the comment block at the top of the command `flexiblepricing/management/commands/configure_for_dedp.py` for an explanation of what your database should have in it after running the command. 

To test full-stop bootstrapping, ensure you do not have any discounts that follow the `DEDP-fa-tier` naming scheme, nor tiers that use them, nor a DEDP program (`program-v1:MITx+DEDP`). Then, run the command. You should have a DEDP program, a set of four DEDP discounts, and a set of four tiers for the program that use the discounts. 

To test partial bootstrapping, run the command after first having any of these conditions:
- an existing DEDP program
- existing DEDP discounts for the current year
- tiers that use the discounts

The command should run successfully and the existing data should be kept if it matches the parameters. 

To test invalidation, run the command after first having any of these conditions:
- a DEDP discount for a prior year (so, for 2022, a discount with a code of `DEDP-fa-tierX-2021` or before)
- tiers that use the prior year discount

The discounts should be invalidated and the tiers should be turned off (`current=False`).

### DIFF
--- a/flexiblepricing/management/commands/configure_for_dedp.py
+++ b/flexiblepricing/management/commands/configure_for_dedp.py
@@ -1,0 +1,196 @@
+"""
+Ensures there's a basic set of records present for financial aid to work for
+DEDP courses. This includes:
+- A DEDP program (required for the tiers to be set up)
+- A set of discounts, specific to DEDP
+- Tiers that are tied to these discounts
+
+The defaults for this are:
+
+Program:
+- Readable ID is `program-v1:MITx+DEDP`
+- Name is `Data, Economics and Development Policy`
+
+Discounts: (year is the current year)
+Code               | Type        | Amount
+-----------------------------------------
+DEDP-fa-tier1-year | dollars-off | 750
+DEDP-fa-tier2-year | dollars-off | 650
+DEDP-fa-tier3-year | dollars-off | 500
+DEDP-fa-tier4-year | percent-off | 0
+-----------------------------------------
+
+Tiers:
+Threshold | Discount
+------------------------------
+0         | DEDP-fa-tier1-year
+25,000    | DEDP-fa-tier2-year
+50,000    | DEDP-fa-tier3-year
+75,000    | DEDP-fa-tier4-year
+------------------------------
+
+If discounts and tiers exist for the current year, this command will leave them
+alone. If they exist for prior years, they will be disabled (current=False for
+tiers and expiration date set to the past for discounts).
+
+If a program exists with the readable ID specified, that will be used.
+
+"""
+from datetime import date, datetime
+from django.core.management import BaseCommand
+from django.contrib.contenttypes.models import ContentType
+from django.conf import settings
+import pytz
+
+from courses.models import Program
+from flexiblepricing.models import FlexiblePriceTier
+from ecommerce.models import Discount
+from ecommerce.constants import REDEMPTION_TYPE_UNLIMITED
+
+
+class Command(BaseCommand):
+    """
+    Sets up some basic records for DEDP Financial Assistance
+    """
+
+    help = "Sets up some basic records for DEDP Financial Assistance"
+    PROGRAM_READABLE_ID = "program-v1:MITx+DEDP"
+
+    def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
+        current_year = date.today().year
+        last_year = datetime(
+            current_year - 1, 1, 1, tzinfo=pytz.timezone(settings.TIME_ZONE)
+        )
+        discounts_and_tiers = [
+            {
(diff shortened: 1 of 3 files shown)


flexiblepricing/management/commands/configure_for_dedp.py
```diff
@@ -0,0 +1,195 @@
...
+            self.stdout.write(self.style.NOTICE(f"Created new program {program.id}"))
+        else:
+            self.stdout.write(self.style.NOTICE(f"Using existing program {program.id}"))
+
+        # Step two: get existing discounts
+        discounts = Discount.objects.filter(
+            discount_code__startswith="DEDP-fa-tier"
+        ).all()
+        self.stdout.write(
+            self.style.NOTICE(f"{len(discounts)} existing discounts for {current_year}")
```
rachellougee, line 118: Is this intended to assume existing discounts are for the current year?

flexiblepricing/management/commands/configure_for_dedp.py
```diff
@@ -0,0 +1,195 @@
...
+                if (
+                    discount.valid_now()
+                    and tier_config["discount"]["discount_type"]
+                    == discount.discount_type
+                    and tier_config["discount"]["amount"] == discount.amount
+                ):
+                    found_discount = discount
+                    self.stdout.write(
+                        self.style.NOTICE(
+                            f"\tFound a discount ({discount.discount_code}"
```
rachellougee, line 143: Minor pick, missing )

rachellougee:
Overall it looks good. I was able to verify the inserted records after the initial run. 
I just have a question regarding the discounts. I manually updated the discount code - DEDP-fa-tier4-2022 to DEDP-fa-tier4-2021, then rerun the command. It didn't update the discount code to the correct year. By looking at the code, these didn't get updated because discounts are still valid (no expiration date), so it's assuming these are still valid even though the text said it's for 2021

```Setting up the program...
Using existing program 1
5 existing discounts for 2022
Looking for 750 dollars-off discount for threshold 0
        Found a discount (DEDP-fa-tier1-2021
        Updated existing tier ID 1 for threshold 0
Looking for 650 dollars-off discount for threshold 25000
        Found a discount (DEDP-fa-tier2-2021
        Updated existing tier ID 2 for threshold 25000
Looking for 500 dollars-off discount for threshold 50000
        Found a discount (DEDP-fa-tier3-2021
        Updated existing tier ID 3 for threshold 50000
Looking for 0 percent-off discount for threshold 75000
        Found a discount (DEDP-fa-tier4-2021
        Updated existing tier ID 4 for threshold 75000
1 discounts expired
0 tiers deactivated
```

flexiblepricing/management/commands/configure_for_dedp.py
rachellougee: ignore me. I misread this text. I think my question was on the different line 135-138

jkachel:
Fixed the output stuff in https://github.com/mitodl/mitxonline/pull/898/commits/00fc4a77b2306be89eb4cc02ed77a6ec6617b778. Good catch on the discount code reuse - I updated the query to explicitly look for discounts for the current year and invalidate the ones that aren't (so if you have some from 2021, say, it will just turn those off). I thought about changing the codes but we don't track revisions of discount codes, so changing them would also change the redemptions for learners, and I don't think we also want to display the wrong year for folks who have been approved for flex pricing either.

rachellougee (approved):

rachellougee:
LGTM. Discount work as expected now

Issue #895, opened 2022-08-25
Bug: Course purchase order error on 100% Discount


aliraza-abbasi:
### Steps to Reproduce

Apply 100% discount coupon and click on `Place your order`

### Stacktrace

- First: `File "/app/./ecommerce/views/v0/__init__.py", line 532, in get
    checkout_payload = api.generate_checkout_payload(request)
  File "/app/./ecommerce/api.py", line 127, in generate_checkout_payload
    USER_MSG_TYPE_PAYMENT_ACCEPTED_NOVALUE,
NameError: name 'USER_MSG_TYPE_PAYMENT_ACCEPTED_NOVALUE' is not defined`

- Second: This error comes up when USER_MSG_TYPE_PAYMENT_ACCEPTED_NOVALUE is imported from constants `File "/app/./ecommerce/views/v0/__init__.py", line 532, in get
    checkout_payload = api.generate_checkout_payload(request)
  File "/app/./ecommerce/api.py", line 124, in generate_checkout_payload
    "response": fulfill_completed_order(
TypeError: fulfill_completed_order() takes 3 positional arguments but 4 were given`

- Third: This error comes up when USER_MSG_TYPE_PAYMENT_ACCEPTED_NOVALUE commented in fulfill_completed_order() argument. Also the user got enrolled in the course but 500 Internal Server Error page is displayed. `Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 188, in _get_response
    self.check_response(response, callback)
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 309, in check_response
    raise ValueError(
ValueError: The view ecommerce.views.v0.CheckoutInterstitialView didn't return an HttpResponse object. It returned None instead.`

pdpinch:
@arslanashraf7 can this be closed?

arslanashraf7:
Yes, Closing it via https://github.com/mitodl/mitxonline/pull/899

PR #899, opened 2022-08-26
fix: basket checkout with zero value

arslanashraf7:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

#895 

#### What's this PR do?
- Adds the missing import
- After Adding that import (I noticed that the `fulfill_completed_order` also had the wrong number of args so i had to fix that too)
- Fixes 500 error when basket is checked out with zero value
- Adds a test to avoid this kind of error with zero value basket in future

- Setup this branch
- Create a discount with 100% or full value off
- Apply the discount at checkout
- Click checkout (User should not see 500 error, instead he should be redirected to dashboard properly)
- You can also follow the testing instruction from [this PR](https://github.com/mitodl/mitxonline/pull/462) that was initially created.

- Try to reproduce this error on the main first

- The zero value checkout was added in https://github.com/mitodl/mitxonline/pull/462/files
- But [another PR](https://github.com/mitodl/mitxonline/pull/817/files#diff-3740341cf4fbf867073cbe39d40a13c4dde5629e6f5cb14a17db83d49d1a9fecL219-L224) changed the code and this usecase was missed. A side effect of https://github.com/mitodl/mitxonline/pull/817

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -18,6 +18,7 @@ from main.constants import (
     USER_MSG_TYPE_ENROLL_DUPLICATED,
     USER_MSG_TYPE_COURSE_NON_UPGRADABLE,
     USER_MSG_TYPE_DISCOUNT_INVALID,
+    USER_MSG_TYPE_PAYMENT_ACCEPTED_NOVALUE,
 )
 
 from mitol.payment_gateway.api import (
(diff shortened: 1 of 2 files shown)


ecommerce/api.py
```diff
@@ -118,13 +119,17 @@ def generate_checkout_payload(request):
...
+        fulfill_completed_order(
+            order, {"amount": 0, "data": {"reason": "No payment required"}}, basket
+        )
         return {
             "no_checkout": True,
-            "response": fulfill_completed_order(
-                order,
-                {"amount": 0, "data": {"reason": "No payment required"}},
-                basket,
-                USER_MSG_TYPE_PAYMENT_ACCEPTED_NOVALUE,
```
arslanashraf7, line 127: This import was missing. So i added it.

arslanashraf7, line 127: `fulfill_completed_order ` was being called with wrong number of args, so i had to refactor it to make it work the way it should.

jkachel (approved):
LGTM 👍

PR #901, opened 2022-08-26
Fixes nav issues when a course date is selected

### LINKED ISSUE #854: More Dates: Course detail page selects last selected date from more dates
### Steps to Reproduce

- Login into MITxOnline RC
- Open a course with multiple dates e.g. https://rc.mitxonline.mit.edu/courses/course-v1:MITx+14.310x/
- Click more dates and select another date (And notice a `#` added to the URL)
- Click back (I won't take the user back to the home page)
- Click back again and open the same course again (Notice that instead of showing the default course run as per the logic that was defined before it would show the last selected course run date again)

### Expected Behavior

- Back button should take the user back to the home page in one go
- When you open the same course again it should show/select the default course run instead of last selected run

### Actual Behavior
- Clicking another date from more dates appends the URL with `#` and then the user has to click twice to go back
- Upon reopening a course the selected date in the course detail page is the one user selected when he/she visited the last time

### Stacktrace
Not a stack trace but I think this is happening because of session storage that we used in [this PR](https://github.com/mitodl/mitxonline/pull/798/files#diff-7f355af766f3260748718866852dc70e469e902235774dd21b82354defff5b85R66-R68).

### Related Issues
https://github.com/mitodl/mitxonline/issues/767

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #854 

#### What's this PR do?

Swaps clickable elements in the course run date picker from `<a>` tags to `<span>` so they don't cause the URL to change. Also removes the dependence on `localStorage` for storing the selected course run (as noted, it should default to the most recent course run). 

Load a course page that has multiple active course runs. Clicking the More Dates link should not cause the URL to change. Selecting a date in the popover should, additionally, not cause the URL to change. The selection should not survive a page refresh.

### DIFF
--- a/frontend/public/src/containers/ProductDetailEnrollApp.js
+++ b/frontend/public/src/containers/ProductDetailEnrollApp.js
@@ -56,16 +56,13 @@ export class ProductDetailEnrollApp extends React.Component<
   }
 
   setCurrentCourseRun = (courseRun: EnrollmentFlaggedCourseRun) => {
-    sessionStorage.setItem('currentCourseRun', JSON.stringify(courseRun))
     this.setState({
       currentCourseRun: courseRun
     })
   }
 
   getCurrentCourseRun = (): EnrollmentFlaggedCourseRun => {
-    const courseRun = sessionStorage.getItem('currentCourseRun')
-    const sessionCourseRun = courseRun ? JSON.parse(courseRun) : null
-    return sessionCourseRun ? sessionCourseRun : this.state.currentCourseRun
+    return this.state.currentCourseRun
   }
 
   renderUpgradeEnrollmentDialog() {
--- a/cms/templates/product_page.html
+++ b/cms/templates/product_page.html
@@ -48,7 +48,7 @@
               </div>
               {% if course_runs|length > 1 %}
                 <strong class="more-dates">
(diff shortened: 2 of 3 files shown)


frontend/public/src/containers/ProductDetailEnrollApp.js
```diff
@@ -56,16 +56,17 @@ export class ProductDetailEnrollApp extends React.Component<
...
     })
   }
 
   getCurrentCourseRun = (): EnrollmentFlaggedCourseRun => {
-    const courseRun = sessionStorage.getItem('currentCourseRun')
-    const sessionCourseRun = courseRun ? JSON.parse(courseRun) : null
-    return sessionCourseRun ? sessionCourseRun : this.state.currentCourseRun
+    // const courseRun = sessionStorage.getItem('currentCourseRun')
+    // const sessionCourseRun = courseRun ? JSON.parse(courseRun) : null
+    // return sessionCourseRun ? sessionCourseRun : this.state.currentCourseRun
```
collinpreston, line 68: Should these be removed?

collinpreston (approved):
Saw some commented out code.  Other than that, it works and looks good to me.

Issue #887, opened 2022-08-19
Versioning of certificate templates


pdpinch:
As a course team member, I'd like to update the signatories on a course certificate, and have it only affect new course certificates created in the future. In particular, I need to replace the outgoing VP's signature with the new VP's signature. 

See https://github.com/mitodl/hq/discussions/131 for a discussion on how to implement this 

#### Acceptance Criteria:

- [ ] write a brief description of the plan
   - [ ] how will migrations be handled?
   - [ ] how will existing PageVersions be addressed?
- [ ] include testing instructions that support changes to the template, learner views and anonymous views.

rhysyngsun:
Blocking #705

PR #903, opened 2022-08-29
Versioning of certificate template

aliraza-abbasi:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [ ] Settings

#887 

- Get or Create two different users
- Enroll them both in any of one course
- The course should have the certificate page in CMS
- Generate certificate for one user and check that user’s certificate.
- Go to the CMS and make some changes in certificate page of that course (e.g. add/remove one or more signatories)
- Generate certificate for second user
- Verify that both users have different certificate template
- One who got the certificate first will have different certificate than a user who got the certificate after making changes in certificate template

Screenshot 2022-08-29 at 2 59 26 PM
Screenshot 2022-08-29 at 3 10 11 PM
Screenshot 2022-08-29 at 3 11 33 PM


pdpinch:
Can you give some more details about what's changed in the template date in the different screenshots? It looks like you changed the CEUs and the signatories?

Also, can you update the admin for certificates so it will display a reference to the pagerendition, ideally displayed with a date.

aliraza-abbasi:
> Can you give some more details about what's changed in the template date in the different screenshots? It looks like you changed the CEUs and the signatories?
> 
> Also, can you update the admin for certificates so it will display a reference to the pagerendition, ideally displayed with a date.

- I've generated the certificate for one user and then changed the CEUs of the certificate
- I generated the second certificate for second user and then changed the CEUs and Institute Text and Signatories
- Then generated the third certifiacate for the third user.

Here you can see three different certificates for one course run. This happens because of their revisions. Revisions are saved with certificate when the certificate is generated, by getting the current live certificate revision.


For second question there is a `History` page at certificate edit page where we can see previous revisions with date and time and can edit revision or compare with another revision (Are you talking about CMS or Django Admin? For CMS We have a page already for django we can create it)

Screenshot 2022-08-30 at 5 46 55 PM
Screenshot 2022-08-30 at 5 47 08 PM

pdpinch:
Thanks for the explanation of the screenshots. They look good to me and we should proceed. 

My question about the admin was about the django admin for certificates, at `/admin/courses/courseruncertificate/` -- that can wait for another PR if you like, but it should be a small change.

aliraza-abbasi:
> Thanks for the explanation of the screenshots. They look good to me and we should proceed.
> 
> My question about the admin was about the django admin for certificates, at `/admin/courses/courseruncertificate/` -- that can wait for another PR if you like, but it should be a small change.

@pdpinch I've added this in admin, here is the screenshot. please have a look
Screenshot 2022-08-31 at 3 06 42 PM

collinpreston:
@aliraza-abbasi , what is the best way that I could test this locally?  I have `devstack` and `mitxonline` running locally.

aliraza-abbasi:
> @aliraza-abbasi , what is the best way that I could test this locally? I have `devstack` and `mitxonline` running locally.

Updated in description

rhysyngsun (commented):
@aliraza-abbasi this generally looks good to me, could you have someone else also do a review plus functional test of your changes since that might take me some time to get around to and I don't want to hold up your work.

courses/api_test.py
```diff
@@ -73,7 +73,8 @@ def course():
 @pytest.fixture()
 def passed_grade_with_enrollment(user):
     """Fixture to produce a passed CourseRunGrade"""
-    course_run = CourseRunFactory.create()
+    course = CourseFactory.create()
```
arslanashraf7, line 76: Ideally, the CourseRunFactory should create the Course object automatically using CourseFactory. Was there anything specific for this change?

cms/models.py
```diff
@@ -155,13 +155,20 @@ def course_certificate(
...
+                if certificate.course_run.course.page
+                else None
+            )
         )
+
         if not certificate_page:
             raise Http404()
 
+        if not certificate.certificate_page_revision:
+            certificate.save()
```
arslanashraf7, line 171: This is fine too but I was thinking of something that might serve us better (This might be even a better use case in xPRO where we have existing certificates in production, in MITxOnline we only have certificates in RC):

How about we create a data migration while we are moving towards the revision-based certificates, e.g. when [this migration](https://github.com/mitodl/mitxonline/pull/903/files#diff-1f2100c0ef3efdd82b047e419dc49af4b0f64c985a6be6e6610a8cf4fe79851aR1) runs, we migrate the existing certificates in the database to use the latest revision of the certificate for the specific course runs present at that moment. 

The benefit is that we will have consistent data right from the start, we won't need to call the save method here every time the revision isn't found. We might even want to set [this field](https://github.com/mitodl/mitxonline/pull/903/files#diff-d5de5fce23898af38cf96ba73fbdae70cb947ffae1e47551a73f5fccbabadaf9R558-R560) as non-nullable.

Before you implement this, Let's here @rhysyngsun 's thoughts on this too?

courses/models.py
```diff
@@ -594,6 +598,17 @@ def __str__(self):
...
+        if not self.certificate_page_revision:
+            certificate_page = (
+                self.course_run.course.page.certificate_page
+                if self.course_run.course.page
+                else None
+            )
+            if certificate_page:
+                self.certificate_page_revision = certificate_page.get_latest_revision()
+        super(CourseRunCertificate, self).save(*args, **kwargs)
+
```
arslanashraf7, line 611: Can you add some test(s) for this?

arslanashraf7 (changes requested):
Just the code review, I will try to also functionality test this too.

courses/api_test.py
aliraza-abbasi: For this test the `page` attribute in course object is required which is created when we call CourseFactory the course created from CourseRunFactory the `page` attribute remains None (this is because many test cases got failed when `page` is created by CourseRunFactory). The `page` attribute was added few days ago as we discussed this issue earlier. Whenever we need `page` attribute for tests then course should be created from CourseFactory not from the CourseRunFactory. If we add `page` to course via CourseRunFactory many of other tests got failed.

cms/models.py
rhysyngsun: I'm fine with enforcing it with a migration instead of an extra write.

courses/models.py
aliraza-abbasi: There are no tests added for this epic yet. I'll add the tests for all certificate apic with different PR

courses/migrations/0020_courseruncertificate_certificate_page_revision.py
```diff
@@ -0,0 +1,53 @@
+# Generated by Django 3.2.15 on 2022-08-29 08:44
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def set_current_certificate_revision(apps, schema_editor):
```
arslanashraf7, line 7: Do we need [this change](https://github.com/mitodl/mitxonline/pull/903/files#diff-64ebda3d82e0c746d2c838699c07acb411fbd29f3b343cf80597b3d1aa24740cR172-R173) after the data migration?

aliraza-abbasi: Yes, This is for an edge case. If the `certificate page` is not yet created at any point for a course at that point if the certificate generated will have no revision. So this will add the Revision whenever the page is available and the certificate is accessed.

courses/api_test.py
arslanashraf7: I feel this behavior is a little inconsistent because this gets confusing at this point, one of the options could have been that the default initialization of `page` variable in the course factory should be set to some default factory page and anyone not needing a page could create the page object like `CourseFactory.create(course__page=None)` Anyway that's not associates with this PR.

But there is a [course fixture](https://github.com/mitodl/mitxonline/blob/4667f071687b999a6c33fa7fe114ffa882954b9a/courses/api_test.py#L69) that you can use instead of creating it here.

arslanashraf7 (approved):
Looks good 👍 I added a small comment regarding using `course` fixture.

courses/api_test.py
aliraza-abbasi: fixed by setting `CourseFactory.create(course__page=None)`

Issue #910, opened 2022-08-30
dashboard doesn't refresh when user unenrolls from course in a program


pdpinch:
### Steps to Reproduce

1. Enroll (audit) in a course in a program 
2. On the dashboard, open the program drawer
3. Using the ⋮ menu, unenroll from the course 

### Expected Behavior

- Similar to the behavior when unenrolling from a course that's _not_ in a program, I expect: 
   - I can read the banner notification confirming my unenrollment 
   - the course should be removed from the dashboard 

### Actual Behavior

- drawer stays open, obscuring the banner notifcation
- the course is still visible in the drawer. Closing and re-opening the drawer does not remove it
- I need to reload the page to have the course removed from the dashboard 

### Related Issues

We've had some feedback that we should have a confirmation message when users unenroll. If it makes sense to include this in the same PR, we can do that.

PR #929, opened 2022-09-02
910 dashboard doesnt refresh when user unenrolls from course in a program

rachellougee:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/910

#### What's this PR do?
Fixing issue when user enrolls from course in a program 
 - Close the program drawer so that user can see the confirmation message
 - Program enrollment card is updated to reflect unenrollment on the dashboard

- Enroll a course from program and a course that's _not_ program

- Unenroll from a course in a program should now looks like this:

(# course is updated as well)
image

- unenroll from a course that is not a program should still work 

(i made this course not part of a program)
image

(Optional)

(Optional)

(Optional)

### DIFF
--- a/frontend/public/src/components/EnrolledItemCard.js
+++ b/frontend/public/src/components/EnrolledItemCard.js
@@ -30,7 +30,8 @@ type EnrolledItemCardProps = {
     enrollmentId: number,
     emailsSubscription: string
   ) => Promise<any>,
-  addUserNotification: Function
+  addUserNotification: Function,
+  closeDrawer:Function,
 }
 
 type EnrolledItemCardState = {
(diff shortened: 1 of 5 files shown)


frontend/public/src/components/EnrolledItemCard.js
```diff
@@ -88,6 +89,9 @@ export class EnrolledItemCard extends React.Component<
       if (isSuccessResponse(resp)) {
         messageType = ALERT_TYPE_SUCCESS
         userMessage = `You have been successfully unenrolled from ${enrollment.run.title}.`
+        if (typeof closeDrawer === "function") {
```
jkachel, line 92: One change for something that wasn't in the initial issue (but is related): can you apply the same change to the email settings notification? Should be largely this same code, but in the `onSubmit` function.

jkachel (changes requested):
Looks good! One change noted inline.

frontend/public/src/components/EnrolledItemCard.js
rachellougee: @jkachel I've applied the same change to the email settings notification. I also moved this code outside if isSuccessResponse as I realized that the drawer should be closed even if there is an error so user can see the message.
Please take a look. thanks

jkachel (approved):
LGTM 👍

PR #936, opened 2022-09-06
add is_self_paced to MicroMaster courserun import script

rachellougee:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

No new ticket, but the closed ticket is https://github.com/mitodl/mitxonline/issues/701

#### What's this PR do?
I ran into an error when running MicroMaster migration queries on RC
```psycopg2.errors.NotNullViolation: null value in column "is_self_paced" violates not-null constraint```

It looks like we added is_self_paced to the CourseRun model as part of [this PR ](https://github.com/mitodl/mitxonline/pull/722). The default value false is handled at django model level, not SQL level, so we would need to add is_self_paced = false in the migration query

Run ```docker-compose run --rm web ./manage.py import_micromasters_data``` and make sure no error

### DIFF
--- a/micromasters_import/management/commands/queries/002_import_courserun.sql
+++ b/micromasters_import/management/commands/queries/002_import_courserun.sql
@@ -12,7 +12,8 @@ INSERT INTO public.courses_courserun(
     live,
     created_on,
     updated_on,
-    course_id)
+    course_id,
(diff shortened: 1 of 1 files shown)


rhysyngsun (approved):
LGTM

PR #947, opened 2022-09-08
Adds a check to make sure flexible pricing forms have the right fields in them

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

n/a so far

#### What's this PR do?

Adds a signal receiver to check FlexiblePricingRequestForms for the required fields when one is published. If the fields aren't in there, the receiver will add them in for you.

The fields in question are the Your Income and Income Currency fields. 

There are three different ways to test:
1. Create a new Flexible Pricing Request Form with no form fields defined. After publishing it, edit it; the required fields should be in there.
2. Create a Flexible Pricing Request Form with only one of the fields defined. After publishing it, edit it; the missing field should be added. 
3. Edit a Flexible Pricing Request Form that has the form fields already in it and rename one of the required fields. No change to the form should be made (but there will be a log message from the signal receiver if your logging is verbose enough).
4. Edit a Flexible Pricing Request Form and remove a required field. Then, save it as draft, and publish it. The required field should be added back. (In local testing, just removing the field and saving it neither removed the field nor triggered the signal. Not sure why that is - maybe a bug in the (old) version of Wagtail we're using currently.) 

This seemed to be a point of confusion and an easily missed step so this should make dealing with flex pricing forms a bit easier.

This does not do anything else with flexible pricing form fields - any additional fields added to the form will still not be displayed. (That seemed like a separate issue.)

### DIFF
--- a/flexiblepricing/utils.py
+++ b/flexiblepricing/utils.py
@@ -1,0 +1,82 @@
+from cms.models import FlexiblePricingRequestForm, FormField
+from django.db.models import Q
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def ensure_flexprice_form_fields(form_instance: FlexiblePricingRequestForm):
+    """
+    Checks the supplied form instance for the proper Flexible Pricing Request Form
+    fields and adds them if it needs to, or logs if it gets confused. This is
+    meant to be called by the signal receiver (see cms/signalreceivers.py) or by
+    a management command.
+
+    Args:
+    - form_instance (FlexiblePricingForm) - the form to work on
+
+    Returns:
+    - boolean - True if the form was OK, False otherwise
+    """
+
+    currency_field = False
+    income_field = False
+
+    relevant_fields = form_instance.form_fields.filter(
+        Q(field_type__in=["country", "number"])
+        | Q(clean_name__in=["your_income", "income_currency"])
+    ).all()
+
+    if len(relevant_fields) != 2:
+        logger.warning(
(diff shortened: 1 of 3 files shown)


flexiblepricing/utils.py
```diff
@@ -0,0 +1,80 @@
...
+    ).all()
+
+    if len(relevant_fields) != 2:
+        logger.warning(
+            f"Improper length {len(relevant_fields)} returned; was expecting 2"
+        )
+
+    for field in relevant_fields:
+        if field.field_type == "country":
+            if field.clean_name == "income_currency":
```
rachellougee, line 37: No need to change, but for me, it looks clearer if two if are combined in one line.

rachellougee:
Fort testing, it works fine with no form fields defined, and it would automatically add two fields.
There seems issue when I only add one field and then publish, I got it to a state where it duplicates the field I originally added
image

When I delete one of the fields, I got a state saying "There is another field with the label Your Income, please change one of them." but there isn't such a field

image

If I remove all the fields, it works as expected as it will just add back two required fields

jkachel:
Testing actually uncovered a couple bugs - I'd swapped the currency field for the income field, but it should now also set the display order on the income field to 1 and the currency field to 2 when it creates them. (If they're already there, it won't do anything with it.)

rachellougee (approved):

rachellougee:
LGTM

Issue #950, opened 2022-09-08
Filter flexbile pricing admin by course/program 


pdpinch:
As a member of the course team who is reviewing flexible pricing requests, I need to filter the list to only see a course/program that is relevant to me. 

#### Acceptance Criteria:

- [ ] On `/staff-dashboard/flexible_pricing`,add a drop-down for choosing from existing courses/programs with financial aid requests

rachellougee:
@pdpinch Just to clarify, Is this referring to Django admin https://rc.mitxonline.mit.edu/admin/flexiblepricing/flexibleprice/
or dashboard https://rc.mitxonline.mit.edu/staff-dashboard/flexible_pricing?

PR #965, opened 2022-09-12
adding course/program as filter on flexible pricing staff dashboard

rachellougee:
#### Pre-Flight checklist

- [x] Testing
  - [ ] Code is tested

https://github.com/mitodl/mitxonline/issues/950

#### What's this PR do?
Adding a static field 'type' to courseware object in /api/flexible_pricing/applications_admin/. (Identity as program or course)
```       
         "courseware": {
                "title": "Data, Economics and Development Policy",
                "readable_id": "program-v1:MITx+DEDP",
                "id": 1,
                "type": "program"
            },
```
Adding a drop-down for choosing from existing courses/programs with financial aid requests

Go to Flexible pricing from the staff dashboard: http://mitxonline.odl.local:8013/staff-dashboard/flexible_pricing
image

The dropdown "Search by Course/Program" should list all the existing courses/programs from flexible pricing requests
image

Search a specific course or program, flexible pricing requests table should only display that course or program
image

The clear icon is also added to deselect a course/program (same as status)
image

### DIFF
--- a/courses/serializers.py
+++ b/courses/serializers.py
@@ -6,7 +6,6 @@ from urllib.parse import urljoin
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.templatetags.static import static
-from flexiblepricing.api import is_courseware_flexible_price_approved
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
 
@@ -14,8 +13,10 @@ from cms.models import CoursePage
 from cms.serializers import CoursePageSerializer
 from courses import models
 from courses.api import create_run_enrollments
+from courses.constants import CONTENT_TYPE_MODEL_COURSE, CONTENT_TYPE_MODEL_PROGRAM
 from ecommerce.models import Product
 from ecommerce.serializers import BaseProductSerializer, ProductFlexibilePriceSerializer
+from flexiblepricing.api import is_courseware_flexible_price_approved
 from main import features
 from openedx.constants import EDX_ENROLLMENT_AUDIT_MODE, EDX_ENROLLMENT_VERIFIED_MODE
 
@@ -45,15 +46,26 @@ def _get_thumbnail_url(page):
 class BaseCourseSerializer(serializers.ModelSerializer):
     """Basic course model serializer"""
 
+    type = serializers.SerializerMethodField(read_only=True)
+
     def to_representation(self, instance):
         data = super().to_representation(instance)
         if not self.context.get("include_page_fields") or not hasattr(instance, "page"):
             return data
         return {**data, **CoursePageSerializer(instance=instance.page).data}
 
+    @staticmethod
+    def get_type(obj):
+        return CONTENT_TYPE_MODEL_COURSE
+
     class Meta:
         model = models.Course
-        fields = ["id", "title", "readable_id"]
+        fields = [
+            "id",
+            "title",
+            "readable_id",
+            "type",
+        ]
 
 
 class ProductRelatedField(serializers.RelatedField):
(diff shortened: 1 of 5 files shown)


courses/serializers.py
```diff
@@ -45,15 +45,26 @@ def _get_thumbnail_url(page):
...
+
     def to_representation(self, instance):
         data = super().to_representation(instance)
         if not self.context.get("include_page_fields") or not hasattr(instance, "page"):
             return data
         return {**data, **CoursePageSerializer(instance=instance.page).data}
 
+    @staticmethod
+    def get_type(obj):
+        return "course"
```
asadali145, line 58: I think we can use the `CONTENT_TYPE_MODEL_PROGRAM` and `CONTENT_TYPE_MODEL_COURSE` constants.

courses/serializers.py
```diff
@@ -242,9 +253,15 @@ class Meta:
 class BaseProgramSerializer(serializers.ModelSerializer):
     """Basic program model serializer"""
 
+    type = serializers.SerializerMethodField(read_only=True)
+
+    @staticmethod
+    def get_type(obj):
+        return "program"
```
asadali145, line 260: As mentioned above, We can use contants. Applies to all usages.

flexiblepricing/views/v0/__init__.py
```diff
@@ -63,6 +62,30 @@ def get_queryset(self):
         if status_search is not None:
             queryset = queryset.filter(status=status_search)
 
+        courseware_search = self.request.query_params.get("courseware")
+        if courseware_search is not None:
```
asadali145, line 66: Can be simplified to `if courseware_search:`.

flexiblepricing/views/v0/__init__.py
```diff
@@ -63,6 +62,30 @@ def get_queryset(self):
...
+                    )
+                )
+            elif courseware_content_type == "program":
+                queryset = queryset.filter(
+                    Q(
+                        courseware_object_id=courseware_object_id,
+                        courseware_content_type=program_type,
+                    )
+                )
+
```
asadali145, line 88: We can simplify this to:
```suggestion
            content_type = ContentType.objects.get(
                app_label="courses", model=courseware_content_type
            )
            queryset = queryset.filter(
                Q(
                    courseware_object_id=courseware_object_id,
                    courseware_content_type=content_type,
                )
            )

```

flexiblepricing/views/v0/__init__.py
```diff
@@ -63,6 +62,30 @@ def get_queryset(self):
         if status_search is not None:
             queryset = queryset.filter(status=status_search)
 
+        courseware_search = self.request.query_params.get("courseware")
+        if courseware_search is not None:
+            courseware_content_type, courseware_object_id = courseware_search.split(":")
+            course_type = models.ContentType.objects.get(
```
asadali145, line 68: `models` is basically flexiblepricing.models and we are using an import of flexiblepricing/models.py. We should import ContentType using `from django.contrib.contenttypes.models import ContentType`.

asadali145 (changes requested):
Everything looks fine. Here are a few things that I noticed.
Haven't tested it locally as well. Will do it with the final review.

flexiblepricing/views/v0/__init__.py
rachellougee: The other 2 params name_search and status_search all have the check 'is not None', it's better to have consistent logic here with is not None IMO

rachellougee:
@asadali145 I've fixed the merge conflict and addressed feedback. Please take a look. Thanks

asadali145 (approved):
Looks good to me. Awesome effort. Added a minor suggestion about using ISort that is up to you. Also, Maybe we should add a bit of responsiveness for the forms as there is no space between filters(SS Attached).
Screenshot 2022-09-13 at 7 11 32 PM

rachellougee:
@asadali145 I've run ISort which auto-fixed the order of import in that file. 
As for responsiveness, it's a bit tricky as it's using ``` <Form layout="inline"``` for the filter, the other option is to set vertical, which will put each filter in its own line, but that also changes how it looks as regular view.
image
I will just keep it the way it is, if this becomes an issue, we can adjust it
Thank you again for all the feedbacks

Issue #942, opened 2022-09-08
unauthorized user can access staff dashboard... 


pdpinch:
... but cannot retrieve any data

### Steps to Reproduce

1. Create a new account, without django staff privileges
2. Visit /staff-dashboard/
3. Click on link to "Flexible Pricing" 

### Expected Behavior

- /staff-dashboard/ should return a 403

### Actual Behavior

- /staff-dashboard/ loads
- unauthorized user can click on "Flexible Pricing"
- "Flexible Pricing" data doesn't load (display spinner) because API returns 403  

### Screenshot or Screencast

<img width="1483" alt="image" src="https://user-images.githubusercontent.com/430126/189023197-a4716285-9319-4b99-a848-ea9e963fd788.png">

PR #969, opened 2022-09-13
942: unauthorized user can access staff dashboard

collinpreston:
#### Pre-Flight checklist

- [x] Testing

https://github.com/mitodl/mitxonline/issues/942

#### What's this PR do?
A user who does not have `is_staff` == `true` will receive an error message and be unable to login to the staff dashboard (http://mitxonline.odl.local:8013/staff-dashboard/).  A user with `is_staff` == `true` will be able to login to the staff dashboard.

This PR also allows the user to refresh the browser tab and not be forced to login again.  This is done by only checking if the user is authenticated after the auth has loaded.  While the auth is loading, the text "loading..." is shown.

1. Configure the refine application in `mitxonline`
2. From within an Incognito window, attempt to login to the refine/staff dashboard with a user who has `is_staff` = `false`.  Verify that you are unable to login and receive an error message.
3. From within a new Incognito window, attempt to login to the refine/staff dashboard with a user who has `is_staff` = `true`.  Verify that you are able to login and don't receive any error message.
4. Ensure that when you refresh the browser tab you are not forced to login again.

Configure refine for `mitxonline`: https://github.com/mitodl/mitxonline/blob/main/docs/source/configuration/refine_admin.rst

This PR adds an additional claim to the `openid` scope.  The additional claim, `is_staff`, is populated with the User's `is_staff` value.  

This new claim is checked at the front-end of the refine/staff dashboard.  If `is_staff` is false, then the user will not be able to login and will be shown an error message.

### DIFF
--- a/frontend/staff-dashboard/src/App.tsx
+++ b/frontend/staff-dashboard/src/App.tsx
@@ -2,7 +2,8 @@ import { Refine } from "@pankod/refine-core";
 import { Icons, notificationProvider } from "@pankod/refine-antd";
 import routerProvider from "@pankod/refine-react-router-v6";
 import "@pankod/refine-antd/dist/styles.min.css";
-import { useAuthProvider } from "hooks/useAuthProvider";
+import {useAuthProvider } from "hooks/useAuthProvider";
+import {useAuth} from "react-oidc-context";
 import {
   Title,
   Header,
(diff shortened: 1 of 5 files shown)


jkachel (approved):
Tested with a non-staff user (going directly to `/staff-dashboard` and after logging in first) and was correctly denied access. Tested with a staff user in same way; worked fine. Tested with admin; that also was OK. Tested the refresh and that also was good. LGTM 👍

PR #974, opened 2022-09-15
data migration for certificate index page

aliraza-abbasi:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested
- [ ] Settings

Data migration fix for #692 #693 

#### What's this PR do?
Data Migration to create or update slug certificate index page

- Open CMS and check if Certificate Index Page
- Edit this page and go to promote tab
- check that if the slug value is `certificate` So this migration will do nothing for testing you can change the slug value and then run the migration to see the slug value again it should be fixed to `certificate`
- If Certificate Index Page is not in CMS run this migration the page should be created and its slug value should be `certificate`

(Optional)

(Optional)

(Optional)


cms/migrations/0026_certificate_index_page_slug.py
```diff
@@ -0,0 +1,59 @@
...
+            if not certificate_index:
+                cert_index_content_type, _ = ContentType.objects.get_or_create(
+                    app_label="cms", model="certificateindexpage"
+                )
+                certificate_index = cms_models.CertificateIndexPage(
+                    title="Certificate Index Page",
+                    content_type_id=cert_index_content_type.id,
+                    slug=CERTIFICATE_INDEX_SLUG,
+                )
+                home_page.add_child(instance=certificate_index)
```
arslanashraf7, line 40: I think we already have a api/utility function to [ensure certificates index pages](https://github.com/mitodl/mitxonline/blob/main/cms/api.py#L192), I'd suggest using that instead of re-writing the whole logic here. Or if there is any specific reason for this, let me know.

arslanashraf7 (changes requested):

cms/migrations/0026_certificate_index_page_slug.py
aliraza-abbasi: fixed

cms/migrations/0026_certificate_index_page_slug.py
```diff
@@ -0,0 +1,44 @@
...
+
+    home_page = Page.objects.filter(
+        content_type=ContentType.objects.get_for_model(cms_models.HomePage), live=True
+    ).first()
+    if home_page:
+        try:
+            home_page.get_specific()
+            ensure_certificate_index()
+        except cms_models.HomePage.DoesNotExist as exc:
+            log.info(
```
arslanashraf7, line 27: We should print it as `log.error`. same for the below error message.

arslanashraf7 (approved):
Looks good 👍 . One minor change and then you can merge it, Would be nice if you could rebase before merge.

cms/migrations/0026_certificate_index_page_slug.py
aliraza-abbasi: fixed

PR #1005, opened 2022-09-16
Fix factory-boy package name and pin

rhysyngsun:
N/A

#### What's this PR do?
This is an attempt to fix the requirements so that #980 can be generated successfully, currently it's conflicting on two points: a) it's pinning to 3.2.1, which conflicts with requirements.txt which is pinned to 3.2.0 b) it's using the legacy `factory_boy` package.

Tests should pass.

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -3,7 +3,7 @@ bpython
 codecov
 ddt
 django-debug-toolbar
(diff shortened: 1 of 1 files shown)


jkachel (approved):
LGTM 👍

Issue #975, opened 2022-09-15
broken link to flexible pricing form on about page pop-up


pdpinch:
### Steps to Reproduce

1. go to https://rc.mitxonline.mit.edu/courses/course-v1:MITxT+8.01.1x/
2. click on "Enroll Now" button
3. click on [Need financial assistance?](https://rc.mitxonline.mit.edu/courses/test-page-for-mechanics-kinematics-dynamics/mitx-mooc-financial-assistance-request/) link in the pop-up.

### Expected Behavior

- link to https://rc.mitxonline.mit.edu/courses/course-v1:MITxT+8.01.1x/mitx-mooc-financial-assistance-request/

### Actual Behavior

- link to https://rc.mitxonline.mit.edu/courses/test-page-for-mechanics-kinematics-dynamics/mitx-mooc-financial-assistance-request/

rachellougee:
@pdpinch Just to confirm if we should always use the link under Price in CMS? 

For example, in some DEDP courses like https://rc.mitxonline.mit.edu/courses/course-v1:MITx+14.310x/, the flexible pricing request currently links to https://rc.mitxonline.mit.edu/courses/course-v1:MITx+14.310x/data-analysis-for-social-scientists-request-flexible-pricing/. But the link provided in CMS (https://rc.mitxonline.mit.edu/cms/pages/10/edit/) is https://rc.mitxonline.mit.edu/programs/program-v1:MITx+DEDP/dedp-micromasters-program-financial-assistance-form/ (for the program). In this case, what's the correct link?

pdpinch:
> @pdpinch <https://github.com/pdpinch> Just to confirm if we should always use the link under Price in CMS?
> 
Yes, that is my expectation, at least for now.

PR #1010, opened 2022-09-19
975 broken link to flexible pricing form on about page pop up

rachellougee:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/975

#### What's this PR do?
This PR is to fix the link to the flexible pricing form on the popup, It should be the same link under Price
- For any course in a program, if the program has multiple flexible assistance forms (See DEDP course on RC. e.g https://rc.mitxonline.mit.edu/courses/course-v1:MITx+14.310x/), it should always link to the form attached to the program
- For any non-program course, should use the flexible assistance form created under this course page
    - Note that the link should be valid even if there is no course selected for this flexible assistance form in this case (e.g. the link in the ticket, it's fixed by selecting a course in cms, but it could work without it?)

Make sure the following pages are created in CMS
- A program page and its flexible assistance form
- A course page for the above program (Link under Price should be to the program's flexible assistance form)
- A non-program course page and its flexible assistance form (Link under Price should be to this flexible assistance form)

For the program course, `Need financial assistance` should always link to the flexible assistance form for the program

For the non-program course, `Need financial assistance` should link to the flexible assistance form for this course
   - it shouldn't be 404 even if there is no course selected for this flexible assistance form

Course in a program
image
image

Non-program course
image
image


rachellougee:
@jkachel Could you review it since we talked about it last friday? thanks

jkachel (approved):
Seems good to me - tested with course and program flex pricing pages with and without a courseware object selected and it grabbed the right form each time. LGTM 👍

Issue #1024, opened 2022-09-21
Bug: deactivating products on RC results in a 500 error


jkachel:
### Steps to Reproduce

See Sentry issue: https://sentry.io/organizations/mit-office-of-digital-learning/issues/3609909359/?alert_rule_id=8724694&alert_timestamp=1663787179172&alert_type=email&environment=rc&project=5864687&referrer=alert_email

1. On RC, use the admin to set the Is Active flag to false for a given product. 
2. Navigate to the course details page for the product's course.

### Expected Behavior

The course details page should render, minus the now-disabled upsell messaging.

### Actual Behavior

You get a 500 error. 

### Stacktrace

https://sentry.io/organizations/mit-office-of-digital-learning/issues/3609909359/?alert_rule_id=8724694&alert_timestamp=1663787179172&alert_type=email&environment=rc&project=5864687&referrer=alert_email

### Related Issues

Noticed this during a sanity check for https://github.com/mitodl/hq/discussions/227#discussioncomment-3702377 .

arslanashraf7:
Closed via #1026

PR #1026, opened 2022-09-22
fix: product discount calculation for inactive product

arslanashraf7:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
- [ ] Settings

#1024 

#### What's this PR do?
- Adds checks for inactive products (None) for discount calculation on course details page

- Follow the instruction on #1024 , along with below:
1. Make sure you have a discount with flexible pricing and associate it with a product discount
2. Make sure to have an approved flexible pricing request for that course run and user

After the above steps, upon visiting the course you should not see any errors

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -862,7 +862,7 @@ class CoursePage(ProductPage):
                 ecommerce_product, request.user
             )
 
-            if discount.check_validity(request.user):
+            if discount and discount.check_validity(request.user):
                 log.debug(
                     f"price is {ecommerce_product.price}, discount is {discount.discount_product(ecommerce_product)}"
                 )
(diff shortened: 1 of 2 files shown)


arslanashraf7:
@jkachel or (Anyone reviewing):

I think the reproduction steps for #1024 were incomplete, because simply making the product inactive doesn't break things, It breaks when a user has an approved flexible pricing request for that course and we try to calculate discount on this product.


There were two possible solutions:

1. We just apply the discount if the product is _Active_ (This is what i did in this PR)
2. We apply the discount for inactive products too

The approach will change for whichever solution we use from above, I used solution#1. I just added checks when `product=None` because we are dealing with active products and in the case of Inactive, the product is None. I think this makes sense too, I think we shouldn't calculate the discounts for the inactive product.

If we want to got for solution#2, we will have to use `all_products` manager for fetching the product for calculating discounts.

Issue #960, opened 2022-09-09
generate a list of paid & enrolled legacy users to upgrade automatically


pdpinch:
As a legacy MicroMasters user who:

1. Is enrolled in a current MITx Online course (3T2022 or 1T2023)
2. has paid
3. has not used all my exam attempts , 

I need to be enrolled in the verified track of my course(s).

As a CS teammember, I'd like to be able to confirm for specific users that contact CS that this will happen.

#### Acceptance Criteria:

- [ ] generate a list of users (emails) and course runs that should be upgraded
- [ ] write a management command or similar for enrolling these users in their courses as verified 
- [ ] since users can enroll, and users can pay, between now and the launch of ecommerce, we will need to be able to run this more than once. 
- [ ] Since these users have paid on micromasters, and AR/Finance has already accounted for their payments, no special tracking is required. 

#### Related issues 

- [ ] test importing of grades into open edX  
- [ ] #961

annagav:
@pdpinch Is the first checkbox, generating the list of users that satisfy all 3 points above must be generated as result of a management command on micromasters? I might have missed this but I don't believe we have any representation of users exam attempts in mitxonline.

pdpinch:
It doesn't matter to me where we run it. I think @rhysyngsun agrees with you that it needs to be run on micromasters. 

Do note though, that I think there's an edge case with how we calculate exam authorizations on micromasters, https://github.com/mitodl/micromasters/issues/5219

pdpinch:
We can write a management command in micromasters to 

1. find everyone who is authorized for the exam
2. upgrade learners with an _existing_ enrollment in mitxonline to verified

pdpinch:
While we're doing this, we should also capture a list of all the users who are authorized but have _not_ enrolled, which can be used as an input to #

annagav:
@pdpinch latter proposed condition is going to give very different result, then the original request on the issue. The second approach, does not involve any current enrollment information. Is that what we want?

pdpinch:
Sorry to distract. I was thinking that the two lists were opposites of each other, and therefor easy to create at the same time. If that's not the case you can ignore for now and we'll have a follow-up for #961.

PR #1039, opened 2022-09-23
Adds management command to generate enrollment codes for legacy learners

### LINKED ISSUE #961: generate enrollment codes for paid but not enrolled legacy users
As a legacy MicroMasters user who:

1. Is **not** enrolled in a current MITx Online course (3T2022 or 1T2023)
2. has paid
3. has not used all my exam attempts , 

I'd like to contact CS to request an enrollment code (100% off coupon code/discount) for future use. 

As a CS teammember, I'd like to be able to distribute enrollment codes for anyone who has paid, so they can enroll in a verified track and take the proctored exam when they like. 

#### Acceptance Criteria:

- [ ] generate a list of users (emails) and course runs of users who have paid, but aren't enrolled in a current run 
- [ ] write a management command or similar to create 100% off discounts on mitxoline for these learners
- [ ] since the codes will be distributed only on request, we could generate this once and not be concerned with updates. Any user paying between now and the launch of ecommerce is expected to enroll. 
- [ ] Since these users have paid on micromasters, and AR/Finance has already accounted for their payments. The discount codes are primarily a convenience for the user, but we should consider a naming scheme that makes it easy for AR to recognize and ignore them. 

#### Related issues 

- [ ] test importing of grades into open edX  
- [x] #960

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #961 

#### What's this PR do?

Adds a management command, `generate_legacy_enrollment_codes`, that will generate a 100% off discount code for learners that have paid for a courserun but aren't enrolled for it. (See #960.) The discount codes are:
* 100% off
* One-time use
* Associated with the specified course run
* Set as a user discount (so, barring other discounts, they will autoapply when the learner attempts to purchase the course)
* Formatted `MM-legacy-<username>-<readable_id>`
The generated codes and the email addresses for the learners they apply to are written to a CSV file.

This also adds a management command, `find_paid_unenrolled_learners`, which will look in the _MITxOnline_ database for users that have purchased a courserun but aren't enrolled in it at all. This outputs in a format that can be ingested by the `generate_legacy_enrollment_codes` command.

The `generate_legacy_enrollment_codes` checks for the existence of the learner and the specified courserun but nothing else, so it can be tested with any learner that exists in the system. 

1. Generate an input CSV file. The file should be in the format `learner email,readable id`. 
2. Run the `generate_legacy_enrollment_codes` command: `./manage.py generate_legacy_enrollment_codes <learner list file> <code output file>`
3. Check that the discount codes are created and set up correctly in Django Admin.

Once the codes are generated, a learner that doesn't have any other automatic discounts (like, for instance, financial assistance) should have the newly generated discount applied when they add the course to their cart automatically. If the learner does have another discount, the code should work to apply it manually. (Financial assistance discounts take precedence over other discounts other than ones that are manually applied.)

### DIFF
--- a/ecommerce/management/commands/generate_legacy_enrollment_codes.py
+++ b/ecommerce/management/commands/generate_legacy_enrollment_codes.py
@@ -1,0 +1,136 @@
+"""
+Generates enrollment (discount) codes for legacy users. 
+
+This is for learners that have paid for a course but aren't enrolled in a 
+course run with a verified enrollment. The list is generated outside this 
+command and is passed to it as a CSV file. The format of the file is:
+
+```
+learner email,course readable id or course readable id or micromasters course ID
+```
+
+and this command will generate single-use per-user discounts for each learner
+and course combination (if a product exists for it). The course must be 
+available for enrollment and must also have a product associated with it. 
+"""
+from django.core.management import BaseCommand
+import csv
+import uuid
+from django.db import transaction
+
+from courses.models import CourseRun, Course
+from ecommerce.models import Discount, DiscountProduct, UserDiscount
+from ecommerce.constants import REDEMPTION_TYPE_ONE_TIME, DISCOUNT_TYPE_PERCENT_OFF
+from users.models import User
+from micromasters_import.models import CourseId
+
+
+class Command(BaseCommand):
+    """
+    Generates enrollment codes for legacy paid learners
+    """
+
+    help = "Generates enrollment codes for legacy paid learners"
+    CODE_PREFIX = "MM-prepaid-"
+
+    def add_arguments(self, parser) -> None:
+        parser.add_argument(
+            "input_file",
+            type=str,
+            help="Input file. See documentation for file format.",
+        )
+        parser.add_argument(
+            "output_file", type=str, help="File to write the results to."
+        )
+
+    def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
+        output_codes = []
+
+        with open(kwargs["input_file"], newline="") as csvfile:
+            reader = csv.reader(csvfile, delimiter=",", quotechar="\\")
+
+            for row in reader:
+                user = User.objects.filter(email=row[0]).first()
+
+                if user is None:
+                    self.stderr.write(
+                        self.style.ERROR(
+                            f"Can't find account for user {row[0]}, skipping row"
+                        )
+                    )
+                    continue
+
+                try:
+                    mmid = int(row[1])
+                    mapped_course = CourseId.objects.filter(
+                        micromasters_id=mmid
+                    ).first()
+
+                    if mapped_course is None:
+                        self.stderr.write(
+                            self.style.ERROR(
+                                f"Can't find mapped MicroMasters course for {mmid}, skipping row"
+                            )
+                        )
+                        continue
+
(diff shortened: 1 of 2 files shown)


rachellougee:
The code looks good to me. For `find_paid_unenrolled_learners` command, it works as expected. 
But I got errors when running `generate_legacy_enrollment_codes`, It looks like some discount code is too long and there is a limit of 50 characters. After I remove a few characters in the code, it then ran fine. so we might want to expand discount_code?

rachellougee (approved):

jkachel:
@pdpinch thoughts on the generated enrollment code format? I erred on the side of more or less human readable but we could do something less so (like, say, `MM-legacy-<user_id>-<courserun_id>` (or something completely random or something else). Or, I can keep it as is and up the length limit on the codes; I don't recall there being a real reason for the code length to be 50 other than maybe that's how it was in MicroMasters.

annagav:
@jkachel I don't think that we can provide course run key for the input file for this task, it's just a course id.

jkachel:
Ok - I updated in https://github.com/mitodl/mitxonline/pull/1039/commits/08edfc5c5596562efda14366b10d4cd0933a2051 to look for a matching course object as well. It'll use the `first_unexpired_run` to determine what courserun to use in that case (since products can't be for a course, and as such neither can discounts).

rachellougee (approved):

rachellougee:
The updated code looks good to me. 
I tested a couple of course IDs, and it was able to generate discount code for unexpired run for course ID. 
Just need to figure out the format of the coupon code in case someone's user name is too long

ecommerce/management/commands/generate_legacy_enrollment_codes.py
```diff
@@ -0,0 +1,106 @@
...
+                        self.style.ERROR(
+                            f"Can't find account for user {row[0]}, skipping row"
+                        )
+                    )
+                    continue
+
+                courserun = CourseRun.objects.filter(courseware_id=row[1]).first()
+
+                if courserun is None:
+                    course = Course.objects.filter(readable_id=row[1]).first()
```
annagav, line 61: I think here you might need to map the `course_id` from MM into `readable_id` in MITxOnline through this mapping model CourseId from micromasters_import/models.py.

annagav (changes requested):

ecommerce/management/commands/generate_legacy_enrollment_codes.py
rachellougee: CourseId in micromasters_import/models.py is the auto-incremented primary key of the course model, not the string field https://rc.mitxonline.mit.edu/admin/micromasters_import/courseid/7/change/. Are we exporting the ID 7 to csv from https://micromasters-rc.odl.mit.edu/admin/courses/course/7/change/?

pdpinch:
We should use something less transparent, but still filterable, like: MM22-<uuid>

jkachel:
https://github.com/mitodl/mitxonline/pull/1039/commits/ca4f92bdc673c023e6cdedc3609f5d9f003c15d0 updates the code format and adds in a check for the numeric ID mapping if the supplied course ID is an integer.

annagav (approved):
👍

PR #1049, opened 2022-09-23
Updates the order fulfillment code to wait for the transaction to complete before sending message

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

n/a

#### What's this PR do?

Adds the call to queue the order receipt email to the post-commit hook for the transaction when fulfilling an order, so the email task gets the order after the state has changed (rather than in the middle, which means it processes it as pre-transaction and throws an error). 

Complete an order. You should get the receipt email.

### DIFF
--- a/ecommerce/mail_api_test.py
+++ b/ecommerce/mail_api_test.py
@@ -17,7 +17,9 @@ def products():
         return ProductFactory.create_batch(5)
 
 
-def test_mail_api_task_called(mocker, user, products, user_client):
+def test_mail_api_task_called(
+    mocker, user, products, user_client, django_capture_on_commit_callbacks
+):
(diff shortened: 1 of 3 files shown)


rachellougee (approved):

rachellougee:
The manual test works for me, Got a receipt after completing the order. Just need to fix the automated test

jkachel:
Requesting re-review for sanity check.. the automated tests should work now.

rachellougee (approved):
Probably don't need  ```as callbacks```, but It's fine as it is
Looks good to me

jkachel:
closed this by accident.. reopening to merge

Issue #1051, opened 2022-09-24
Don't display "active" on the dashboard when it is past the course run's `course_end` date


pdpinch:
### Steps to Reproduce

1. Create a course without an `enrollment_end` date, and with an `course_end` date in the past
2. enroll a user manually, for example by using the `create_enrollment` management command

### Actual Behavior

- the course card displays "Active" followed by the course start date

### Expected Behavior

- The course card should display "Ended" followed by the course end date

### Screenshot or Screencast

<img width="1130" alt="image" src="https://user-images.githubusercontent.com/430126/192116726-2fe7ba72-4a0a-48cb-96d0-3fab235e53a4.png">

PR #1057, opened 2022-09-26
1051: Don't display "active" on the dashboard when it is past the course run's course_end date

collinpreston:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/1051

#### What's this PR do?
When viewing the dashboard, any course cards for a course which have an `end_date` which is prior to the current date will display "Ended" and then the date on which they ended.

1. Create a Course Run with an `end_date` in the past.
2. enroll your test user in the Course Run.
3. Go to the Dashboard page.
4. Verify that the course card on the dashboard shows "Ended" and the `end_date` from the Course Run.

Screen Shot 2022-09-26 at 15 05 28
Screen Shot 2022-09-26 at 15 05 34

### DIFF
--- a/frontend/public/src/components/EnrolledItemCard.js
+++ b/frontend/public/src/components/EnrolledItemCard.js
@@ -17,7 +17,7 @@ import { partial } from "ramda"
 
 import { ALERT_TYPE_DANGER, ALERT_TYPE_SUCCESS } from "../constants"
 import GetCertificateButton from './GetCertificateButton'
-import { isFinancialAssistanceAvailable, isLinkableCourseRun, generateStartDateText } from "../lib/courseApi"
+import { isFinancialAssistanceAvailable, isLinkableCourseRun, generateStartDateText, courseRunStatusMessage } from "../lib/courseApi"
 import { isSuccessResponse } from "../lib/util"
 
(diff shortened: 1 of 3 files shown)


frontend/public/src/components/EnrolledItemCard.js
```diff
@@ -297,6 +298,22 @@ export class EnrolledItemCard extends React.Component<
     const pageLocation = enrollment.run.page
     const menuTitle = `Course options for ${enrollment.run.course.title}`
 
+    let courseRunStatusMessage = null
+
+    if (startDateDescription !== null) {
+      if (startDateDescription.active) {
+        if (moment(enrollment.run.end_date).isBefore(moment())) {
```
arslanashraf7, line 305: In my opinion, this can be a reusable function, Should we convert this into a util function just like [isWithinEnrollmentPeriod](https://github.com/mitodl/mitxonline/blob/main/frontend/public/src/lib/courseApi.js#L28-L37) and then use that function here? What do you think?

arslanashraf7 (approved):
Looks great 👍 . I also tested it and it works well. I just added a small comment regarding some refactoring. Feel free to merge after that.

PR #1059, opened 2022-09-27
Upgrade legacy learners that paid and are enrolled, have exam attempt

annagav:
Fix https://github.com/mitodl/mitxonline/issues/960

#### What's this PR do?
This is a management command that take in a file and upgrades learners in their current enrollment

Create a csv file that has data of a learner in your local mitxonline environment, that is auditing a a current course.
In the csv file add a record: <learner_email>, <course_run_key>
Run the command `./manage.py upgrade_eligible_users input_file.csv`
The enrollment should get upgrades to 'verified'.

### DIFF
--- a/courses/management/commands/upgrade_eligible_users.py
+++ b/courses/management/commands/upgrade_eligible_users.py
@@ -1,0 +1,85 @@
+"""
+Upgrades legacy users in the enrollment.
+This is for learners that have paid for a course and are enrolled in an upcoming
+course run with a remaining exam attempt.
+The list is generated outside this
+command and is passed to it as a CSV file. The format of the file is:
+```
+learner email,courserun readable id
+```
+"""
+from django.core.management import BaseCommand
+import csv
+
+from courses.api import create_run_enrollments
+from courses.models import CourseRun, CourseRunEnrollment
+from ecommerce.models import Discount, DiscountProduct, UserDiscount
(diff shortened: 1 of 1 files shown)


courses/management/commands/upgrade_eligible_users.py
```diff
@@ -0,0 +1,85 @@
...
+
+                if course_run is None:
+                    self.stderr.write(
+                        self.style.ERROR(
+                            f"Can't find courserun for {row[1]}, skipping row"
+                        )
+                    )
+                    continue
+                upgrade_count = 0
+                enrollment = CourseRunEnrollment.objects.filter(
```
jkachel, line 62: missing the fetch part - this returns a QuerySet

jkachel (changes requested):
Algorithm looks good and `create_edx_enrollments` is the right way to go about this (since that'll also perform the upgrade in edX); there's a simple error that needs to be resolved but otherwise is OK.

jkachel (changes requested):
One more change regarding handling failed enrollments

annagav:
@jkachel I added a check for edx_request_success to log out failed enrollments. Thanks

jkachel (approved):
LGTM - ready to merge once tests pass (looks like it needs a black run) 👍

Issue #1044, opened 2022-09-23
When referring to a course in email, don't include the full course id


pdpinch:
courses in email should always be formatted as “<number> <title>”. 

In  this example of a refund email, it should be “Test101 
Engineering Test Course” 

![image](https://user-images.githubusercontent.com/430126/192004638-fff51157-490d-499b-a6f7-c27001e4b965.png)

collinpreston:
@pdpinch I assume that a course referenced in the Flexible Price emails should also follow the described formatting?

pdpinch:
> I assume that a course referenced in the Flexible Price emails should
also follow the described formatting?

yes

PR #1000, opened 2022-09-16
Adds an email message that is sent when an order is refunded.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested
  - [X] Changes have been manually tested

#### What are the relevant tickets?

Fixes #957

#### What's this PR do?

Adds a new email message that gets sent when an order is refunded. The text of the messages is in the issue (#957). 

#### How should this be manually tested?

Automated tests should work.

The easiest way to trigger the refund is through a Django shell; this method does not require that your system be able to accept callbacks from CyberSource. (Refunding through the Django Admin does attempt to perform the refund through CyberSource.) 

From a shell, load a FulfilledOrder, then call the "refund" method:
```
from ecommerce.models import FulfilledOrder

order = FulfilledOrder.objects.first() # change this accordingly

response_data = { "id": "an id" } # the only required field here is id 

order.refund(api_response_data=response_data, amount="50", reason="testing") # change amount and reason accordingly
```
This should trigger the email. Note that this will change the status of the order in the system. 

If your system can receive callbacks from CyberSource, you can post a refund through the EBC. That should also trigger the refund email.

PR #1060, opened 2022-09-27
1044: when referring to a course in email don't include the full course

collinpreston:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/1044

#### What's this PR do?
Formats all of the emails from `mitxonline` where the courseware title is shown to include the courseware number in the format of <courseware number> <courseware title>.

This change impacts the following emails:

1. Refund email
2. Flexible Price email (accepted, reset, pending manual approval) 

Trigger a refund email following the instructions listed on this [PR](https://github.com/mitodl/mitxonline/pull/1000).
Trigger flexible price emails by setting up a Flexible Price following [these instructions ](https://github.com/mitodl/mitxonline/blob/main/docs/source/ecommerce/flexible_pricing.rst).  Then submit a flexible price request for your user and change the flexible price request status through the [refine-staff dashboard](http://mitxonline.odl.local:8013/staff-dashboard/).

Ensure that wherever the courseware title is shown, the courseware number is also shown.  You should also test that this works with a Flexible Price which is associated to a Program and a Course without a Program.

Screen Shot 2022-09-27 at 14 47 00
Screen Shot 2022-09-27 at 14 46 52

### DIFF
--- a/ecommerce/mail_api.py
+++ b/ecommerce/mail_api.py
@@ -87,9 +87,7 @@ def send_ecommerce_refund_message(order_record):
                 recipient,
                 {
                     "order": order_record,
-                    "readable_id": line.purchased_object.readable_id
-                    if line and len(line.purchased_object.readable_id.split("+")) < 2
-                    else line.purchased_object.readable_id.split("+")[1],
+                    "readable_id": line.purchased_object.course_number,
                     "title": line.purchased_object.title if line else None,
                     "transaction_amount": transaction.amount.quantize(Decimal("0.01")),
                 },
--- a/ecommerce/mail_api.py
+++ b/ecommerce/mail_api.py
@@ -87,7 +87,7 @@ def send_ecommerce_refund_message(order_record):
                 recipient,
                 {
                     "order": order_record,
-                    "readable_id": line.purchased_object.readable_id if line else None,
+                    "readable_id": line.purchased_object.readable_id.split('+')[1] if line else None,
                     "title": line.purchased_object.title if line else None,
                     "transaction_amount": transaction.amount.quantize(Decimal("0.01")),
                 },
--- a/ecommerce/mail_api.py
+++ b/ecommerce/mail_api.py
@@ -87,7 +87,9 @@ def send_ecommerce_refund_message(order_record):
                 recipient,
                 {
                     "order": order_record,
-                    "readable_id": line.purchased_object.readable_id.split('+')[1] if line else None,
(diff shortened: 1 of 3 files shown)


ecommerce/mail_api.py
```diff
@@ -87,7 +87,9 @@ def send_ecommerce_refund_message(order_record):
                 recipient,
                 {
                     "order": order_record,
-                    "readable_id": line.purchased_object.readable_id if line else None,
+                    "readable_id": line.purchased_object.readable_id
+                    if line and len(line.purchased_object.readable_id.split("+")) < 2
+                    else line.purchased_object.readable_id.split("+")[1],
```
rachellougee, line 92: Does it make sense to use the existing property course_number from the Course model? something like line.purchased_object.course_number

rachellougee (commented):
The functionality looks good. Have a comment on the added formatting code.
Is the flexiblepricing change part of this request? If so, are we not changing send_financial_assistance_request_denied_email?

collinpreston:
> "readable_id": line.purchased_object.readable_id

I interpreted the original issue as saying courses in an email (or any email) should be formatted as described and update the Flexible Pricing emails based on that.

ecommerce/mail_api.py
collinpreston: Yeah this is a good idea!  It will work well for the CourseRun model, but I don't think it can be used for the Course model.

collinpreston: Good find!

collinpreston: @rachellougee I also added the course_number property to the Course model since it previously only existed on the Course_Run model (even though that just referenced the Course model).

ecommerce/mail_api.py
```diff
@@ -87,7 +87,7 @@ def send_ecommerce_refund_message(order_record):
                 recipient,
                 {
                     "order": order_record,
-                    "readable_id": line.purchased_object.readable_id if line else None,
+                    "readable_id": line.purchased_object.course_number,
```
rachellougee, line 90: You might want to do the same thing here to check if it's CourseRun before accessing course_number here:
 line.purchased_object.course_number if line and isinstance(line.purchased_object, CourseRun)

Right now it's fine because we don't have the program run as product.

rachellougee (approved):
Works for me. I left a comment, but feel free to merge in

PR #1070, opened 2022-09-28
Added data model for ProgramRequirement

### LINKED ISSUE #1066: Program Requirements: Data Models
We need a data model to represent program requirements. These take the form of required courses and electives

See the [previous discussion for more detail](https://github.com/mitodl/hq/discussions/3).

An additional case that is related and probably best supported by our solution to this is the determination of which program(s) a course is in. Currently this is limited to a single program because the relationship is just a FK on `Course` to `Program`. In the future we'll need to be able to have a course in multiple programs so we should try to address this as part of this work. Additionally, we don't want to have this represented in two different places, because this can cause issues if these two disagree.

Given both of the above, the best approach is probably going to be to use the treebeard library mentioned in the discussion. I'll work a bit on trying to verify this works as I think it would.

rhysyngsun:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes #1066 

#### What's this PR do?
Adds a data model for program requirements

- Migrations should run and tests should pass
- You should be able to edit the requirements for a program:
  - Add, remove, reorder top-level sections (e.g. "Required Courses" in the screenshot)
  - Add, remove, reorder second-level requirements (courses or operators)
  - Add, remove, reorder third-level courses under second-level operators

#### Screenshots

Screenshot 2022-10-26 at 11-26-00 no-id Data Economics and Development Policy Change program Django site admin


rhysyngsun:
@pdpinch 

I think I've got this where I want it. Some examples of how this would get used:

```python
# which programs is a course in?
Program.objects.filter(requirements__course=course).distinct()

# which courses are in a program?
Course.objects.filter(in_programs__program=program).distinct()

# has a learner passed all the requirements for a certificate?
def has_earned_program_cert(user, program):
    passed_courses = Course.objects.filter(
        in_programs__program=program,
        courserungrade__user=user,
        courserungrade__passed=True,
    )
    root = ProgramRequirement.get_root_nodes().get(program=program)

    def _has_earned(node):
        if node.is_root() or node.is_all_of_operator:
            # has passed all of the child requirements
            return all(_has_earned(child) for child in node.get_children())
        elif node.is_min_number_of_operator:
            # has passed a minimum of the child requirements
            return len(filter(_has_earned, node.get_children())) >= int(node.operator_value)
        elif node.is_course:
            # has passed the reference course
            return node.course in passed_courses

    return _has_earned(root)

has_earned_program_cert(user, program)
```

Compare `has_earned_program_cert` to the function in MicroMasters: https://github.com/mitodl/micromasters/blob/master/grades/api.py#L243-L270

A few things to note when comparing these:
- Some logic in MicroMasters is in MMTrack, so it's actually a little more complicated than at-a-glance
- The MicroMasters implementation doesn't support mutally exclusive electives and would require another data model and more logic to support it

rachellougee (approved):
Code looks good. I've applied the schema and added some test rows, but I am not sure about how it's been used. I ran into errors when calling ```has_earned_program_cert```, but that's different issue

jkachel (commented):
LGTM - this seems pretty understandable to me. 👍

pdpinch:
> The MicroMasters implementation doesn't support mutually exclusive electives and would require another data model and more logic to support it

How are mutually exclusive electives configured in this model?

To be more specific, how would we model the (current) DEDP requirements, described here: https://micromasters.mit.edu/dedp/course-overviews/ 

Since this is pretty complex functionality, does it make sense to include a migration or a management command that sets up the DEDP requirements to make development easier?

rhysyngsun:
@pdpinch mutually exclusive courses are modelling here in the tests: https://github.com/mitodl/mitxonline/pull/1070/files#diff-3a704dc028b439e999fab8f01530da708a1eaafe3440a2303460953226b06f26R50-R55

The current DEDP requirements would be modeled like the remainder of that test fixture.

courses/models_test.py
```diff
@@ -15,13 +16,56 @@
...
 from users.factories import UserFactory
 
 pytestmark = [pytest.mark.django_db]
 
 
+@pytest.fixture
+def program_with_requirements():
+    program = ProgramFactory.create()
+    required_courses = CourseFactory.create_batch(3, program=program)
+    elective_courses = CourseFactory.create_batch(3, program=program)
```
pdpinch, line 30: Probably not important, but DEDP currently has 4 electives:

14.009x
14.73x
14.740x
14.750x

courses/models_test.py
```diff
@@ -15,13 +16,56 @@
...
+    mut_exclusive_courses = CourseFactory.create_batch(3, program=program)
```
pdpinch, line 31: I'm not sure I understand how mutually exclusive works. 

In DEDP, if you take both 14.73x and 14.009x you will only get credit towards the program certificate for one of them. 

So would this be 

```suggestion
    mut_exclusive_courses = CourseFactory.create_batch(2, program=program)
```

courses/models.py
```diff
@@ -892,3 +903,127 @@ class Meta:
...
+    required_courses = root.add_child(operator=ProgramRequirement.Operator.all_of, title="Required Courses")
+    required_courses.add_child(course=course1)
+    required_courses.add_child(course=course2)
+
+    # at least two must be taken
+    elective_courses = root.add_child(
+        operator=ProgramRequirement.Operator.MIN_NUMBER_OF,
+        operator_value=2,
+        title="Elective Courses"
+    )
```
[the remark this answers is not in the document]
rhysyngsun: Not sure I follow completely but I think yes

courses/models_test.py
rhysyngsun: So to summarize the logic here in psuedocode:
```python
has_cert = (
   # required
   # ALL_OF
   all(
      # must take all of these
      [r.passed for r in required_courses]
   )
   # at least min_elective_count electives passed
   # MIN_NUM_OF, value=5
   and sum(
      [1 for r in elective_courses if r.passed] +
      # mutually exclusive, only 1 counts
       # MIN_NUM_OF, value=1
      [1 if any([r.passed for r in mut_exclusive_courses]) else 0]
   ) > 5
)
```

The structure here is very flexible, it can support fairly complex things like this (I've seen things like this in residential programs):

```python
has_cert = (
   # ALL_OF
   all(
       # must take all of these
       [r.passed for r in required_courses] + 
       # any of these, but you MUST take one
       # MIN_NUM_OF, value=1
       [any([r.passed for r in mut_exclusive_required_courses1])]
       # any of these, but you MUST take one
       # MIN_NUM_OF, value=1
       [any([r.passed for r in mut_exclusive_required_courses2])]
   )
   # at least min_elective_count electives passed
   # MIN_NUM_OF, value=5
   and sum(
      [1 for r in elective_courses if r.passed] +
      # mutually exclusive, only 1 counts
      # MIN_NUM_OF, value=1
      [1 if any([r.passed for r in mut_exclusive_courses1]) else 0]
      # mutually exclusive, only 1 counts
      # MIN_NUM_OF, value=1
      [1 if any([r.passed for r in mut_exclusive_courses2]) else 0]
   ) > 5
)
```

rhysyngsun: Very broadly, the data model here representation a logical tree of boolean checks and for MIN_NUM_OF a check of a minimum threshold being met.

rhysyngsun: A more concrete example is probably more useful to illustrate:

```python
program = ProgramFactory.create(name="DEDP")

root_node = ProgramRequirement.add_root(program=program)

required_courses_node = root_node.add_child(
    operator=ProgramRequirement.Operator.ALL_OF, title="Required Courses"
)
for required in [
    CourseFactory.create(readable_id="course-v1:MITxT+14.100x"),
    CourseFactory.create(readable_id="course-v1:MITxT+JPAL102x"),
    CourseFactory.create(readable_id="course-v1:MITxT+14.310x"),
]:
    required_courses_node.add_child(course=required)

elective_courses_node = root_node.add_child(
    operator=ProgramRequirement.Operator.MIN_NUMBER_OF,
    operator_value=2,
    title="Elective Courses",
)
for elective in [
    CourseFactory.create(readable_id="course-v1:MITxT+14.750x"),
    CourseFactory.create(readable_id="course-v1:MITxT+14.740x"),
    [
        CourseFactory.create(readable_id="course-v1:MITxT+14.73x"),
        CourseFactory.create(readable_id="course-v1:MITxT+14.009x"),
    ]
]:
    if isinstance(elective, list):
        elective_courses_node.add_child(
            operator=ProgramRequirement.Operator.MIN_NUMBER_OF, operator_value=1
        )
    else:
        elective_courses_node.add_child(course=elective)
```

rhysyngsun:
> Code looks good. I've applied the schema and added some test rows, but I am not sure about how it's been used. I ran into errors when calling `has_earned_program_cert`, but that's different issue

That was sorta pseudocode, I didn't test it thoroughly because I want to keep this PR just to the data model but the general idea there of walking the tree will work.

courses/models_test.py
pdpinch: Ok. That wasn't what I was expecting, but it makes sense to me, with the tree structure. 

What does this look like through the django admin? How do we add or change requirements?

rhysyngsun: Good question on how to add or change requirements, so a question for you:

Will we have the need (similar to certificates) to display what the historical requirements were when a learner earned a certificate?

courses/models_test.py
```diff
@@ -645,3 +705,197 @@ def test_program_certificate_start_end_dates_and_page_revision(user):
...
+def test_program_requirements_is_min_number_of_operator(operator, expected):
+    """Test is_min_number_of_operator"""
+    assert ProgramRequirement(operator=operator).is_min_number_of_operator is expected
+
+
+def test_courses_in_program(program_with_requirements):
+    """Test CourseQuerySet.courses_in_program"""
+    CourseFactory.create_batch(4)
+
+    courses = Course.objects.courses_in_program(program_with_requirements.program)
```
collinpreston, line 833: Is this test to ensure that an Admin can only add course requirements for courses that are associated with the program, or can the Admin add any course?

courses/serializers.py
```diff
@@ -6,6 +6,7 @@
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.templatetags.static import static
+from mitol.common.utils.collections import filter_dict_by_key_set
```
collinpreston, line 9: Can this import be removed?

requirements.txt
```diff
@@ -166,14 +166,16 @@ django-taggit==1.5.1
...
+    #   -r requirements.in
+    #   wagtail
 django-user-agents==0.4.0
     # via -r requirements.in
 django-webpack-loader==1.4.1
     # via
     #   -r requirements.in
     #   mitol-django-common
-djangorestframework==3.12.4
+djangorestframework==3.14.0
```
collinpreston, line 178: Was this upgrade intentional?

courses/serializers.py
```diff
@@ -500,3 +502,133 @@ def get_certificate(self, user_program_enrollment):
         """
         certificate = user_program_enrollment.get("certificate")
         return ProgramCertificateSerializer(certificate).data if certificate else None
+
+
+class ProgramRequirementDataSerializer(StrictFieldsSerializer):
```
collinpreston, line 507: Are there any caveats that would limit the future reuse of these serializers when displaying program requirements to learners?  I'm guessing no, but figured I might be overlooking something and now is a good time to ask.

requirements.txt
rhysyngsun: It was but it's not necessary anymore, i'll revert.

courses/serializers.py
rhysyngsun: I don't think we'd want to because the data structure is a little more complicated than it needs to be since it's how `MP_Node.dump_bulk` operates.

courses/models_test.py
rhysyngsun: No, this is to verify that the queryset function returns courses that show up in a program's requirements.

collinpreston (approved):
Looks good.  Thanks for answering my curiosities.  auto-checks may need another run

Issue #1069, opened 2022-09-28
Upsell card displays negative values if the discount is greater than the purchase price of the course


jkachel:
### Steps to Reproduce

1. Configure a course with a low enough value so that the financial assistance discount will result in the course price being less than zero. 
2. Apply for and approve a financial assistance request for that course. 
3. Navigate to the course details page.
4. Click Enroll.

On RC, this can be tested with Test External Checkout Course - https://rc.mitxonline.mit.edu/courses/course-v1:MITx+CHECKOUT/ . 

### Expected Behavior

The price displayed should be the course price minus the discount, with a floor of $0.00. 

### Actual Behavior

The price displayed doesn't have a floor, so the price displayed may be less than $0. 

### Related Issues

Found while working through #1006 .

### Screenshot or Screencast

Post-approved flex pricing page:
![image](https://user-images.githubusercontent.com/945611/192870476-a78dea8b-012d-4881-9bb0-07ba31ad1cd7.png)

Upsell card with wrong price:
![image](https://user-images.githubusercontent.com/945611/192870515-de4c77a2-4d66-47b5-8899-a0bc65c5e985.png)

pdpinch:
So far, it seems like everyone wants to use percent-off discounts for financial assistance. Can this happen with percent-off discounts?

pdpinch:
Oh wait. I'm wrong. I'm going to delete my comment.

arslanashraf7:
Closing via #1079

PR #1079, opened 2022-09-30
fix: show zero for negative prices

arslanashraf7:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

#1069 

#### What's this PR do?
 - Fixes the frontend when discount price is more then actual price and the price shows negative value.

Good steps mentioned in #1069 

#### For Reviewer
Are there any other places you think this should be fixed too?

image

image

### DIFF
--- a/frontend/public/src/lib/util.js
+++ b/frontend/public/src/lib/util.js
@@ -227,7 +227,7 @@ export const getErrorMessages = (
 }
 
 export const formatLocalePrice = (amount: number | null) => {
-  if (amount === null) amount = 0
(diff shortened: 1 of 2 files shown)


arslanashraf7:
@jkachel In this specific case we are actually calculating the prices on the frontend, I fixed this in the display format function to make it applicable in all the places that it uses. Does this make sense, or there might be some other way you would have liked this to be fixed?

jkachel (approved):
LGTM 👍 - this should take care of it, I think

PR #1080, opened 2022-09-30
Adding duplicate checking and expiration date on generated enrollment codes

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

none but was mentioned in https://github.com/mitodl/hq/discussions/281

#### What's this PR do?

Adds two features to the `generate_legacy_enrollment_codes` management command:
- It will now check to see if a code exists for the given user (matching on discount type, redemption type, and amount; product; and user) - if it does, it will skip code creation and emit a message saying as much
- You can now specify the expiration date by passing `--expires <date>` to the command. Dates should be in ISO format (YYYY-MM-DD) though this uses the `dateutil` package so there is some flexibility here. Passed dates are converted to the default timezone (UTC usually). 

1. Generate some codes in the normal fashion. 
2. Re-run the same input file. No new codes should be generated.
3. Delete the generated codes (either directly through the DB or through Django Admin).
4. Run the input file again, but add `--expires <date>` to the command. The generated codes should now have an expiration date.

### DIFF
--- a/ecommerce/management/commands/generate_legacy_enrollment_codes.py
+++ b/ecommerce/management/commands/generate_legacy_enrollment_codes.py
@@ -14,15 +14,19 @@ and course combination (if a product exists for it). The course must be
 available for enrollment and must also have a product associated with it. 
 """
 from django.core.management import BaseCommand
+from django.db import transaction
+
 import csv
 import uuid
-from django.db import transaction
+import dateutil
+import pytz
 
 from courses.models import CourseRun, Course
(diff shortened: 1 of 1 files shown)


rachellougee (approved):
Code and test look good to me. It successfully blocks creating duplicates code and --expires works as expected
@jkachel there is "This branch had an error being deployed" error, I couldn't tell what it is. feel free to merge in if it's not related

jkachel:
merging this - I don't know what the odlbot failure is but I don't think it means anything quite yet

PR #1081, opened 2022-09-30
Fix a small bug in upgrade_eligible_users

annagav:
Related to #1059

#### What's this PR do?
Fix a small bug related to the final print out message counting the number of upgraded users.

The same as in #1059


jkachel (approved):
LGTM 👍

Issue #1086, opened 2022-10-04
Performance: django admin for order editing is selecting all users


rhysyngsun:
### Steps to Reproduce

This is most easily observed in production where we have a large enough number of users for it to manifest. View a particular order and note how slow it is to load.

### Expected Behavior

The page loads in a second or so.

### Actual Behavior

The page takes up to 10 seconds because it's querying for and rendering an html select of all users. See the screenshot.

We should probably make most/all fields on this page readonly because there's no reason for the data to be altered manually - if we did need to do something we'd implement a management command or a refine interface to do it.

### Screenshot or Screencast

![Screenshot from 2022-10-04 12-38-16](https://user-images.githubusercontent.com/28598/193877587-766de5eb-41ef-418a-95a8-b90b306780d2.png)

asadali145:
Just FYI @rhysyngsun @rachellougee and to document,

Actions for an order are also disabled as a result of this ticket. Not sure if we need those or not. It throws 403 forbidden if we try to move using the action buttons.

<img width="1058" alt="Screenshot 2022-10-12 at 4 51 14 PM" src="https://user-images.githubusercontent.com/52656433/195335994-0cb7384e-db58-4c6b-823f-56e0beb9d666.png">
<img width="1394" alt="Screenshot 2022-10-12 at 4 53 36 PM" src="https://user-images.githubusercontent.com/52656433/195336033-004ac899-ab12-40ac-a660-5ca2331f681a.png">

asadali145:
Got it! Thanks

PR #1087, opened 2022-10-04
make order admin page readonly

rachellougee:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/1086

#### What's this PR do?
This PR is to make the Order Admin pages as view only. It affects Order, Fulfilled Order, Pending Order and Refunded Order
image

Go to the order admin page http://mitxonline.odl.local:8013/admin/ecommerce/order/
Open any order, it should be readonly

image

### DIFF
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -153,6 +153,9 @@ class BaseOrderAdmin(FSMTransitionMixin, TimestampedModelAdmin):
     inlines = [OrderLineInline, OrderDiscountInline, OrderTransactionInline]
     readonly_fields = ["reference_number"]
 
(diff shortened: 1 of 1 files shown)


rhysyngsun (changes requested):
I'd suggest a slightly different approach just to future proof this a bit:

https://docs.djangoproject.com/en/4.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.has_change_permission

Implement that and have it return `False` for both the admins here.

rhysyngsun (approved):
We should probably hide the state change buttons at the bottom but I'm going to approve this for now because you properly get a 403 when you try.

PR #1095, opened 2022-10-05
Updates the call to subscribe to edX emails to be in a post-commit hook

### LINKED ISSUE #1093: CourseRunEnrollment.DoesNotExist: CourseRunEnrollment matching query does not exist.
Again. I'm not sure on the steps to reproduce. 

I was able to complete a purchase (using PayPal) but when I returned to mitxonline, I got the banner notification "Unknown error trying to complete order" on my dashboard. 

This is the only error that I see, related to my enrollment. 

Sentry Issue: [MITXONLINE-2D9](https://sentry.io/organizations/mit-office-of-digital-learning/issues/3615579419/?referrer=github_integration)

```
CourseRunEnrollment.DoesNotExist: CourseRunEnrollment matching query does not exist.
  File "courses/tasks.py", line 39, in subscribe_edx_course_emails
    enrollment = CourseRunEnrollment.objects.select_related("user", "run").get(
  File "django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
```

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #1093

#### What's this PR do?

Updates the `create_run_enrollments` call to put its `subscribe_edx_course_emails` call inside a post-commit hook. Sometimes this function runs inside a transaction, so putting the email call inside a hook ensures that the enrollment will actually exist when the task processes. (If there's no transaction going on, it just runs immediately, [per the Django docs.](https://docs.djangoproject.com/en/3.2/topics/db/transactions/#module-django.db.transaction:~:text=If%20you%20call%20on_commit()%20while%20there%20isn%E2%80%99t%20an%20active%20transaction%2C%20the%20callback%20will%20be%20executed%20immediately.)) 

Automated tests should work.

Complete a purchase, ideally using PayPal as the funding source if you can. You should get the normal Success banner and there shouldn't be any log messages resulting from the task running. (However, since this uses PayPal and may require the callback URL to work, this may be hard to test locally.)


rachellougee (approved):
Looks good to me. Tested and I don't get any error after paying for it

PR #1096, opened 2022-10-05
Save users with no enrollment into file

annagav:
None

#### What's this PR do?
Update the management command `upgrade_eligible_users` to save users that are not currently enrolled in mitxonline, so that they can receive coupons.

The input file should look something like:
```
useremail@example.com,course-v1:MITxT+14.310x+3T2022
```
To test this PR you would need an existing user in your local instance but not enrolled in this course. Then you
run the command `./manage.py upgrade_eligible_users <file_name.csv>`.
It should create a file with your record that is missing the enrollment.

### DIFF
--- a/courses/management/commands/upgrade_eligible_users.py
+++ b/courses/management/commands/upgrade_eligible_users.py
@@ -9,12 +9,11 @@ learner email,courserun readable id
 ```
 """
 from django.core.management import BaseCommand
+from django.conf import settings
 import csv
 
 from courses.api import create_run_enrollments
 from courses.models import CourseRun, CourseRunEnrollment
-from ecommerce.models import Discount, DiscountProduct, UserDiscount
-from ecommerce.constants import REDEMPTION_TYPE_ONE_TIME, DISCOUNT_TYPE_PERCENT_OFF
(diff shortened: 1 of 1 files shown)


courses/management/commands/upgrade_eligible_users.py
```diff
@@ -62,6 +62,7 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
                     user=user, run=course_run
                 ).first()
                 if enrollment is None:
+                    users_no_enrollment.append((row[0], course_run.course_id))
```
jkachel, line 65: This should be the `courseware_id` for the course run - if it's numeric the `generate_legacy_enrollment_codes` command will look it up in the MicroMasters import CourseId table, but it'll look it up based on the MicroMasters ID so coupon generation later will fail. Using the readable ID for the run will work, though.

jkachel (changes requested):
Small change for IDs but otherwise looks OK.

annagav:
@jkachel I addressed your comment. Thanks for the review!

jkachel (approved):
LGTM 👍

PR #1108, opened 2022-10-07
Updates the generate_legacy_enrollment_codes command to match on username or email

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

none

#### What's this PR do?

Updates the `generate_legacy_enrollment_codes` to make it match learners based on username or email address. This should allow for creating enrollment codes based on a user's social auth record rather than relying on email address (which may be different). 

Generate an input file in the expected format: `<username or email>,<course readable id or MicroMasters import id>`
Then run `generate_legacy_enrollment_codes <input file> <output file>` (and optionally add `--expires` if you want to test expiration dates). The output should contain valid codes for users that are found within MITxOnline.


annagav (approved):
LGTM 👍

PR #1110, opened 2022-10-07
Updates `configure_for_dedp` command to make it more generic

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

none

#### What's this PR do?

Updates the `configure_for_dedp` command in the following ways:
- Renames it to `configure_tiers`
- Adds some options to specify the program's readable ID, title, and abbreviation (for use in the discount codes)
- Adds an option to specify a CSV file with the tier data in it

The defaults remain the same - without options, it'll set up tiers for DEDP - but specifying the options will allow you to generate tiers for other programs (and will create the program in the system if necessary). 

Additionally, it makes some minor styling changes so the text isn't red for everything. 

Automated tests should pass.

_To test defaults (i.e., DEDP):_ run `configure_tiers` with no options.

_To test changing program options:_ run `configure_tiers` with `--program <new readable ID>`, `--program-name "Program Name"`, and/or `--program-abbrev XXXX". The program should be created with the values specified. Omitting any of these will use the defaults:

| Option | Default |
|--------|---------|
| `--program` | `program-v1:MITx+DEDP` |
| `--program-name`  |  `Data, Economics and Development Policy` |
| `--program-abbrev`  | `DEDP` |

Note that the command will _update_ if it finds a matching readable ID for the program, so if you don't supply `--program` but supply either of the other two options, it will change your existing DEDP program record. 

_To test custom tiers:_ 

1. Create a CSV file with the following fields: `threshold,discount type,discount amount`
    Do not include a header. `discount type` should be one of `percent-off`, `dollars-off`, or `fixed-price`. A tier with threshold 0 must exist - the command will refuse to run if one isn't specified. Specifying tier data completely overrides the default tier settings, so you must provide all the tiers you wish to create.
2. Run the command with the `--tier-info <filename>` option. (You can combine this with program attributes if you want.) The command should create tiers as specified. 
 
As per usual, if the program/tier data exists, it should reuse existing discounts and tiers and all that. (It did this previously; it should continue to do that.)


flexiblepricing/management/commands/configure_tiers.py
```diff
@@ -0,0 +1,283 @@
...
+        # Step zero: if the user supplied a defaults file, then parse that
+
+        if "tier_info" in kwargs and kwargs["tier_info"] is not None:
+            discounts_and_tiers = []
+            found_zero_tier = False
+
+            self.stdout.write("Reading tier info from file...")
+
+            with kwargs["tier_info"] as tierfile:
+                csvreader = csv.reader(tierfile)
```
asadali145, line 174: For better readability of code, We could also use csv.DictReader
```suggestion
                csvreader = csv.DictReader(tierfile)
```

flexiblepricing/management/commands/configure_tiers.py
```diff
@@ -0,0 +1,283 @@
...
+                                "discount_code": f"{program_abbrev}-fa-tier{idx+1}-{current_year}",
+                                "discount_type": row[1],
+                                "amount": row[2],
+                            },
+                        }
+                    )
+
+                    if int(row[0]) == 0:
+                        found_zero_tier = True
+
```
asadali145, line 192: With DictReader we could have this as:
```suggestion
                    self.stdout.write(
                        f"Read tier: threshold {row['threshold_amount']}, discount type {row['discount_type']}, value {row['discount_amount']}"
                    )
                    discounts_and_tiers.append(
                        {
                            "tier": {"threshold": row['threshold_amount']},
                            "discount": {
                                "discount_code": f"{program_abbrev}-fa-tier{idx+1}-{current_year}",
                                "discount_type": row['discount_type'],
                                "amount": row['discount_amount'],
                            },
                        }
                    )

                    if int(row['threshold_amount']) == 0:
                        found_zero_tier = True


```

flexiblepricing/management/commands/configure_tiers.py
```diff
@@ -0,0 +1,283 @@
...
+The defaults can be changed:
+- Specify --program <courseware-id> to use a different program (this will be
+created if it's not there already)
+- Specify --program-name "<name>" to use a different program name (required if
+you specify --program)
+- Specify --program-abbrev "<abbreviation>" to use a different program
+abbreviation in the discount codes (defaults to DEDP, required if you specify
+--program) 
+- Specify --tier-info <filename> to set different tier levels. This expects a 
+CSV file with "threshold amount,discount type,discount amount" as columns.
```
asadali145, line 47: If we opt to use csv.DictReader, we can replace the space in the column names with an underscore as well.

asadali145 (approved):
LGTM! Works like a charm. A few minor suggestions that we may consider and can ignore as well.

jkachel:
Resolved comments in https://github.com/mitodl/mitxonline/pull/1110/commits/78bd6ea119d0c232d1aaf467b94b1b6bb7a86c57 - note that even though it now uses DictReader, there should not be a header row in the input file

asadali145 (approved):
LGTM!

Issue #705, opened 2022-07-15
Story: legacy DEDP learners' course certificates 


pdpinch:
As a returning DEDP learner who earned a course certificate on micromasters.mit.edu, I can link to my past course certificates on my dashboard.

Course certificates will be hosted on mitxonline going forward, so this has a dependency on https://github.com/mitodl/mitxonline/issues/691

Note that the source of truth for DEDP course certificates is the micromasters app. The source of truth for other MIcroMasters programs is edx.org. 

#### Designs and Mockups

same as #691 

#### required MM data 

- [ ] Micromasters course certificates

#### out of scope

- program certificates are a separate story

rhysyngsun:
Tagging this as blocked because we haven't sorted out how to associate certificates with certain signatures.

pdpinch:
Ok, so this is blocked by #887 then.

aliraza-abbasi:
#887 is merged and deployed,
cc: @pdpinch @rhysyngsun

rachellougee:
@pdpinch @rhysyngsun How should we determine which course run to generate the certificate for on MITxOnline?

In MM, the certificate is per course. In MITxOnline, it's per course run

For example, this student has a certificate for "[Data Analysis for Social Scientists"](https://micromasters.mit.edu/admin/grades/micromasterscoursecertificate/1723/change/?_changelist_filters=q%3Dasiroky) in MM
He has passed grades for two course runs on MITxOnline

- [course-v1:MITx+14.310x+2T2020 | Data Analysis for Social Scientists ](https://mitxonline.mit.edu/admin/courses/courserungrade/26458/change/)
- [course-v1:MITx+14.310x+2T2018 | Data Analysis for Social Scientists](https://mitxonline.mit.edu/admin/courses/courserungrade/25518/change/)

When we migrate his certificate, should we generate two certificates (course-v1:MITx+14.310x+2T2020 and course-v1:MITx+14.310x+2T2018), Or should we pick one?

pdpinch:
Oh, I forgot about that. I think Anna reminds me of it frequently. 

We should pick the one with the highest grade, if it's not too much trouble.

rachellougee:
Keep it open until we run it on production

rachellougee:
@pdpinch Please let me know when we should run the migration to bring over the MM certificates.

pdpinch:
This looks good to me. I opened a small issue about spacing in the UI #1206 but it's already been fixed.

[...]

PR #1111, opened 2022-10-07
add course certificate migration

rachellougee:
#### Pre-Flight checklist

- [ ] Migrations
- [ ] Testing
  - [ ] Code is tested
  
https://github.com/mitodl/mitxonline/issues/705

#### What's this PR do?
It's to migrate the course certificate from MM for DEDP courses. 
There are a couple of other changes:
 - Modify the `import_micromasters_data` command to accept argument `--num` so it runs a specific file
 - Creates a new mapping table for `CourseRun` to  'current' `PageRevision` (It will be populated outside PR)

- Run migration to generate mapping table

- Go to /admin/micromasters_import/courseruncertificaterevisionid/ to add some records for CourseRun -> PageRevision mapping. (This requires setup Certificates in CMS)

- Run migration script 
   (It requires a foreign data wrapper created locally see #859)
```./manage.py import_micromasters_data --num 009```

(Optional)

(Optional)

### CODE REVIEW

**Review by collinpreston — APPROVED** [review]
Looks good to me.  Thank you for all the help today!

**Comment by rachellougee** [comment]
@rhysyngsun @pdpinch Do either of you want to take a look before I merge? Thanks

File: micromasters_import/management/commands/import_micromasters_data.py [hunk 1]
```diff
@@ -24,18 +24,33 @@ class Sql:
...
         self.stdout.write(f"Gathering SQL queries:")
 
-        for root, _, files in os.walk(SQL_FILES_DIR):
-            for file_name in sorted(files):
-                with open(os.path.join(root, file_name), "r") as sql_file:
-                    self.stdout.write(file_name)
-                    sqls.append(self.Sql(file_name, sql_file.read()))
+        if file_num:
+            for file in os.listdir(SQL_FILES_DIR):
+                if file_num in file:
```
rhysyngsun, line 44: This is probably open to user error issues, for example if I passed `--num 2`, there's no validation of the length of the input being a zero padded value, so this means any script with '2' in the filename `e.g. 020, 152, etc` would get run. I think you want to either validate the input length or zero-pad it in the command to support either style. Then I think you probably want `file_name.startswith` instead of an `in` check.

Also I think you could simplify this into a single file walk but moving the `if file_num` into the `os.walk` below and making it a:

```python
if file_num and not file_name.startswith(file_num):
    continue
```

rhysyngsun (changes requested):
The query is a little complicated to follow, would you be able to put some code comments in there?

micromasters_import/management/commands/import_micromasters_data.py
rachellougee: I didn't think too much of validation since it will be used once. But I see your point here. I will make the change

rachellougee:
@rhysyngsun I addressed your feedback and add some comments for queries. Please take a look

rhysyngsun (approved):
LGTM

PR #1117, opened 2022-10-10
Adds additional error reporting; accepts transactions with status code 100

### LINKED ISSUE #1116: "Unknown error" occurs when completing a transaction
When purchasing a course _on production_, the system records an unknown error. The transaction succeeds, but the user is redirected back to the cart with an error message. It appears the transaction result is falling out the bottom of the `process_cybersource_payment_response` call. In looking at the transaction details in CyberSource, it seems like it's sending back a response status of `TRANSMITTED` - the code doesn't know what to do with that, so it displays the generic error message. It should handle this status and, additionally, it should log something when it hits this status so we can have some logs when/if this happens again. The `TRANSMITTED` state is code 100, which indicates that the transaction was successful; we should process this state like an accepted transaction. 

This may require a PR in the ol-django repo to add a constant for the `TRANSMITTED` status. 

I'm not 100% sure about this but CyberSource's ability to process webhooks may have some bearing on why this hasn't popped up on RC or in local instances. 

### Steps to Reproduce

1. Add a course to the cart and attempt to check out.
2. Complete the purchase process. (Payment method does not matter here - it fails with either PayPal or credit card transactions.)

### Expected Behavior

You are returned to the dashboard with a success message indicating the purchase was successful.

### Actual Behavior

You are returned to the cart with an error message. However, you have been enrolled in the course successfully - it will appear in your dashboard and you'll have a `verified` enrollment. 

### Related Issues

#1114 is probably related to this.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #1116

#### What's this PR do?

Adds a status code check alongside the reason code check when processing the CyberSource payload post-payment. If the reason is `ACCEPT` _or_ the status code is `100`, the `process_cybersource_payment_response` call should fulfill the order. 

This also updates the status code check to log an error message if it falls through the bottom, so we can see what it's doing. (The "unknown error" only occurs if the order isn't in one of the states that we expect, and it was only logging to info so errors weren't being captured on production.) 

Submit an order. It should work as expected - you should be redirected to the dashboard and you should see the successfully enrolled message. You should only receive one set of enrollment and receipt emails.

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -341,7 +341,10 @@ def process_cybersource_payment_response(request, order):
         if basket and basket.compare_to_order(order):
             basket.delete()
 
-    elif processor_response.state == ProcessorResponse.STATE_ACCEPTED:
+    elif (
+        processor_response.state == ProcessorResponse.STATE_ACCEPTED
+        or reason_code == 100
+    ):
         # It actually worked here
         basket = Basket.objects.filter(user=order.purchaser).first()
         try:
@@ -359,6 +362,9 @@ def process_cybersource_payment_response(request, order):
 
         return_message = order.state
     else:
+        log.error(
+            f"Unknown state {processor_response.state} found: transaction ID {transaction_id}, reason code {reason_code}, response message {processor_response.message}"
+        )
         order.cancel()
         order.save()
         return_message = order.state
--- a/ecommerce/views/v0/__init__.py
+++ b/ecommerce/views/v0/__init__.py
@@ -455,7 +455,7 @@ class CheckoutCallbackView(View):
                     processor_response = PaymentGateway.get_formatted_response(
                         ECOMMERCE_DEFAULT_PAYMENT_GATEWAY, request
                     )
-                log.info(
+                log.error(
                     "Checkout callback unknown error for transaction_id %s, state %s, reason_code %s, message %s, and ProcessorResponse %s",
                     processor_response.transaction_id,
                     order.state,


arslanashraf7 (approved):
Looks good 👍.

I was able to test the scenario mentioned in the PR description.

You'd already have that planned but just as a reminder, Before merging this we'd need to push a new release of payment gateway on PyPI and then update the version here.

jkachel:
Merging this - after some thought I think the `TRANSMITTED` status was a red herring, so want to get this up to get the logging change up to see if we can identify the problem. (This change doesn't depend on the ol-django change anymore either.)

ecommerce/api.py
```diff
@@ -359,6 +362,9 @@ def process_cybersource_payment_response(request, order):
 
         return_message = order.state
     else:
+        log.error(
+            f"Unknown code {processor_response.state} found: transaction ID {transaction_id}, reason code {reason_code}, response messager {processor_response.message}"
```
arslanashraf7, line 366: Maybe we want to change it a bit from code to state? but I'll let you decide. Other than this there is a typo in `message`.
```suggestion
            f"Unknown state {processor_response.state} found: transaction ID {transaction_id}, reason code {reason_code}, response message {processor_response.message}"
```

arslanashraf7 (approved):
Just a minor suggestion.

arslanashraf7 (approved):
👍 feel free to merge when the tests pass.

PR #1122, opened 2022-10-11
Updates the checkout redirect code to handle processed orders

### LINKED ISSUE #1116: "Unknown error" occurs when completing a transaction

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #1116 (moreso)

#### What's this PR do?

If the learner's order is already processed (as it might be if the callback is hit before the learner completes loading the redirect back to MITxOnline from CyberSource), the `CheckoutCallbackView` should now redirect the user to the dashboard with the proper result message, rather than getting confused (since the order's no longer Pending in that case). 

Automated tests should pass. Callbacks really need to work for this to be tested manually. 

Attempt to purchase a course. You should see the correct message when you've completed your order.

### DIFF
--- a/ecommerce/views/v0/__init__.py
+++ b/ecommerce/views/v0/__init__.py
@@ -391,6 +391,72 @@ class CheckoutCallbackView(View):
 
         self.logger = logging.getLogger(__name__)
 
+    def post_checkout_redirect(self, order_state, order, request):
+        """
+        Redirect the user with a message depending on the provided state.
+
+        Args:
+            - order_state (str): the order state to consider
+            - order (Order): the order itself
+            - request (HttpRequest): the request
+
+        Returns: HttpResponse
+        """
+        if order_state == Order.STATE.CANCELED:
+            return redirect_with_user_message(
+                reverse("cart"), {"type": USER_MSG_TYPE_PAYMENT_CANCELLED}
+            )
+        elif order_state == Order.STATE.ERRORED:
+            return redirect_with_user_message(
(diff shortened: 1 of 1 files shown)


rhysyngsun (approved):
LGTM

rachellougee (approved):

Issue #1109, opened 2022-10-07
management command to create verified enrollment


pdpinch:
As an admin I need to create verified enrollments sometimes: 
- to address customer support issues
- to test the system
- to enroll staff 

These verified enrollments need to be accounted for properly, so they should be created with an enrollment code.

Sometimes I need to do this after the `enrollment_end` date has passed, or the `upgrade_deadline` has passed. 

#### Acceptance Criteria:

- [ ] create a verified enrollment in mitxonline and in open edX 
- [ ] if possible, use a `-f` switch to force the creation after the `enrollment_end` date and after the `upgrade_deadline` date

pdpinch:
This could be an addition to the existing create_enrollment management command

aliraza-abbasi:
@pdpinch the `-f` switch should also create enrollment on edx after `enrollment_end` date?
I'm currently looking into OpenEdx if it supports enrollment after the enrollment end date

pdpinch:
If it's easier to break this into two PRs I would be OK with that. The first one would create verified enrollments for courses that are open, and the second would be the `-f` switch for creating verified enrollments in courses that are closed. 

Regarding your approaches, I agree that #3 is best. Is it a REST API that could be used by mitxonline?

aliraza-abbasi:
Sure, I'll raise the first PR without expired courses enrollment and after that, I'll make the follow-up PR to support expired courses enrollment. But before that, I want to confirm a few things.

- The `enrollment_code` that we pass in the command should be assumed to be already created and `raise` the `CommandError` if not created or do we have to create the `enrollment_code` on run time if it is not created?
- Should I need to verify the `DiscountProduct` relation or it is ok to use the `enrollment_code` even if there is no relation between the `product` and `enrollement_code`?
- Do I need to verify that the given `enrollment_code` should be 100% off or it should be the same amount as the product price, or not?
- If the given `enrollement_code` is expired can we use it through the command or not?

In current context the `enrollment_code` can be used with any `product` without `DiscountProduct` relation if it is not `for_flexible_pricing` if it is `for_flexible_pricing` then the `enrollement_code` requires a relation with `product` in DiscountProduct model

aliraza-abbasi:
> Sure, I'll raise the first PR without expired courses enrollment and after that, I'll make the follow-up PR to support expired courses enrollment. But before that, I want to confirm one thing.
> 
[...]

PR #1129, opened 2022-10-12
feat: User verified course enrollment

aliraza-abbasi:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [ ] Migrations
- [x] Testing
  - [x] Code is tested
- [ ] Settings

#1109 

#### What's this PR do?
This creates a verified enrollment for courses on edx and mitxonline

- Fetch and check out to this branch and make sure edx is configured
- Create a course on Edx and the same on mitxonline
- Add verified course mode in `course modes` on Edx
- Add a `discount code` on mitxonline and run the following command
- `./manage.py create_verified_enrollment --user user_email --run course_run --code enrollment_code`
- make sure to replace `user_email`, `course_run` and `enrollment_code` with appropriate values
- Validate the error messages
- After enrollment validate entries in `Order`, `FulfilledOrder`, `Transaction`, `PaidCourseRun`, `DiscountRedemption` and `CourseRunEnrollment` tables in Django admin

### DIFF
--- a/courses/management/commands/create_verified_enrollment.py
+++ b/courses/management/commands/create_verified_enrollment.py
@@ -31,10 +31,7 @@ class Command(BaseCommand):
             required=True,
         )
         parser.add_argument(
-            "--code",
-            type=str,
-            help="The enrollment code for the course",
-            required=True
+            "--code", type=str, help="The enrollment code for the course", required=True
         )
         parser.add_argument(
             "-k",
@@ -62,9 +59,7 @@ class Command(BaseCommand):
                 "Could not find course run with courseware_id={}".format(options["run"])
             )
 
-        product = Product.objects.filter(
-            object_id=run.id
-        ).first()
+        product = Product.objects.filter(object_id=run.id).first()
         if product is None:
             raise CommandError(
                 "No product found for that course with courseware_id={}".format(
@@ -79,7 +74,9 @@ class Command(BaseCommand):
             )
         if not discount.check_validity(user):
             raise CommandError(
-                "That enrollment code {} is not valid for user {}".format(options["code"], options["user"])
+                "That enrollment code {} is not valid for user {}".format(
+                    options["code"], options["user"]
+                )
             )
 
         with transaction.atomic():
--- a/courses/management/commands/create_verified_enrollment.py
+++ b/courses/management/commands/create_verified_enrollment.py
@@ -20,7 +20,7 @@ from django.db import transaction
 
 from courses.api import create_run_enrollments
 from courses.models import CourseRun, PaidCourseRun
-from ecommerce.api import fulfill_completed_order
+from ecommerce.api import fulfill_completed_order, zero_payment_data
 from ecommerce.discounts import DiscountType
 from ecommerce.models import Order, PendingOrder, Product, Discount
 from openedx.constants import EDX_ENROLLMENT_VERIFIED_MODE
@@ -78,25 +78,10 @@ class Command(BaseCommand):
                     options["run"]
                 )
             )
-
-        if run.course.is_country_blocked(user):
+        if PaidCourseRun.fulfilled_paid_course_run_exists(user, run):
             raise CommandError(
-                "Enrollment is blocked of this course with courseware_id={} for user {}".format(
-                    options["run"], options["user"]
-                )
-            )
-
-        # PaidCourseRun should only contain fulfilled or review orders
-        # but in order to avoid false positive passing in order__state__in here
-        paid_course_run = PaidCourseRun.objects.filter(
-            user=user,
-            course_run=run,
-            order__state__in=[Order.STATE.FULFILLED, Order.STATE.REVIEW],
-        ).first()
-        if paid_course_run:
-            raise CommandError(
-                "User {} already enrolled in this course with courseware_id={}\nEnrollment order:{}".format(
-                    options["user"], options["run"], paid_course_run.order
+                "User {} already enrolled in this course with courseware_id={}".format(
+                    options["user"], options["run"]
                 )
             )
 
@@ -115,9 +100,7 @@ class Command(BaseCommand):
                 "That enrollment code {} does not exist".format(options["code"])
             )
 
-        if discount.products.exists() and not (
-            discount.products.filter(product=product).exists()
-        ):
+        if not discount.check_validity_with_products([product]):
             raise CommandError(
                 "That enrollment code {} is invalid for course with courseware_id={}".format(
                     options["code"], options["run"]
@@ -134,7 +117,14 @@ class Command(BaseCommand):
         discounted_price = DiscountType.get_discounted_price([discount], product)
 
         if discounted_price > 0:
-            raise CommandError("Discount code is not 100% off")
+            raise CommandError("Enrollment code is not 100% off")
+
+        if run.course.is_country_blocked(user):
+            raise CommandError(
+                "Enrollment is blocked of this course with courseware_id={} for user {}".format(
+                    options["run"], options["user"]
+                )
+            )
 
         with transaction.atomic():
             successful_enrollments, edx_request_success = create_run_enrollments(
@@ -146,11 +136,8 @@ class Command(BaseCommand):
             if not successful_enrollments:
                 raise CommandError("Failed to create the enrollment record")
             order = PendingOrder.create_from_product(product, user, discount)
-            fulfill_completed_order(
-                order,
-                {"amount": 0, "data": {"reason": "No payment required"}},
-                already_enrolled=True,
-            )
+            payment_data = zero_payment_data()
+            fulfill_completed_order(order, payment_data, already_enrolled=True)
 
         self.stdout.write(
             self.style.SUCCESS(
--- a/courses/management/commands/create_verified_enrollment.py
+++ b/courses/management/commands/create_verified_enrollment.py
@@ -4,8 +4,10 @@ from django.core.management.base import BaseCommand, CommandError
 from django.db import transaction
 
 from courses.api import create_run_enrollments
-from courses.models import CourseRun
-from ecommerce.models import DiscountRedemption, Product, Discount, DiscountProduct
+from courses.models import CourseRun, PaidCourseRun
+from ecommerce.api import fulfill_completed_order
+from ecommerce.discounts import DiscountType
+from ecommerce.models import Order, PendingOrder, Product, Discount, DiscountProduct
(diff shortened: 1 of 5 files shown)


asadali145 (commented):
Added a few suggestions but overall LGTM. Just did the code review. Going to test it now.

courses/management/commands/create_verified_enrollment.py
```diff
@@ -0,0 +1,157 @@
...
+        if product is None:
+            raise CommandError(
+                "No product found for that course with courseware_id={}".format(
+                    options["run"]
+                )
+            )
+
+        purchased_object = product.purchasable_object
+
+        if purchased_object.course.is_country_blocked(user):
```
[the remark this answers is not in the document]
arslanashraf7: An example of that could be https://github.com/mitodl/mitxonline/blob/main/courses/management/commands/manage_certificates.py

courses/management/commands/create_verified_enrollment.py
```diff
@@ -0,0 +1,157 @@
...
+            dest="force",
+            help="If provided, Enroll user in expired courses",
+        )
+        super().add_arguments(parser)
+
+    def handle(self, *args, **options):
+        """Handle command execution"""
+        user = fetch_user(options["user"])
+
+        run = CourseRun.objects.filter(courseware_id=options["run"]).first()
```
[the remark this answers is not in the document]
aliraza-abbasi: This is like we use `--user` and we take `user_email` as an argument and then we get the user by email. The same goes here `--run` is an argument for course run and takes `courseware_id` which is the id for Course Run. There is nothing confusing here. It's like a convention that is being already used in mitxpro and in the previous command which is `create_enrollment` in mitxonline

courses/management/commands/create_verified_enrollment.py
```diff
@@ -0,0 +1,157 @@
...
+                    )
+                )
+
+        if (
+            not options["force"]
+            and isinstance(purchased_object, CourseRun)
+            and not purchased_object.is_upgradable
+        ):
+            raise CommandError(
+                "The course with courseware_id={} is not upgradeable or the upgrade deadline has been passed".format(
```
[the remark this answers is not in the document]
aliraza-abbasi: That'll be added in the follow-up PR when the support of enrollment in expired courses on edx will be added. In fact, I'll remove this `-f` check from this PR

asadali145:
LGTM while testing. Nice work!

courses/management/commands/create_verified_enrollment.py
```diff
@@ -0,0 +1,157 @@
...
+                    options["run"]
+                )
+            )
+
+        discount = Discount.objects.filter(
+            for_flexible_pricing=False, discount_code=options["code"]
+        ).first()
```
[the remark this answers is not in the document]
aliraza-abbasi: `discount.check_validity(user)` already performing these checks, No need to check them here in the query

courses/management/commands/create_verified_enrollment.py
```diff
@@ -0,0 +1,157 @@
...
+        if not discount:
+            raise CommandError(
+                "That enrollment code {} does not exist".format(options["code"])
+            )
+
+        if discount.products.exists() and not (
+            DiscountProduct.objects.filter(product=product, discount=discount).exists()
+        ):
+            raise CommandError(
+                "That enrollment code {} is not valid for course with courseware_id={}".format(
```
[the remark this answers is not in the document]
aliraza-abbasi: I think the above message is more convenient as I've used a similar message that is already being used on mitxpro

courses/management/commands/create_verified_enrollment.py
```diff
@@ -0,0 +1,157 @@
...
+            if PaidCourseRun.objects.filter(
+                user=user,
+                course_run=purchased_object,
+                order__state__in=[Order.STATE.FULFILLED, Order.STATE.REVIEW],
+            ).exists():
+                raise CommandError(
+                    "User {} already enrolled in this course with courseware_id={}".format(
+                        options["user"], options["run"]
+                    )
+                )
```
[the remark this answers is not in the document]
aliraza-abbasi: I think we don't need the `Order` or `PaidCourseRun` detail here. but @pdpinch or @arslanashraf7 can verify this

asadali145 (approved):
LGTM!

courses/management/commands/create_verified_enrollment.py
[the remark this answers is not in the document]
arslanashraf7: I don't think `PaidCourseRun` might be needed but we can log the `order` in case we want to look at the order/transaction details of that enrollment.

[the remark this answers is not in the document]
aliraza-abbasi: Added order info

courses/management/commands/create_verified_enrollment.py
```diff
@@ -0,0 +1,161 @@
...
+            )
+
+        # PaidCourseRun should only contain fulfilled or review orders
+        # but in order to avoid false positive passing in order__state__in here
+        paid_course_run = PaidCourseRun.objects.filter(
+            user=user,
+            course_run=run,
+            order__state__in=[Order.STATE.FULFILLED, Order.STATE.REVIEW],
+        ).first()
```
arslanashraf7, line 95: This is a duplicate code and we should avoid it, and it is also very less likely that this would behave differently in different contexts. What I think is we should convert this code into a function and use it here and `ecommerce/models.py` so that any future changes in this logic apply everywhere.

arslanashraf7: My second suggestion here is a bit related to the sequence of the checks we have in place. We are checking for product and blocked country above however if a user already has an enrollment those checks would be running as extra.

How about we take this check above along with the passed discount that we have below? so that we only check for country blocked or product of the user doesn't have enrollment already and the discount code is also valid?

courses/management/commands/create_verified_enrollment.py
```diff
@@ -0,0 +1,161 @@
...
+            raise CommandError(
+                "That enrollment code {} for course with courseware_id={} is invalid for user {}".format(
+                    options["code"], options["run"], options["user"]
+                )
+            )
+
+        discounted_price = DiscountType.get_discounted_price([discount], product)
+
+        if discounted_price > 0:
+            raise CommandError("Discount code is not 100% off")
```
arslanashraf7, line 137: This will keep it standardized with the naming we are using above. We're using the Enrollment code everywhere I think.

```suggestion
            raise CommandError("Enrollment code is not 100% off")
```

courses/management/commands/create_verified_enrollment.py
```diff
@@ -0,0 +1,161 @@
...
+        # but in order to avoid false positive passing in order__state__in here
+        paid_course_run = PaidCourseRun.objects.filter(
+            user=user,
+            course_run=run,
+            order__state__in=[Order.STATE.FULFILLED, Order.STATE.REVIEW],
+        ).first()
+        if paid_course_run:
+            raise CommandError(
+                "User {} already enrolled in this course with courseware_id={}\nEnrollment order:{}".format(
+                    options["user"], options["run"], paid_course_run.order
```
arslanashraf7, line 99: I'd like to mention that I suggested this in a discussion but looking at the `str` override for the Order model, I don't think the order info printed here (email, and type) would benefit since this we already print this info with the user. So we can remove it.

arslanashraf7 (changes requested):
Took a quick look and I have a few suggestions. Let me know if you have any queries.

aliraza-abbasi:
@arslanashraf7 please review it now

courses/management/commands/create_verified_enrollment.py
```diff
@@ -0,0 +1,148 @@
...
+from openedx.constants import EDX_ENROLLMENT_VERIFIED_MODE
+from users.api import fetch_user
+
+User = get_user_model()
+
+
+class Command(BaseCommand):
+    """creates an enrollment for a course run"""
+
+    help = "Creates an enrollment for a course run"
```
arslanashraf7, line 35:
```suggestion
    help = "Creates a verified enrollment for a course run"
```

ecommerce/api.py
```diff
@@ -234,14 +228,24 @@ def apply_user_discounts(request):
...
-    order.fulfill(payment_data)
+def fulfill_completed_order(order, payment_data, basket=None, already_enrolled=False):
+    order.fulfill(payment_data, already_enrolled=already_enrolled)
     order.save()
 
     if not order.is_review and (basket and basket.compare_to_order(order)):
         basket.delete()
 
 
+def zero_payment_data():
```
arslanashraf7, line 239: This could just be a constant. The function isn't needed

ecommerce/models.py
```diff
@@ -291,6 +288,22 @@ def check_validity(self, user: User):
...
         return self.valid_now()
 
+    def check_validity_with_products(self, products: list):
+        """
+        Enforces the redemption rules for a given discount.
+
+        Args:
+            - products (list): List of products.
+        Returns:
+            - boolean
```
arslanashraf7, line 298: The docstrings need to be standardized for all the functions that you added. An example of that can be seen [here](https://github.com/mitodl/mitxonline/blob/main/courses/api.py#L128). This will apply to other functions in this PR too.

courses/models.py
```diff
@@ -900,3 +900,27 @@ class Meta:
 
     def __str__(self):
         return f"Paid Course Run - {self.course_run.courseware_id} by {self.user.name}"
+
+    @classmethod
+    def fulfilled_paid_course_run_exists(cls, user: User, run: CourseRun):
```
arslanashraf7, line 905: generally, a class method should deal with only class variables and the outside ones, We could have moved it to api.py without having to resolve circular dependency too but this is fine for now. keep it as is.

arslanashraf7 (changes requested):
A few changes.

ecommerce/models.py
aliraza-abbasi: I just followed [this](https://github.com/mitodl/mitxonline/blob/421f2b9d3707cd72d135f0fa8ae375f231780c8b/ecommerce/models.py#L261) one docstring format. I'll change it according to the format you mentioned

arslanashraf7: It should be standardized everywhere. Like in this PR different functions had different formats.

arslanashraf7 (approved):
Thanks for making all the changes, just as a reminder, rebase before you merge this PR.

Issue #1139, opened 2022-10-14
Customer support discount enrollment codes for future course runs


pdpinch:
As a customer supporter, I'd like to have (small) pools of coupons for a course, so I can offer it to users who have paid but can't complete the course. 

This is most common with DEDP, for example this request for JPAL102x: [ZenDesk 137496](https://odl.zendesk.com/agent/tickets/137496)

#### Acceptance Criteria:

- [ ] a short list of enrollment codes that can be drawn down over time
- [ ] that expire in two years
- [ ] for simplicity, it would be good if they were good for any course run within the expiration period. If that's too hard, CS will just have to manage them more carefully

jkachel:
Created a management command to make these in #1142, then used it to generate 25 codes with `CS-` prefix for this purpose. These are all one-time use 100% off codes that expire on 2024-10-17. These got put into a Google Sheet in the shared drive for MITxOnline and sent explicitly to @pdpinch as well. I can rerun this at any time to add more codes as necessary (and the command offers a few other options too if we want to generate some for some other reason). 

For completeness, this was the command I ran:
```
heroku run -a mitxonline-production ./manage.py generate_discount_code --one-time --expires 2024-10-17 --amount 100 --discount-type percent-off --prefix CS- --count 25
```

jkachel:
Closing this - has been done for a bit now.

PR #1142, opened 2022-10-14
Adds a management command to create discount code(s) via command line

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#1139 

#### What's this PR do?

Adds a command to create discount/enrollment codes via the command line. Optionally, it can create batches of codes.

Run the management command `generate_discount_code --help` to see the options, then test with various permutations:

* The defaults are to create a `percent-off` code with unlimited redemptions for the specified amount with the code specified. (You can specify multiple codes on the command line to generate several codes.) 
* You can add an expiration date by specifying `--expires`, or make the code a one-time redemption code by using `--one-time`
* The discount type can be specified with `--discount-type`
* Use `--prefix` and `--count` to generate a batch of codes. In this mode, the code specified on the command line will be ignored, and it will instead use your specified prefix and a UUID as the code. 

The command will write its output to `generated-codes.csv`. This file will be overwritten on each run.


ecommerce/management/commands/generate_discount_code.py
```diff
@@ -0,0 +1,175 @@
...
+        parser.add_argument(
+            "--one-time",
+            help="Make the resulting code(s) one-time redemptions (otherwise, default to unlimited)",
+            nargs="?",
+            const=True,
+            default=False,
+        )
+
+        parser.add_argument(
+            "codes",
```
asadali145, line 77: IMO we should add a bit more detail about using codes. Maybe in the module docs or here about using codes.

asadali145 (changes requested):
Works fine, Mentioned a few things that I noticed in this.

asadali145:
Also FYI @jkachel, I am not sure about that 2 years thing mentioned in slack and in the ticket `that expire in two years` as well.

jkachel:
Addressed comments in https://github.com/mitodl/mitxonline/pull/1142/commits/3272508579256d3ba7d9c2dc3facd261ae3ea77c . The expiration date can be set with `--expires`, so to complete the ticket I would just specify the expiration date on the command line.

asadali145 (approved):
LGTM! Awesome work...

Issue #1148, opened 2022-10-17
Course enrollment upgrading is not ever synchronized with edx if the original update request fails


collinpreston:
### Steps to Reproduce

1. Enroll a user into the free version of a course run which has a matching product in `mitxonline`.
2. Verify that the enrollment is visible in edx (`admin/student/courseenrollment/`) with a `Mode` of `audit`
3. Before step 4, you will need to sever the tie between `mitxonline` and edx so that any change to the course run enrollment on `mitxonline` will initially fail to synchronize with edx.  When performed locally, I just did a `docker-compose stop` from the root directory of `devstack`.  I'm not completely confident on how you could do this on RC (maybe by removing the `verified` course_mode associated with the course being used in step 1).
4. Now, through `mitxonline`, upgrade the user and pay for the course run from step 1.
5. Perform step 2 again and verify that the `Mode` is still `audit`.
6. There is a task which runs periodically called `retry-failed-edx-enrollments`.  In production this runs every 30 minutes.  Either trigger this task to run through the Django admin or wait until it runs agains and verify step 2 once more.

### Expected Behavior

We would want the course enrollment record in edx to show the same value for `Mode` as `mitxonline`.  If the original attempt at synchronization fails, then we should retry the synchronization during the `retry-failed-edx-enrollments` task.

### Actual Behavior

The course enrollment record in edx will always show the value for `Mode` as `audit` until the user make another change to their course run enrollment on `mitxonline` which triggers another synchronization request to edx.

### Related Issues

https://sentry.io/organizations/mit-office-of-digital-learning/issues/3077938630/?project=5864687&query=is%3Aunresolved

### Extra information
It appears that the `retry-failed-edx-enrollments` task checks for any course_run_enrollments which match the following https://github.com/mitodl/mitxonline/blob/3eb7ade6cc3a311fce44164f414fa9e9abfa2c99/openedx/api.py#L549-L551.  

So it appears that after a user completes the checkout process to upgrade their enrollment, we send only a single request to edx to update the course enrollment record with the `Mode` value of `verified`.  If that fails, then maybe we should set `edx_enrolled` for the course_run_enrollment record in `mitxonline` to `False`.

PR #1151, opened 2022-10-18
1148: course-enrollment-upgrading-is-not-ever-synchronized-with-edx-if-the-original-update-request-fails

collinpreston:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested
 
https://github.com/mitodl/mitxonline/issues/1148

#### What's this PR do?
Sets the course_run_enrollment's `edx_enrolled` value equal to False when the edx API call, to modify an existing course_enrollment record's `mode` from `audit` to `verified`, fails.  This ensures that the edx API call is attempted again when the `retry-failed-edx-enrollments` task is performed periodically.  Ultimately this ensure that we synchronize the enrollment mode between `mitxonline` and edx even if the edx API calls fail.

Tester should have `devstack` and `mitxonline` running locally.

- Before starting `mitxonline`, I modified this section [here](https://github.com/mitodl/mitxonline/blob/7f4246df904f3557e28f3c662f2780bfd810b100/main/settings.py#L797-L801) so that the task to sync course modes runs more often.   
- Create a course, course_run, associated product, associated course CMS page all within `mitxonline`.  To make things easier, I used the default course that is created in `devstack` ([E2E Test Course](http://localhost:18000/courses/course-v1:edX+E2E-101+course).
- Enroll your user through `mitxonline` into the free version of the course.
- Ensure that both `mitxonline` and `devstack` show a course mode value for the user's enrollment with a value of `audit`.
- Perform a `docker-compose stop` on your `devstack`.
- Upgrade your user's enrollment through `mitxonline` by paying for the course.
- Perform a `make dev.up.large-and-slow on your `devstack` to start it again.
- Wait until the `retry_failed_edx_enrollments` task is run again.  You should be able to see this mention in the `celery` logs for `mitxonline`.
- Once the `retry_failed_edx_enrollments` task is run, verify the course mode for the enrollment on `devstack` has changed from `audit` to `verified`.

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -275,6 +275,7 @@ def create_run_enrollments(
                 # Case (Upgrade): When user was enrolled in free mode and now enrolls in paid mode (e.g. Verified)
                 # So, User has an active enrollment and the only changing thing is going to be enrollment mode
                 if enrollment.active and enrollment_mode_changed:
+                    enrollment.edx_enrolled = edx_request_success
                     enrollment.update_mode_and_save(mode=mode)
 
                 elif not enrollment.active:
--- a/courses/api.py
+++ b/courses/api.py
@@ -272,14 +272,13 @@ def create_run_enrollments(
 
             if not created:
                 enrollment_mode_changed = mode != enrollment.enrollment_mode
+                enrollment.edx_enrolled = edx_request_success
                 # Case (Upgrade): When user was enrolled in free mode and now enrolls in paid mode (e.g. Verified)
                 # So, User has an active enrollment and the only changing thing is going to be enrollment mode
                 if enrollment.active and enrollment_mode_changed:
-                    enrollment.edx_enrolled = edx_request_success
                     enrollment.update_mode_and_save(mode=mode)
 
                 elif not enrollment.active:
-                    enrollment.edx_enrolled = edx_request_success
                     if enrollment_mode_changed:
                         enrollment.enrollment_mode = mode
                     enrollment.reactivate_and_save()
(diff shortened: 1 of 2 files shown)


courses/api.py
```diff
@@ -275,6 +275,7 @@ def send_enrollment_emails():
                 # Case (Upgrade): When user was enrolled in free mode and now enrolls in paid mode (e.g. Verified)
                 # So, User has an active enrollment and the only changing thing is going to be enrollment mode
                 if enrollment.active and enrollment_mode_changed:
+                    enrollment.edx_enrolled = edx_request_success
```
rachellougee, line 278: This line could be moved to line 275  as if and elif blocks now have the same exact line

courses/api_test.py
```diff
@@ -30,6 +30,11 @@
     ENROLL_CHANGE_STATUS_DEFERRED,
     ENROLL_CHANGE_STATUS_REFUNDED,
 )
+from openedx.constants import (
+    EDX_DEFAULT_ENROLLMENT_MODE,
+    EDX_ENROLLMENT_AUDIT_MODE,
+    EDX_ENROLLMENT_VERIFIED_MODE,
```
rachellougee, line 36: There is already import on line 51

rachellougee (commented):
Just small comments. I will do a functional test

courses/api.py
collinpreston: Nice!  Good catch

rachellougee (approved):
LGTM

PR #1156, opened 2022-10-19
Changes staff dashboard to use Django sessions rather than OAuth2

### LINKED ISSUE #1089: Refine: switch to using built-in django sessions for auth
We currently have oauth setup through OIDC: https://github.com/mitodl/mitxonline/blob/main/frontend/staff-dashboard/src/index.tsx#L11

This requires that when you go to the staff dashboard you must click a button to login. This login state also gets reset if you do a reload of the full page (not an in-app refresh button).

Ideally when the user hits the refine app we do some checks that the user is logged in and that they have the appropriate permissions. I think this is just staff but whoever works on this should verify how the APIs are guarded.

Reference: https://refine.dev/docs/advanced-tutorials/access-control/

Note: although I think implementing some of the libraries referenced in the docs above would be a good thing across the whole app, for now let's just do whatever is simplest and minimal to get the job done.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #1089 

#### What's this PR do?

Updates the Staff Dashboard Refine application to use Django sessions rather than having its own OAuth2 integration:
- This eliminates the need to log into the Staff Dashboard separately from the main MITx Online application. (To that end, the login screen is removed, and the logout functionality logs you out of MITx Online entirely. We may want to revisit that.) 
- Some minor changes were made to the User serializer to add the necessary authorization flags to the session check endpoint (`/api/users/me`). 
- Some more minor changes were made to the User serializer to report back the Django permissions set for the current user. We don't use this as of yet but this provides the ability to use the standard Django permissions system to define access for the Staff Dashboard app (amongst other things).
- The OAuth2 code was removed from the app. (The libraries are still there, but the app doesn't use them.) 

With regard to access control: the app uses the standard session check call to determine the state of your `is_staff` and `is_superuser` flags. Superusers are unrestricted in the app. Staff can only view the Dashboard and the Flexible Pricing screen. The app only considers those two flags right now; it would be pretty trivial to update the `accessControlProvider` to make it look at the Django permissions that are now passed through the session check API. We should ideally do this with more of a plan, though, both with regard to an authorization system and to the scope of the Staff Dashboard application. 

The [original issue](https://github.com/mitodl/mitxonline/issues/1089) links to the Refine docs for authorization controls, which lists a few authorization control libraries that could be used for RBAC/ABAC. For the short term, I've made this work by using the flags noted above; one of these systems seemed to be a bit much for this one PR (and would be something I'd rather have some consensus via a discussion post rather than just going with, say, Casbin, especially since there can also be a server-side component to these). 

Use the staff dashboard as normal. You should no longer be required to log in separately, and the Log Out button should completely log you out of the app (not just from the Staff Dashboard). 

If you test with a staff user, you should additionally only have the Dashboard and Flexible Pricing options available to you. A superuser account should also see Discounts. 

Because we're now loading a profile for the currently logged in user, your username should be shown at the top right of the screen. (This is a Refine thing - we just didn't load a profile before now, so it didn't do it.)

### DIFF
--- a/frontend/staff-dashboard/src/App.tsx
+++ b/frontend/staff-dashboard/src/App.tsx
@@ -1,9 +1,8 @@
-import { Refine } from "@pankod/refine-core";
+import { Refine, useGetIdentity } from "@pankod/refine-core";
 import { Icons, notificationProvider } from "@pankod/refine-antd";
 import routerProvider from "@pankod/refine-react-router-v6";
 import "@pankod/refine-antd/dist/styles.min.css";
 import { useAuthProvider } from "hooks/useAuthProvider";
-import { useAuth } from "react-oidc-context";
 import {
   Title,
   Header,
@@ -35,6 +34,7 @@ axiosInterface.interceptors.request.use((config: any) => {
   return config;
 }, (error: any) => Promise.reject(error));
 
+const _ = require("lodash");
 
 const { RouterComponent : RefineRouterComponent } = routerProvider;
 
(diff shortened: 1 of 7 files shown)


collinpreston (approved):
Nice!  Looks robust to me

PR #1160, opened 2022-10-20
Adding simple Varnish config file and service block to the Compose environment

### LINKED ISSUE #205: add HTTP caching to local development stack
As a developer, I need to be able to reproduce the caching behavior of our production system locally. 

According to @blarghmatey we should be able to do this by adding a Varnish proxy to docker-compose, with the appropriate settings.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #205 

#### What's this PR do?

Adds a [Varnish cache](https://varnish-cache.org/) container that sits in front of the nginx container for development. The Varnish cache is to replicate the Fastly cache that exists on the RC and production instance. It only caches Django content (so we don't muck with any reloading stuff that happens with the JavaScript/React assets). 

This also increases the `start_period` and `interval` for the `watch` and `refine` containers - on my machine at least, either one of these can take over 2 minutes to start up cold so this should ensure everything actually starts up. 

After restarting (you will need to bring your stuff down and then re-`docker compose up` it), you should be able to load your local instance normally. Inspecting the network tab in devtools should show the Varnish headers in place (`Via: 1.1 varnish (Varnish/7.1)` for instance). The app should work normally otherwise.

### DIFF
--- a/config/default.vcl
+++ b/config/default.vcl
@@ -1,0 +1,22 @@
+# specify the VCL syntax version to use
+vcl 4.1;
+
+# import vmod_dynamic for better backend name resolution
+import dynamic;
+
+# we won't use any static backend, but Varnish still need a default one
+backend default none;
+
+# set up a dynamic director
+# for more info, see https://github.com/nigoroll/libvmod-dynamic/blob/master/src/vmod_dynamic.vcc
+sub vcl_init {
+        new d = dynamic.director(port = "8013");
+}
+
+sub vcl_recv {
+	# force the host header to match the backend (not all backends need it,
+	# but example.com does)
+	set req.http.host = "nginx";
+	# set the backend
+	set req.backend_hint = d.backend("mitxonline.odl.local");
+}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,8 +37,6 @@ services:
   nginx:
     build:
       context: ./nginx
-    ports:
-      - "8013:8013"
     links:
       - web
     volumes:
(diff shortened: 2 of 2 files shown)


collinpreston (approved):
Looks good and works for me.

config/default.vcl
```diff
@@ -0,0 +1,22 @@
...
+sub vcl_init {
+        new d = dynamic.director(port = "8013");
+}
+
+sub vcl_recv {
+	# force the host header to match the backend (not all backends need it,
+	# but example.com does)
+	set req.http.host = "nginx";
+	# set the backend
+	set req.backend_hint = d.backend("mitxonline.odl.local");
```
collinpreston, line 21: Is this line how you restrict the cache for only local development?

jkachel: It's kinda restricted by virtue of the fact that it's in the Docker mitxonline-default network - those lines set the target host to "nginx" (which is the name of the web server container and also the DNS name inside the Docker network) and additionally sets the host header correctly so it doesn't get confused.

config/default.vcl
```diff
@@ -0,0 +1,22 @@
...
+# set up a dynamic director
+# for more info, see https://github.com/nigoroll/libvmod-dynamic/blob/master/src/vmod_dynamic.vcc
+sub vcl_init {
+        new d = dynamic.director(port = "8013");
+}
+
+sub vcl_recv {
+	# force the host header to match the backend (not all backends need it,
+	# but example.com does)
+	set req.http.host = "nginx";
```
arslanashraf7, line 19: @jkachel I think this might have added a side effect on local environment, I was testing the payment while working on [another PR](https://github.com/mitodl/ol-django/pull/96) and the callback after the payment takes me to something like `nginx/callback/result` (can't remember exactly) and this URL wasn't resolvable for obvious reasons.

I'm not sure if I missed anything but I rebuilt containers `docker-compose build` and I could see `varnish` one too when it happend. Or maybe i'm missing something else.

jkachel: Noticed this too.. working on a fix for it.

PR #1163, opened 2022-10-21
Some changes to the Varnish config

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Re-fixes #1160

#### What's this PR do?

Updates the local Varnish config - Django uses the host setting to build URLs and that was being set wrong, causing failures elsewhere.

In my testing, the easiest way to check would be to create a new user account - you'll get an email from the system, and you should be able to verify that the hostname is correct.


arslanashraf7 (approved):
Looks good 👍 , Tested it too e.g. registration, login, checkout, etc.

Issue #1152, opened 2022-10-18
A management command for populating the db for development


annagav:
Write a management command that creates:

- [ ] test users
- [ ] programs, courses, course runs, enrollments
- [ ] products, discounts, 
- [ ] flexible price forms

PR #1168, opened 2022-10-25
Adds a management command to create and optionally enroll a user

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Part 1 of #1152 

#### What's this PR do?

This adds a management command that will create and optionally enroll a new user in the system. The enrollment is done using the existing `create_enrollment` command.

Run the management command: 
`./manage.py create_user <username> <email> <firstname> <lastname> <displayname> <countrycode> [--enroll <courseware>]`

This should create the user and, if you've supplied one or more `--enroll` flags, enroll them in the specified courseware object (either a course run or a program run).

### DIFF
--- a/users/management/commands/create_user.py
+++ b/users/management/commands/create_user.py
@@ -1,0 +1,120 @@
+"""
+Creates a new user in MITx Online, and their associated edX account.
+
+Note that this depends on the `create_enrollment` command - this will call that
+to generate the enrollments if specified (rather than reinventing the wheel).
+"""
+
+from argparse import RawTextHelpFormatter
+from getpass import getpass
+
+from django.core.management import BaseCommand
+from django.core.management import call_command
+from django.db.models import Q
+
+from mail.api import validate_email_addresses
+from users.models import User, LegalAddress, validate_iso_3166_1_code
+
+
+class Command(BaseCommand):
+    """
+    Creates a new user in MITx Online, and their associated edX account.
+    """
+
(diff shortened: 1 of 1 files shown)


rachellougee (changes requested):
The overall functionality works ok. There are a couple of small things for improvement if it's easy to implement
1. Got stack trace errors on 'duplicate key value violates unique constraint "users_user_username_key"', could we print out a readable message if a duplicate is found?
2. got "TypeError: object of type 'NoneType' has no len()" when running the command without --enroll argument
3. got stack trace errors on 'openedx.exceptions.NoEdxApiAuthError', this will happen whenever new user is created since this new user isn't auth in edx, could we create auth token if it's easy to do? if not, maybe just print out a readable message?

jkachel:
Addressed comments in https://github.com/mitodl/mitxonline/pull/1168/commits/965dc5caf72beea37cefbe9029bd6520bd601136 - 1 and 2 I resolved, but 3 I don't know that I can do a whole lot about. The underlying `create_run_enrollments` call logs that particular exception and that's what you're actually seeing there. The enrollment persists on the MITxOnline side (since I'm specifying `keep_failed_enrollments=True`) but not sure I want to change the actual enrollment code for this.

rachellougee (approved):
LGTM. didn't notice keep_failed_enrollments flag but agree no need to touch enrollment code

Issue #1167, opened 2022-10-24
PaidCourseRun for legacy orders


rhysyngsun:
We've brought over legacy micromasters order data but one thing we've missed is populating `PaidCourseRun` based on this data.

Since we'll need this for any new data, it probably makes sense to update our existing scripts.

rhysyngsun:
> Yes, we do have an existing migration script 0013_backfill_paid_courserun.py, we can just re-use it to backfill PaidCourseRun

We can't reuse that exact migration because ti's already been run, but you can definitely reuse the code elsewhere.

rachellougee:
Yes, that's what I meant by reuse, not exactly the same script

PR #1169, opened 2022-10-25
add a migration to backfill PaidCourseRun for the legacy orders

rachellougee:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/1167

#### What's this PR do?
This PR is to back-fill PaidCourseRun for any migrated MicroMasters orders.

Run `./manage.py import_micromasters_data --num 010`
Then check http://mitxonline.odl.local:8013/admin/courses/paidcourserun/, it should list all the paid courseruns for _imported_ MicroMaster orders where is in fulfilled

### DIFF
--- a/micromasters_import/management/commands/queries/010_backfill_paidcourserun.sql
+++ b/micromasters_import/management/commands/queries/010_backfill_paidcourserun.sql
@@ -1,0 +1,21 @@
+----This script is to back-fill courses_paidcourserun for the imported MicroMaster orders (course runs)
+
+INSERT INTO public.courses_paidcourserun (
+    created_on,
+    updated_on,
+    course_run_id,
+    order_id,
+    user_id
+)
+SELECT
+    mo_order.created_on,
+    mo_order.updated_on,
+    mo_line.purchased_object_id,
+    mo_order.id,
+    mo_order.purchaser_id
+FROM public.ecommerce_order AS mo_order
+JOIN public.ecommerce_line AS mo_line
+    ON mo_order.id = mo_line.order_id
+WHERE mo_order.reference_number LIKE 'MM%'
+    AND mo_order.state = 'fulfilled'
+ON CONFLICT DO NOTHING;


courses/migrations/0021_backfill_paidcourserun_for_legacy_orders.py
```diff
@@ -0,0 +1,30 @@
...
+def backfill_paidcourserun(apps, schema_editor):
+    """Backfill PaidCourseRun for mirgated/existing orders"""
+    PaidCourseRun = apps.get_model("courses", "PaidCourseRun")
+    CourseRun = apps.get_model("courses", "CourseRun")
+    Order = apps.get_model("ecommerce", "Order")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+
+    for order in Order.objects.filter(state__in=["fulfilled", "review"]):
+        content_type = ContentType.objects.get_for_model(CourseRun)
+        for run_obj in order.lines.filter(purchased_content_type=content_type):
```
asadali145, line 15: This is not really a course-run object here in the loop. Wouldn't it be better to rename it as `order_line`?

asadali145 (approved):
Overall LGTM! Just a small suggestion added. Also, could you please refresh this branch from the main and recreate this migration? There is already a migration starting with 0021 in the main?

rhysyngsun (changes requested):
I don't think we should be doing this in another migration, because we'll continue to bring in more records from micromasters over time and so we'll be back to square 1 with those. I'd suggest adding it to the set of scripts we're running in the import command, filtering to orders prefixed with the micromasters prefix.

courses/migrations/0021_backfill_paidcourserun_for_legacy_orders.py
rachellougee: You are right. The variable name is a bit confusing, I think `purchased_line` or `order_line` works better than `run_obj`

rachellougee:
@rhysyngsun Can you elaborate on what you meant by adding to the import command? I thought we won't be running order import from MicroMaster anymore. Are there any more orders we need to bring outside legacy order import? if so, I can change it to be re-useable

rhysyngsun:
@rachellougee we'll probably still need to run it occasionally for learners who haven't created mitxonline accounts yet

rachellougee:
@rhysyngsun I've added it to import_micromasters_data command so that it can be re-used when next time we need to run order migration again. Please take a look. thanks

rachellougee:
@rhysyngsun Do you think you can review this today before your vacation?

collinpreston (approved):
Looks good.  Tested this with some dummy data locally.

PR #1172, opened 2022-10-25
Adding management command to create a really basic courseware about page

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Part 2 of #1152 

#### What's this PR do?

Adds a command in to make a really simple about page for the specified courseware object. The object must exist, as well as the index page that it should reside under. The Title and Description will be pulled from the object and the page will be created as a draft. The user will have to go in and fill out the rest of the info on the page to make it live (there's only so much content data in the a course or program record). 

Automated tests should pass.

1. Create a new course and/or program. (This doesn't have to exist in edX. Don't create a CMS page manually for it.)
2. Run `create_courseware_page <coursewareid>`. It should create the courseware page successfully.
3. Run the same command again. It should fail as there should already be a courseware page set up.
4. Check the CMS - the new page should exist, but should be a Draft.

### DIFF
--- a/cms/api.py
+++ b/cms/api.py
@@ -1,10 +1,11 @@
 """API functionality for the CMS app"""
 import logging
-from typing import Tuple
+from typing import Tuple, Union
 from urllib.parse import urljoin, urlencode
 
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ValidationError
 
 from django.utils.text import slugify
 from wagtail.core.blocks import StreamValue
@@ -14,6 +15,7 @@ from cms import models as cms_models
 from cms.exceptions import WagtailSpecificPageError
 from cms.models import Page
 from cms.constants import CERTIFICATE_INDEX_SLUG
+from courses.models import Course, Program
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


annagav:
I think this PR looks good. But I think it automates a very small chunk of the setup. Thinking in terms of usefulness to someone who doesn't have data set up some reason, or for a person who is just getting mitxonline up and running having a command that would  be called `populate_test_data`, that would set up the basics for viewing a user's dashboard. Since there are these many little things that need to be created: in order to see something a dashboard you need to have a Program, a course, a course run, a user, enrollment of the user in course, etc.

In micromasters for example, we had a command that created two programs, one non-financial aid, the other with financial aid. Then each program had three courses created, each course had 3 course runs, with different starting dates. The same command also created many test users. Then you had another command that would allow to enroll users in courses, have them passed the course, failed, not paid and many variations from here.

So maybe we could combine this functionality that you added with also setting up all the Program, course, course run objects?

jkachel:
> So maybe we could combine this functionality that you added with also setting up all the Program, course, course run objects?

Yes, that's the plan. These last 3 PRs I've opened (this one, https://github.com/mitodl/mitxonline/pull/1168 , and https://github.com/mitodl/mitxonline/pull/1173 ) all do a small part and the next one will add in a command that does everything in one fell swoop using those commands and the ones that already existed for other parts of it.

annagav (commented):
Is there a way to also add the created cms page to home "featured" productson the home page?

jkachel:
https://github.com/mitodl/mitxonline/pull/1172/commits/81cc1d015f2e4b9f916fb47b683dfa4d85e1ca1a adds the page to the Featured Products section. (It also adds a flag to make the page live - by default, the page is a draft since it's mostly empty.)

annagav (approved):
LGTM 👍

PR #1173, opened 2022-10-26
Adds a management command to create courseware objects

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Part 3 of #1152 

#### What's this PR do?

Adds a management command to generate courseware objects. These can be programs, courses (optionally with a course run), or a standalone course run.

To create a program: `create_courseware program <program id> <program title> [--live]`
To create a course: `create_courseware course <course id> <course title> [--live] [--program <program id>] [--program-position <int>]`
.. with a course run, append: `--create-run <run tag> [--run-url <course URL>] [--self-paced]`
To create a course run: `create_courseware courserun <course id> <run title> --run-tag <tag> [--run-url <course URL>] [--self-paced]`

This command does nothing in edX/devstack - you will need to create the courses there. (However, it doesn't allow for specifying the dates because the assumption is that you'll do that there and they'll come over via the normal sync process.) 

Run through the options. The easiest way will be to do it in order - create a program, then create a course for the program, then create a course run in the course, then create a combined course and course run, etc. Each should be successful, and the system should error out if you specify an ID that already exists (or, in the case of course runs, doesn't).

### DIFF
--- a/courses/management/commands/create_courseware.py
+++ b/courses/management/commands/create_courseware.py
@@ -1,0 +1,180 @@
+"""
+Creates a courseware object. This can be a program or a course (and optionally
+a course run).
+"""
+from django.core.management import BaseCommand
+
+from courses.models import Program, Course, CourseRun
+
+
+class Command(BaseCommand):
+    """
+    Creates a courseware object.
+    """
+
+    help = "Creates a courseware object (a program or course, with or without a courserun)."
+
+    def create_course_run(self, course, **kwargs):
+        run_id = f"{course.readable_id}+{kwargs['create_run']}"
+
+        if CourseRun.objects.filter(
+            run_tag=kwargs["create_run"], course=course
+        ).exists():
+            self.stderr.write(
+                self.style.ERROR(
+                    f"Course run for {course} with ID {kwargs['create_run']} already exists."
+                )
+            )
+            exit(-1)
+
+        course_run = CourseRun.objects.create(
+            course=course,
+            title=kwargs["title"],
+            run_tag=kwargs["create_run"],
+            courseware_id=run_id,
+            courseware_url_path=kwargs["run_url"],
+            live=kwargs["live"],
+            is_self_paced=kwargs["self_paced"],
+        )
+
+        self.stdout.write(
+            self.style.SUCCESS(f"Created course run {course_run.id}: {course_run}")
+        )
+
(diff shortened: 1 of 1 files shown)


courses/management/commands/create_courseware.py
```diff
@@ -0,0 +1,191 @@
...
+                )
+                exit(-1)
+
+            program = None
+
+            if "program" in kwargs and kwargs["program"] is not None:
+                try:
+                    program = Program.objects.filter(pk=kwargs["program"])
+                except:
+                    program = Program.objects.filter(readable_id=kwargs["program"])
```
collinpreston, line 138: Should this be `get` instead of filter?  I was able to create a Course and specified a program value that didn't exist.  The course was created with a value in "position in program" but no "program" value.

courses/management/commands/create_courseware.py
```diff
@@ -0,0 +1,191 @@
...
+
+            if "create_run" in kwargs and kwargs["create_run"] is not None:
+                run_id = f"{new_course.readable_id}+{kwargs['create_run']}"
+
+                if CourseRun.objects.filter(
+                    run_tag=kwargs["create_run"], course=new_course
+                ).exists():
+                    self.stderr.write(
+                        self.style.ERROR(
+                            f"Course run for {new_course} with ID {kwargs['create_run']} already exists."
```
collinpreston, line 164: Is it possible to hit this since we just created a the Course and so there would be no prior CourseRuns existing?

collinpreston (changes requested):
Just a few comments.  This was helpful to use since my local database just got wiped randomly again

courses/management/commands/create_courseware.py
jkachel: no, the indention on this got munged up.. the `first` call was supposed to be outside the if. If you specify a program that doesn't exist, it should revert to None, though.

courses/management/commands/create_courseware.py
jkachel: Seems pretty unlikely indeed. I culled that in https://github.com/mitodl/mitxonline/pull/1173/commits/d5452a3851bfec808fd96f644d353462984e4843 .

collinpreston (commented):
Looks good sir!

collinpreston (approved):

PR #1174, opened 2022-10-26
Fix fmt and fmt:check commands

rhysyngsun:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

N/A

#### What's this PR do?
This fixes the `fmt` and `fmt:check` commands which were no-ops when run via `yarn workspaces foreach run COMMAND`  because of the `$PWD` prefix, which ended up being the project root in that case.

This PR includes the fix to those commands and the resulting format fixes.

Tests should pass


jkachel (approved):
LGTM 👍

PR #1176, opened 2022-10-27
Adds wrapper command to bootstrap a fresh MITxOnline instance

### LINKED ISSUE #1152: A management command for populating the db for development

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #1152 (this is the last part of that issue)

#### What's this PR do?

Adds a management command that will bootstrap a fresh MITxOnline instance. This involves:
* Running `configure_wagtail`
* Creating a new DEDP program
* Creating two courses to match the default ones you get with devstack (one of which gets added to the DEDP program), with a run for each, and then syncing that up with devstack (if that's configured)
* Creating product pages for the program and courses
* Creating a financial assistance form for the program
* Creating a new learner account and enrolling them in one of the courses
* Creating products for the course runs

So, the only steps you would need to do after checking out MITxOnline for the first time would be:
* Running migrate
* Running createsuperuser
* Performing the setup to talk to devstack (this can't really be automated via a management command, as you need to make records in the edX database)

By default, the learner account will have these parameters:
| Param | Setting |
|--------|--------|
| username | testlearner |
| email | testlearner@mitxonline.odl.local |
| name | Test Learner |
| password | you'll be prompted for this | 
| enrollments | `course-v1:edX+DemoX+Demo_Course` |

You can optionally have it not create the enrollment by supplying `--dont-enroll`. 

The courses that come with devstack are:
| Course Title | Courseware ID | Run Tag | Price |
| --- | --- | --- | --- | 
| Demonstration Course | course-v1:edX+DemoX | Demo_Course | $999 |
| E2E Test Course | course-v1:edX+E2E-101 | course | $999 |

The Demonstration Course is added to the DEDP program `program-v1:MITx+DEDP` when it's created. 

This uses a number of commands internally to accomplish its functions. See the `configure_instance.py` code for details. If you need to perform any of these tasks individually on the command line, you can do that by running just that command. 

*In addition*, this adds a `create_product` command that creates a product for the specified courseware. This works on a course run, and will default to creating an active product with the specified price and with the description set to the courseware ID. The product can be optionally be made inactive, or you can manually specify the description.

(For testing, I used the `dumpdata` command to grab a snapshot of my actually useful database first before testing, as you really need to have a clear database to test this. You may want to use `dumpdata` to grab OAuth2 config separately and re-apply it to your database after running migrations so you don't have to totally re-do the edX integration.)

1. Clear your database - I usually start up `dbshell` and run `drop schema public cascade; create schema public;` 
2. Run `migrate`
3. (Optionally) run `createsuperuser`
4. Run `configure_instance`
5. When prompted, enter the password you want for the test learner account. 
6. Review output. Depending on what the state of your devstack integration was, you may have some errors; in my case, the service account worker was already set up OK so the course run syncing worked fine. If that's not set up, though, course run sync will fail. 
7. Check out the site - you should be able to log in as the learner, they should have one enrollment on their dashboard, and you should be able to see the mostly empty about pages for the demo courses.  

To test `create_product`:

1. Run `create_product`, specifying a course run ID and price. (A product can exist for the course run - the system allows for multiple products for a course run.) 
2. Re-run 1, but specify `--description` to set the product description and/or `--inactive` to make the product inactive by default.

#### Other thoughts

* Should this maybe go ahead and do `migrate` and/or `createsuperuser`? It could also generate the MITxOnline side of the OAuth2 config for devstack too (but that'd be all, it wouldn't be able to make changes on the edX side).

### DIFF
--- a/ecommerce/management/commands/create_product.py
+++ b/ecommerce/management/commands/create_product.py
@@ -1,0 +1,67 @@
+"""
+Creates a product for the given courseware ID. This only supports course runs
+for right now (since we don't really do program runs).
+"""
+from django.contrib.contenttypes.models import ContentType
+from django.core.management import BaseCommand
+
+from courses.models import CourseRun
+from ecommerce.models import Product
+
+
+class Command(BaseCommand):
+    """
+    Creates a product for the given courseware ID.
+    """
+
+    help = "Creates a product for the given courseware ID."
+
+    def add_arguments(self, parser) -> None:
+        parser.add_argument(
+            "courseware",
+            type=str,
+            help="The courseware object to make the product for.",
+        )
+
+        parser.add_argument("price", type=str, help="The product's price.")
+
+        parser.add_argument(
+            "--description",
+            "-d",
+            nargs="?",
+            type=str,
+            help="The product's description. Defaults to the courseware object's readable ID.",
+            metavar="description",
+        )
+
+        parser.add_argument(
+            "--inactive", action="store_false", help="Make the product inactive."
+        )
+
+    def handle(self, *args, **kwargs):
+        try:
+            courserun = CourseRun.objects.filter(
+                courseware_id=kwargs["courseware"]
+            ).get()
+        except Exception as e:
+            self.stderr.write(f"Could not retrieve {kwargs['courseware']}: {e}")
+            exit(-1)
+
+        content_type = ContentType.objects.filter(
+            app_label="courses", model="courserun"
+        ).get()
+
+        description = courserun.readable_id
+
+        if "description" in kwargs and kwargs["description"] is not None:
+            description = kwargs["description"]
+
+        product = Product.objects.create(
+            object_id=courserun.id,
+            content_type=content_type,
+            price=kwargs["price"],
+            description=description,
+            is_active=kwargs["inactive"],
+        )
+
+        self.stdout.write(f"Created product {product}.")
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Run through those steps **including the addition of `/etc/hosts` aliases and the
 
 ### Configure mitxonline and Open edX
 
-See [Configure Open edX](docs/source/configuration/open_edx.rst)
+See [MITx Online Quick Start](docs/source/configuration/quickstart.rst) and [Configure Open edX](docs/source/configuration/open_edx.rst)
 
 ### Configuring the CMS
 
--- a/docs/source/commands/configure_instance.rst
+++ b/docs/source/commands/configure_instance.rst
@@ -1,0 +1,19 @@
+``configure_instance``
+======================
+
+Configures a fresh MITx Online instance. For more information, see :doc:`MITx Online Quick Start<../configuration/quickstart>`.
+
+Syntax
+------
+
+``configure_instance <platform> [--dont-enroll|-D] [--dont-create-superuser|-S] [--edx-oauth-client <client id>] [--edx-oauth-secret <client secret>] [--gateway <ip>]``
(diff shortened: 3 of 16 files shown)


collinpreston:
@jkachel , I think it would be good to add some documentation on this command and some of the other individual commands you recently added.

jkachel:
> @jkachel , I think it would be good to add some documentation on this command and some of the other individual commands you recently added.

Added a bunch of docs to the repo in https://github.com/mitodl/mitxonline/pull/1176/commits/3a31d38e39f226e048d474a754d4c61d7c1d32df . Each command also had some documentation in the code itself and help options available too. (That commit also adds a create_product function, which I'd forgotten about.)

jkachel:
https://github.com/mitodl/mitxonline/pull/1176/commits/97c767dac5970e2d0d0a84be7085517a4625c48d contains a handful of changes from suggestions from @collinpreston :
- The course pages should now be live - otherwise they get added to the homepage and you can't actually go to them. (You should still modify these after running the command - they'll be very bare.) 
- I've added a call to `createsuperuser` and some code to create the OAuth2 app record. You can turn both of these off, and the OAuth2 record can be created with an existing client ID and secret (useful if you've already set up devstack - just feed it the ones it already has). 
- Documentation was updated to reflect the above changes and I also standardized the product name to "MITx Online".

ecommerce/management/commands/create_product.py
```diff
@@ -0,0 +1,67 @@
+"""
+Creates a product for the given courseware ID. This only supports course runs
+for right now (since we don't really do program runs).
+"""
+from django.core.management import BaseCommand
+
+from ecommerce.models import Product
+from courses.models import CourseRun
+from django.contrib.contenttypes.models import ContentType
```
asadali145, line 9: Could we do isort?

ecommerce/management/commands/create_product.py
```diff
@@ -0,0 +1,67 @@
...
+            courserun = CourseRun.objects.filter(
+                courseware_id=kwargs["courseware"]
+            ).get()
+        except Exception as e:
+            self.stderr.write(f"Could not retrieve {kwargs['courseware']}: {e}")
+            exit(-1)
+
+        content_type = ContentType.objects.filter(
+            app_label="courses", model="courserun"
+        ).get()
```
asadali145, line 52: We could also do `.get(app_label="courses", model="courserun")` for simplicity.

asadali145 (commented):
Looks Nice! Added a few questions/suggestions.

jkachel:
Code changes are up in https://github.com/mitodl/mitxonline/pull/1176/commits/5b4d06fcb21431a7ae03223dcd000bb8c6f46ab7 . 

Did some testing with the rST files in PyCharm - it doesn't parse these correctly in the previewer. (For what it's worth, neither does VSCode and the GitHub previewer does better but also gets it wrong.) We use Sphinx to build the docs so there's a bunch of stuff in there that won't get picked up by the previewers, including the TOC blocks. If you build the docs, however, it's fine (or at least it should be - I tested the built version of the docs).

asadali145 (approved):
LGTM! Actually, this is some awesome stuff. I faced this issue multiple times when I accidentally removed all the containers and my data got wiped. You are the real savior 🙌🏻

jkachel:
Thanks! but to be sure, this was @annagav 's idea, I just implemented. :smiley:

PR #1198, opened 2022-11-08
Re-groups and tags enrollments in the program drawer

jkachel:
#### Pre-Flight checklist

- [X] Screenshots and design review for any changes that affect layout or styling
  - [X] Desktop screenshots
  - [X] Mobile width screenshots
- [X] Testing
  - [X] Code is tested
  - [X] Changes have been manually tested

#### What are the relevant tickets?

Fixes #1190 
Fixes #1193 (the code doesn't work in that way anymore so that issue is moot)

#### What's this PR do?

This PR adds grouping of courses in the program drawer according to the program's set requirements. It also resets some of the styling attributes for the enrolled item cards and program drawer to better match the InVision design here: https://projects.invisionapp.com/d/main?origin=v7#/console/21535972/468431707/preview?scrollOffset=0 

To summarize, this:
* Changes the program drawer to display a "Program overview" that consists of a total course count, not enrolled count, and passed course count
* Groups the available courses in the program by their status as an elective or a requirement in the course
* Updates the item cards in the program drawer to display an Enroll message when the user isn't enrolled in the course
* Adds tags (back) to the cards (on both the regular dashboard and in the program drawer):
   * "In progress" for courses whose start date is before now
   * "Enrolled in certificate track" for courses where the learner has a verified enrollment
   * "Course passed" for courses that the learner has a passing grade for 
* Updates the upsell message (either enroll or get certificate) to be formatted in a stacked manner _only if the card is displayed in the program drawer_

This PR also adds a serializer for grades, some helper functions to mark courses in a program as requirements or electives, and updates the `create_courseware` command to support that same tagging.  

#### How should this be manually tested?

Automated tests should pass.

Testing the `create_courseware` management command: See the documentation for details. 
* The `--required` and `--elective` flags are new and will flag the course appropriately within the program. 
* The command will now error if you swap the readable ID and title fields.
* It will also ignore `--required` and `--elective` if the course you're creating won't be live. (Having a not-live course in the requirements tree causes Django Admin to have issues.) 

Testing the program drawer: 
* You will need to either turn the feature flag for the program UI on or use the `?enable_programs` flag on the URI to see the program drawer.
* The tags should display accordingly for the status of the course.
* Opening the program drawer should present a UI that looks like the InVision layout in desktop mode.
* The program drawer should span the width of the screen on a mobile device (< 1024px horizontally). 
* The program drawer should list the total number of courses in the program, the number you've passed, and the number you've not enrolled in.
[...]

PR #1212, opened 2022-11-15
Re-groups enrollments in the program drawer and adds tags back to enrollments

jkachel:
This is a re-do of [the prior PR for this ](https://github.com/mitodl/mitxonline/pull/1198) - see that for details; forgoing the normal template because of that. (I'm reopening this with no code changes because the prior merge went away somehow.)


collinpreston (approved):
Approved.

PR #1218, opened 2022-11-16
Adds some additional options and docs for some management commadns

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

none

#### What's this PR do?

Adds a couple things to the management commands:
- `create_courseware` can accept dates 
- `create_discount_code` can now create one-time or one time per user discounts
- Added some documentation for `create_discount_code`

* Attempt to create a course run using `create_courseware` while specifying dates. It should work as long as you provide an actual date. You'll get an exception if you don't provide a valid date.
* Attempt to create a discount code with the one-time or once-per-user flag set. These should be created as expected. 
* The documentation for `create_discount_code` should not have errors.


collinpreston (commented):
Adding code review comments prior to testing

collinpreston (commented):
I think we should also update the documentation here: https://github.com/mitodl/mitxonline/blob/0988612476949dae9709aed103489778eada1ec6/docs/source/commands/create_courseware.rst#L15

collinpreston (approved):
Nice, works well!

Issue #1219, opened 2022-11-16
Certificates: program certificate creation should use `ProgramRequirement` tree


rhysyngsun:
Currently the logic for determining when a learner has earned a program certificate walks the `Course` -> `Program` relationship, but this is being replaced with the new `ProgramRequirement` data structure. This will enabled support for more complicated elective requirements also.

rhysyngsun:
@aliraza-abbasi yes, one of the big reasons for doing things this way was to support a course being in multiple programs.

PR #1239, opened 2022-11-30
fix: program certificate creation should use ProgramRequirement tree

aliraza-abbasi:
#### Pre-Flight checklist

  - [ ] Desktop screenshots
- [x] Migrations
  - [x] Migration is backward-compatible with the current production code
- [x] Testing
  - [x] Code is tested
- [ ] Settings

#1219 

#### What's this PR do?
(Required)

(Required)


courses/utils.py
```diff
@@ -54,7 +56,16 @@ def generate_program_certificate(user, program):
...
-    courses_in_program_ids = set(program.courses.values_list("id", flat=True))
+    courses_in_program_ids = set(
+        program.get_requirements_root()
+        .get_children()
+        .filter(operator=ProgramRequirement.Operator.ALL_OF)
+        .first()
+        .get_children()
+        .filter(node_type=ProgramRequirementNodeType.COURSE)
+        .values_list("course_id", flat=True)
+    )
```
rhysyngsun, line 67: So it won't be as straightforward as just changing this query. I think what your want to do is:

- Do this query, although you need to modify it a bit both to simplify it and also handle `MIN_NUMBER_OF` operators:
```suggestion
    courses_in_program_ids = set(
        program.get_requirements_root()
        .get_descendants()
        .filter(node_type=ProgramRequirementNodeType.COURSE)
        .values_list("course_id", flat=True)
    )
```
Then, you'll need to walk the tree of requirements checking that against the course certificates. I wrote some pseudocode for this [here](https://github.com/mitodl/mitxonline/pull/1070#issuecomment-1262632406) a while back, although it incorrectly uses grades.

rhysyngsun (changes requested):

rhysyngsun:
To fix the github actions test failures, I think you need a rebase now that https://github.com/mitodl/mitxonline/pull/1241 is merged.

courses/utils.py
aliraza-abbasi: Added `MIN_NUMBER_OF` check and updated query, please have a look

courses/utils.py
```diff
@@ -30,6 +33,31 @@ def is_grade_valid(override_grade: float):
     return 0.0 <= override_grade <= 1.0
 
 
+# has a learner passed all the requirements for a certificate?
+def has_earned_program_cert(user, program):
+    passed_courses = Course.objects.filter(
+        in_programs__program=program,
+        courseruns__courserungrade__user=user,
+        courseruns__courserungrade__passed=True,
```
rhysyngsun, line 41: This should be checking for certificates instead.

courses/utils.py
```diff
@@ -63,7 +126,9 @@ def generate_program_certificate(user, program):
         .count()
     )
 
-    if len(courses_in_program_ids) > num_courses_with_cert:
+    if (
+        len(courses_in_program_ids) - not_required_courses_count
+    ) > num_courses_with_cert or not has_earned_program_cert(user, program):
         return None, False
```
rhysyngsun, line 132: I'm a bit confused by this, but I don't think you need the course count stuff anymore, it should be purely based on `has_earned_program_cert()`:
```suggestion
    if not has_earned_program_cert(user, program):
        return None, False
```

courses/utils.py
```diff
@@ -30,6 +33,31 @@ def is_grade_valid(override_grade: float):
     return 0.0 <= override_grade <= 1.0
 
 
+# has a learner passed all the requirements for a certificate?
+def has_earned_program_cert(user, program):
```
rhysyngsun, line 37: Could you give this a proper docstring?

courses/utils.py
aliraza-abbasi: I added this because in `has_earned_program_cert()` there was a check for `courserun grade` So I added this check to verify if there are all required certificates are available. but as per your above [comment](https://github.com/mitodl/mitxonline/pull/1239#discussion_r1041439683) this not required anymore. So I removed it

courses/utils.py
aliraza-abbasi: fixed

courses/utils.py
```diff
@@ -30,6 +33,42 @@ def is_grade_valid(override_grade: float):
     return 0.0 <= override_grade <= 1.0
 
 
+def has_earned_program_cert(user, program):
```
rhysyngsun, line 36: Could you actually move this and the next two methods into `api.py` instead? They definitely fit in there more than in `utils.py`

rhysyngsun (approved):
The functionality looks good, one comment on where some of the code should live but feel free to merge once you've addressed that.

courses/utils.py
aliraza-abbasi: moved

PR #1248, opened 2022-12-05
Fixes the URL in the partner school email

### LINKED ISSUE #1247: Learner record partner school sharing email redirects to API endpoint
The email generated and sent to a selected partner school from a learner record should contain a link to the learner's record, but the link goes to the API endpoint rather than the frontend. 

### Steps to Reproduce

Learner records are disabled by default; you will need to enable both it and the program UI to test. Append `?enable_programs&enable_lr` to the URL to do this. Additionally, at least one partner school record will need to be added. (I added one on RC but it goes to me so you'll probably want to add a second one if you're testing on RC.)

1. Go to the learner record.
2. Enable sharing if it's not already enabled.
3. Send the record to a partner school.
4. Wait for the email. It should arrive with a link to the school record.

### Expected Behavior

The learner record should open in the app.

### Actual Behavior

You get a link to the API endpoint instead.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Fixes #1247 

#### What's this PR do?

Fixes the URL that gets put into the partner school sharing email so it goes to the right place.

Send a partner school sharing email (see #1247 for details). The link should go to the right URL, and not to DRF.

### DIFF
--- a/courses/mail_api.py
+++ b/courses/mail_api.py
@@ -82,11 +82,7 @@ def send_partner_school_sharing_message(learner_record):
                 learner_record.partner_school.email,
                 {
                     "learner_record": learner_record,
-                    "record_link": SITE_BASE_URL
-                    + reverse(
-                        "shared_learner_record_from_uuid",
-                        kwargs={"uuid": learner_record.share_uuid},
-                    ),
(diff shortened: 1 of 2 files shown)


annagav (approved):
👍

PR #1249, opened 2022-12-06
Adds "reference_number" to the searchable fields in order admins

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

n/a

#### What's this PR do?

Adds the reference number field to the searchable fields in `BaseOrderAdmin` and `FulfilledOrderAdmin` (which doesn't subclass `BaseOrderAdmin`). This should make it less annoying to try to find orders based on the reference number. 

In an instance that has orders:
1. Navigate to Django Admin
2. Test that you can search by the reference number


collinpreston (approved):
Looks good and works for me

Issue #1235, opened 2022-11-26
Create 2023 DEDP courses on RC for testing


pdpinch:
As a developer, I'd like a robust and representative testing environment on RC. 

In production, there are 7 DEDP courseruns scheduled for 2023. There are only two on RC. The necessary course runs should already existing on studio-qa.mitxonline.mit.edu but they don't have courserun records.

#### Acceptance Criteria:

- [ ] Follow the necessary steps (7-9) in https://github.com/mitodl/hq/discussions/306 to create the course runs
- [ ] update the [discussion](https://github.com/mitodl/hq/discussions/306) with suggestions on streamlining and automating the process

jkachel:
Looking at this but I think I need some clarification - there are only 3 "2023" courses in Studio on RC:
- 14.740x (twice; one's MITx and one's MITxT)
- 14.009x

Do the remaining courses need to be created in Studio on RC as well? They're definitely there in production so I can just export them from there and import them into RC - really just want to make sure I won't be intruding on any other process before I do it.

pdpinch:
There should be 5 for 1T2023. I created the missing ones by clicking the "rerun" in studio for the missing ones. Then I set the dates [based on this sheet](https://docs.google.com/spreadsheets/d/1FGJPXhuWUGs-YPj2HFEYaQK4Sf7lO2oYf0V0ycKYjfY/edit#gid=0).

course-v1:MITxT+14.310x+1T2023
course-v1:MITxT+14.73x+1T2023
course-v1:MITxT+14.740x+1T2023
course-v1:MITxT+JPAL102x+1T2023
course-v1:MITx+14.750x+1T2023

PR #1256, opened 2022-12-07
Adds courserun importing from edX (like sync_courserun, but moreso)

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

No ticket for this but was inspired by #1235 

#### What's this PR do?

Adds a management command, `import_courserun`, that will create course runs in MITx Online and import the relevant data from edX in one fell swoop. Optionally, it will create CMS pages for courses that don't exist. Also optionally, it will walk a program and attempt to find all the courses in edX that exist for a given run tag and set them all up at once (so if you create a chunk of courses for a new semester, you can just tell the command the program and the new semester run tag and it'll take care of the rest). 

(Added later:) It will also optionally create products for the new courseruns. 

1. Create some courses in your local devstack instance that don't have corresponding courseruns in MITx Online. Ideally, make some that are in a program and some that are outside that program so you can test individual run generation as well as whole program generation.
2. For the individual courserun, run `manage.py import_courseware --courserun <course run ID>`. The course run should be created and its information should match the course on edX.
3. For the program, run `manage.py import_courseware --program <program> --run-tag <run tag>`. It should attempt to find all the courses that are in the program on edX and, if it finds them, then attempt to create courseruns for them. Depending on how your devstack instance is set up, you may get a bunch of errors here due to courses not existing in devstack. 
4. Apply the `--live` flag and make sure the course runs are marked as live.
5. Attempt to sync a courserun for a course that doesn't have a CMS page, and specify `--create-cms-page`. The command should create a CMS page for you.
6. The command should fail gracefully if:
   7. The course run(s) already exist
   8. The course doesn't exist 
   9. The CMS page for the course already exists
   (i.e. you should be able to re-run this for a program and have it add new things but not touch things that were already extant)

In addition, run these commands specifying `--price <price>` and ensure a product is created for the courserun. It will only create products for courseruns that it's creating, so in my testing the easiest way to test this was to delete the courseruns it created in mitxonline and re-run the command.


pdpinch:
I could use this right away to help with https://odl.zendesk.com/agent/tickets/141442

pdpinch:
Does this address step 7 & 10 in https://github.com/mitodl/hq/discussions/306 ? Is it possible to include any others?

jkachel:
It can take care of the product creation.. that should be pretty simple to add so will do that today.

jkachel:
Product creation added in https://github.com/mitodl/mitxonline/pull/1256/commits/62a9685d9327818077af1ae88657f1bc8cbd6c73

courses/management/commands/import_courserun.py
```diff
@@ -0,0 +1,215 @@
...
+create_courseware_page command for the course run.
+"""
+
+from decimal import Decimal
+from urllib import parse
+
+from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
+from django.core.management import BaseCommand
+from django.db.models import Q
```
rachellougee, line 27: Is this import used?

courses/management/commands/import_courserun.py
```diff
@@ -0,0 +1,215 @@
...
+        elif kwargs["program"] is not None and kwargs["run_tag"] is not None:
+            if kwargs["program"].isnumeric():
+                program = Program.objects.filter(pk=kwargs["program"]).all()
+            else:
+                program = Program.objects.filter(readable_id=kwargs["program"]).all()
+
+            if len(program) > 1:
+                self.stdout.write(
+                    self.style.ERROR(
+                        f"Program ID {kwargs['program']} is ambiguous - {len(program)} results found."
```
rachellougee, line 117: How could we get multiple rows when filtering on program ID or readable_id since these are unique? Not sure if I am missing something

rachellougee (commented):
Left a couple of comments. I am going to do a functional test

courses/management/commands/import_courserun.py
jkachel: Both of these are related; I was using a Q object to pull the program info. This has been totally refactored in https://github.com/mitodl/mitxonline/pull/1256/commits/e4f82bfe2db26e18d1510a30e6b05f14e4e0055b .

jkachel:
Updates in https://github.com/mitodl/mitxonline/pull/1256/commits/e4f82bfe2db26e18d1510a30e6b05f14e4e0055b include fixing the bug with CMS pages (live should reflect whether or not you pass `--live`; it was always "draft" before) and updating the single courserun code to create the course if it also doesn't exist. 

For the course creation - you can now specify `--courserun` and `--program` together; in this case, the program is the program that the new course should be added to. Don't specify `--run-tag` here - it'll be ignored. The run tags is part of the courserun readable ID so it'll get the run tag from that.

rachellougee (approved):
LGTM. Tested courserun and program with different flag and worked for me

PR #1268, opened 2022-12-08
Flipping the default for `for_flexible_pricing` from True to False

### LINKED ISSUE #1265: Make "for flexible pricing" default to false for new discount codes
#### Discussed in https://github.com/mitodl/hq/discussions/358

<div type='discussions-op-text'>

<sup>Originally posted by **jkachel** November 28, 2022</sup>
There was an issue recently where a discount code was created for Black Friday/Cyber Monday in MITx Online, and the discount code ended up failing when the promotion started because the "For Flexible Pricing" flag was set on the code. This was easily fixed but there's some process improvement that could be done to prevent this scenario.

Two ideas:
* **Simply setting the flag to default to false:** if you're creating a code via Django Admin, you're more than likely making a code that is intended to be used by learners directly, so the flag should be set false. (Generally, financial assistance codes need to be paired with tiers, so you'd be better off creating them using the `configure_tiers` command, which does everything for you.) 
* Eliminating the flag in favor of calculating the discount's status: This is a bigger change but this flag could be replaced by checking for the code's association with a financial assistance tier. The point of having the flag is to prevent learners from being able to manually apply a financial assistance discount, so we can just check if the code is being used in a tier when a learner attempts to apply it to their order. 

Thoughts?</div>

---

Per discussion post, the discount code's flexible pricing flag should default to False. This may require some changes elsewhere (like in `create_tiers` management command) to make sure it's explicitly setting the flag to True where necessary. 

#### Acceptance Criteria:

- [ ] Creating a new discount code in Django Admin defaults to making it a general-use code (flexible pricing = False)
- [ ] Creating a new discount code via the command line also defaults to a general-use code

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

Closes #1265 

#### What's this PR do?

Makes the default for `for_flexible_pricing` in discounts False instead of True.

Try to create a discount code via django admin. It should default to not being for flexible pricing.


arslanashraf7 (approved):
Looks good 👍

PR #1281, opened 2022-12-13
Updates the program drawer to use the requirements tree

### LINKED ISSUE #1238: Program drawer should use titles from the program's requirements tree
The program drawer groups courses into Requirements and Electives based on a flattened representation of the requirements tree that is passed via the API call it uses. This should be changed so that the grouping is done by walking the requirements tree itself. This would allow us to set the titles for the program's groupings (as the tree nodes have title fields). We'd also get better support of nested operators (for example, when an elective is two of three things, and one of the things can be one of two courses). 

#### Acceptance Criteria

- [ ] The Program APIs should return back the requirements tree. There are already serializers for this.
- [ ] The section headings for Required and Electives in the program drawer should reflect those actual node names in the database. (Usually, they'll be Required and Elective Courses but they should be updatable.) 
- [ ] The course listings should be derived by walking the tree provided by the API. 

#### Nice to Haves

- [ ] We could display the operator info for each course group in the drawer. (E.g., adding a header like "All Of:" and then listing out courses, until we hit another operator node, etc.)

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

Closes #1238 

#### What's this PR do?

Updates the program drawer to use the requirements tree for titles and course listings rather than the flattened requirements in the API call. Updates the program serializer to include the requirements tree in its output.

Automated tests should pass.

The program drawer should work as it has. The titles should now be the titles from the nodes of the requirements tree, instead of "REQUIRED" and "OPTIONAL". 

(No actual UI changes here so just desktop screenshots.)

Before:
image

After:
Screen Shot 2022-12-13 at 3 52 12 PM

### DIFF
--- a/courses/serializers.py
+++ b/courses/serializers.py
@@ -283,6 +283,7 @@ class ProgramSerializer(serializers.ModelSerializer):
     enrollment_start = serializers.SerializerMethodField()
     topics = serializers.SerializerMethodField()
     requirements = serializers.SerializerMethodField()
+    req_tree = serializers.SerializerMethodField()
 
     def get_courses(self, instance):
         """Serializer for courses"""
@@ -370,6 +371,9 @@ class ProgramSerializer(serializers.ModelSerializer):
 
         return formatted_reqs
 
+    def get_req_tree(self, instance):
+        return ProgramRequirementTreeSerializer(instance.requirements_root).data
+
     class Meta:
         model = models.Program
         fields = [
@@ -382,6 +386,7 @@ class ProgramSerializer(serializers.ModelSerializer):
             "enrollment_start",
             "topics",
             "requirements",
+            "req_tree",
         ]
 
 
--- a/courses/serializers.py
+++ b/courses/serializers.py
@@ -372,7 +372,12 @@ class ProgramSerializer(serializers.ModelSerializer):
         return formatted_reqs
 
     def get_req_tree(self, instance):
-        return ProgramRequirementTreeSerializer(instance.requirements_root).data
+        req_root = instance.get_requirements_root()
+
+        if req_root is None:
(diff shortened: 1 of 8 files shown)


rachellougee (approved):
This looks good to me. 
Just a side note, the number next to "Program overview" are different from the total number of required and elective courses. That's still pulling from courses from program rather than from program requirement. Is that something we will update later?

jkachel:
https://github.com/mitodl/mitxonline/pull/1281/commits/4ae382defa8af3740afaa75bb6fd4e537d38c2a3 updates to have that overview section be generated off of the requirements tree.

rachellougee (approved):
program drawer looks good now.
Also noticed that the number on dashboard might need to be updated. But leave it to you if that should be a separate issue since this PR is for program drawer
image

jkachel:
https://github.com/mitodl/mitxonline/pull/1281/commits/a269342a55bb2ee2febc0929c05f3152e75377c3 fixes the count on the dashboard - this also ended up being a bit of a thing; if there was a program with no requirements tree at all in here, the serializer would end up using the requirements tree from the prior program in the list. (So, I had 3 programs - the first two had a requirements tree but the third didn't have one at all as it predated the requirements code and I hadn't done anything to bring it up to date. But, if I looked in the data in `/api/program_enrollments`, the 3rd one had a requirements tree, and it was the tree from course # 2.) The last commit(s) fixes that. Might not have been an actual issue in practice - we only have one actual program right now and new ones would definitely have a req tree - but needed fixing anyway.

rachellougee (approved):
Looks good to me. count are all matched up.

Issue #1257, opened 2022-12-07
Allow course team to specify date for certificates


pdpinch:
As an instructor, I need more time between when my course ends and when I'm ready for certificates to be created. I use this time to review and adjust grades, and to review any proctored exams that were flagged. 

#### Acceptance Criteria:

- [ ] Research how to turn on the "Certificates Display Behavior" and "Certificates Available Date" settings in Studio (it's not available on studio.mitxonline.mit.edu, but it is available on studio.edx.org 

**Syncing Steps**
- [ ] Sync the certificates available date from Studio to the mitxonline app
- [ ] Create certificates after the certificates available date, instead of after the course end date. 

The above criteria acceptance might be divided into a couple of PRs as mentioned below (Orderly):

1. A PR in MITxOnline to add the `certificate_available_date` field, and change all the certificate generation logic from `end_date` to `certificate_available_date`. (This enables us to add certificate available date manually until we have our PR merged in edX. Already in progress WIP-https://github.com/mitodl/mitxonline/pull/1286)
2. An upstream PR to edX to include `certificate_available_date` to course details API.
3. A PR in [edx-api-client](https://github.com/mitodl/edx-api-client) to enable fetching `certificate_available_date` in conjunction with another PR in MITxOnline that uses this edx-api-client and syncs the `certificate_available_date` from edX.

arslanashraf7:
Update:

So I was looking around this and this is what I've got.

>Research how to turn on the "Certificates Display Behavior" and "Certificates Available Date" settings in Studio (it's not available on studio.mitxonline.mit.edu, but it is available on studio.edx.org

To enable the Certificate's display behavior and certificate available dates we need to do the following:
1. Add a waffle switch in `admin/waffle/switch/1/change/` with value `http://localhost:18000/admin/waffle/switch/1/change/`  ([Ref](https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/certificates/config.py#L11) from edX).
2. Add a Django setting/feature in CMS's `FEATURES` dictionary `"ENABLE_V2_CERT_DISPLAY_SETTINGS": True`. ([Ref](https://github.com/openedx/edx-platform/blob/master/cms/envs/common.py#L472) from edX).

After the above steps, the certificate display behavior and date options should be visible in `studio`>`your course`>`Settings`>`Schedule & Details`.


> Sync the certificates available date from Studio to the mitxonline app

[...]

PR #1286, opened 2022-12-15
feat: Change certificate generation based on certificate available date

arslanashraf7:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Migrations
- [x] Testing
  - [x] Code is tested

Part of https://github.com/mitodl/mitxonline/issues/1257

#### What's this PR do? (As of now, this PR enables us to at least add the certificate end dates manually until we sync them with edX)
- Adds a new field `certificate_available_date`
- Bases the certificate generation functionality on `certificate_available_date` instead of `end_date`
- Moving on we need to sync this date with edX (See https://github.com/mitodl/mitxonline/issues/1257#issuecomment-1346463078)
 

You will need a few configurations and setting up things to test this PR end-to-end.

- Try generating the certificate with steps mentioned in https://github.com/mitodl/mitxonline/pull/722 with `certificate_available_date` instead of `end_date`.
- The behavior that was once based on the course run's `end_date` should now be transferred to `certificate_available_date`. The existing behavior is mentioned in detail at https://github.com/mitodl/mitxonline/pull/722

Configure your edX instance with MITxOnline Locally

We added the polling and certificate generation in https://github.com/mitodl/mitxonline/pull/722 where we based the conditions on course end_date

**NOTE**
Probably, there are going to be some other PRs as well that will enable us to sync the `certificate_available_date` with edX. But we are not yet ready to start on them, See the comments (https://github.com/mitodl/mitxonline/issues/1257#issuecomment-1354905830, https://github.com/mitodl/mitxonline/issues/1257#issuecomment-1355023307)

image


pdpinch:
How does this work with self-paced courses?

courses/migrations/0029_add_certificate_available_date.py
```diff
@@ -0,0 +1,23 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name="courserun",
+            name="certificate_available_date",
+            field=models.DateTimeField(
+                blank=True,
+                db_index=True,
+                help_text="The day certificates should be available to users. This value is synced automatically with edX studio.",
+                null=True,
```
pdpinch, line 20: If we set `certificate_available_date` to null, does that mean the certificates will never be generated?

Could we set it to the same value as the course end date, in the migration?

arslanashraf7: >If we set certificate_available_date to null, does that mean the certificates will never be generated?

Yes, just like run's `end_date` it would be like the `certificate_available_date` is open-ended and hasn't arrived yet.

>Could we set it to the same value as the course end date, in the migration?

Yes, we can. I can do that change tomorrow.

arslanashraf7:
> How does this work with self-paced courses?

With `self-paced` courses it would work the same as before, e.g. generating the certificates whenever a user has completed the course w/ passed grades and the enrollment is `paid/verified`. `certificate_available_date` won't matter in this case.

arslanashraf7:
This has been reviewed as part of https://github.com/mitodl/mitxonline/pull/1317.

arslanashraf7:
Closing this PR since https://github.com/mitodl/mitxonline/pull/1317 was merged recently that was rebased on this PR and has now been reviewed and merged so this PR can now be closed.

PR #1295, opened 2022-12-19
Updates courseware URL generation in import_courseware command

### LINKED ISSUE #1280: update Courseware URL path set by`import_courserun` management command
Courseware URL path determines the URL on the mitxonline dashboard that links users to their open edX courseware. The code management command that imports courses generates a link that goes to the unused course "about" page, instead of the course home page. 

https://github.com/mitodl/mitxonline/blob/da3cf35b6e78b9a72e3ac7c242ed76bc9a9be7a4/courses/management/commands/import_courserun.py#L181-L183

It would be more direct to use links with the format:

`f"/learn/{edx_course.course_id}/home",`

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes #1280

#### What's this PR do?

Changes the generated URL for `courseware_url` to include `/home` when importing a course run from edX, so the learner will be taken into the course rather than to the about page on edX. 

Import a course run - the instructions from https://github.com/mitodl/mitxonline/pull/1256 or the [documentation for the command](https://mitodl.github.io/mitxonline/commands/import_courserun.html) would be helpful here. The resulting course run in MITx Online should have a URL that ends in `/home` and activating it should take you into the course.

### DIFF
--- a/courses/management/commands/import_courserun.py
+++ b/courses/management/commands/import_courserun.py
@@ -180,7 +180,7 @@ class Command(BaseCommand):
                     is_self_paced=edx_course.is_self_paced(),
                     courseware_url_path=parse.urljoin(
                         settings.OPENEDX_API_BASE_URL,
-                        f"/courses/{edx_course.course_id}",
+                        f"/courses/{edx_course.course_id}/home",
                     ),
                 )
 


arslanashraf7 (approved):
👍

pdpinch:
Uh, does this actually work? 

I'll see if I can catch it on RC.

arslanashraf7:
> Uh, does this actually work?
> 
> I'll see if I can catch it on RC.

Ah, I see, I think I get what you are saying. And you are right! This might not work. While reviewing the change was pretty straightforward and my edX learning MFE was having problems with localhost OAuth and thought we can take on it. But I think the `/learn` part should be necessary here so that it generates the URLs like `/learn/course-v1:edX+DemoX+Demo_Course/home` instead of `/courses/course-v1:edX+DemoX+Demo_Course/home`.

arslanashraf7:
Now that i think of it, There's an even better solution to this.

Instead of hardcoded `/learn` here which is our public_path for learning MFE, I think we should use something like `/course` e.g. `/courses/course-v1:edX+DemoX+Demo_Course/course`. It's better because edX will automatically handle the MFE redirection in this case and we won't have to worry about this hardcoded `/learn` since it can change.

arslanashraf7:
My apologies and https://github.com/mitodl/mitxonline/pull/1305 is the PR in which i shifted it to `/course` instead as per my most recent comment above. @pdpinch @jkachel

pdpinch:
I hadn't thought about it, but I agree that using `/courses/course-v1:edX+DemoX+Demo_Course/course` would be more robust since the path of the MFE is configurable (we use `/learn/`)

PR #1299, opened 2022-12-20
Fixes the course model to round the grade (fixes a flaky test)

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

n/a

#### What's this PR do?

Updates `grade_percent` to return a Decimal that's rounded half even to 2 places rather than the default. This was causing a test to fail randomly. 

The test that fails is `courses/serializers_test.py::test_serialize_course_run_enrollments_with_grades` and the problem is that the grade is stored as a decimal and the grade calculation (which then uses a `Decimal` object and divides it) ends up adding precision where it's not really needed. (So, grade is .78 and then it gets multiplied by 100 and due to floating point math ends up being 77.99999 or something.)

### DIFF
--- a/courses/models.py
+++ b/courses/models.py
@@ -4,6 +4,9 @@ Course models
 import logging
 import operator as op
 import uuid
+from decimal import ROUND_HALF_EVEN
+from decimal import Context as DecimalContext
+from decimal import Decimal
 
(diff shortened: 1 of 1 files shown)


annagav (approved):
👍

PR #1305, opened 2022-12-20
fix: Fix courseware URL in command

arslanashraf7:
Fix for https://github.com/mitodl/mitxonline/pull/1295 (Please take a look at the comments in this PR)

#### What's this PR do?
Actually, fixes a value in courseware path generation that we missed in https://github.com/mitodl/mitxonline/pull/1295

Same steps as mentioned in https://github.com/mitodl/mitxonline/pull/1295


jkachel (approved):
LGTM 👍

PR #1307, opened 2022-12-21
Fixes import_courserun to handle courses that exist already

### LINKED ISSUE #1306: import_courseware fails if the course exists when importing an individual courserun
When using `--courserun`, the command will create the course if it's not there; however, it doesn't use `defaults` right, so if the course exists, it'll fail. 

### Steps to Reproduce

Import a new run of an existing course as a one-off (using `--courserun`). 

### Expected Behavior

The courserun is created successfully.

### Actual Behavior

You get a duplicate key error.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes #1306 

#### What's this PR do?

Updates the call to `get_or_create` so that single courseruns create the underlying course successfully. If you tried to import a course run for a course that existed, this would fail because the call to `get_or_create` to retrieve the course was malformed.

Create a new course run for an existing course in edX, then import using `import_courserun --courserun`. The command should complete successfully.

Create an entirely new course in edX, then import using `import_courserun --courserun`. The command should complete successfully. 

Noticed this reviewing #1305

### DIFF
--- a/courses/management/commands/import_courserun.py
+++ b/courses/management/commands/import_courserun.py
@@ -151,13 +151,14 @@ class Command(BaseCommand):
             course_readable_id = edx_course.course_id.removesuffix(f"+{courserun_tag}")
 
             try:
-                (course, created) = Course.objects.filter(
-                    readable_id=course_readable_id
-                ).get_or_create(
-                    program=program,
-                    title=edx_course.name,
+                (course, created) = Course.objects.get_or_create(
                     readable_id=course_readable_id,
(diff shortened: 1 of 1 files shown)


arslanashraf7 (approved):
👍 Looks good.

I was able to reproduce this error on the main branch.

Tested both with a new course and a new run of an existing course I was able to import both successfully. I also tried with other args e.g. `--price, --create-cms-page` etc and it works.

Issue #1326, opened 2023-01-05
decimal grades on the learner record


pdpinch:
Two significant digits are enough 

### Steps to Reproduce

Look up the learner record for the user with ZenDesk [#144252](https://odl.zendesk.com/agent/tickets/144252) on production

### Expected Behavior

- The grade for 14.750x should be 58, not 57.99999999999999
- Don't change the data from edX, just change the display to only show integers

### Screenshot or Screencast

<img width="1145" alt="image" src="https://user-images.githubusercontent.com/430126/210888938-d2ba4768-37a3-4ea2-887c-988fa6963153.png">

PR #1331, opened 2023-01-09
1326: decimal grades on the learner record

collinpreston:
#### Pre-Flight checklist

  - [x] Desktop screenshots
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/1326

#### What's this PR do?
Sets the grade returned from the `LearnerRecordSerializer` to have only 2 decimal places.  That serializer is used on the program records page (http://mitxonline.odl.local:8013/records/1/)

1. Have `mitxonline` up and running locally.
2. Set `ENABLE_LEARNER_RECORDS=True` in the .env file.
3. Have a user enrolled into a course run for a course that is a requirement of a program.  Create a course run grade for the course run, and set the grade value as a repeating decimal.
4. Create a partner school record through Django Admin.
5. Navigate to the `mitxonline` Dashboard (http://mitxonline.odl.local:8013/dashboard/?enable_programs&enable_lr).
6. Go to the programs tab and open the program drawer by clicking on the program card.
7. Click "view program record" from the drawer.
8. Ensure that the grade shown for the course is only showing 2 decimal places and not the repeating decimal.

Screenshot 2023-01-09 at 15 35 08

### DIFF
--- a/courses/models.py
+++ b/courses/models.py
@@ -1118,7 +1118,7 @@ class CourseRunGrade(TimestampedModel, AuditableModel, ValidateOnSaveMixin):
     @property
     def grade_percent(self):
         """Returns the grade field value as a number out of 100 (or None if the value is None)"""
-        TWOPLACES = Decimal(10) ** -2
+        TWOPLACES = Decimal(10)
         return (
             Decimal(self.grade * 100).quantize(TWOPLACES)
             if self.grade is not None
--- a/courses/models.py
+++ b/courses/models.py
@@ -4,7 +4,7 @@ Course models
 import logging
 import operator as op
 import uuid
-from decimal import ROUND_HALF_EVEN
+from decimal import ROUND_HALF_EVEN, getcontext
 from decimal import Context as DecimalContext
 from decimal import Decimal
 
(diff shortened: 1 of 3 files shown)


jkachel (changes requested):
FWIW tested by giving myself a grade in mitxonline of 0.5467666 and still got a rounding error on the program record.

courses/serializers.py
```diff
@@ -760,6 +760,7 @@ def to_representation(self, instance):
                 .order_by("-grade")
                 .first()
             )
+            grade.grade = round(grade.grade, 2)
```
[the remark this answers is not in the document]
collinpreston: Ha, I used .3333333 and that worked.  Thanks for trying something different and catching that!

courses/models.py
```diff
@@ -1119,7 +1119,7 @@ def to_dict(self):
     def grade_percent(self):
         """Returns the grade field value as a number out of 100 (or None if the value is None)"""
         return (
-            Decimal(self.grade * 100, DecimalContext(prec=2, rounding=ROUND_HALF_EVEN))
+            Decimal(self.grade * 100).quantize(exp=Decimal(1), rounding=ROUND_HALF_EVEN)
```
collinpreston, line 1122: @jkachel, do you think we should still stick with `ROUND_HALF_EVEN` (ref: https://docs.python.org/3/library/decimal.html#rounding-modes).  I would think we would want something like `ROUND_HALF_UP`.

jkachel: I was a bit conflicted about this when I added the rounding in.. half even is more "fair" over a long timescale (across all grades recorded) but less so for an individual learner so half up may be the better choice.

jkachel (approved):
LGTM 👍

Issue #1349, opened 2023-01-13
Set blocked country when importing new courseruns 


pdpinch:
As an admin, it's time-consuming and error-prone to set up country blocking. We should make this simpler and more robust. 

Currently, everytime I create a new **_course_** in mitxonline, I need to add a new record for each blocked course and country pair through the django admin `/admin/courses/blockedcountry/`

**And**, in open edX studio, when we create a new **_courserun_**, I need to add a new record for each course run in the django admin, `/admin/embargo/restrictedcourse/`

#### Acceptance Criteria:

- [ ] Add a parameter to the `import_courserun` management command for setting the blocked countries. This should be pretty simple for creating a `blockedcountry` record in mitxonline
- [ ] research if there is a way to get and set `restrictedcourse` in open edX studio.

asadali145:
> research if there is a way to get and set restrictedcourse in open edX studio.

There is no way in edX to get and set `restrictedcourse` data for a course. Maybe we can create our own app and APIs if needed in [open-edx-plugins](https://github.com/mitodl/open-edx-plugins)

Ferdi:
>  in open edX studio, when we create a new courserun, I need to add a new record for each course run in the django admin, /admin/embargo/restrictedcourse/

Is this necessary  if we are blocking users in mitxonline app ?

pdpinch:
> in open edX studio, when we create a new courserun, I need to add a new record for each course run in the django admin, /admin/embargo/restrictedcourse/

> Is this necessary if we are blocking users in mitxonline app ?

We miss some edge cases, like users who report they are from one country, but use an IP address from another country. However, I don't (small) risk justifies the effort and complexity.

PR #1352, opened 2023-01-17
feat(import_courserun): add ability to block countries

asadali145:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes https://github.com/mitodl/mitxonline/issues/1349

#### What's this PR do?

- Adds a new flag `--block_countries` to create `BlockedCountry` objects.
- `--block_countries` is a comma-separated list of country names to block enrollments.

- Make sure that your integration with edX is working fine.
- Create a new course in edX locally.
- Import the course using `import_courserun` command and provide an argument to block countries like `--block_countries='pakistan,united states of america'`. You should see success messages for the blocked countries.
- Visit `/admin/courses/blockedcountry/`. Blocked Country objects should be created.

### DIFF
--- a/courses/management/commands/import_courserun.py
+++ b/courses/management/commands/import_courserun.py
@@ -83,7 +83,7 @@ class Command(BaseCommand):
         parser.add_argument(
             "--block_countries",
             type=str,
-            help="Comma separated list of countries to block enrollments.",
+            help="Comma separated list of countries to block enrollments. Both Country Name and ISO code are supported",
             nargs="?",
         )
 
--- a/courses/management/commands/import_courserun.py
+++ b/courses/management/commands/import_courserun.py
@@ -234,25 +234,33 @@ class Command(BaseCommand):
                             )
                         )
 
-                for country_name in kwargs["block_countries"].split(","):
-                    country_code = countries.by_name(country_name)
-                    if country_code:
-                        BlockedCountry.objects.get_or_create(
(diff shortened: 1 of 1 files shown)


arslanashraf7 (approved):
Looks great 👍 

Please confirm with @pdpinch before merging this.

courses/management/commands/import_courserun.py
```diff
@@ -79,6 +80,13 @@ def add_arguments(self, parser) -> None:
             nargs="?",
         )
 
+        parser.add_argument(
+            "--block_countries",
+            type=str,
+            help="Comma separated list of countries to block enrollments.",
```
pdpinch, line 86: Does this work for names or 2-letter ISO codes? We should include an example in the help message.

pdpinch (commented):
just one questions:

What happens if the course is already blocked for the country specified? It seems like you just display the success message again, which is fine. I just want to confirm the logic.

courses/management/commands/import_courserun.py
asadali145: It works for both. It looks for the name and if not found then it looks for the code. I will update the docs with an example.

asadali145:
> What happens if the course is already blocked for the country specified? It seems like you just display the success message again, which is fine. I just want to confirm the logic.

If a course is already blocked for a country then we just display the success message. If not then we block and display the message.

PR #1361, opened 2023-01-18
Removes check for values before rendering the create discount form

jkachel:
The create form was checking for values before rendering, causing it to not to render on RC.

#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

n/a

#### What's this PR do?

Removes a check for values before rendering the discount form.

Try to create a discount using the staff dashboard. You should get an empty form, and you should be able to create the new discount successfully.

Editing an existing discount should also work as expected.


asadali145 (approved):
LGTM!

PR #1363, opened 2023-01-19
Updates discount application code to strip whitespace

### LINKED ISSUE #1362: Discount codes aren't applied if the user adds whitespace
This originated in https://odl.zendesk.com/agent/tickets/144920 - a learner had cut and pasted a discount code into the field but had grabbed a bunch of blank space before the code, which caused it to fail to apply. 

### Steps to Reproduce

1. Grab/generate a valid discount code. 
2. Set up a cart that the discount code can be applied to. 
3. Attempt to apply the discount, but put spaces before or after the code.

### Expected Behavior

The system considers only the actual code, and strips any spaces added before or after.

### Actual Behavior

It doesn't, and the code isn't applied. You get an error message instead.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes #1362

#### What's this PR do?

Updates the code that processes discount codes to strip whitespace when looking for the discount code.

Attempt to apply a discount code with spaces at the front and/or back of the code. It should be applied successfully. 

Tests should additionally pass.

### DIFF
--- a/ecommerce/views/v0/__init__.py
+++ b/ecommerce/views/v0/__init__.py
@@ -427,7 +427,7 @@ class CheckoutApiViewSet(ViewSet):
 
         try:
             discount = Discount.objects.filter(for_flexible_pricing=False).get(
-                discount_code=request.data["discount"]
+                discount_code=request.data["discount"].strip()
             )
 
             if not api.check_discount_for_products(discount, basket):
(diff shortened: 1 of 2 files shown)


asadali145 (approved):
LGTM!

PR #1367, opened 2023-01-20
Makes it easier to cancel an order in the Review state

jkachel:
We've gotten a few orders that have hit the Review state due to payment processor issues. In these cases, we have to clear the discount redemptions that may exist for the order and change the order status to `canceled`; there wasn't a way to do this via the Django Admin, so you had to drop into a database shell. This adds in a Cancel Order button to the Django admin for these orders, which will do that process for you. 

#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

n/a (but there are 3-4 Zendesk tickets that spurned me on to do this)

#### What's this PR do?

* Adds a new Review Order page under Ecommerce in the Django Admin, which will just show orders in Review state
* For orders in Review, adds a button to the button bar titled "Cancel Order" that delivers you to an interstitial page to confirm cancelling the order.
* Model support for this functionality.

1. Complete an order. For a full test, complete two orders and ensure one's got a discount applied to it (doesn't matter what kind).
2. Manually set the status the order to "review". You will need to do this directly using a database shell. 
3. Find Review Order in Django Admin. Orders in Review state should be displayed.
4. Open the order you changed in step 2. You should see a Cancel Order button.
5. Click Cancel Order. You should be sent to an interstitial page listing some of the order details.
6. Click Cancel Order. You should be redirected back to the list page for Review Orders and you should see a message saying the order has been cancelled. 
7. If the order had discounts applied, check the Discount Redemptions table to ensure those redemptions are gone.

Applicable Zendesk tickets: 145785, 145591, 144756, 143051

image
image
image

### DIFF
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -1,7 +1,5 @@
 """Admin management for Ecommerce module"""
 
-import logging
-
 from django.contrib import admin, messages
 from django.contrib.admin.decorators import display
 from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -270,8 +270,6 @@ class ReviewOrderAdmin(TimestampedModelAdmin):
         "reference_number",
     ]
     list_display = ["id", "state", "get_purchaser", "total_price_paid"]
-    list_fields = ["state"]
-    list_filter = ["state"]
     inlines = [OrderLineInline, OrderDiscountInline, OrderTransactionInline]
     readonly_fields = ["reference_number"]
 
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -1,5 +1,7 @@
 """Admin management for Ecommerce module"""
 
+import logging
+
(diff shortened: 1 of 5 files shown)


ecommerce/admin.py
```diff
@@ -256,6 +257,44 @@ def get_queryset(self, request):
...
+
+    search_fields = [
+        "id",
+        "purchaser__email",
+        "purchaser__username",
+        "reference_number",
+    ]
+    list_display = ["id", "state", "get_purchaser", "total_price_paid"]
+    list_fields = ["state"]
+    list_filter = ["state"]
```
collinpreston, line 274: Nit-pick but should the filter be changed to something different since all the orders will have a status of review?

collinpreston (approved):
Looks good! Tested a few times without issue.  Just a UX comment.

jkachel:
Good point - I just pulled the filtering. If there's more than a couple records sitting in Review, we have problems so not a whole lot of use for filtering. (I don't think this is a change that requires re-review so just going ahead with merging.)

PR #1371, opened 2023-01-23
Updates repair_faulty_edx_user to reconnect edX users

### LINKED ISSUE #258: Add or update management command to sync existing edX users
We've experienced quite a few cases where a user creates their account, the equivalent user is successfully created on the edX side, but due to some error, an `OpenEdxUser` record is not created. As a result, when then script is run to repair users that do not have `OpenEdxUser`/`OpenEdxApiAuth` records, we get errors because the script is trying to create a new user on the edX side, but a user with the same username/email already exists. We should update the logic for repairing these users so that if a user already exists on the edX side with the same username and email, we simply create the equivalent `OpenEdxUser` on the mitxonline side.

Just to help guide implementation, here is a script that I ran to find users in this state. This code doesn't need to be used – it's just here as a guide. A couple things to note:
- It uses an admin user for the API client. I don't think we do that anywhere in the app currently
- It uses an edX API endpoint that the edX API client library does not support. If we wanted to use that endpoint, we should add support for it in the client first
- The first hint that a user already exists in edX is that the initial request to create the user comes back with a 409 response. This is probably what we should use to determine which users are candidates for this type of repair. Note that we get a 409 if _only_ the username _or_ the email match, and not both. In those cases, we should _not_ create a new `OpenEdxUser` record

```python
from urllib.parse import urljoin
from django.db.models import Count, Q
User = get_user_model()
from openedx.constants import PLATFORM_EDX
from openedx.models import OpenEdxApiAuth, OpenEdxUser
from openedx.api import get_edx_api_client
#
admin_user = User.objects.filter(is_superuser=True).first()
admin_client = get_edx_api_client(admin_user)
users_to_repair = User.objects.all().annotate(openedxusers=Count("openedx_users")).filter(openedxusers=0)
#
missing_local = []
for user_to_repair in users_to_repair:
    resp = admin_client.user_info.requester.get(urljoin(admin_client.user_info.base_url, f"/api/user/v1/accounts/{user_to_repair.username}"))
    if resp.status_code != 200:
        continue
    resp_dict = resp.json()
    if user_to_repair.username == resp_dict["username"] and user_to_repair.email == resp_dict["email"]:
        missing_local.append(user_to_repair)
print(f"Users without a local OpenEdxUser that do exist in edX: {len(missing_local)}")
```

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes #258 

#### What's this PR do?

Makes some changes to the `repair_faulty_edx_users` API call: 
* If the call to create the user account in edX fails with a 409 specifically, this should ignore that error and continue. (Otherwise, it should re-raise the exception that is generated.)
* After creating the edX auth tokens, it should now check to see if there's an OpenEdxUser record as well, and should create a new one if there isn't one. 

1. Create two users manually (via Django Admin) in both MITx Online and Open edX. 
   * Note that the Open edX account _must_ have a profile for this to work. (Specify anything here; it's enough for it to have the record at all.) If you're testing a failure mode, then don't specify a profile.
   * The Open edX account password doesn't matter. 
   * The usernames and email addresses must match between MITx Online and Open edX. 
   * The users must be active.
2. Run the `repair_missing_courseware_records` management command. It should find the user you created, then try to repair it. It should successfully create auth tokens. 

If you're running the devstack Open edX deployment, it comes with a bunch of accounts set up for various purposes; these are handy to test with as well. (For example, you should be able to make an "edx@example.com" account in mitxonline and it should then connect to the devstack one.)  

The failure mode here is that the user accounts get created in both systems, but the auth tokens don't get stored. When this happens and the repair command is run, the `repair_faulty_edx_users` API call first tries to re-create the user in edX. edX notices that there's already a user with a matching username and email, so it errors out, and the `repair_faulty_edx_users` call wasn't catching the exception when this happened so the entire process stopped for that user. However, in these cases, we already have users on both sides, so really we just need to make the respective auth tokens and go from there, which is what the changes here do.

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -26,27 +26,26 @@ from authentication import api as auth_api
 from courses.constants import ENROLL_CHANGE_STATUS_UNENROLLED
 from main.utils import get_partitioned_set_difference
 from openedx.constants import (
-    EDX_ENROLLMENT_AUDIT_MODE,
     EDX_DEFAULT_ENROLLMENT_MODE,
+    EDX_ENROLLMENT_AUDIT_MODE,
     EDX_ENROLLMENT_VERIFIED_MODE,
     OPENEDX_REPAIR_GRACE_PERIOD_MINS,
     PLATFORM_EDX,
     PRO_ENROLL_MODE_ERROR_TEXTS,
 )
-
 from openedx.exceptions import (
+    EdxApiEmailSettingsErrorException,
     EdxApiEnrollErrorException,
+    EdxApiRegistrationValidationException,
     NoEdxApiAuthError,
     OpenEdXOAuth2Error,
     OpenEdxUserCreateError,
+    UnknownEdxApiEmailSettingsException,
     UnknownEdxApiEnrollException,
     UserNameUpdateFailedException,
-    EdxApiEmailSettingsErrorException,
-    UnknownEdxApiEmailSettingsException,
-    EdxApiRegistrationValidationException,
 )
 from openedx.models import OpenEdxApiAuth, OpenEdxUser
-from openedx.utils import edx_url, SyncResult
+from openedx.utils import SyncResult, edx_url
 from users.api import fetch_user
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


asadali145 (approved):
LGTM!

PR #1376, opened 2023-01-24
Incorporates changes to make MITxOnline work with Tutor

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

n/a

#### What's this PR do?

Documentation updates and some minor code changes to facilitate using MITx Online with Tutor. 

This also nerfs the old Refine Admin docs; they were out of date (we don't have a specific OIDC setup for that anymore). 

**Documentation/Tutor Deployment:**
Note that Open edX's Tutor distribution replaces devstack - you can run them at the same time (they expose different ports) but I at least found that to be a bit confusing, not to mention resource intensive. If you're running this on an arm64/Apple Silicon machine, [there are specific docs for that](https://docs.tutor.overhang.io/tutorials/arm64.html) (you'll have to rebuild the `openedx` image anyway but there are other things that need to be done). 

Build the documentation and then read it: 
```
./pants docs ::
docker run --rm -ti --name mitxdocs -p 7888:80 -v `pwd`/dist/sphinx:/usr/share/nginx/html nginx
```

This will build the documentation, then run a nginx container on port 7888 pointed at the built docs (or just open it from `dist/sphinx`, that works too).

Follow the documentation at `docs/source/configuration/tutor.rst` (or http://localhost:7888/configuration/tutor.html if you did the Docker thing above). It should work.

**Support:**

* The `configure_instance` command now accepts a `--tutor` flag - if you specify this, it will use "local.overhang.io" instead of "edx.odl.local:18000" for URLs in the appropriate spots. (This will get tested as part of the Tutor deployment docs but if you want to test it manually, you can - just check for different URLs in the OAuth2 application records.)
* MITx Online containers should be able to resolve `local.overhang.io`, and it should resolve to what you've set `OPENEDX_IP` to rather than what's in DNS. 

[HQ thread](https://github.com/mitodl/hq/discussions/421)

[devstack deprecation](https://github.com/openedx/devstack/issues/907)


asadali145:
@jkachel Once this PR is merged and approved, can we still use the Devstack based edX installation with MIT xOnline?

jkachel:
> @jkachel Once this PR is approved and merged, can we still use the Devstack-based edX installation with MIT xOnline?

Yes; the actual changes to the app are additive.

rachellougee (approved):
Great work! I was able to follow the documentation to get MITx Online app to work with tutor local, and ran a couple of commands to sync data.

Issue #1365, opened 2023-01-20
course.title and courserun.title don't match


pdpinch:
### Steps to Reproduce

1. There are a couple of different ways to see this, but the easiest is to look at the API results at 
https://mitxonline.mit.edu/api/courses/ and search for the course id `course-v1:MITxT+10.50.CH01x`

Presumably this is a result of the title being updated in Studio, which is then synced with courserun.title but not course.title (see https://github.com/mitodl/mitxonline/issues/106) 

### Expected Behavior

- course.title and courserun.title should be the same

### Actual Behavior

- course.title and courserun.title are not the same
- If there are multiple courserun.titles, we could set the course title to the one with the latest start date

### Screenshot or Screencast

<img width="1126" alt="Screenshot 2023-01-20 at 10 22 37 AM" src="https://user-images.githubusercontent.com/430126/213738104-ebdd98ca-ee2b-469c-801e-c54f0d946287.png">

pdpinch:
@asadali145 what is your plan with this? 

There was some discussion in Slack at https://mitodl.slack.com/archives/C02649X2P1V/p1674227568748509

In particular, I liked this suggestion from @rhysyngsun 
> so an option might be to update the course title from the page title, although right now we don't enforce 1 product page per course/program so we'd want to address that too

asadali145:
@pdpinch I was about to implement it as we have mentioned in the ticket i.e. sync course.title with courserun.title. But it seems like the suggestion above by @rhysyngsun is a better option to go with.

> although right now we don't enforce 1 product page per course/program so we'd want to address that too

I don't think there is any issue with this part of the suggestion as we **_cannot_** create multiple CMS Pages for a single course. So, It does not require anything else apart from syncing the CMS Page title with course.title.

asadali145:
@pdpinch Could you please have a look at these changes and see if you have any suggestions? https://github.com/mitodl/mitxonline/pull/1382

pdpinch:
Your PR looks like what @rhysyngsun and I had in mind.

mbertrand:
This seems to be an issue still on production:

```json
    {
        "id": 45,
        "title": "Chapter 2: Mathematical Formulation",
        "readable_id": "course-v1:MITxT+10.50.CH02x",
        "courseruns": [
            {
                "title": "Analysis of Transport Phenomena: Mathematical Formulation",
                "start_date": "2023-03-01T14:00:00Z",
[...]

PR #1382, opened 2023-01-25
feat: Sync courseware title with CMS page title

asadali145:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes https://github.com/mitodl/mitxonline/issues/1365

#### What's this PR do?
Syncs courseware title with CMS Product page title.

- Update CMS Product page i.e. CoursePage or ProgramPage
- Verify that the title is synced with the Courseware object title either by visiting Django Admin or through the APIs.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -4,7 +4,7 @@ import logging
 import re
 from datetime import datetime, timedelta
 from json import dumps
-from urllib.parse import quote_plus, urljoin
+from urllib.parse import quote_plus
 
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
@@ -14,7 +14,6 @@ from django.db import models
 from django.forms import ChoiceField, DecimalField
 from django.http import Http404
 from django.template.response import TemplateResponse
-from django.templatetags.static import static
 from django.urls import reverse
 from django.utils.text import slugify
 from mitol.common.utils.datetime import now_in_utc
@@ -58,7 +57,7 @@ from cms.constants import (
 )
 from cms.forms import CertificatePageForm
 from courses.api import get_user_relevant_course_run, get_user_relevant_course_run_qset
-from courses.models import Course, CourseRunCertificate, ProgramCertificate, Program
+from courses.models import Course, CourseRunCertificate, Program, ProgramCertificate
 from flexiblepricing.api import (
     determine_auto_approval,
     determine_courseware_flexible_price_discount,
@@ -73,7 +72,6 @@ from flexiblepricing.models import (
     FlexiblePrice,
     FlexiblePricingRequestSubmission,
 )
-from main import features
 from main.views import get_base_context
 
 log = logging.getLogger()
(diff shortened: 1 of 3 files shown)


cms/models.py
```diff
@@ -748,6 +746,20 @@ class Meta:
...
+        super().save(clean=clean, user=user, log_action=log_action, **kwargs)
+
+        courseware_object = None
+        if self.is_course_page:
+            courseware_object = getattr(self, "course")
+        elif self.is_program_page:
+            courseware_object = getattr(self, "program")
+        courseware_object.title = self.title
+        courseware_object.save()
+
```
arslanashraf7, line 762: I think this might be better to do in an atomic transaction since this by any means if one of them fails we don't want the other one to be updated and be out of sync.

arslanashraf7 (changes requested):

cms/models.py
asadali145: After a discussion, we decided to change the ordering of saving the objects instead doing an atomic transaction.

cms/models_test.py
```diff
@@ -539,3 +539,20 @@ def test_certificate_for_program_page():
         assert signatory.value.title_2 == "Title_2"
         assert signatory.value.organization == "Organization"
         assert signatory.value.signature_image.title == "Image"
+
+
+@pytest.mark.parametrize("test_course", [True, False])
+def test_courseware_title_synced_with_product_page_title(test_course):
+    """Tests that Courseware title is synced with the Course Page title"""
```
arslanashraf7, line 546:
```suggestion
    """Tests that Courseware title is synced with the Course Page title from CMS"""
```

arslanashraf7 (approved):
Looks good 👍  Just a minor suggestion.

Please confirm with @pdpinch before merging, since this might be a change people might want to know about.

rhysyngsun (commented):
Do you think we should make the course/program titles readonly in django admin (perhaps with a note/link on where to change it) if there's a CMS page to prevent that from getting out of sync by a change in that part of the application?

asadali145:
@rhysyngsun Could you please review this PR now? 9e43731d10dd7de7dcf06e53ec393ddae45696d0 makes the course title read only for Django Admin.

asadali145:
@arslanashraf7 Could you please review the latest commit?

rhysyngsun (changes requested):
Check for `is not None` and also need to check that there's a page otherwise I get an error on django-admin.

courses/admin.py
```diff
@@ -68,6 +69,32 @@ class CourseAdmin(admin.ModelAdmin):
         models.CharField: {"widget": TextInput(attrs={"size": "80"})}
     }
 
+    def get_readonly_fields(self, request, obj=None):
+        """
+        Adds `title` as readonly field while editing an existing object.
+        """
+        if obj:
```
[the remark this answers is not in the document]
asadali145: I have used `hasattr` as I was getting No attribute error when I was doing `obj.page`.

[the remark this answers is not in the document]
rhysyngsun: Another more concise way you can do this to avoid the extra `and` is with `getattr(obj, "page", None) is not None`, because `getattr(None, "page", None) is None`. I'm nitpicking at this point so this is a take-it-or-leave-it kind of comment.

asadali145:
@rhysyngsun Should I merge this one or do you have some more suggestions?

rhysyngsun (approved):
LGTM

PR #1389, opened 2023-01-26
fix: Fix edX username validation to avoid username collision

asadali145:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Somewhat https://github.com/mitodl/mitxonline/issues/259 and this slack [discussion](https://mitodl.slack.com/archives/C02649X2P1V/p1674669465296569).
Fixes https://github.com/mitodl/mitxonline/issues/1393

#### What's this PR do?
Fixes the username validation with edX.

- We implemented username validation in https://github.com/mitodl/mitxonline/pull/757. We were rooting on a validation message.
- edX updated validation messages under https://github.com/openedx/edx-platform/pull/30937.
- This PR fixes the validation and also makes it generic so that any issue with the username can be validated.

- Make sure that your edX integration works and `FEATURES["ENABLE_UNICODE_USERNAME"] = True`
- Create an account in MITxOnline with accented characters.
- Verify that the account is created in MitxOnline and in edX.
- Now try to create another account with the same username but unaccented characters.
- You should see an error message that `A user already exists with this username. Please try a different one.`

Most of the details are available in the slack [thread](https://mitodl.slack.com/archives/C02649X2P1V/p1674669465296569) and https://github.com/mitodl/mitxonline/issues/259

Here are the alternative approaches that I considered:

- Perform an unaccented search in MITxOnline in such cases by following https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/lookups/#unaccent but that requires some setup in MitxOnline and Postgres.
- Perform an accented search in edX, but that might need some collation settings with MySQL that we are sure of.


users/serializers.py
```diff
@@ -141,9 +145,12 @@ def validate_username(self, value):
         if not re.fullmatch(USERNAME_RE, trimmed_value):
             raise serializers.ValidationError(USERNAME_ERROR_MSG)
 
-        openedx_username_taken = False
+        openedx_validation_msg = ""
         try:
-            openedx_username_taken = check_username_exists_in_edx(trimmed_value)
+            openedx_validation_msg = validate_username_with_edx(trimmed_value)
+            openedx_validation_msg = OPENEDX_USERNAME_VALIDATION_MSGS_MAP.get(
```
arslanashraf7, line 151: So I was thinking. If edX returns some error message for the username while validating account registration and the message is not related to the existing username that we had in `OPENEDX_USERNAME_VALIDATION_MSGS_MAP` we will get None for `openedx_validation_msg` even though there can be some other error for username?

In the above case, I think the user creation would still succeed, whereas I think if we get any error for the username field we should block the registration process, and in this case, we can send a generic message to the frontend for the user that the registration can't proceed.

arslanashraf7: I think it would be good to take @rhysyngsun's opinion on this too.

asadali145: I have used the `.get()` method of the dictionary and provided the msg returned by edX as default. So, in case the msg does not exist in our `OPENEDX_USERNAME_VALIDATION_MSGS_MAP` then it will return the msg returned by edX as I am doing `OPENEDX_USERNAME_VALIDATION_MSGS_MAP.get(openedx_validation_msg, openedx_validation_msg)`

rhysyngsun: The approach here is fine to pass through the openedx error message.

rhysyngsun (changes requested):
In addition to the changes I'm suggesting here, we'd talked about making the username validation required (meaning if openedx fails, we don't allow account creation to proceed).

To accomplish this I think we should actually move the username validation out of `validate_username` and into the general purpose `validate()`. Then in that method we can update this so that if the API call to openedx fails, return an object-level error (make sure the form displays this) that we're unable to create an account at this time and to please reach out to support.

This is a bit more invasive than your original changes but I think we need to stop soft-failing when there's an openedx outage because then we're right back to having accounts get created that need username repairs, but if we block on openedx being available, if there's an outage there's no further effort required on our part once that system is available again.

users/serializers.py
```diff
@@ -180,6 +170,26 @@ def validate(self, data):
...
+                openedx_validation_msg = validate_username_with_edx(username)
+                openedx_validation_msg = OPENEDX_USERNAME_VALIDATION_MSGS_MAP.get(
+                    openedx_validation_msg, openedx_validation_msg
+                )
+            except (
+                HTTPError,
+                RequestsConnectionError,
+                EdxApiRegistrationValidationException,
+            ):
+                raise serializers.ValidationError(
```
rhysyngsun, line 186: Could you log this exception so we'll know registrations are failing?

rhysyngsun (changes requested):
I'm dealing with rebuilding devstack at the moment, so I wasn't able to functionally test. One piece of feedback though.

rhysyngsun:
@asadali145 if your PR is ready and @arslanashraf7 is able to functionally test it, don't let me own functional testing hold up merging this.

arslanashraf7 (approved):
Functionality works fine 👍

Issue #1074, opened 2022-09-29
tracking types of discounts 


pdpinch:
As a member of the marketing team, I'd like to be able to filter financial reports on marketing discounts, so I can focus my analysis on the transactions where marketing coupons were used. 

As a member of the MITx team, I'd like to be able to filter on financial assistance discounts, so I can analyze and project how much it costs. 

in xPRO we have various coupon "types." These have proven helpful for filtering data in reports
https://github.com/mitodl/mitxpro/blob/2dfd014f70abf56b3bde856160ad5d2686bfa6d3/ecommerce/models.py#L506-L515

#### Acceptance Criteria:

- [ ] Attach a type to coupons. Suggested types: 
  1. marketing (sometimes called "B2C") 
  2. customer support (enrollment codes for medical issues, etc.) 
  3. flexible pricing / financial assistance 
  4. sales (sometimes called "B2B", for the future)
- [ ] #1427

asadali145:
Removing the `On Hold` tag as it seems we can start working on this one as it is moved to TODO.

asadali145:
As we are adding a new choice field that will also indicate whether a discount is for financial assistance or not. We will also need to remove the usages of the existing boolean field `for_flexible_pricing`. This includes writing a data migration and some refactoring to use the new field in place of `for_flexible_pricing`. 

@jkachel WDYT how we should approach replacing `for_flexible_pricing`? Should all this be a part of a single PR? Or should we create a separate ticket/PR for replacing the usages and creating a data migration?

We (me & @arslanashraf7) think that we should go with a single PR.

asadali145:
> include the type in AR/Financial reports

@pdpinch What do I need to do for this? Sorry, but I do not know about these changes. Is this something that I need to work on?

jkachel:
> @jkachel WDYT how we should approach replacing `for_flexible_pricing`? Should all this be a part of a single PR? Or should we create a separate ticket/PR for replacing the usages and creating a data migration?

I think this is fine to roll into the open PR, since it's all affecting the same parts of the code.

pdpinch:
> > include the type in AR/Financial reports

> @pdpinch What do I need to do for this? Sorry, but I do not know about these changes. Is this something that I need to work on?

I opened a separate issue for this, and put it on hold. You can close this issue when #1390 is merged.

PR #1390, opened 2023-01-26
feat: Add discount payment types

asadali145:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

Closes https://github.com/mitodl/mitxonline/issues/1074

#### What's this PR do?

- Adds a new field `payment_type` in the discounts model.
- Replaces all usages of `for_flexible_pricing` with `payment_type`.
- Adds data migration to set the payment types for the existing discounts.
- Updates the `generate_discount_code` management command to add a new argument.

There are a few things that we need to test.
- Test that the old discounts are updated per the data migration logic `/admin/ecommerce/discount/`.
  - If `for_flexible_pricing=True` then `payment_type=financial-assistance`
  - if the discount code starts with `CS-` then `payment_type=customer-support`
  - if the discount code starts with `JPAL-` then `payment_type=staff`
  - if the discount code starts with `MITALUM15-`, `14750x-`, `14740x-`, `1473x-`, `14310x-`, `JPAL102x-` or `CYB2022-` then `payment_type=marketing`
  - If discount code is `CYBER2022` then `payment_type=marketing`
- Test that the new field is added in the Django Admin list and detail view `/admin/ecommerce/discount/`.
- Test that the new field is added in refine dashboard. The discounts filter works fine and the new field is also working fine in discounts create, view, and edit forms.
- Test that the financial assistance request flow is fine. i.e Once a request is approved, the discount is applied automatically on the checkout page.
- Test that the discounts work fine when applying manually on the checkout page.


ecommerce/migrations/0031_discount_payment_type.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('ecommerce', '0030_alter_discount_for_flexible_pricing'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='discount',
+            name='payment_type',
+            field=models.CharField(choices=[('B2C', 'Marketing'), ('B2B', 'Sales'), ('FA', 'Financial Assistance'), ('CS', 'Customer Support')], max_length=3, null=True),
```
pdpinch, line 16: I need to confirm this list. Please don't merge until I've signed off.

pdpinch:
How should we handle updating existing coupons?

asadali145:
> How should we handle updating existing coupons?

I am not really sure about that. I have created a data migration for financial assistance discounts. I am not sure about all the other types of discounts. Maybe if have some kind of a pattern then we can create a data migration for other discounts as well.

For now, I have kept the new field nullable. So, It will be null for all the discounts apart from the ones that are related to financial assistance.

Do you have any information about any pattern that we used for the discounts that we can use and write a data migration?

ecommerce/migrations/0031_discount_payment_type.py
asadali145: @pdpinch Do you have an update on this?

pdpinch:
> I have created a data migration for financial assistance discounts.

great. 

> It will be null for all the discounts apart from the ones that are related to financial assistance.

Good idea. That's fine. 

> Do you have any information about any pattern that we used for the discounts that we can use and write a data migration?

@jkachel may have some suggestions for some of the coupons he created. 

To be clear, it's not necessary that we backfill all of them. I just want to update the easy ones.

ecommerce/migrations/0031_discount_payment_type.py
pdpinch: Need to add one more: "staff"

jkachel:
We've only had to do a few runs of these so it shouldn't be too bad to do. The patterns I've been using are:
* Customer Service - always starts with `CS-`
* JPAL staff discounts - starts with `JPAL-`
* MIT Alumni discounts - starts with `MITALUM15-`
* There were also 100 codes each for $100 fixed price for the individual DEDP courses that start with an abbreviated version of the course number (i.e. `14750x-`) - I can't find the ticket that we made these for though. (These are different than the JPAL discounts, which also have the course number in them.) 

The `create_discount_code` management command also needs to be updated to support this. Didn't see that in the code but batches of codes (and thus codes that would generally need to be tagged in this manner) are generated using that command.

asadali145:
> The create_discount_code management command also needs to be updated to support this.

Yeah, I will be updating a couple of commands, The work is in progress for that.

> MIT Alumni discounts - starts with MITALUM15-

What should be the type for these discounts?

> There were also 100 codes each

What should be the type for these course discounts?

asadali145:
Hey @pdpinch & @jkachel,
I am doing the following mapping for the new payment_type field. Let me know if there is any issue with this.

- Discounts with `for_flexible_pricing=True` will have `payment_type=financial-assistance`
- Discounts with discount codes starting with `CS-` will have `payment_type=customer-support`
- Discounts with discount codes starting with `JPAL-` will have `payment_type=staff`


Also, I am not sure about these 2 discounts. What should be the payment_type for these 2 types of discounts?

> MIT Alumni discounts - starts with MITALUM15-

> There were also 100 codes each for $100 fixed price for the individual DEDP courses that start with an abbreviated version of the course number (i.e. 14750x-) - I can't find the ticket that we made these for though. (These are different than the JPAL discounts, which also have the course number in them.)

jkachel:
> > MIT Alumni discounts - starts with MITALUM15-
> 
> What should be the type for these discounts?
> 
> > There were also 100 codes each
> 
> What should be the type for these course discounts?

After some thought, I think both of these are B2C. 

I did find out the purpose for the last group of codes I mentioned. They're for the lottery - https://odl.zendesk.com/agent/tickets/145641 is the Zendesk ticket for that (which you'll probably not be able to get into, just including for completeness) and this is it in HQ: https://github.com/mitodl/hq/issues/558. The relevant info from the ticket is:
> These coupons are used for a lottery that's run every term for DEDP. The specific products change each term because not every course is offered every term. 

In addition, I also found a couple more groups that will need backfilling: 

- `CYB2022-` (batch) and `CYBER2022` (singular) - we have a handful of these codes that were for this past Cyber Monday sale - those are definitely B2C. (There was an issue with the actual `CYBER2022` singular code I'd set up for this so there are also a batch of codes to hand out to folks who had issues with the code.)
- `MM-prepaid-` - there are a bunch of these from the MicroMasters migration. I'm not sure if these are B2C or CS, though, or if we should leave these as-is because this is a one-off special case.

asadali145:
@jkachel There are also some discount codes starting with `JPAL102x-`. Are they the same as `JPAL-`? Or are they more like lottery discounts? What should be the type of these? `staff` or `marketing`?

asadali145:
> MM-prepaid- - there are a bunch of these from the MicroMasters migration. I'm not sure if these are B2C or CS, though, or if we should leave these as-is because this is a one-off special case.

@pdpinch What do you suggest about these?

jkachel:
> @jkachel There are also some discount codes starting with `JPAL102x-`. Are they the same as `JPAL-`? Or are they more like lottery discounts? What should be the type of these? `staff` or `marketing`?

`JPAL102x-` are lottery codes and `JPAL-` are for JPAL staff.

jkachel (approved):
Everything looks good - staff dashboard looked OK, management command is good, workflow is fine on the frontend. Approving conditionally because of two things:
- waiting on resolution for `MM-prepaid` codes from @pdpinch (my thought is to leave them blank because we're never going to do anything like that again probably, or at least not without it being a whole thing that gets tracked like a project)
- can you update the docs for `generate_discount_code` in `docs/source/commands`? It needs the new flag added. 

Other than those two things, LGTM 👍

jkachel (changes requested):
Rescinding previous approval - did not notice but the financial assistance emails are broken. I can't link to a Sentry log for this because it's on my local Sentry instance but here's the relevant stack trace:

image

asadali145:
> Rescinding previous approval - did not notice but the financial assistance emails are broken. I can't link to a Sentry log for this because it's on my local Sentry instance but here's the relevant stack trace:

@jkachel Could you please verify if this is a legit issue? It seems like a migrations issue to me. Not sure if you tried to change something there. Everything is working fine for me. I am getting the emails. I have also tried to reapply all the migrations as well.

Screen Shot 2023-02-06 at 11 39 25 AM

jkachel:
> @jkachel Could you please verify if this is a legit issue? It seems like a migrations issue to me. Not sure if you tried to change something there. Everything is working fine for me. I am getting the emails. I have also tried to reapply all the migrations as well.

After some checking I have no idea. The local copy I use most of the time has issues (even with a fresh database and the migrations applied), but I tested it on a separate system and it works OK there. So, this does seem like a local-to-me issue or maybe at worst a macOS issue.

jkachel (approved):
LGTM 👍

pdpinch:
> waiting on resolution for MM-prepaid codes from @pdpinch (my thought is to leave them blank because we're never going to do anything like that again probably, or at least not without it being a whole thing that gets tracked like a project)

I think we should make a new category for these, "legacy"

asadali145:
23f0b3d1393994a61f44093e0affd5d2bcaa1695 adds a new payment type `legacy`.
@jkachel Could you have one last look?

PR #1401, opened 2023-01-31
Add a workflow for new issues

rhysyngsun:
#### Pre-Flight checklist
N/A

N/A

#### What's this PR do?
This adds a github action workflow to add new issues to a [global project on the hq repo](https://github.com/orgs/mitodl/projects/21/views/1) so we can triage them properly before starting on them.

Can't test until it's on `master`, but you can see the same workflow setup on this test repo:

https://github.com/mitodl/workflow-testing/blob/main/.github/workflows/new-issues.yml

And a successful run of that here:

https://github.com/mitodl/workflow-testing/actions/runs/4048796451

Note: the issue this ran for has since been removed from the project.

### DIFF
--- a/.github/workflows/new-issues.yml
+++ b/.github/workflows/new-issues.yml
@@ -1,0 +1,12 @@
+---
+name: Add new issues to mitodl/hq
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  add-to-hq:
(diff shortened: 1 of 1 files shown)


rhysyngsun:
Got a verbal :+1: from Peter on this so merging.

PR #1416, opened 2023-02-06
Log any exception thrown by hubspot task helpers

mbertrand:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Closed https://github.com/mitodl/mitxonline/issues/1414

#### What's this PR do?
If any kind of exception occurs when a hubspot task helper function calls a celery task, log the exception instead of raising it.

- Set up hubspot integration as described in the README (you can use the same token as RC)
- With all your docker instances running, try this in a shell (assuming you already have 2 `Order` objects):
  ```python
  from hubspot_sync.task_helpers import *
  from ecommerce.models import Order
  for order in Order.objects.all()[:2]:
    sync_hubspot_deal(order)
  ```

- The above should work without errors or log statements.
- Kill your redis instance (`docker kill mitxonline_redis_1`)
- Run the following:
  ```python
  for order in Order.objects.all()[:2]:
    sync_hubspot_deal(order)
  ```

- The above should generate two exception log statements with the message `Exception calling sync_deal_with_hubspot for order <Order.id>`

### DIFF
--- a/hubspot_sync/task_helpers.py
+++ b/hubspot_sync/task_helpers.py
@@ -1,7 +1,13 @@
 """ Task helper functions for ecommerce """
+import logging
+
 from django.conf import settings
 
 from hubspot_sync import tasks
+
+# pylint:disable-bare-except
+
+log = logging.getLogger(__name__)
 
 
 def sync_hubspot_user(user):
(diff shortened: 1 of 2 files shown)


rachellougee (approved):
👍 It works for me

PR #1423, opened 2023-02-09
Adds methods to check pending orders for resolution through CyberSource

### LINKED ISSUE #1415: Check pending orders for resolution through CyberSource
A scenario that's happened [at least once](https://github.com/mitodl/hq/issues/586) is that a learner may complete their order, but the order remains in Pending state in MITx Online. (In these cases, the callback that CyberSource hits fails for whatever reason.) This prevents the learner from seeing their receipt and leaves their course enrollment as `audit`. We should proactively check pending orders to see if there's a resolution in CyberSource for them, and cancel them otherwise. 

#### Acceptance Criteria:

- [ ] There should be code added to use the REST API to check transaction status. (The Secure Acceptance API does not have a method for checking on transaction status.) 
   - The Transaction Details and Transaction Search explorers here may be of use: https://developer.cybersource.com/api-reference-assets/index.html#transaction-details_transaction-details_retrieve-a-transaction
   - This code may be best kept in the [ol-django payment gateway](https://github.com/mitodl/ol-django) library. 
- [ ] A management command should exist to manually check the status of an order at will. 
- [ ] A scheduled task should exist to check the status of all pending orders. 
   - This should ideally make some change to the transaction if there's no matching CyberSource order - orders will sit in Pending state if they're abandoned carts, for instance, so the scheduled task shouldn't hit the API repeatedly for an order that will never be fulfilled. 
- [ ] If the check does find a successful fulfillment, it should complete the order as normal - there should be a transaction line, the receipt in MITx Online should work, and the learner should receive a receipt email

jkachel:
(This depends on https://github.com/mitodl/ol-django/pull/109 so it's in draft mode until that's done.)

#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

Closes #1415

#### What's this PR do?

Adds a management command and a task to find pending orders and check for their resolution through CyberSource.

Automated tests should work.

The easiest way to get an order to be both pending and completed in CyberSource is to go through an order and then cancel it as it's redirecting you back into MITx Online. This will be difficult to do on RC or prod but should be doable on a local instance, because your browser will (probably) complain about the form being submitted to a non-secure endpoint. You will need one of these to test this functionality. 

In addition, you may want to add an offset to the generated reference number in the Orders model. On like 483 in `ecommerce/models.py`:
```python
        return f"{REFERENCE_NUMBER_PREFIX}{settings.ENVIRONMENT}-{<some number> + self.id}"
```

This way you'll be more certain to get an actual unique order ID, especially if you're using the MIT test CyberSource account. 

To test for denied orders, you will need to adjust the code to treat reason code 100 as a failure rather than a success. 

Test the management command by running `check_pending_order --all` or `check_pending_order --order <order ID>`. It should run successfully (the order should be fulfilled).

Schedule the `process_pending_order_resolutions` task. It should run successfully (the order should be fulfilled).

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -472,3 +472,64 @@ def unenroll_learner_from_order(order_id):
                     )
             except ObjectDoesNotExist:
                 pass
+
+
+def check_pending_orders_for_resolution():
+    """Checks pending orders for resolution."""
+
+    gateway = PaymentGateway.get_gateway_class(ECOMMERCE_DEFAULT_PAYMENT_GATEWAY)
+
+    pending_orders = PendingOrder.objects.filter(
+        state=PendingOrder.STATE.PENDING
+    ).values_list("reference_number")
+
+    if len(pending_orders) == 0:
+        return
+
+    log.info(f"Resolving {len(pending_orders)} orders")
+
+    results = gateway.find_and_get_transactions(pending_orders)
+
+    if len(results.keys()) == 0:
+        log.info(f"No orders found to resolve.")
+        return
+
+    for result in results:
+        payload = results[result]
+        if int(payload["reason_code"]) == 100:
+            try:
+                order = PendingOrder.objects.filter(
+                    state=PendingOrder.STATE.PENDING,
+                    reference_number=payload["req_reference_number"],
+                ).get()
+
+                order.fulfill(payload)
+
+                log.info(f"Fulfilled order {order.reference_number}.")
+            except Exception as e:
+                log.error(
+                    f"Couldn't process pending order for fulfillment {payload['req_reference_number']}: {str(e)}"
+                )
+        else:
+            try:
+                order = PendingOrder.objects.filter(
+                    state=PendingOrder.STATE.PENDING,
+                    reference_number=payload["req_reference_number"],
+                ).get()
+
+                order.cancel()
+                order.transactions.create(
+                    transaction_id=payload["transaction_id"],
+                    amount=order.total_price_paid,
+                    data=payload,
+                    reason=f"Cancelled due to processor code {payload['reason_code']}",
+                )
+                order.save()
+
+                log.info(
+                    self.style.SUCCESS(f"Cancelled order {order.reference_number}.")
+                )
+            except Exception as e:
+                log.error(
+                    f"Couldn't process pending order for cancellation {payload['req_reference_number']}: {str(e)}"
+                )
--- a/ecommerce/management/commands/check_pending_order.py
+++ b/ecommerce/management/commands/check_pending_order.py
@@ -1,0 +1,138 @@
+"""
+Looks up pending orders in CyberSource, and changes the status of the order if 
+necessary. 
+
+Occasionally, there may be a breakdown of communication between MITx Online and
+CyberSource, and orders that have gone through on the CyberSource end may not
+have their updated status reflected in MITx Online. This command will find those
+orders (or, alternatively, look at the specified order) and then will either 
+fulfill or cancel the order as necessary.
+
+"""
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management import BaseCommand
+from mitol.payment_gateway.api import PaymentGateway
+
+from ecommerce.models import PendingOrder
+from main.settings import ECOMMERCE_DEFAULT_PAYMENT_GATEWAY
+
+
+class Command(BaseCommand):
+    """
+    Looks up pending orders in CyberSource, and changes the status of the order if necessary.
+    """
+
+    help = "Looks up pending orders in CyberSource, and changes the status of the order if necessary."
+
+    def add_arguments(self, parser) -> None:
+        parser.add_argument(
+            "--order",
+            type=str,
+            help="The order ID to look for (mitxonline-prod-1).",
+            required=False,
+        )
+
+        parser.add_argument(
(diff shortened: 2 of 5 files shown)


ecommerce/api.py
```diff
@@ -472,3 +472,67 @@ def unenroll_learner_from_order(order_id):
...
+        else:
+
+            try:
+                order = PendingOrder.objects.filter(
+                    state=PendingOrder.STATE.PENDING,
+                    reference_number=payload["req_reference_number"],
+                ).get()
+
+                order.cancel()
+                order.transactions.create(
```
arslanashraf7, line 524: The logic here seems to duplicate the task and the command. I think we should introduce a common method that handles this thing and we call that function from within the loop here and in the command.

ecommerce/tasks.py
```diff
@@ -32,3 +32,10 @@ def send_order_refund_email(order_id):
     order = Order.objects.get(pk=order_id)
 
     send_ecommerce_refund_message(order)
+
+
+@app.task(acks_late=True)
+def process_pending_order_resolutions():
```
arslanashraf7, line 38: I don't think we are calling/scheduling this task from anywhere. I'm assuming that we want to schedule this task but I might be wrong. In case, we want to schedule it, is it planned for another PR or we are going to manually run this task as of now?

ecommerce/api.py
```diff
@@ -472,3 +472,67 @@ def unenroll_learner_from_order(order_id):
                     )
             except ObjectDoesNotExist:
                 pass
+
+
+def check_pending_orders_for_resolution():
```
arslanashraf7, line 477: It's fine too but since this function and the command both not only check the order status but also fix them too. Would it be a good idea to rename things e.g. `process_pending _orders` or something similar?

I'm fine if you leave it as is.

ecommerce/management/commands/check_pending_order.py
```diff
@@ -0,0 +1,138 @@
...
+    Looks up pending orders in CyberSource, and changes the status of the order if necessary.
+    """
+
+    help = "Looks up pending orders in CyberSource, and changes the status of the order if necessary."
+
+    def add_arguments(self, parser) -> None:
+        parser.add_argument(
+            "--order",
+            type=str,
+            help="The order ID to look for (mitxonline-prod-1).",
```
arslanashraf7, line 32:
```suggestion
            help="The order reference number to look for (mitxonline-prod-1).",
```

arslanashraf7 (changes requested):
Added some suggestions. 

Additionally, while testing I think I was able to reproduce the scenario as mentioned in the PR description and here are my testing results:
1. Running the command e.g. with a specific reference number e.g. `./manage.py check_pending_order --order=mitxonline-dev-83` results in success but the order stays in `Pending` state although the transaction gets created successfully:
image

2. There is a difference between the transaction data of a regular successful order and the order processed with a command in this PR. Maybe we need to process the data somehow? Anyways I noticed these:
    - Some of the fields are empty in order transaction data when fixed by the command in the PR. In my case, these fields were `signature, request_token, card_type_name`, and more.
    - I don't know if it matters but the formatting of amount fields in both the responses is not the same as well. The amount in the transaction response of this command is without decimals.

jkachel:
> There is a difference between the transaction data of a regular successful order and the order processed with a command in this PR. Maybe we need to process the data somehow? Anyways I noticed these:
> 
>     * Some of the fields are empty in order transaction data when fixed by the command in the PR. In my case, these fields were `signature, request_token, card_type_name`, and more.
> 
>     * I don't know if it matters but the formatting of amount fields in both the responses is not the same as well. The amount in the transaction response of this command is without decimals.

This is (somewhat) documented more in the [ol-django code](https://github.com/mitodl/ol-django/blob/main/src/mitol/payment_gateway/api.py#L821-L826) but the REST API delivers the transaction details in a _very_ different format from what we get via Secure Acceptance. The ol-django code re-formats it into something resembling the SA payload, and it gets pretty close but it's not a 100% match. Some of those fields (especially the card related ones) don't have counterparts in the REST API result.

ecommerce/api.py
jkachel: I refactored this; I had them separate because the command did output during the process but info logs are fine I think. This also fixes the other issue you noted - I don't think it saved the changes it made. The API call one definitely does.

ecommerce/tasks.py
jkachel: We aren't. We'll need to figure out a timeframe to do it and maybe make some changes - I think we want to leave a pending order alone if it's below an age threshold, since there's potential that the order could be actively in progress (i.e. the learner's filling out the payment form when the task runs). I don't want that discussion to be blocking the rest of it, though, especially since we need to be able to resolve a bunch of pending orders sooner rather than later.

ecommerce/api.py
jkachel: I changed it

ecommerce/tasks.py
arslanashraf7: Thanks for the clarification, I think it makes sense and I agree that this needs further discussion within a separate thread.

arslanashraf7 (approved):
Looks great 👍. Thanks for making all the changes.

Just a last piece of suggestion if it matters, Should we rename the command as well just the method where we renamed the method to `check_and_process_pending_orders_for_resolution`.

Issue #1438, opened 2023-02-17
User.name field allows adding unlimited characters, while edX name field is restricted to 255 characters


arslanashraf7:
### Steps to Reproduce

- Go to https://rc.mitxonline.mit.edu/create-account/
- In the registration form's Full name field, Add a test/name that has more than 255 characters
- You won't see enroll button on any course, And there will be no edX user against this user so you should see EdX API Auth errors.
- You should see edX errors complaining about user data being too long.
- Also notice the header in UI, the name becomes too long and the drawer hides

### Expected Behavior

- User should be created in both MITxOnline/edX and all the integration things with edX should work.

### Actual Behavior

- The user is only created in MITxOnline and he/she can't enroll in any courses.
- The UI also breaks, see screenshot below.

### Stacktrace
_Related Exceptions from Sentry_

In MITxOnline: https://mit-office-of-digital-learning.sentry.io/issues/2514390299/?project=5864687&query=is%3Aunresolved&referrer=issue-stream
In edX: https://mit-office-of-digital-learning.sentry.io/issues/3943612059/?project=5939801&query=is%3Aunresolved&referrer=issue-stream

### Related Issues
We had a related issue in xPRO and fixed it https://github.com/mitodl/mitxpro/issues/2575

### Screenshot or Screencast
<img width="1792" alt="image" src="https://user-images.githubusercontent.com/34372316/219708180-b1490329-f16c-46d5-93f9-586b5d5d6bc9.png">

**Additional Notes:**
We already have registration validation API integrated in MITxOnline, but i think we only check that for username so this issue actually bypasses that API too. Maybe we should look at that with full name's perspective too but that might not be in the scope of this ticket,

arslanashraf7:
Fixed via #1440, Deployed to Production as well.

PR #1440, opened 2023-02-20
fix: limit user full name to 255 characters

arslanashraf7:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

#1438 

#### What's this PR do?
- Limits user's full name to have a max of 255 characters
- changes user name filed from Text to Char
- Adds 255 character limit to user profile form on the frontend
- adds a migration to trim existing full names longer than 255 characters

- Setup on master and create an account that has a full name containing more than 255 characters
- Shift to this branch, build and run the containers again, and make sure the new migration runs successfully and it trims the full name in the above point to 255 characters
- Register a new account, and check that you see a validation error message for the `Full Name` when you enter more than 255 characters

- Setup on master, register an account

#### Screenshots
**Before**
image

**After**
image

image


asadali145 (approved):
LGTM!

arslanashraf7:
@pdpinch Could you also take a look at mostly the design part? When the user's full name was very long it would break the UI. The issue was noted in https://github.com/mitodl/mitxonline/issues/1438. In this PR I have limited the display limit of the name to `40 characters` at max with extra text to use an ellipsis (Screenshot attached in PR description). 
Let me know what you feel about this.

pdpinch:
> In this PR I have limited the display limit of the name to 40 characters at max with extra text to use an ellipsis 

This is a good approach. How did you decide on 40 characters?

How many characters are visible when the user sets their full name in the profile form?

arslanashraf7:
> > In this PR I have limited the display limit of the name to 40 characters at max with extra text to use an ellipsis
> 
> This is a good approach. How did you decide on 40 characters?

This limit seemed reasonable enough and looked fine. How this looks with 40 characters, There is a screenshot in the PR description. We can change this too, let me know if you want to try some other limits and i can get their screenshots for you.

> 
> How many characters are visible when the user sets their full name in the profile form?

In the profile form, all the characters would be visible, The display change only affects the name in the header.

arslanashraf7:
The name in the profile form looks like below and it's horizontally scrollable within the field, if that's what you meant in your question.

image

Issue #1425, opened 2023-02-10
Add new profile fields to Profile page


pdpinch:
As a user, I need to be able to update my profile at any time. 

The profile page form should allow editing of all profile fields from #1399 and #1424 

If it's simpler, we can display all the "professional" fields no matter what the user type is set to.

Issue #1424, opened 2023-02-10
Profile: Pop-up request for optional profile information


pdpinch:
part of #1398 

As a marketing team member, I'd like to collect optional profile information to target marketing messages better. 

As an educational researcher, I'd like to know more about our users to improve research and analysis. 

As a user, I don't want to share more information unless I know what value I'm getting for it. 

### description 

When the user take an action that demonstrates interest in the platform, like enrolling in a course, we will display a form that requests additional, optional profile information. We'll only show this once, even if the user has chosen not to fill out any of the additional fields. 

### acceptance criteria 

- [ ] pop-up form when a user visits a course from their dashboard
- [ ] include an introduction: "To help us with our educational research missions, please tell us more about yourself"
- [ ] only show the pop-up once (if they user wants to change their data, they can use the profile page) 
- [ ] basic, optional fields:
  - [ ] Gender Identity (choose one)
     - Woman
     - Man
     - Transgender
     - Non-binary / non-conforming
     - Prefer not to respond 
  - [ ] Highest level of education: choose one, [same values as xPRO](https://github.com/mitodl/mitxpro/blob/b6305e2b60bb335fd99bb7ed4905d16462419642/static/js/constants.js#L98-L108)
  - [ ] user type:
     - label: "Are you a: "
     - values: student, professional, educator, other
     - multi-select
- [ ] if user-type includes "professional" display additional fields
   - [ ] Company: (free text)
   - [ ] Job Title: (free text) 
   - [ ] Industry: same values as xPRO
   - [ ] Function: same values as xPRO
   - [ ] Company size: same values as xPRO
   - [ ] Years of experience: same values as xPRO
   - [ ] Leadership Level: same values as xPRO
- [ ] include a note at the end "You can update this information anytime in your User Profile."

pdpinch:
Please include a feature flag for these changes. I'd like to be able to show it some more people on RC, and maybe make some tweaks, before we release it to production.

Issue #1399, opened 2023-01-31
Add (minimal) fields to the user registration form


pdpinch:
part of #1398 

As a user registering, I don't want to spend a lot of time filling out information before I'm sure I really want to enroll in a course. 

As an admin, I want to gather enough information that we can create an account and do the necessary compliance checks for age, OFAC (country) and SDN (name) 

### acceptance criteria 

- [ ] add a field for age and validate that the user is at least 13 years old
- [ ] if the user is in the US, display a field for their state
  - partial addresses help us avoid false positives from SDN checks

jkachel:
Per discussion, I'm starting with:
* State field - if the user is in the United States, show a dropdown field with US states in it and make the field required
* Year of Birth - adding the field in; will also need a bit of validation to block registration if age is < 13
* Gender - adding optional field
This will be implemented in the signup flow and the profile view/edit flow. (This is not a comprehensive description of what needs to be done, just notating first steps for me.)

PR #1443, opened 2023-02-21
Fixes some issues with validation for new profile fields; adds extended profile fields

jkachel:
#### Pre-Flight checklist

- [X] Migrations
- [X] Testing
  - [X] Code is tested

Related to #1399 and #1425 and #1424

#### What's this PR do?

There were some bugs in validation and saving for users that had existing accounts:
- In Edit Profile, the Gender and Year of Birth fields weren't validating at all
- Saving with these extra fields caused an error if the user didn't already have a UserProfile record (which they won't if they're old)

In addition, this adds the extra fields and updates the tests and serializers for #1424. All the fields are nullable and I've cleaned up the UserProfileSerializer a bit too. (I wanted to get these fixes out before being fully done with the UI stuff since it's blocking the release of the minimal fields from #1399.)

New user registration should work as it has.

As an existing user that _does not_ have an existing UserProfile record, edit and save your profile. You should get a validation error if you leave the Year of Birth field blank, and the data should save successfully. 

As an existing user that _did have_ an existing UserProfile record with a blank Year of Birth field, test editing the profile. Validation should work as expected. Your UserProfile record should have the Year of Birth field set to `3000` (which should ensure it fails validation).


frontend/public/src/components/forms/ProfileFormFields.js
```diff
@@ -222,12 +234,73 @@ export const LegalAddressFields = ({
...
+        </label>
+        <Field
+          component="select"
+          name="user_profile.year_of_birth"
+          id="user_profile.year_of_birth"
+          className="form-control"
+          aria-describedby="user_profile.year_of_birth_error"
+        >
+          <option value="f">-----</option>
+          {reverse(range(seedYear - 120, seedYear - 14)).map((year, i) => (
```
pdpinch, line 291: Currently, on RC, this renders the range 1903 - 2008. 

image

I think 1903 - 2009 would be more correct, since someone born 12/31/2009 would be over 13 years old today. 

I don't know how the range function handles boundary conditions, but maybe this should be:

```suggestion
          {reverse(range(seedYear - 120, seedYear - 13)).map((year, i) => (
```

users/migrations/0020_add_addl_profile_fields.py
```diff
@@ -0,0 +1,151 @@
...
+                    ("Other education", "Other education"),
+                ],
+                default="",
+                max_length=60,
+                null=True,
+            ),
+        ),
+        migrations.AddField(
+            model_name="userprofile",
+            name="industry",
```
collinpreston, line 75: Should this have options like xpro?

collinpreston: This question also applies to job_function and leadership_level.

users/utils.py
```diff
@@ -134,3 +138,21 @@ def ensure_active_user(user):
...
+        12,
+        31,
+        hour=23,
+        minute=59,
+        second=59,
+        tzinfo=pytz.timezone(settings.TIME_ZONE),
+    )
+
+    return math.floor(
+        (datetime.now(tz=pytz.timezone(settings.TIME_ZONE)) - approx_dob).days / 365
```
collinpreston, line 157: Can this instead be
```suggestion
        (datetime.now(tz=pytz.timezone(settings.TIME_ZONE)).year - year)
```

collinpreston (approved):
Looks good and worked when testing.  Left a couple of questions.

users/utils.py
jkachel: Simplified this per suggestion.

PR #1444, opened 2023-02-21
Updates the command to include the enrollment mode when retrying enrollments

### LINKED ISSUE #1439: retry_edx_enrollment doesn't sync the enrollment mode
This was discovered while working through https://github.com/mitodl/hq/issues/742 - in this case, the learner's mode was set to `verified` in MITx Online but that change wasn't reflected in edX after forcing the enrollment resync using `retry_edx_enrollment`. This should retry the enrollment with the learner's enrollment mode and at current it does not consider that at all. 

### Steps to Reproduce

1. Enroll a learner in a course as `audit`. 
2. Manually upgrade them to `verified` using the Django Admin.
3. Attempt to push this change to edX by running `retry_edx_enrollment`.

### Expected Behavior

The learner's enrollment in edX matches what's in MITx Online.

### Actual Behavior

The edX enrollment will remain at `audit`.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes #1439

#### What's this PR do?

Updates the call to `enroll_in_edx_course_runs` to include the enrollment mode when it attempts to retry the enrollment, so the user's current status (as stored by MITx Online) is reflected in edX.

1. Enroll a user in a course in the audit mode.
2. Manually change their enrollment to verified in Django Admin in MITx Online.
3. Run `retry_edx_enrollment` for the user, and specify the `-f` flag to force it.
4. Check the resulting enrollment in edX. It should be verified.

### DIFF
--- a/openedx/management/commands/retry_edx_enrollment.py
+++ b/openedx/management/commands/retry_edx_enrollment.py
@@ -66,7 +66,9 @@ class Command(BaseCommand):
             user = enrollment.user
             course_run = enrollment.run
             try:
-                enroll_in_edx_course_runs(user, [course_run])
+                enroll_in_edx_course_runs(
+                    user, [course_run], mode=enrollment.enrollment_mode
+                )
             except Exception as exc:  # pylint: disable=broad-except
                 self.stderr.write(self.style.ERROR(f"{str(exc)}"))
             else:


collinpreston (approved):
Looks good

PR #1450, opened 2023-02-24
Bump root python image version to 3.9.16

jkachel:

n/a

#### What's this PR do?

Bumps the version of the `python` image referenced in the Dockerfile to 3.9.16. I was having some issues with the prior 3.9.6 image not working (across a couple x86 Linux machines); bumping it to 3.9.16 fixed it and probably wouldn't hurt to have this use the current 3.9 release anyway. This only affects local development.

Rebuild your image and run the standard test suites. Tests should all pass and everything should work as expected.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9.6 as base
+FROM python:3.9.16 as base
 LABEL maintainer "ODL DevOps <mitx-devops@mit.edu>"
 
 # Add package files, install updated node and pip


arslanashraf7 (commented):
I was looking through this and I noticed that I probably forgot to click the submit review or it probably didn't work at that time.

After @pdpinch 's comment, My take is that we can now close this PR for 2 reasons:
1) runtime.txt was removed as part of [poetry integration](https://github.com/mitodl/mitxonline/pull/1766/files#diff-550a968ef70ab0363efff4b7bdfb9737fa8ccfd118ff5fff6540c8cff2ba26b5)
2) The Python version is already on [3.9.18](https://github.com/mitodl/mitxonline/blob/main/Dockerfile#L1). Although, We can try considering Python upgrade in another PR at some other time so that we' don't stay far behind the latest version 3.12.x as of now.

PR #1456, opened 2023-03-01
Fixing a call to `set_rollback` that was incorrect; adding a test

### LINKED ISSUE #1432: When PaymentGateway does not return success raise an exception
Related Sentry error: 
https://mit-office-of-digital-learning.sentry.io/issues/3915111706/?environment=production&project=5864687&query=is%3Aunresolved+set_rollback%28%29&referrer=issue-stream&statsPeriod=14d

Currently when PaymentGateway does not raise an exception we just log an error and rollback the refund manually: [here](https://github.com/mitodl/mitxonline/blob/93043bf4f2962cfeb3c34c30049ac97684982cdb/ecommerce/api.py#L443)

Instead we want to throw an error and show the error message.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes #1432

#### What's this PR do?

Corrects a call to `set_rollback`; sets it to just roll back the transaction when a refund attempt is processed and the payment gateway returns an error. Raises an exception in this case as well, rather than just returning false. 

Automated tests should pass.

You would have to mock out the payment gateway API to return an error (as this hits a CyberSource API to do its job) but if you have a transaction that has an error and was returning `Error: set_rollback() missing 1 required positional argument: 'rollback'`, it shouldn't do that anymore. 

There are transactions on prod that have this - they're regular credit card transactions, not PayPal ones, so not sure what the actual error is (because the syntax error ends up taking precedence).

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -442,9 +442,10 @@ def refund_order(*, order_id: int = None, reference_number: str = None, **kwargs
                 )
                 # PaymentGateway didn't raise an exception and instead gave a Response but the response status was not
                 # success so we manually rollback the transaction in this case.
-                transaction.set_rollback()
-                # Just return here if the refund failed, no matter if unenroll was requested or not
-                return False
+                transaction.rollback()
+                raise Exception(
+                    f"Payment gateway returned an error: {response.message}"
+                )
 
         except RefundDuplicateException:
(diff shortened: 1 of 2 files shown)


collinpreston (approved):
Looks good to me, nice

PR #1462, opened 2023-03-03
Moves Highest Level of Education field up

### LINKED ISSUE #1457: Profile: Highest Level of Education should be displayed on all profile forms
Currently, "Highest Level of Education" is only displayed for users who select the "Professional" user type. 

"Highest Level of Education" should be displayed to all users. 

It's probably best to move it above the user type selection.

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

Closes #1457

#### What's this PR do?

Two things:
* Moves the Highest Level of Education field above the "Are you a..." field in the extended profile fields (and thus moves it up out of the conditionally-displayed fields)
* In the edit profile form's onSubmit handler, changes the page refresh method to use window.location.refresh rather than pushing a new browser history entry. (This should prevent people from hitting back and getting the profile page over and over again, and also makes it so that the feature flag keeps working if you specify it in the URL.) 

These fields are hidden behind a feature flag; you'll either need to set `FEATURE_ENABLE_ADDL_PROFILE_FIELDS` or append `?enable_addl_profile_fields` to the URL.

1. Load the Profile page.
2. You should see the Highest Level of Education field below the Gender and Year of Birth fields.
3. Selecting Professional should have no effect on the Highest Level of Education field. 
4. If you're using the URL flag, saving should not clear the flag.

image

### DIFF
--- a/frontend/public/src/components/forms/ProfileFormFields.js
+++ b/frontend/public/src/components/forms/ProfileFormFields.js
@@ -416,6 +416,31 @@ export const ProfileFields = () => (
 export const AddlProfileFields = ({ values }: AddlProfileFieldsProps) => (
   <React.Fragment>
     <div className="form-group">
+      <div className="row">
+        <div className="col">
+          <label
+            htmlFor="user_profile.highest_education"
+            className="font-weight-bold"
+          >
+            Highest Level of Education
+          </label>
+          <Field
+            component="select"
+            name="user_profile.highest_education"
+            id="user_profile.highest_education"
+            className="form-control"
+          >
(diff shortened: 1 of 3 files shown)


annagav (changes requested):
It looks like I can select multiple:
Screen Shot 2023-03-03 at 3 01 44 PM

jkachel:
Yes, we allow you to choose any of those options. There's a set of fields that only appears if you've chosen Professional, though, and the Highest Level of Education field was incorrectly in that group.

PR #1463, opened 2023-03-03
Changes refund_order to let exceptions bubble up, removes duplicate as a successful state

### LINKED ISSUE #1461: Refund requests resolve with no error if there's a duplicate
Noticed when working on https://github.com/mitodl/hq/issues/616: in that ticket, there appears to be some sort of state issue on the CyberSource side with the refund in question. There does appear to be a refund in progress or errored out on that side, though, so re-running the refund from MITx Online results in it responding with a duplicate request error. This is explicitly caught and ignored by the `refund_order` API in MITx Online, so the order isn't ever touched and the Google sheet doesn't show anything wrong. I think it'd be better if it raised an exception if it encounters an error of any kind from CyberSource so we can see when things don't process as we expect them to.

### Steps to Reproduce

This is really hard to reproduce on a local instance, since it'd require that you have an order you can pull that's in an error state in CyberSource, but it can be replicated on production by attempting to re-process line 41 in the Google sheet.

### Expected Behavior

The row is marked as an error in the Google sheet. 

### Actual Behavior

The row is marked as completed in the Google sheet, as it would if the refund had processed successfully. The order remains in the Fulfilled state.

### Related Issues

#1432

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested
 
Closes #1461

#### What's this PR do?

Changes the behavior of `refund_order` to prevent it from trying to catch exceptions that happen during the refund process. 

Changes the list of refund success states to remove duplicate as as successful state.

While this is really meant to be run from within the Google sheets processing code, you can call `refund_order` from a Django shell if you'd rather not set all that up. 

1. Complete a transaction in the system.
2. Refund your transaction. 
3. Attempt to process the refund transaction again. An exception should be raised - CyberSource won't let you refund a transaction twice. 
4. If you do have Google Sheets integration set up, you should see an error message in the refunds sheet. 

I think we regarded "duplicate" as a success state to allow the Google automation the ability to re-process rows that were completed. (In other words, if it processed row 12 successfully in a prior run, when it processes row 12 in a future run it should not report errors just because it's already been done.) However, in practice, we keep track of what rows have been processed so instead I feel like it's better that we see that an error has occurred - it won't re-process rows (unless you do some finagling to force that to happen).

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -421,35 +421,27 @@ def refund_order(*, order_id: int = None, reference_number: str = None, **kwargs
             ECOMMERCE_DEFAULT_PAYMENT_GATEWAY, transaction_dict
         )
 
-        try:
-            response = PaymentGateway.start_refund(
-                ECOMMERCE_DEFAULT_PAYMENT_GATEWAY,
-                refund_gateway_request,
+        response = PaymentGateway.start_refund(
+            ECOMMERCE_DEFAULT_PAYMENT_GATEWAY,
+            refund_gateway_request,
+        )
+
+        if response.state in REFUND_SUCCESS_STATES:
+            # Record refund transaction with PaymentGateway's refund response
+            order.refund(
+                api_response_data=response.response_data,
+                amount=transaction_dict["req_amount"],
+                reason=refund_reason,
             )
-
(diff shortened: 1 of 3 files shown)


annagav (approved):
LGTM

Issue #1455, opened 2023-02-28
Block search crawlers on CI, RC


pdpinch:
As a google search user, it's confusing to see search results from MicroMasters RC.

on MicroMasters, we added a library django-robots (https://github.com/mitodl/micromasters/pull/5280) which made it easy to customize the robots.txt file on each instance of the app. 

Acceptance Criteria:
 - [ ] Add django-robots
 - [ ] When deployed to CI and RC, block all robots
 - [ ] When deployed to production, confirm that the expected pages are still crawlable

collinpreston:
Hi @pdpinch , so when researching this approach further I found a few things out:
1. `mitxonline` is already using `django-robots`
2. Based on Google's documentation (ref: https://developers.google.com/search/docs/crawling-indexing/robots/intro#what-is-a-robots.txt-file-used-for) the best way to block robots is not actually by hiding or modifying the robot.txt file but actually with a `<meta>` tag containing a reference to `noindex`.  This ensures that search engines don't index the pages even if other pages or sites link to the pages.

pdpinch:
Is it set up right on CI and RC?

https://rc.mitxonline.mit.edu/robots.txt 

returns:

```
User-agent: *
Disallow:
```

Shouldn't it be 

```
User-agent: *
Disallow: /
```

I could be wrong. 

As for a metadata tag, I'm not sure how to apply that to every page on CI and RC and none on production. Robots.txt seems simpler when we want to block the whole site.

PR #1468, opened 2023-03-07
1455: Adds templatetag for noindex in non-prod

collinpreston:
#### Pre-Flight checklist

- [x] Testing

https://github.com/mitodl/mitxonline/issues/1455

#### What's this PR do?
Blocks search engines from indexing `mitxonline` pages in non-production environments using a `<meta>` tag with the `noindex` value described here: https://developers.google.com/search/docs/crawling-indexing/block-indexing.  

The `<meta>` tag is included in the header of each `mitxonline` page if the value of the `MITX_ONLINE_ENVIRONMENT` environment variable does not equal 'prod' or 'production'.  The default value for this environment variable is 'dev'

1. Pull down this branch.
2. If `MITX_ONLINE_ENVIRONMENT` has not been added to your `.env` file, then start `mitxonline`.  Otherwise, set `MITX_ONLINE_ENVIRONMENT` equal to `dev` in your `.env`.
3. Visit any page in your local instance of `mitxonline` and view the page source.  Ensure that you see `<meta name="robots" content="noindex">` in the header of the page when `MITX_ONLINE_ENVIRONMENT` equal to `dev`.
4. Stop `mitxonline`.  Set `MITX_ONLINE_ENVIRONMENT` equal to `prod` in your `.env` file.
5. Start `mitxonline`
6. Visit any page and view the source code.  Ensure that you do not see `<meta name="robots" content="noindex">` in the header of the page when `MITX_ONLINE_ENVIRONMENT` is equal to `prod`.

When this change is deployed to QA or Production, we can also test this functionality using the steps described here: https://developers.google.com/search/docs/crawling-indexing/block-indexing#debugging-noindex-issues

### DIFF
--- a/main/templatetags/noindex_meta.py
+++ b/main/templatetags/noindex_meta.py
@@ -12,5 +12,5 @@ def noindex_meta():
     return (
         mark_safe("""<meta name="robots" content="noindex">""")
         if settings.ENVIRONMENT not in ("production", "prod")
-        else None
+        else ""
     )
--- a/main/templatetags/noindex_meta.py
+++ b/main/templatetags/noindex_meta.py
@@ -1,0 +1,16 @@
+"""Blocks search engine indexing for non-production environments."""
+from django import template
+from django.utils.safestring import mark_safe
+from django.conf import settings
+
+register = template.Library()
+
+
+@register.simple_tag()
+def noindex_meta():
+    """Adds in noindex for non-production environments."""
+    return (
+        mark_safe("""<meta name="robots" content="noindex">""")
+        if settings.ENVIRONMENT not in ("production", "prod")
+        else None
+    )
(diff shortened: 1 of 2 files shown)


main/templatetags/noindex_meta.py
```diff
@@ -0,0 +1,16 @@
...
+register = template.Library()
+
+
+@register.simple_tag()
+def noindex_meta():
+    """Adds in noindex for non-production environments."""
+    return (
+        mark_safe("""<meta name="robots" content="noindex">""")
+        if settings.ENVIRONMENT not in ("production", "prod")
+        else None
```
asadali145, line 15: It works fine when we return the meta tag but has some issues when we return `None`. It adds `None` as a string and moves it with all the meta tags and the title below it to the `<body>`. It works fine if we return an empty string instead of None.
Screen Shot 2023-03-08 at 2 33 35 PM

asadali145 (changes requested):

main/templatetags/noindex_meta.py
collinpreston: Nice catch!

asadali145 (approved):
LGTM!

PR #1470, opened 2023-03-07
Adds friendlier error message when State is unset

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

https://github.com/mitodl/hq/issues/886

#### What's this PR do?

Adds a `typeError` to the validation for the State field that says "State is a required field" rather than the default yup error (which is very programmery). 

1. Log in with an account that either doesn't have a country/state set or has one set to a country that doesn't have states (so not US or Canada)
2. Set the country to US or Canada
3. Open the State field, but don't select anything. Instead, click to a different field.
4. Validation should run and you should see the "State is a required field" error message.

https://share.cleanshot.com/167R3ykMmtl8vrKPS5lt

### DIFF
--- a/frontend/public/src/components/forms/ProfileFormFields.js
+++ b/frontend/public/src/components/forms/ProfileFormFields.js
@@ -56,8 +56,11 @@ export const legalAddressValidation = yup.object().shape({
       .string()
       .label("State")
       .when("country", {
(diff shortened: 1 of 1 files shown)


asadali145 (approved):
LGTM!

PR #1471, opened 2023-03-08
Declining an order clears discounts associated with the order

### LINKED ISSUE #1467: Denied transactions should clear discount redemptions
Noticed in https://github.com/mitodl/hq/issues/959 - the learner here had redeemed a discount code, and the resulting transaction was declined. They were unable to re-use the code. Declined is a terminal state for orders, so when they hit this state, any discounts redemptions that existed for the order should be cleared so that the learner can resolve the issue and use the code again.

### Steps to Reproduce

1. Redeem a discount for an order and follow the process to the extent that you get redirect to CyberSource to complete the order. (You do not need to actually complete the order.)
2. Decline the order through the Django shell.

### Expected Behavior

You should be able to use the code again.

### Actual Behavior

You're prevented from using the discount code again.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes #1467 

#### What's this PR do?

Adds some code to remove discount redemptions when a pending order is declined. 

1. Start an order that has a discount attached to it and progress into the CyberSource part of it. (Don't complete the order). 
2. From a Django shell, load the order and call `decline` on it.
3. The order should change status to Declined, and the discount redemptions should be cleared for the order.

### DIFF
--- a/ecommerce/api_test.py
+++ b/ecommerce/api_test.py
@@ -1,14 +1,14 @@
 """Tests for Ecommerce api"""
 
-import json
 import random
 import uuid
+from datetime import datetime
 
 import pytest
+import pytz
 import reversion
 from CyberSource.rest import ApiException
 from django.conf import settings
-from django.http import HttpRequest
 from django.urls import reverse
 from mitol.payment_gateway.api import ProcessorResponse
 from reversion.models import Version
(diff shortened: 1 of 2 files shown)


collinpreston (approved):
Looks good and tested well!

PR #1474, opened 2023-03-08
Allows verified learners the ability to unenroll; adjusts flow for refunds

### LINKED ISSUE #1466: Allow users to unenroll from a course at any time
We've received many customer supports tickets for users who want more control over unenrolling and refunds, which could be self-service if: 

1) We let users unenroll themselves whenever they want to 
2) When refunding a user, we don't automatically unenroll them

#### Designs and Mockups

No new design is needed; the existing Unenroll and Refund UX is adequate.

#### Acceptance Criteria:

- [ ] Allow enrolled users to unenroll at any time
- [ ] Update the refund workflow so that users _**are not unenrolled**_ when they get a refund. 
   - Customer Support will instruct users to unenroll, if they choose to do so. 
   - refunded learners should still have there enrollment mode/course mode changed from "verified" to "audit" 
- [ ] Confirm that unenrolled users can still request a refund

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots (but no mobile since it's just copy changes)
- [X] Testing
  - [X] Code is tested

Closes #1466 

#### What's this PR do?

- If the learner's payment is refunded, we now bump the enrollment to audit instead of completely unenrolling them.
- Learners can now unenroll from certficate track (verified) enrollments at will (but this won't result in a refund).

Learners unenrolling from the certificate track will unenroll from the entire course, _and_ this will clear their `PaidCourseRecord` entry for the course so they can enroll again. (The checkout process checks the `PaidCourseRecord` table for an existing purchase of the course to prevent duplicate purchases.)

Testing unenrollments:
1. Enroll in and upgrade your enrollment for a course. 
2. From the Dashboard, select Unenroll from the menu.
3. You should get a dialog asking you to confirm. The dialog should note that unenrolling will not refund your money. 
4. Continuing through the dialog should unenroll you from the course. 
5. You should be able to re-buy into the course. 

Testing refunds:
For this to work, you'll need to have REST API credentials for your CyberSource integration set up. 
1. Enroll in a course and upgrade the enrollment.
2. Refund the transaction. You can do this either in a Django shell by calling `ecommerce.api.refund_order` or you can process a row in the Google sheet if you have that integration set up. 
3. The refund should be successful and the enrollment should be converted to an `audit` enrollment.
4. The dashboard should reflect the change.

The text in the unenroll confirmation dialog was amended a bit to include refund text:
image

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -277,7 +277,11 @@ def create_run_enrollments(
             enrollment, created = CourseRunEnrollment.all_objects.get_or_create(
                 user=user,
                 run=run,
-                defaults=dict(edx_enrolled=edx_request_success, enrollment_mode=mode),
+                defaults=dict(
+                    change_status=None,
+                    edx_enrolled=edx_request_success,
+                    enrollment_mode=mode,
+                ),
             )
 
             if run.course.program is not None:
--- a/courses/models.py
+++ b/courses/models.py
@@ -990,6 +990,17 @@ class CourseRunEnrollment(EnrollmentModel):
         """
         return cls.objects.filter(user=user, run__course__program=program)
 
+    def deactivate_and_save(self, change_status, no_user=False):
+        """
+        For course run enrollments, we need to clear any PaidCourseRun records
+        for this enrollment (if any) so they can re-enroll later.
+        """
+        from courses.tasks import clear_unenrolled_paid_course_run
+
+        clear_unenrolled_paid_course_run.delay(self.id)
+
+        return super().deactivate_and_save(change_status, no_user)
+
     def to_dict(self):
         return {**super().to_dict(), "text_id": self.run.courseware_id}
 
--- a/courses/tasks.py
+++ b/courses/tasks.py
@@ -11,6 +11,7 @@ from courses.models import (
     CourseRun,
     CourseRunEnrollment,
     LearnerProgramRecordShare,
+    PaidCourseRun,
     Program,
 )
 from main.celery import app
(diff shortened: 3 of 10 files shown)


collinpreston (commented):
Code looks good and I was able to verify the steps you provided.  A couple things I was curious about:

1. I did the following: enrolled in a course as a verified learner.  Unenrolled from the course and then ran `refund_order()`.  I then re-enrolled into the course as an audit learner.  This left me with a `course_run_enrollement` record indicating an enrollment_mode of `audit` as well as a status of `unenrolled`.  Will the `unenrolled` status have an impact elsewhere?
2. This is probably more of a product consideration, but if a user has unenrolled as a verified learner, should that user be able to re-enroll as a verified learner without paying a second time?

jkachel:
> I did the following: enrolled in a course as a verified learner. Unenrolled from the course and then ran refund_order(). I then re-enrolled into the course as an audit learner. This left me with a course_run_enrollement record indicating an enrollment_mode of audit as well as a status of unenrolled. Will the unenrolled status have an impact elsewhere?

https://github.com/mitodl/mitxonline/pull/1474/commits/2c22e792f28d85fcbb036aa1fddc153fa56454de adds some code to clear the change_status flag when re-enrolling (good catch, it should have been doing that)

> This is probably more of a product consideration, but if a user has unenrolled as a verified learner, should that user be able to re-enroll as a verified learner without paying a second time?

Agree - this might be something @pdpinch should weigh in on but my thought is that if they're re-buying in they probably need to talk to CS about their circumstances. (In other words, I figure folks that are unenrolling and expecting a refund are going to talk to CS about doing that rather than just flipping the switch themselves.)

collinpreston (approved):
Looks good!

Issue #1224, opened 2022-11-17
Can not create a program from django admin


annagav:
### Steps to Reproduce

Go to /admin/courses/program/add/ and try to create a program

### Expected Behavior

successfuly create a program

### Actual Behavior
Throws an error 
```
ValueError at /admin/courses/program/add/
save() prohibited to prevent data loss due to unsaved related object 'program'.
```

pdpinch:
@rhysyngsun @annagav is this still an issue?

jkachel:
Can confirm that this is still a problem - I've been creating them in a Django shell as a workaround.

pdpinch:
@rhysyngsun I'm going to take you off this, so someone else can work on it. Let me know if that's an issue. 

I'd like to be able to make a test program on RC soon-ish, for DEDP 2, the US version.

annagav:
I think this was fixed in https://github.com/mitodl/mitxonline/pull/1477

asadali145:
Yeah, Sorry I missed assigning myself as I came to know about this ticket after my PR.

asadali145:
Fixed with the linked PR.

PR #1477, opened 2023-03-10
fix: Fix program admin to add a new program

asadali145:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

Fixes https://mit-office-of-digital-learning.sentry.io/issues/3777726522/events/?project=5864687&referrer=issue-stream
https://github.com/mitodl/mitxonline/issues/1224

#### What's this PR do?
Fixes program creation from Django Admin.

The program admin was throwing 500 when we try to create a new program. The new program instance was not saved as the model admin does not commit the object while saving the form and we were trying to create program requirements. Stacktrace is attached
Screen Shot 2023-03-10 at 3 08 52 PM

- Visit /admin/courses/program/
- Add a new program with some requirements
- Program should be created

`ModelAdmin` has a `save_form` method which saves the form to get the unsaved new program instance. It sets `commit=False`. So, when we try to create requirements, It raised an exception that the related object `program` is unsaved.

### DIFF
--- a/courses/forms.py
+++ b/courses/forms.py
@@ -220,7 +220,15 @@ class ProgramAdminForm(ModelForm):
     def save(self, commit=True):
         """Save requirements"""
         program = super().save(commit=commit)
+        transaction.on_commit(self.save_requirements)
+        return program
+
+    def save_requirements(self):
+        """
+        Save related program requirements.
+        """
         with transaction.atomic():
+            program = self.instance
             root = program.get_requirements_root(for_update=True)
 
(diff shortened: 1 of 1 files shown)


collinpreston:
Hey @asadali145 , I think you can link this PR to this issue: https://github.com/mitodl/mitxonline/issues/1224

asadali145:
@collinpreston Thanks for that.

collinpreston (approved):
Nice clean solution!

PR #1496, opened 2023-03-17
Resolve issue when repairing user's edx synchronised records

collinpreston:
#### Pre-Flight checklist

- [x] Migrations
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/1493

#### What's this PR do?
An issue occurs when we attempt to repair mitxonline User's edx OAuth accounts.  This issue occurs when a user meets the following criteria:
- User has an `OpenEdxUser` record in mitxonline.
- No user records exist for the user in Edx.

1. Have mitxonline and devstack up and running locally, and connected.
2. You will need to run a `makemigrations` in mitxonline.
3. Create a user through the mitxonline UI.
4. Delete the user's record, created during step 2, from devstack: http://localhost:18000/admin/auth/user/
5. From the Django shell, run:
```
from users.models import User
from openedx import api
user = User.objects.get(id=<USER_ID_FROM_STEP_2>)
result = api.repair_faulty_edx_user(user)
```

6. Verify that a user record is created in devstack: http://localhost:18000/admin/auth/user/
7. Verify that the OpenEdxUser record in mitxonline is marked as synced: http://mitxonline.odl.local:8013/admin/openedx/openedxuser/


jkachel (changes requested):
This seems to work if I remove the `OpenEdxUser` record but it doesn't if it's still there. Line 314 in `openedx/api.py` looks for that record so it seems like it it's skipping the call to regen the auth tokens?

collinpreston:
> This seems to work if I remove the `OpenEdxUser` record but it doesn't if it's still there. Line 314 in `openedx/api.py` looks for that record so it seems like it it's skipping the call to regen the auth tokens?

Nice find.  I updated the PR to now make an API call after retrieving the existing token: `client.user_info.get_user_info()` .  An Exception should be thrown in the `try` if the user either doesn't have a valid `OpenEdxApiAuth` record and edx responds to the creation of a new `OpenEdxApiAuth` record with an error since the user does not exist, or if the user has a valid `OpenEdxApiAuth` then edx will respond to a call to `get_user_info()` with an Exception.

jkachel (approved):
Tried repairing in the way that I was and was successful. LGTM! 👍

PR #1499, opened 2023-03-21
Requests additional information from the learner when they register.

### LINKED ISSUE #1494: Request profile info earlier in registration flow
As a marketer, I'd like to get more profile information sooner, when the user registers and before they have enrolled. Knowing users' profiles allows me to target them for enrollment messaging. 

#### Acceptance Criteria:

- [ ] Add a second screen to the registration flow, similar to xPRO. 
- [ ] On the second screen, include the message about collecting data for research purposes
- [ ] Level of Education, the Type checkboxes and the conditional fields should appear on the 2nd screen. All should be optional
- [ ] When a user clicks into a course from their dashboard, display a pop-up if they haven't responded to level of education and type. Now these fields should be required. 
- [ ] Only show the pop-up once to users.

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Testing
  - [X] Code is tested

Closes #1494

#### What's this PR do?

Updates the registration workflow to include a step for the learner to provide additional demographic information about themselves. This is the same info that we request when they navigate to a course for the first time and is optional during the registration flow. 

This also makes a small change to make the normal profile form set the additional detail flag when saving there (otherwise learners might get prompted to supply data that they've already supplied). 

As this flow involves page transitions, you'll need to set the `FEATURE_ENABLE_ADDL_PROFILE_FIELDS` flag in your `.env` file to test - using the URL flag won't be sufficient.

1. Register a new account. 
2. After confirming your email, fill out the new user form.
3. You should be see a request for more information.
   1. Submitting the form without supplying data should work. In this case, the additional field flag in the user's profile should be clear.
   2. Submitting the form with data should also work. In this case, the additional field flag should be set.
   3. In both cases, you should be redirected to the dashboard.
4. If you _did not_ supply demo data, you should see the modal popup requesting demo data when you try to navigate into a course.  

image
image


collinpreston (approved):
Nice, tested with screen reader as well.

jkachel:
Thanks - rerequesting review because I missed something. If you get the popup request, we should require the Highest Level of Education and "Are you a" fields. This is in commit https://github.com/mitodl/mitxonline/pull/1499/commits/ef559b27930e8a561b163b984eaf6061a241ae68 .

collinpreston (commented):
Looks good.  Just two comments:

1. Less important: I noticed that if I am using keyboard controls on the additional details modal, I receive the validation error message after I move from the first option to the next option.  I think this is similar to https://github.com/mitodl/hq/issues/886 where we probably want to display the error message only after the user has submitted the form.

The suggestions here worked when messing around on that prior issue: https://stackoverflow.com/questions/56742376/react-formik-trigger-validation-only-on-form-submit

2. If I click on the course title from the dashboard (to navigate to the course on edx) I am shown the modal as expected with the required fields.  If I close that modal and then click the course title again, I am not shown the modal.  My user will not have data for the required fields but will have `Addl field flag` set to True.  Is this expected?

jkachel:
These last two things should be fixed in this commit: https://github.com/mitodl/mitxonline/pull/1499/commits/3c874107bcafccfa95b1c185203c83eef66b5a78 (totally forgot about the "Enrolled √" button)

PR #1503, opened 2023-03-23
Removing unnecessary alt texts from images

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes mitodl/hq#653
Closes mitodl/hq#659

#### What's this PR do?

Removes the alt text from the course images in these locations:
- Course detail page (was course title)
- Cart page (was course title)
- Dashboard (was "Preview image")

It does not remove the alt text from the featured item cards on the homepage as they were already an empty string.

Removes the alt text from the MIT logo on the certificate page. 

Load any of the pages in question: course detail page, cart, dashboard, or certificate. The affected images should have an empty string as the alt text and screen readers should not attempt to describe the image.

### DIFF
--- a/cms/templates/certificate_page.html
+++ b/cms/templates/certificate_page.html
@@ -62,7 +62,7 @@
                     </div>
                     <span class="institute-text">Massachusetts Institute of Technology</span>
                     <div class="institute-logo">
-                        <img src="{% static 'images/certificates/certificate-logo.png' %}" alt="MIT">
+                        <img src="{% static 'images/certificates/certificate-logo.png' %}" alt="">
                     </div>
                     <span class="certify-text">This is to certify that</span>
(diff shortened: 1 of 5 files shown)


collinpreston (approved):
Looks good!

jkachel:
Got it in https://github.com/mitodl/mitxonline/pull/1503/commits/3fb2c2b90a4ef5a026a04a115d5706b343a8dce8

FWIW we do have a Title in the Featured Image section in the CMS so that may be something we can use in future - a potential issue with this (outside of getting the title, which is in the model but I don't know that we expose it anywhere) is having titles in there that are worthwhile. For example, there's a bunch of images that get used for Featured Images for the newer math courses on prod and they all have terrible titles like "5-brown_2.2". This is a training issue, though.

Issue #1522, opened 2023-04-03
Remove ErrorMessage components from 'required' form fields


collinpreston:
### Steps to Reproduce

Our intention is to only display error messages to users after they have submitted a form, as opposed to displaying error messages for a field once a user moves focus off the field (example; Clicks on the field, then clicks on another field).  Because of this design choice, there is no need to include formik `ErrorMessage` components for these required fields.

Reference: https://formik.org/docs/api/errormessage

PR #1526, opened 2023-04-03
1522: Remove instances of ErrorMessage for required fields

collinpreston:
#### Pre-Flight checklist
- [x] Testing

https://github.com/mitodl/mitxonline/issues/1522

#### What's this PR do?
In order to improve accessibility, we have added the `required` attribute to required form fields.  Because of this, users are not able to submit forms if any of these `required` fields are blank.  This means that many of the formik `<ErrorMessage>` components never being accessible when the field's validation only checks that a value is defined for that field.

This PR also:
- Ensures forms are only validated when the user presses the submit button
- Adds more `autoComplete` where possible.

Visit the following forms within the application and verify that you cannot submit the form if any of the required fields are blank.  Verify that any validation error messages, unrelated to the field being required, are displayed (example: email format invalid).

Change email form:
- Validations are only displayed when the form is submitted.
- Ensure you cannot submit the form with an empty email field.
- Ensure a validation tool-tip is shown when submitting the form with an invalid email format.  The form should not allow the user to submit.
- Ensure validation message is shown if the user uses their existing email for the email field upon submitting the form.
- Ensure you cannot submit the form with an empty password field.
- Ensure validation message is shown if the password is less than 8 characters upon submitting the form.
- The email field should autocomplete if possible.
- The current password field should autocomplete if possible.

Change password form:
- Validations are only displayed when the form is submitted.
- Ensure you cannot submit the form with any empty fields.
- Ensure validation message is shown if the new password is less than 8 characters upon submitted the form.
- Ensure validation message is shown if the new password and confirmation password fields don't match upon submitting the form.
- The current password field should autocomplete if possible.
- The new password fields should autocomplete if possible.

Email form during sign in:
- Validations are only displayed when the form is submitted.
- Ensure you cannot submit the form with an empty email field.
- Ensure a validation tool-tip is shown when submitting the form with an invalid email format.  The form should not allow the user to submit.

Password form during sign in:
- Validations are only displayed when the form is submitted.
- Ensure you cannot submit the form with an empty password field.
- All existing validation messages are shown when the form is submitted
- The password field should autocomplete if possible.

Profile form:
- Validations are only displayed when the form is submitted.
- Ensure you cannot submit the form with an empty required field.
- Ensure that you cannot submit the form with an empty "state" field if the country is "Unite States". 
- The date of birth field should autocomplete if possible.
- Company name field should autocomplete if possible.
- Title field should autocomplete if possible.
- All other validations should perform as expected.

Register email form:
- Validations are only displayed when the form is submitted.
- Ensure you cannot submit the form with an empty required field.
- Ensure a validation tool-tip is shown when submitting the form with an invalid email format.  The form should not allow the user to submit.

Register password form:
- New password field should autocomplete if possible.

Additional profile information modal:
- Validations are only displayed when the form is submitted. 
- Ensure you cannot submit the form with an empty Highest level of education field.


collinpreston:
> Looks pretty good but found a weird spot in the change email form. If I pop the same email error, fix that, and then try to submit with another error, the same email error doesn't clear. See screengrab: https://share.cleanshot.com/rdTLs4jfSMs4sQjbDQMW
> 
> Everything else seemed to work as expected (and the profile State field pops the error right if you select Canada as the country too).

Good call.  That is occurring since your second validation error (empty password field) is not allowing you to submit the form, and the message associated with the email field is only validated once the form is submitted.  It's a battle of two different validation methods; HTML's `required` attribute ensuring the form is not submitted if fields are empty, and formik/yup ensuring the email is different from the current email only after the form is submitted.

I'll need to think about how we can resolve this.

collinpreston:
@jkachel so I think we have a couple options to address your comment:
1. Perform Yup validations when the user clicks the submit button, even though the form is not submitting.  We would need to do some javascript trickery for this.
2. Perform Yup validations when the user shifts focus from the field, which is the opposite direction from where we are heading.
3. Leave as is, only perform Yup validations after the form is submitted.

Curious what you think.

jkachel:
> Perform Yup validations when the user clicks the submit button, even though the form is not submitting. We would need to do some javascript trickery for this.

I think this might work - not 100% sure on how this will work out though but if it can at least clear the Yup validation out before the browser validation happens, that would do it. 

I did have another idea, though: There's a pattern match option for the in-built browser validation - https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern You'd need to generate a regex that'd match on the user's existing email address, which shouldn't be too hard. I'm not sure how this would work from an accessibility standpoint, though, and haven't tested this at all so not sure how well it works. (If you already discounted this one, though, feel free to ignore this.)

collinpreston:
> > Perform Yup validations when the user clicks the submit button, even though the form is not submitting. We would need to do some javascript trickery for this.
> 
> I think this might work - not 100% sure on how this will work out though but if it can at least clear the Yup validation out before the browser validation happens, that would do it.
> 
> I did have another idea, though: There's a pattern match option for the in-built browser validation - https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern You'd need to generate a regex that'd match on the user's existing email address, which shouldn't be too hard. I'm not sure how this would work from an accessibility standpoint, though, and haven't tested this at all so not sure how well it works. (If you already discounted this one, though, feel free to ignore this.)

@jkachel nice find with `pattern`!  I have updated many of our existing validations which used Yup, to now use `pattern`.  I The regex that `pattern` relies upon is placed in `validation.js`.  I removed a lot of unit tests which focused on the validation messages.  I couldn't find a way to test the validation messages that `pattern` displays, but I made sure to write unit tests for the regex.

jkachel:
This looks good for the change email form but I have a couple of things:
- On the change email form actually, the regex doesn't escape `+` so emails that involve that don't pass validation.
- On the edit profile form, filling out First Name with a name fails validation. The test was for a test learner account that did not have a name specified and the name I was putting in was "Test". Not sure what the problem is - it seems to match in RegexPal but Firefox complains about it. I tried manually entering a legal address in for the user (via dbshell) and it also did not validate properly there. I have a screengrab here: https://share.cleanshot.com/156K78k55gxSnXcjR9w1
   - Do we need to validate the name fields this much? We need to strip stuff out before it goes to edX but not sure about limiting the learner's input here.

collinpreston:
@jkachel I'm not sure about what the driver is for the strict name validation.  There are also some Regex validations which take place on the API side, but could be done on the front end as well, such as username: https://github.com/mitodl/mitxonline/blob/3d5464b2831c4e4ed5ec7b6c4e39fe3d1d787962/users/serializers.py#L38.  After this PR I will create an issue to investigate these comments.

jkachel (approved):
Tested and looks good, including the things that I noted earlier. LGTM 👍

PR #1535, opened 2023-04-06
Order History Page table makeover

annagav:
#### Pre-Flight checklist

  - [x] Desktop screenshots
Fix https://github.com/mitodl/hq/issues/718

#### What's this PR do?
Updates the order history page to display a proper table so that screen readers work.

Go to orders history page. Make sure it looks good. Then use a screed reader to go over the order history structure.

Screen Shot 2023-04-06 at 12 34 16 PM

### DIFF
--- a/frontend/public/src/containers/pages/checkout/OrderHistory.js
+++ b/frontend/public/src/containers/pages/checkout/OrderHistory.js
@@ -85,7 +85,7 @@ export class OrderHistory extends React.Component<Props> {
         <td>
           <a
             className="link-text"
-            aria-label={`View order details for ${orderTitle}`}
+            aria-label={`View order details for ${order.titles}`}
             onClick={() => this.renderOrderReceipt(order.id)}
             href="#"
             role="link"
--- a/frontend/public/src/containers/pages/checkout/OrderHistory.js
+++ b/frontend/public/src/containers/pages/checkout/OrderHistory.js
@@ -75,6 +75,34 @@ export class OrderHistory extends React.Component<Props> {
       parseDateString(order.created_on)
     )
     return (
+      <tr scope="row" key={`ordercard_${order.id}`}>
+        <td className="d-flex">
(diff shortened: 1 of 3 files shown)


collinpreston (changes requested):
I am able to use the screen reader to access and hear the table, columns, and rows.  However, there are a couple of issues introduced:
1. The mobile view of the order history page needs refactoring.
2. The "view" link in the far right column is not accessible with the tab key.  I think this requires `role="link"`
3. It might be helpful to add `title` to the table and reference the `h1` header "Order History"

frontend/public/src/containers/pages/checkout/OrderHistory.js
```diff
@@ -144,30 +147,31 @@ export class OrderHistory extends React.Component<Props> {
...
+                    {orderHistory && orderHistory.results.length > 0
+                      ? orderHistory.results.map(this.renderOrderCard.bind(this))
+                      : null}
+                    <tr className="border-top border-2 border-dark">
+                      <td>
+                        Page {this.offset + 1} of{" "}
+                        {orderHistory ? Math.ceil(orderHistory.count / 10) : 0} |{" "}
+                        {orderHistory ? orderHistory.count : "0"} orders total
+                      </td>
+                      <td /> <td /> <td />
```
collinpreston, line 166: I think these are causing an error to be thrown to the console in Google Chrome: `Whitespace text nodes cannot appear as a child of <tr>. Make sure you don't have any extra whitespace between tags on each line of your source code.`

annagav:
@collinpreston this is ready for another review, thanks!

collinpreston (changes requested):
Looks good, I would just add an aria label to the "View" link for each item in the table.  Something like "View order for {item}" would then be spoken like "View order for E2E Test Course link" for those using screen readers.  Everything else looks good!

collinpreston (approved):
Nice, I _think_ that additional aria-label will be helpful for those who need it.  Looks good!

PR #1537, opened 2023-04-07
715: ecommerce pressing pay jumps back to dashboard without focus on alert

collinpreston:
#### Pre-Flight checklist

- [x] Testing

#### What's this PR do?
Updates bootstrap and reactstrap in order to improve accessibility of alerts for users with screen readers (refer to associated issue).

1. Ensure that you can focus on the close button of alerts when using only the keyboard to navigate.
2. When an alert is triggered, ensure that the alert is read via screen reader.
3. When an alert is triggered, ensure that the focus is set to the alert before any other item on the page (after pressing "Skip to main content".
4. While running this branch locally, visit each page of the application and compare it with RC to ensure that the UI is the same between both.  I did not supply screenshots as this change should not change the current UI in any way.

#### Additional comments
Use the commits and their comments to help understand what each change in the PR does.


annagav (changes requested):
Screen Shot 2023-04-10 at 2 20 16 PM
The button is too big.
Also, I am not able to see the alerts anymore.

annagav (approved):
👍

Issue #1538, opened 2023-04-07
Users are still able to log in using a retired email account


cachob:
### Steps to Reproduce

We came across a ticket where the learner requested to have their accounts deleted / removed from the system:

[les030127@foxmail.com](mailto:les030127@foxmail.com)
[mit030127@gmail.com](mailto:mit030127@gmail.com)

We followed the standard retire users process by, running the retire_users management command unchecking the "Active" status on each of the accounts in Django:

~ $ python manage.py retire_users -u les030127@foxmail.com
Retiring user: User username=mit0127 email=les030127@foxmail.com
User: 'User username=mit0127 email=les030127@foxmail.com' is already deactivated in MIT mitX Online
~ $ python manage.py retire_users -u mit030127@gmail.com
Retiring user: User username=OP0127 email=mit030127@gmail.com
User: 'User username=OP0127 email=mit030127@gmail.com' is already deactivated in MIT mitX Online

### Expected Behavior

The learner will not be able to log into the platform with the retired email accounts. The email accounts will be marked as retired in Django when looked up in https://mitxonline.mit.edu/admin/users/user/

![image](https://user-images.githubusercontent.com/116296444/230659977-e3fb033a-3501-4dc4-9b3a-212f245e94f3.png)

### Actual Behavior

This learner is still able to log into the system, and the user account email does not get marked as retired:

![image](https://user-images.githubusercontent.com/116296444/230660123-0a46c838-ce37-4296-a98f-40e60ed10a52.png)

![image](https://user-images.githubusercontent.com/116296444/230660172-92cd6ce4-5be1-4574-b7f2-9b89f7c20b17.png)

https://mitxonline.mit.edu/admin/users/user/54903/change/?_changelist_filters=q%3Dsiuuuuu

### Related Issues

(https://odl.zendesk.com/agent/tickets/152865)

collinpreston:
@cachob you should be able to deactivate these users now.

Issue #1536, opened 2023-04-07
Display an error message to the user if their password does not log them in


collinpreston:
### Steps to Reproduce

Attempt to login using an incorrect password.

### Expected Behavior

In addition to the current behavior, the user is informed that their email and password combination does not match.

### Actual Behavior

The password field is red.  Focus is made on the password field and it's highlighted red.

collinpreston:
Will be done alongside https://github.com/mitodl/mitxonline/issues/1538

PR #1539, opened 2023-04-10
1538 users are still able to log in using a retired email account

collinpreston:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/1538
https://github.com/mitodl/mitxonline/issues/1536

#### What's this PR do?

1. User records that have `is_active = False` should not allow the user to login.
2. For the password and email login screens, if a valid user record matching the email is not found or their password does not match the user record, an error message should be displayed to the user.

1. Go through the registration process for a new user.
3. Once the user is created, go through Django admin and set `is_active = False`.  Logout with that user.
4. Attempt to log back in with that user.  The email login screen should present an error that the user was not found.
5. Through Django admin and set `is_active = True`
6. Attempt to log back in with that user but stop once you reach the password login screen.
7. Through Django admin and set `is_active = False`
8. Enter the password for the user and attempt to login.  The password login screen should present an error that the password did not match the user.
9. Through Django admin and set `is_active = True`
10. Login with the user in order to create an active session.
11. Execute `python manage.py retire_users -u <USER_EMAIL>`
12. Refresh the page where the user is logged in - you should be redirected to the login page.
13. Attempt to login with the user again.  The email login screen should present an error that the user cannot be found.
14. Verify that you can register a new user with the same email address you used in step 1.

### DIFF
--- a/authentication/pipeline/user.py
+++ b/authentication/pipeline/user.py
@@ -106,6 +106,7 @@ def create_user_via_email(
         )
 
     data["email"] = kwargs.get("email", kwargs.get("details", {}).get("email"))
+    data["is_active"] = True
     serializer = UserSerializer(data=data, context=context)
 
     if not serializer.is_valid():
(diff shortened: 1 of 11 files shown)


annagav (approved):
LGTM 👍

PR #1544, opened 2023-04-12
715: ecommerce pressing pay jumps back to dashboard without focus on alert (2)

collinpreston:
#### Pre-Flight checklist
- [x] Testing

#### What's this PR do?

- Makes more `data-*` elements Bootstrap v5 compliant by modifying to `data-bs-*`.  
- Fixes the "more dates" popover shown on a course about page with multiple future course runs.
- Fixes the links in navigation menu (top right) for the mobile view.

1. Have a course with multiple course run records where the enrollment period is now and the course start date is in the future.
2. Visit the course page in mitxonline for these courses.  Verify that the "More dates" option is shown, you can select it and view the popover, and you can select a different date.
3. Test the navigation links in the drop-down menu in the top right of the mitxonline application.  Verify that menu works as expected for both authenticated and unauthenticated users.


annagav (approved):
Works for me 👍

PR #1545, opened 2023-04-12
fix: remove codecov because it's gone from PyPI, the codecov action would do it anyway

arslanashraf7:
None, The package is gone from PyPI, and we should remove it.
More info on [this slack thread](https://mitodl.slack.com/archives/G3A5RSFDJ/p1681318942371839) and the following conversation.

#### What's this PR do?
- Removed the codecov package from `test_requirements.txt`

Make sure the CI checks pass.

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,6 +1,5 @@
 black
 bpython
-codecov
 ddt
 django-debug-toolbar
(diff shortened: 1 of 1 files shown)


collinpreston (approved):
Looks like it works

PR #1558, opened 2023-04-20
Update requests package

collinpreston:
Also gets rid of the chardet dependency.

#### What's this PR do?
Updates the `requests` package.  This removed the dependency on the `chardet` package which is out of date (https://github.com/mitodl/mitxonline/pull/993).  

Tests should pass.  All uses of request should continue to function as expected - this can be tested by creating a user and ensuring the `openedx.models.OpenEdxApiAuth` record is created and matching the associated record in edx.  When in RC, additional verifications such as recaptcha and refunding orders can be easily tested.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,8 @@ amqp==5.0.9
     # via kombu
 anyascii==0.2.0
     # via wagtail
+appnope==0.1.3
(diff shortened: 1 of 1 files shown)


annagav (approved):
👍

PR #1561, opened 2023-04-20
1295: learner menu stops functioning at a particular width range

collinpreston:
#### Pre-Flight checklist
- [x] Testing

https://github.com/mitodl/hq/issues/1295

#### What's this PR do?
The hamburger menu was not functional when screens were _medium_ width (https://getbootstrap.com/docs/5.0/layout/breakpoints/#available-breakpoints)

1. Set the width of your browser screen to each of the breakpoints mentioned in the link shared above.  Verify that you can view whichever menu is shown based on the browser screen size.

This PR removes the css transition that comes with Bootstrap navbars.  Previously, we wouldn't display the animation and would only shown the navbar options after the animation completed.  Removing the animation speeds this up.

There is also some general refactoring to cleanup reused code.

### DIFF
--- a/frontend/public/scss/top-app-bar.scss
+++ b/frontend/public/scss/top-app-bar.scss
@@ -9,7 +9,7 @@ header.site-header {
     justify-content: space-between;
     padding: 0;
 
-    @include media-breakpoint-down(sm) {
+    @include media-breakpoint-down(md) {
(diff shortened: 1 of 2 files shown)


jkachel (approved):
LGTM 👍

PR #1578, opened 2023-04-27
Updates fields that are sent to edX and adds profile sync

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

https://github.com/mitodl/hq/issues/1118

#### What's this PR do?

Three things:
- Updates the fields that are sent when a learner's account is created in edX. Prior to this, the code always sent the user's country as "US"; now it sends the code that the learner has specified and the state they're in if they're in the US. 
- Adds an API that allows for this data to be re-sent to edX. This syncs name, country, and state (if in the US) and can be amended to sync other data if we feel the learner profile in edX needs to be more complete.
- Adds a management command to sync a particular user, or all users. 

Initial field sync:
1. Create a new account.
2. Specify a country and (optionally) state (depending on country selection).
3. Verify the account information in the LMS Django Admin. The country and state should be filled correctly.

Re-sync:
1. Make sure you have user accounts in MITx Online that also exist in the LMS. 
2. Furthermore, make sure these accounts have country data in them.
3. Run the `update_edx_profiles` command. Specify either `--user <username>` to sync a single user, or `--all` to sync all of them. 
4. Check the synced accounts in the LMS Django Admin and ensure the data is correct.

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -143,73 +143,6 @@ def create_edx_user(user):
         return True
 
 
-def update_edx_user_profile(user):
-    """
-    Updates the specified user's profile in edX. This only changes a handful of
-    fields; it's mostly for syncing demographic data.
-
-    Args:
-        user(user.models.User): the user to update
-    """
-
-    # Step 1
-    with requests.Session() as req_session:
-        # Step 2
-        django_session = auth_api.create_user_session(user)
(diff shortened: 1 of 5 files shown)


annagav (approved):
LGTM 👍

Issue #1573, opened 2023-04-25
Sync extended profile data to hubspot


pdpinch:
The marketing team would like to be able to segment the mitxonline audience based on all the profile fields that we capture. 

According to [this comment](https://github.com/mitodl/hq/issues/905#issuecomment-1522151973), we need to add the new extended profile fields to this list:

https://github.com/mitodl/mitxonline/blob/93eeba8b675ed53e2c493e88c9a00c463258e133/hubspot_sync/api.py#L45-L51

Any fields that don't already exist in HubSpot (for example, "Leadership Level") we need to run a management command, https://github.com/mitodl/mitxonline/blob/main/hubspot_sync/management/commands/configure_hubspot_properties.py

pdpinch:
After 1573 merges, what else will need to be done, to make sure that all our existing users in mitxonline have their profile data in hubspot? 

Please feel free to edit the issue description. I'd like a checklist of some kind.

PR #1579, opened 2023-04-28
1573 sync extended profile data to hubspot

collinpreston:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/1573

#### What's this PR do?
Adds all of the user profile fields from MITx Online to Hubspot.

1. You'll need some new .env settings:

```
MITOL_HUBSPOT_API_PRIVATE_TOKEN=<ask_me, and I can also add you to the hubspot account>
MITOL_HUBSPOT_API_ID_PREFIX=<your_initials>-mitxonline-dev
HUBSPOT_PIPELINE_ID=19817792
```
2. Register a user through your local MITx Online instance
3. View the Hubspot contact record created for the user in step 2 (https://app.hubspot.com/contacts/23586992/objects/0-1/views/all/list?redirectFrom=crm-records-ui) and verify that the contact record contains properties of all the populated MITx Online user profile attributes.
4. Make an update to the user's MITx Online profile and verify that the Hubspot contact properties reflect the update.


pdpinch:
> You'll need some new .env settings:

Does this need more/better documentation in the repo?

mbertrand (commented):
Looks great.  One question: should state be synced (when applicable)?  Currently it isn't.

mbertrand (commented):
We might need to adjust the mgmt command for updating existing users (a separate PR for this would be fine) so as to allow for specifying certain batches of users (by first/last name starting with a specific letter, or users updated before/after a certain date, for example).  There are currently over 200K users on production and batch updating of hubspot contacts is pretty constrained (only 10 allowed per batch, and only so many requests allowed per second/hour).  I'm afraid the task might take days and potentially could get interrupted by dyno restarts.  But it's probably worth trying as is for now and seeing how it goes.

mbertrand (approved):
👍 

Separate issue (bug?) - Noticed that gender is not shown in either form when registering a new user for the 1st time, it's only available from the profile edit page.

Issue #1580, opened 2023-04-28
Two accounts with the same e-mail


pdpinch:
### Steps to Reproduce

1. Enter an email, with lowercase letters in the address, at https://rc.mitxonline.mit.edu/create-account/
2. In another tab, open https://rc.mitxonline.mit.edu/create-account/ again and enter the same email, but all uppercase 
3. You should get two verification emails with different links. 
4. Open both links in separate tabs
4. Fill out one of the registration forms and  complete the first screen of the account creation process
5. In the second tab, fill out the registration form again. In my testing, I used a different username to avoid the username validation check. 

### Expected Behavior

- you should get an error when filling out the second form because an account with that email already exists

### Actual Behavior

- two accounts are created
- You can see where I've done this at https://rc.mitxonline.mit.edu/admin/users/user/?q=2023-04-28
- I'm not aware of this having happened in production on mitxonline, but we should check

### Related Issues

- xPRO has the same bug, https://github.com/mitodl/mitxpro/issues/2633
- I imagine that bootcamps does as well.

jkachel:
This is on production now; probably a similar approach can be used for the other two (I know MITx Online lifted a lot of the code from xPro so that's probably the easiest to do it in).

jkachel:
Closing this; this has been on prod for about a week.

PR #1586, opened 2023-05-01
Throw an error if the user manages to get to the registration screen with the same email

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

https://github.com/mitodl/mitxonline/issues/1580

#### What's this PR do?

Updates the UserSerializer to throw a validation error if the email address isn't actually unique. While there was a database constraint on this, that considers case differences as unique. So, the email validation function has been expanded to do a case insensitive search for the email address, and to also strip whitespace before doing the check. 

This also adds some code to the RegisterDetailsPage component to display an email error using the standard user notification system (since we don't expose the email field on the form). 

1. Open two separate browsers or browser sessions (incognito, etc.)
2. Navigate to the Create Account page in each.
3. Supply an email address in one in all lowercase.
4. Supply the same email address in the other one, but in all caps. 
5. When the verification emails come through for each, navigate to the create account page in each browser. You will need to have the details page open in both before continuing - do not fill the form out yet. 
6. Fill out the details form in one browser and complete the signup process. This should be successful.
7. Fill out the details form in the other browser. This should fail with an error message. 

In addition, editing the profile and changing the email address of an existing account should be successful.

### DIFF
--- a/frontend/public/src/containers/pages/register/RegisterDetailsPage.js
+++ b/frontend/public/src/containers/pages/register/RegisterDetailsPage.js
@@ -116,6 +116,17 @@ export class RegisterDetailsPage extends React.Component<Props> {
           // field errors.
           if (field_errors) {
             setErrors(field_errors)
+
+            if (field_errors["email"]) {
+              addUserNotification({
+                "registration-failed-status": {
+                  type:  ALERT_TYPE_DANGER,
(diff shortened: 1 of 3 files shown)


collinpreston (approved):
Looks good and works for me!

PR #1589, opened 2023-05-01
Fixes us_state to return None if there's no state; adds a test for that

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

#1588 

#### What's this PR do?

If the user's in the US and has not specified a state, the `us_state` property in `legal_address` should just be None. 

1. Update a user account to have the country set to US and no state specified (you may need to set this up in Django Admin).
2. Attempt to update the edx profile for that user with the `update_edx_profiles --user` command. This should succeed.


collinpreston (approved):
Looks good and works well

PR #1600, opened 2023-05-04
Changes the login workflow to ignore case in the email field

### LINKED ISSUE #1592: Add more user email normalization to lowercase
Noticed this while working through https://github.com/mitodl/mitxonline/pull/1586 - didn't address it in that PR though. Searches for email addresses in the user object should be case insensitive, as email addresses themselves are not case sensitive. I noted this because I created an account with all caps and then tried to log in without that capitalization and was denied access. (Oddly, there's some of this already - the _domain_ is normalized to lowercase but the username part is not. I was able to log into RC as `APPTEST+DUPTEST@JKACHEL.COM` but this is stored as `APPTEST+DUPTEST@jkachel.com` in the database.) 

We should probably store the email address as specified by the user and instead normalize interactions through the system to lowercase to compensate for insufficiently smart email servers that may actually treat the username part of the email address as case-sensitive. (In less words, login should use a case insensitive search but in my example it should be stored as it was.) 

### Steps to Reproduce

1. Create a user account using an email address in all capitals. 
2. Attempt to log in with that same email address, but input it in lowercase.

### Expected Behavior

I should be able to supply the password and complete the login process.

### Actual Behavior

I get the "Can't find your email" message and cannot log in.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes #1592 

#### What's this PR do?

Changes some of the email lookups that the social auth login code uses to use case insensitive ones, so users do not have to match case when logging in.

1. Sign up with an email address using all capitals. This should be successful.
2. Log out.
3. Log back in with the same email address, but specify it in all lowercase. This should be successful.

FWIW the social auth code will create a separate row for each permutation of the email address that it sees - these all will still point to the correct User object.


collinpreston (approved):
Nice!

Issue #1598, opened 2023-05-03
Program Menu accessibility updates


pdpinch:
## Remove live region markup

On both the "courses" tab and "my programs" tab:

```
<div id="enrollment-items" class="enrolled-items" aria-live="polite">
<div><h1 class="hide-element">My Courses</h1>
...
```

This will read the entire course or program list. Instead, shift focus to the first focusable item. 

## Submenus?

I see three controls here, each marked up as a submenu (`aria-haspopup`). None of them display a menu, a dialog (as value of aria-haspopup seems to suggest), nor anything else I can detect. 

However, looking at the DOM, there is indeed a modal:

```
<div class="nav-drawer open" id="program_enrollment_drawer" role="dialog" tabindex="-1" aria-modal="true" aria-label="Data, Economics and Development Policy" aria-description="6 courses, 0 passed" aria-hidden="false">
...
</div>
```

The reason it isn't appearing for screen reader users is because of the use of `aria-hidden`:

```
<div class="std-page-body dashboard container" aria-hidden="true">
...
</div>
```

__This attribute should never be used__!

THere are very few instances in which it is necessary to use this attribute, and as shown by this example, it causes issues that are difficult to find by sighted developers who don't test with screen readers because it's effects are undetectable unless one uses a screen reader.

When hiding something via css or the HTML `hidden` attribute, it is hidden for screen readrs as well, so there is __absolutely not reason to duplicate functionality via `aria-hidden`__!!

collinpreston:
I think this may have been included with the work done for https://github.com/mitodl/hq/issues/703.  But maybe I missed something or didn't get it right.

jkachel:
Think so - closing

PR #1602, opened 2023-05-08
Removes a stray aria-hidden attribute

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

https://github.com/mitodl/mitxonline/issues/1598

#### What's this PR do?

There was a stray `aria-hidden` attribute on the program drawer; this removes it. (This was the only one I could find via grep.) 

The issue with the program drawer noted in https://github.com/mitodl/mitxonline/issues/1598 should be resolved - there's a lot there so it's better to reference the original issue. 

I couldn't find the first issue - I didn't see any `aria-live` attributes anywhere other than in the notification container (which would actually need one). The ticket for this came out pretty soon after the PR @collinpreston mentioned was merged so this might have been fixed already.


collinpreston (approved):
Nice

PR #1621, opened 2023-05-19
Expands edX profile sync.

### LINKED ISSUE #1620: Expand upon edX profile sync functionality
As an instructor (or other admin type), I would like to see more demographic data from reports that are generated within edX. Specifically, the gender, level of education, and year of birth fields should be present within the reports.

As a learner, when I update my profile, I would like those changes to be represented in edX as well. 

See HQ ticket https://github.com/mitodl/hq/issues/1118 and also https://github.com/mitodl/mitxonline/pull/1578 - we initially implemented syncing the country and state data, but did not opt at that point to sync anything else.

#### Acceptance Criteria:

- [ ] Gender, level of education, and year of birth fields from the extended profile should be synced with edX. This sync should be one-way. 
- [ ] The sync should happen when a learner registers.
- [ ] The sync should happen when the `update_edx_profiles` command is issued.
- [ ] The sync should happen when a learner updates their profile from within MITx Online.
   - [ ] The sync should queue a task to perform this sync. 
   - [ ] Learners should not experience any delays from this new sync (i.e. clicking save shouldn't take any longer than it does now).

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes #1620
Addresses issues raised in https://github.com/mitodl/hq/issues/1118

#### What's this PR do?

- Adds gender, year of birth, and level of education fields to the list of fields we now sync with edX.
- Updates the user profile API to sync the profile data on save. 

You will need to have a working edX instance for this - either Tutor or devstack should work (tested on Tutor). 

Manual sync:
1. Change an account's user profile gender, year of birth, or highest education fields via Django Admin/dbshell/other method that's not the normal frontend.
2. Run `manage.py update_edx_profiles --user <username>` where `<username>` is the username you edited. 
3. Check the data in edX - I found this was easiest to do using the Swagger interface (ex: http://local.overhang.io:8000/api-docs/#/user/user_v1_accounts_read if you're using Tutor)

Automatic sync:
1. Update an account's user profile gender, year of birth, or highest education fields in the normal way through the frontend.
2. Check the data in edX - the changes should be present.

Account creation:
1. Create an account and fill out the gender, year of birth, and highest education fields. 
2. Check the new account's data in edX. The changes should be present.

Both the gender and the level of education fields in the edX user model use flags. We use the same flags for the gender field in mitxonline but the highest education field is both named differently and does not use flags, so there's a new list to map highest education to level of education in the User model to avoid writing a database migration to mass-update the data and changing it elsewhere in the code.

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -242,6 +242,13 @@ def update_edx_user_profile(user):
             name=user.name,
             country=user.legal_address.country if user.legal_address else None,
             state=user.legal_address.us_state if user.legal_address else None,
+            gender=user.user_profile.gender if user.user_profile else None,
+            year_of_birth=user.user_profile.year_of_birth
+            if user.user_profile
+            else None,
+            level_of_education=user.user_profile.highest_education
+            if user.user_profile
+            else None,
         ),
         headers={
             "Authorization": f"Bearer {auth.access_token}",
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -128,6 +128,13 @@ def create_edx_user(user):
                 name=user.name,
                 country=user.legal_address.country if user.legal_address else None,
                 state=user.legal_address.us_state if user.legal_address else None,
+                gender=user.user_profile.gender if user.user_profile else None,
+                year_of_birth=user.user_profile.year_of_birth
+                if user.user_profile
+                else None,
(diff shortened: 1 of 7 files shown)


annagav (approved):
Works for me 👍

Issue #1627, opened 2023-05-23
Improve HubSpot sync script to handle large numbers of users


collinpreston:
When new fields/attributes are added to HubSpot objects such as Contacts, we will need to sync the User records in MITx Online with the related Contacts in HubSpot in order for the new attributes to be populated with data from MITx Online. The `sync_db_to_hubspot` script should be able to effectively handle large numbers of users to ensure that all Production users are synced with HubSpot.

#### Acceptance Criteria:

- [ ] Ensure that the script can resume if terminated without starting from the same records each time.
- [ ] Ensure that the script is automated.
- [ ] Ensure that the script includes logging to indicate progress.

PR #1628, opened 2023-05-24
1627: improve hubspot sync script to handle large numbers of users

collinpreston:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/1627

#### What's this PR do?

1. Creates a scheduled celery task to sync all MITx Online User records with HubSpot.  Performs upsert to HubSpot contacts.
2. Adds a new attribute, `hubspot_sync_datetime`, to the `User` model in MITx Online.  This is used to record the datetime at which the last create or update to the related HubSpot contact was made.  Default value is null.  This field is visible through Django Admin.
3. Adds logging to the `batch_update_hubspot_objects_chunked` method which is called by the task created in bullet #1 and the `sync_db_to_hubspot` management command.  The logging will display the HubSpot user ID's updated in each chunk as well as the percentage completed after each chunk.

**Test creating a new MITx Online user with a successful HubSpot call**
1. Configure HubSpot in your `.env`
2. Register a new user through MITx Online and verify, through Django Admin, that the user has a populated `hubspot_sync_datetime` value and an associated HubSpot object exists.

**Test creating a new MITx Online user with an unsuccessful HubSpot call**
1. Remove or comment out the HubSpot configuration in your `.env`
2. Register a new user through MITx Online and verify, through Django Admin, that the user does not have a populated `hubspot_sync_datetime` value and an associated HubSpot object does not exist.

**Test the sync_db_to_hubspot command**
For this test, you will refer to the user created during "Test creating a new MITx Online user with an unsuccessful HubSpot call"
1. Configure HubSpot in your `.env`
2. Run the `sync_db_to_hubspot` management command from the web container like so: `python manage.py sync_db_to_hubspot --users update`.
3. Verify that you see the new log messages described (they will display in the Celery container).
4. Verify, through Django Admin, that the user has a populated `hubspot_sync_datetime` value and an associated HubSpot object exists.

**Test the sync-hubspot-user-contacts scheduled celery task**
1. Remove or comment out the HubSpot configuration in your `.env`
2. Register a new user through MITx Online.
3. Stop your MITx Online instance.  
4. Add back or uncomment your HubSpot configuration in your `.env`.
5. Open `main/settings.py` and update the `sync-hubspot-user-contacts` entry to be:
```
    "sync-hubspot-user-contacts": {
        "task": "hubspot_sync.tasks.sync_all_contacts_with_hubspot",
        "schedule": crontab(
            minute="*",
            hour="*",
            day_of_week="*",
            day_of_month="*",
            month_of_year="*",
        ),
    },
```
6. Restart your MITx Online instance.
7. View the logs for the Celery container and verify that the `sync-hubspot-user-contacts` task runs and the new log messages are displayed.
8. Verify the user you created in step 2 now has a populated `hubspot_sync_datetime` value and an associated HubSpot object exists.

**Updating an existing user's profile information**
1. Use an existing MITx Online user who has a populated `hubspot_sync_datetime` value and an associated HubSpot object exists.
2. Login with that user and update their first name or any other profile information.
3. Verify that the `hubspot_sync_datetime` value has been updated to reflect a more recent datetime.


hubspot_sync/tasks.py
```diff
@@ -229,13 +232,12 @@ def batch_create_hubspot_objects_chunked(
...
-                    )
+                    user = User.objects.filter(
+                        email__iexact=result.properties["email"], is_active=True
+                    ).first()
+                    user.hubspot_sync_datetime = datetime.now()
+                    user.save()
+                    object_id = user.id
                 else:
                     object_id = result.properties["unique_app_id"].split("-")[-1]
                 HubspotObject.objects.update_or_create(
```
mbertrand, line 243: I still kind of feel that the better place to store this might be in `HubspotObject`, because eventually we might run into the same problem if we need to update all existing deals in hubspot.  There aren't nearly as many deals as contacts (yet) but that might change once we start creating deals for free enrollments.

So could do this here:

```python
HubspotObject.objects.update_or_create(
                    content_type=content_type,
                    hubspot_id=result.id,
                    object_id=object_id,
                    defaults: {
                        "hubspot_sync_datetime": datetime.now()
                    }
                )
```

And also update the `mitol.hubspot_api.api.upsert_object_request` function:
```python
    if object_id:
        HubspotObject.objects.update_or_create(
            object_id=object_id, 
            content_type=content_type, 
            hubspot_id=hubspot_id or result.id, 
            defaults: {
                "hubspot_sync_datetime": datetime.now()
            }            
        )

File: hubspot_sync/tasks.py [hunk 2]
```diff
@@ -299,14 +301,26 @@ def batch_update_hubspot_objects_chunked(
...
             last_error_status = ae.status
             still_failed = handle_failed_batch_chunk(
                 [item[0] for item in chunk], hubspot_type
             )
             if still_failed:
                 errored_chunks.append(still_failed)
+        log.info("Updated the following HubSpot ID's %s", updated_ids)
+        percent_complete = (len(updated_ids) / len(object_ids)) * 100
+        log.info("%i%% complete updating HubSpot ID's", percent_complete)
+        if ct_model_name == "user":
```
mbertrand, line 316: Continuing from previous comment, could track for all objects, not just users:

```python
 HubspotObject.objects.filter(
    content_type = ContentType.objects.exclude(app_label="auth").get(model=ct_model_name),
    hubspot_id__in=updated_ids,
).update(hubspot_sync_datetime=datetime.now)
```

hubspot_sync/tasks.py
```diff
@@ -344,9 +369,11 @@ def batch_upsert_hubspot_objects(  # pylint:disable=too-many-arguments
             id__in=[id[0] for id in synced_object_ids]
```
mbertrand, line 369:
```python
.order_by("hubspot_sync_datetime")
```
here instead of below for users only?

mbertrand (commented):
Adding the field to `User` is fine if it's unlikely we'll ever have the same issue with updating other hubspot objects like deals.  But I suspect we might have to update lots of deals at some point in the future (though at least we can do 100+ per request for those instead of just 10).  What do you think?

hubspot_sync/tasks.py
collinpreston: @mbertrand My other worry would be that updating applications other than MITx Online which directly call HubSpot would need to be prioritized for after this and the other MITx Online HubSpot work is completed, on June 12th.  I don't think that will cause problems, but there's the chance other work is brought forward prioritized ahead of this.

mbertrand: It might be possible to move that direct bulk update call to ol-django too, at least partially.  I think the June 12th work involves adding all enrollments to hubspot as deals, and there are over 100K enrollments on prod, so it's possible we might have to do the same for those as for contacts to get them all created/updated.  But since the process of updating ol-django then adding the latest version to each app is cumbersome and 6/12 is fast approaching, I'll accept the current approach.  Will test functionality next.

users/migrations/0023_user_hubspot_sync_datetime.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("users", "0022_backfill_userprofile_records"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="hubspot_sync_datetime",
+            field=models.DateTimeField(null=True),
```
mbertrand, line 15: You might want to set a default value for this, or try an [approach like that described here](https://isotoma.com/blog/2015/11/23/sorting-querysets-with-nulls-in-django/) to get nulls returned first in the list.  When trying out locally, users with null values always come after users with non-null values.

hubspot_sync/tasks.py
```diff
@@ -299,14 +301,26 @@ def batch_update_hubspot_objects_chunked(
...
+        log.info("Updated the following HubSpot ID's %s", updated_ids)
+        percent_complete = (len(updated_ids) / len(object_ids)) * 100
+        log.info("%i%% complete updating HubSpot ID's", percent_complete)
+        if ct_model_name == "user":
+            user_ids = HubspotObject.objects.filter(
+                content_type=ContentType.objects.get_for_model(User),
+                hubspot_id__in=updated_ids,
+            ).values_list("object_id", flat=True)
+            User.objects.filter(id__in=user_ids).update(
+                hubspot_sync_datetime=datetime.now(pytz.timezone(settings.TIME_ZONE))
```
mbertrand, line 322:
```python
...
...
hubspot_sync_datetime=now_in_utc()
```

mbertrand (changes requested):

users/migrations/0023_user_hubspot_sync_datetime.py
collinpreston: Good catch.  I found this https://docs.djangoproject.com/en/4.2/ref/models/expressions/#using-f-to-sort-null-values as well.

hubspot_sync/tasks.py
collinpreston: I will create a new issue in ol-django to support bulk create/update as well as adding the datetime attribute to the HubSpotObject.  This will ensure that no other apps need to call HubSpot directly and can instead rely on the ol-django methods.

PR #1631, opened 2023-05-24
Adds bulk code generation to staff-dashboard

### LINKED ISSUE #1623: Add bulk discount code generation to staff dashboard
As a staff user, I'd like to be able to generate blocks of discount codes for various purposes without involving an engineer.

We'll often get requests to create blocks of discount codes for MITx Online. (This is most often a block of 50 100% off one-time codes that are used by customer service, but there's other reasons to generate these as well.) These are pretty easy to generate using the command line `generate_discount_code` command but this requires it be done by someone with the necessary expertise and Heroku access. It would be quicker to create a separate screen in the staff-dashboard app to handle this, so these requests can be (more) self-service.

#### Designs and Mockups

This would follow the design of the existing Create page in the staff-dashboard, except behind a different button (Create Batch). Discount Code would change to "Discount Code Prefix" and be limited to 12 characters (technically it's 13, but we'll want to programmatically add a `-` between the prefix and the code), and I think it'd be good to add a "bulk" flag to the discount model to signify codes that are created in a batch so they can more easily be revoked later. (Revoking is a thing that happens more on the xPro side but it's feasible that we'd do this on mitxonline too.)

#### Acceptance Criteria:

- [ ] A new "Create Batch" button is added to the staff dashboard's Discount screen alongside the existing "Create" button.
- [ ] Selecting "Create Batch" opens a page much like the "Create Discount" screen with some minor changes.
   - [ ] Some copy is added to inform the user that they are creating a _batch_ of codes, and not just a single one.
   - [ ] The Discount Code field is changed to "Discount Code Prefix" and is limited to 12 characters.
   - [ ] A Count field is added. This is a numeric field that specifies the number of codes to create.
   - [ ] All other fields remain the same.
- [ ] Activating the Save button creates a batch of codes to the specification given.
   - [ ] This should work like running `generate_discount_code` with the `--count` option specified - the codes should be the prefix with `-` and a UUID appended; there should be the number specified; etc. 
   - [ ] The `batch` flag should be set on these codes.
- [ ] The batch of codes is then displayed to the user separately.
   - [ ] Unlike the Create Discount page, we would display an interstitial page with just the new codes in it.
   - [ ] The user would be able to download the code as a CSV file for use elsewhere or input into a Google Doc (using PapaParse and js-file-download most likely).
- [ ] The `generate_discount_code` command should be updated to set the `batch` flag if it's creating a batch of codes as well

#### Out of Scope

- Revocation - I think this is a good idea to add but not in this issue.

jkachel:
#### Pre-Flight checklist

  - [X] Desktop screenshots
- [X] Migrations
- [X] Testing
  - [X] Code is tested

Closes #1623

#### What's this PR do?

* Ups the maximum discount code length to 100 characters (since bulk codes use UUIDs, we were left with 13 characters for the bulk code prefix, which has been sort of a problem)
* Refactors bulk code generation into `api.py`
* Adds a Create Bulk Discount form to the staff dashboard to allow staff users to create codes on their own
   * The form works like the existing Create Discount form, with a couple of small modifications to support bulk generation
   * Generated codes are displayed after creation and the option is given to download the codes as a CSV file

Note that both your watch and refine containers should probably be restarted for testing this as there are new JavaScript dependencies. 

1. Load the staff dashboard and go to Discounts.
2. Click the Bulk Create button. 
3. Fill out the form appropriately. 
   4. Note that the prefix does not automatically append a separator between it and the code itself, so if you want one (our convention is to use a `-`) you need to specify it explicitly.
   5. If you fill out the date fields, make sure to also specify a time. Times are in UTC.
6. Click Save to create.
7. You should see the list of discount codes. 
8. Click Download CSV to download the generated codes in a CSV file.
9. The CSV file should be correct.

Screenshot 2023-05-24 at 12 36 38 PM
Screenshot 2023-05-24 at 12 36 04 PM
Screenshot 2023-05-24 at 12 35 27 PM

(I changed the button title from Create Batch to Bulk Create for consistency but did not redo the screenshot.)


rachellougee (commented):
Can we add error handlings on the front end? I've received 400 error when missing a field, I can see `This field may not be blank` in the response if I inspect, it just doesn't show in the frontend. I've also got 500 error if I put 1 as Codes to Create. I guess whoever use it would not just create 1 code, but it would be great if we can handle 500 error and display missing field error messages.

Should we name "Codes to Create" to "Number of Codes to Create" to be clear?

It works as expected if everything fill up correctly.

jkachel:
Added some validation rules in https://github.com/mitodl/mitxonline/pull/1631/commits/6b48b1e037f30da518f2ce46ce85c96842e5ca10 . I updated the discount code count field to require a minimum of 2 - if you're only making 1 you should just be using the regular form, I think; it wasn't worth refactoring the api for this to accommodate it.

rachellougee (approved):
LGTM. Validation rules work as expected.

PR #1636, opened 2023-05-26
Add row header to order history table

annagav:
#### Pre-Flight checklist

  - [x] Desktop screenshots

Fix https://github.com/mitodl/hq/issues/718

#### What's this PR do?
Add row header to order history table

Nothing should break.
Use a screen reader to make sure that it reads the row index.

Screen Shot 2023-05-26 at 7 51 29 AM


collinpreston (approved):
Looks good and tested correctly.  When using VoiceOver, if the focus is on a single cell such as the far right column "view", and you then move the focus to the "view" column for a row above or below, the screen reader will pronounce the row header (order title) before reading the content of the focused cell.

PR #1667, opened 2023-06-08
Handle HubSpot exceptions when unenrolling from a course

collinpreston:
#### Pre-Flight checklist

- [x] Testing
  - [x] Code is tested

https://github.com/mitodl/mitxonline/issues/1666

#### What's this PR do?

1. Handles the return from the filter query for Line objects correctly.
2. Handles any exception returned by HubSpot as to not hinder the user from unenrolling from the course.

- Enroll a user into a course.
- Verify that an Order object is created in MITx Online for the enrollment.
- Unenroll from the course.
- Verify no errors are being displayed in the logs and the user is able to unenroll.
- Enroll a user into a course again.
- Remove or comment out your HubSpot settings from your `.env` file.
- Restart MITx Online and attempt to unenroll from the course again.
- Verify that the user is not hindered from unenrolling from the course.

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -365,7 +365,7 @@ def deactivate_run_enrollment(
         CourseRunEnrollment: The deactivated enrollment
     """
     from ecommerce.models import Line, Order
-    from hubspot_sync.api import sync_line_item_with_hubspot
+    from hubspot_sync.task_helpers import sync_hubspot_line_by_line_id
 
     try:
         unenroll_edx_course_run(run_enrollment)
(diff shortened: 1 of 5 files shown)


courses/api.py
```diff
@@ -393,8 +393,15 @@ def deactivate_run_enrollment(
         order__purchaser=run_enrollment.user,
     )
     if line:
-        line_id = line.id
-        sync_line_item_with_hubspot(line_id)
+        line_id = line.first().id
+        try:
+            sync_line_item_with_hubspot(line_id)
```
mbertrand, line 398: Another option would be to create an autoretrying task for this, like there is for deals, contacts, etc.  Or just call `hubspot_sync.tasks.sync_deal_with_hubspot` instead since that will also sync the lines.

```python
@app.task(
    acks_late=True,
    autoretry_for=(TooManyRequestsException, BlockingIOError),
    max_retries=3,
    retry_backoff=60,
    retry_jitter=True,
)
@raise_429
@single_task(10, key=task_obj_lock)
def sync_line_with_hubspot(line_id: int) -> str:
    """
    Sync a Line with a hubspot line

    Args:
        line_id(int): The Line id

    Returns:
        str: The hubspot id for the line
    """
    return api.sync_line_with_hubspot(line_id).id
```

mbertrand (approved):
👍

PR #1670, opened 2023-06-09
1669: Handle enrollment for users with multiple PendingOrder records

collinpreston:
#### Pre-Flight checklist

- [ ] Testing
  - [ ] Code is tested

https://github.com/mitodl/mitxonline/issues/1669

#### What's this PR do?
Adds logic to handle the enrollment and creation of a PendingOrder for users who already have multiple prior PendingOrders for the same product.

1. Enroll a user into a course run
2. You should see a new pending Order created here: http://mitxonline.odl.local:8013/admin/ecommerce/order/
3. Manually create a second pending Order through Django Admin that matches the pending Order from step 2.
4. Unenroll from the course.
5. Enroll into the course again.
6. Verify that there exists only 2 pending Orders for the user.
7. Verify that the corresponding Line object in HubSpot is updated for only one of the PendingOrders

Prior to recent changes, PendingOrders were created whenever a user would start the checkout process, usually by visiting the cart page after clicking "get certificate" buttons in the application.  This has resulted in a some users having multiple PendingOrders for the same product.  Recent changes have changed this functionality so that PendingOrders are reused and only one is ever created.

### DIFF
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -609,16 +609,23 @@ class PendingOrder(FulfillableOrder, Order):
             product_content_types.append(product.content_type_id)
 
         # Get or create a PendingOrder
-        order, _ = Order.objects.select_for_update().get_or_create(
+        orders = Order.objects.select_for_update().filter(
             lines__purchased_object_id__in=product_object_ids,
             lines__purchased_content_type_id__in=product_content_types,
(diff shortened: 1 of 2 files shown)


annagav (commented):
The code looks good. Going to test functionality now

annagav (approved):
👍

PR #1672, opened 2023-06-09
Adding date checks to Discount model

### LINKED ISSUE #1630: Discount codes can be created with expiration dates in the past
### Steps to Reproduce

1. Generate a code (method doesn't necessarily matter) with the expiration date in the past.

### Expected Behavior

You're disallowed from saving the code.

### Actual Behavior

You're allowed to create the code, and thus can make a useless code.

### Related Issues

https://github.com/mitodl/hq/issues/1498

### Other Notes

The codes referenced in the related issue were bulk generated but this fix should go somewhere (like the model) where it applies universally. There's not really a good reason to create a discount code that has expired so if someone runs into an edge case where this needs to be done, they should be talking to engineering in the first place, and I think it's acceptable to require the change to be made directly in the database.

jkachel:
#### Pre-Flight checklist

- [X] Testing
  - [X] Code is tested

Closes #1630 
https://github.com/mitodl/hq/issues/1498

#### What's this PR do?

Adds some date checking to the `save` method in the Discount model:
- Expiration date, if specified, should be after activation date (if it too is specified)
- Expiration date, if specified, should not be in the past

Create a Discount with an expiration date in the past. You should receive an error.
Create a Discount with an activation date in the future, and an expiration date before that. You should receive an error.
Create a Discount with no activation or expiration date. This should be successful.
Create a Discount with either an activation date or an expiration date in the future. This should be successful.

### DIFF
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -265,7 +265,7 @@ class Discount(TimestampedModel):
             pytz.timezone(TIME_ZONE)
         ):
             raise ValidationError(
-                f"Expiration date {self.expiration_date} is in the past."
+                f"Expiration date {self.expiration_date} must be in the future."
             )
 
         if (
@@ -274,7 +274,7 @@ class Discount(TimestampedModel):
             and self.activation_date > self.expiration_date
         ):
             raise ValidationError(
-                f"Expiration date {self.expiration_date} is before the activation date {self.activation_date}."
+                f"Expiration date {self.expiration_date} must be before the activation date {self.activation_date}."
             )
 
         return True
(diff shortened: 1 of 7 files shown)


collinpreston:
Should "Expiration date, if specified, should be in the past" in your description be "Expiration date, if specified, should **not** be in the past"?

jkachel:
> Should "Expiration date, if specified, should be in the past" in your description be "Expiration date, if specified, should **not** be in the past"?

yup, fixed that

jkachel:
Refactored a bit in https://github.com/mitodl/mitxonline/pull/1672/commits/176c38626df3653ba4cb2ad68da08a03043c89cd

jkachel:
> Looks good. Nit-picky, but the only thing I would change is the error message to say:

Good call - did that in https://github.com/mitodl/mitxonline/pull/1672/commits/d42263d4ced946a55c93c3e531f09aba6efb688c (this is just a text change so will merge once it passes tests).

Issue #1664, opened 2023-06-08
Management command to create Orders for existing audit enrollments


collinpreston:
As a HubSpot user, I want to be able to view enrollments prior to the release of https://github.com/mitodl/mitxonline/pull/1644.

#### Acceptance Criteria:

- [ ] Enrollments created prior to the release of https://github.com/mitodl/mitxonline/pull/1644 should have an associated Order created.
- [ ] Once all enrollments in MITx Online have an associated Order created, those Orders should be synced with HubSpot.

#### Suggestions:

- When a user enrolls into the audit track, their Order is created here: https://github.com/mitodl/mitxonline/blob/08bae5abf6fbc98e751d2fd6981e4621b13917f5/courses/views/v1/__init__.py#L200
- This management command can be used to sync Orders with HubSpot: https://github.com/mitodl/mitxonline/blob/4121fe8f50b0be1cec8a798a2e1a6ccd8a926b5e/hubspot_sync/management/commands/sync_db_to_hubspot.py

PR #1644, opened 2023-06-01
904: tracking audit enrollments for mitx online courses in hubspot

collinpreston:
#### Pre-Flight checklist
- [x] Testing
  - [x] Code is tested
  - [x] Changes have been manually tested

#### What are the relevant tickets?
https://github.com/mitodl/hq/issues/904

#### What's this PR do?
1. When a user enrolls into a course, a Deal is created in HubSpot along with Line items that contain information on the `enrollment_mode` and `change_status` related to the user's enrollment in the associated course run.
2. When a user proceeds to upgrade or pay for the certificate track (verified) for a course run, the existing `PendingOrder` is used for initializing the checkout workflow.
3. At the time of checkout, if there are no PendingOrders associated with the user that contain Line items matching the products found in the cart, then a new PendingOrder is created.
4. When a user unenrolls from a course run, the Line item in HubSpot is updated.
5. When a user has a FulfilledOrder but their associated CourseRunEnrollment is not synced with Edx, and the user attempts to enroll into the course again (From the course about page) then no new PendingOrder shall be created.

#### How should this be manually tested?

1. Have a Course, CourseRun, User, Product, Discount, link the Discount with the Product.
2. Enroll your user into the Course from the course about page.
3. Verify that a PendingOrder is created after selecting an option on the upsell modal.
4. Verify that a HubSpot Deal is created and the associated Line displays the correct enrollment information for enrollment mode and change status.
5. Ensure that Edx is not running or is not connected with your MITx Online instance.
6. Upgrade and pay for your course enrollment using a Discount code.
7. Verify that a FulFilledOrder exists.
8. Verify that the HubSpot Deal from step 4 has been updated to reflect a closed deal.
9. Verify that the HubSpot Deal's Line from step 4 has been updated to reflect the `verified` enrollment mode.
10. From the course about page, attempt to enroll in the course again.
11. Verify that no PendingOrder is created from step 10.
12. Unenroll from your course.
13. Ensure that the HubSpot Deal's Line from step 4 has been updated to reflect the change_status of the CourseRunEnrollment.

#### Where should the reviewer start?

1. Have the required access to HubSpot for testing MITx Online locally.  Ask if you need it.
14. Configure HubSpot with your local instance of MITx Online.

#### Any background context you want to provide?

- There are a few places where I've removed unused imports that were not related to this work.
- I refactored `create_from_basket()` and `create_from_product()` to both call a private method since much of their functionality was shared.

PR #1674, opened 2023-06-12
1664: management command to create Order from enrollment

collinpreston:
#### Pre-Flight checklist

- [x] Testing

https://github.com/mitodl/mitxonline/issues/1664

#### What's this PR do?
A recent PR ([ref](https://github.com/mitodl/mitxonline/pull/1644)) now creates an Order object when a user enrolls into a course.  The Order object's information is synced with HubSpot which allows Marketing team members to view user enrollment information.  **This PR** intends to add a management command that will create an Order for each of the existing audit enrollments.

1. Setup a Course, CourseRun, Product, User, and the CMS page for the course.
2. Configure the required HubSpot environment variables in your `.env` file.
3. Enroll your user into the Audit track of the course by clicking the "Enroll now" button but not purchasing the course.
4. Through Django Admin, delete the Order associated with your enrollment: http://mitxonline.odl.local:8013/admin/ecommerce/order/
5. In a terminal and from the root of the project, run `docker-compose run web ./manage.py create_order_from_course_run_enrollments`
6. Verify that an Order is created for your enrollment.
7. Run step 5 again and confirm that only a single Order exists for your enrollment.
8. Verify that the Order details are synced with HubSpot.
9. Upgrade your course enrollment to the Verified track by paying for the course.
10. Run step 5 again and confirm that only a single fulfilled Order exists for your enrollment: http://mitxonline.odl.local:8013/admin/ecommerce/fulfilledorder/

Users that have a "verified" enrollment, will already have an Order associated with their enrollment since, prior to the recent change of Orders being created during enrollment, Orders were historically created at the time of initiating the checkout process.

A second management command that already exists in MITx Online, `sync_db_with_hubspot`, can be used to sync all of the Orders with HubSpot.

### DIFF
--- a/hubspot_sync/management/commands/create_order_from_course_run_enrollments.py
+++ b/hubspot_sync/management/commands/create_order_from_course_run_enrollments.py
@@ -7,6 +7,7 @@ from django.core.management import BaseCommand
 
 from courses.models import CourseRun, CourseRunEnrollment
 from ecommerce.models import Order, PendingOrder, Product
+from hubspot_sync.task_helpers import sync_hubspot_deal
 from reversion.models import Version
 from openedx.constants import EDX_ENROLLMENT_AUDIT_MODE
 
--- a/hubspot_sync/management/commands/create_order_from_course_run_enrollments.py
+++ b/hubspot_sync/management/commands/create_order_from_course_run_enrollments.py
@@ -6,7 +6,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.management import BaseCommand
 
 from courses.models import CourseRun, CourseRunEnrollment
-from ecommerce.models import FulfilledOrder, Order, PendingOrder, Product
+from ecommerce.models import Order, PendingOrder, Product
 from reversion.models import Version
 from openedx.constants import EDX_ENROLLMENT_AUDIT_MODE
 
@@ -33,8 +33,8 @@ class Command(BaseCommand):
                 product_version = Version.objects.get_for_object(product).first()
                 product_object_id = product.object_id
                 product_content_type = product.content_type_id
-                existing_fulfilled_order = FulfilledOrder.objects.filter(
-                    state=Order.STATE.FULFILLED,
+                existing_fulfilled_order = Order.objects.filter(
+                    state__in=[Order.STATE.FULFILLED, Order.STATE.PENDING],
                     purchaser=course_run_enrollment.user,
                     lines__purchased_object_id=product_object_id,
                     lines__purchased_content_type_id=product_content_type,
--- a/hubspot_sync/management/commands/create_order_from_course_run_enrollments.py
+++ b/hubspot_sync/management/commands/create_order_from_course_run_enrollments.py
@@ -46,3 +46,4 @@ class Command(BaseCommand):
                     PendingOrder.create_from_product(
                         product, course_run_enrollment.user
                     )
+        self.stdout.write("'sync_db_to_hubspot --deals create' should be run next. \n")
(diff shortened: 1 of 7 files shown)


mbertrand (commented):
One thing I noticed though is that `PendingOrder.create_from_product` is called for every audited enrollment, even if a pending order already exists. Although it doesn't create a dupe order (just gets the existing one), it still triggers a hubspot sync.  Probably not a big deal, just maybe a bit inefficient.  However, assuming there are thousands of missing orders on production, I wonder if there's a chance we might go over the hubspot rate limit.  That can be fixed later though by calling sync_db_to_hubspot to batch create/update them.

hubspot_sync/management/commands/create_order_from_course_run_enrollments.py
```diff
@@ -0,0 +1,47 @@
...
+                product_object_id = product.object_id
+                product_content_type = product.content_type_id
+                existing_fulfilled_order = FulfilledOrder.objects.filter(
+                    state=Order.STATE.FULFILLED,
+                    purchaser=course_run_enrollment.user,
+                    lines__purchased_object_id=product_object_id,
+                    lines__purchased_content_type_id=product_content_type,
+                    lines__product_version=product_version,
+                )
+                if not existing_fulfilled_order:
```
mbertrand, line 43: Maybe query for a PendingOrder too and then:

```python
if not existing_fulfilled_order and not existing_pending_order:
```

?

mbertrand: Or are there just not that many and this will probably just be run once anyway?

collinpreston:
> One thing I noticed though is that `PendingOrder.create_from_product` is called for every audited enrollment, even if a pending order already exists. Although it doesn't create a dupe order (just gets the existing one), it still triggers a hubspot sync. Probably not a big deal, just maybe a bit inefficient. However, assuming there are thousands of missing orders on production, I wonder if there's a chance we might go over the hubspot rate limit. That can be fixed later though by calling sync_db_to_hubspot to batch create/update them.

@mbertrand Yeah, that's a good point.  Do you think there is a way to avoid that?  Right now it's triggering the HubSpot sync when the Order is saved.

mbertrand:
> Do you think there is a way to avoid that? 

Too avoid the hubspot syncing, maybe `Order.save()` could take an optional_kwarg `hubspot_sync: bool = True` also add that kwarg to `PendingOrder._get_or_create` to pass it on from there, but with `hubspot_sync=True`

collinpreston:
@mbertrand Okay, so your comment had me recognize that `Order.save()` was being called twice within `PendingOrder._get_or_create()`, which would result in `sync_deal_with_hubspot()` being called twice.  After reading one of Nathan's recent discussion posts, I also began to think that calling `sync_deal_with_hubspot()` within `Order.save()` was restricting developers (as we ran into with my original design for this PR) but also stuffing more non-related functionality into `Order.save()` which should really remain pretty basic.  

So I removed `sync_deal_with_hubspot()` from within `Order.save()` and added calls to `sync_deal_with_hubspot()` wherever Orders were created or updated in more outer layer/view level methods.  While this requires more responsibility on the developer to ensure that `sync_deal_with_hubspot()` is called when Orders are created or updated, it also keeps much of our business logic in the controller level of our apps.  

This also avoids calling `sync_deal_with_hubspot()` for each Order created when running the create_order_from_course_run_enrollments command.

When testing this, I did notice that running `sync_db_to_hubspot --orders create` is throwing an error: `VALIDATION_ERROR`.  However that error is also occurring with the main branch of MITx Online.

mbertrand:
Sounds like a good approach to me!  

For your local instance, try running `sync_hubspot_ids_to_db` first before running `sync_db_to_hubspot`, also make sure that `MITOL_HUBSPOT_API_ID_PREFIX` in your .env file is not the same value as on RC

pdpinch:
How far back in time is this going to go?

collinpreston:
> How far back in time is this going to go?

All the way back to the first enrollment.

Issue #1643, opened 2023-06-01
Dashboard and Program record are showing different things


annagav:
### Steps to Reproduce

The dashboard of the user in ticket https://github.com/mitodl/hq/issues/1512 is showing nothing for the course 14.310 while they do have a certificate for it.  As a result the user cannot view the certificate.
The grade records are showing the course grade correctly.

<img width="574" alt="Screen Shot 2023-06-01 at 8 45 23 AM" src="https://github.com/mitodl/mitxonline/assets/7574259/c8ad4b05-ce4c-4a01-bb26-7e91e3e0799b">
<img width="1206" alt="Screen Shot 2023-06-01 at 8 45 42 AM" src="https://github.com/mitodl/mitxonline/assets/7574259/c32eda49-e505-4810-8fa8-2613f7d558c9">

annagav:
I found another case (in Microeconomics course) with this bug https://github.com/mitodl/hq/issues/1549

rachellougee:
Looked into the discrepancies between Dashboard and Program record

For [mitodl/hq/issues/1512](https://github.com/mitodl/hq/issues/1512), this user is currently enrolled in two runs  `course-v1:MITx+14.310x+2T2021` and `course-v1:MITxT+14.310x+1T2022` for course 14.310x, but there is only grade for 1T2022 (not 2T2021), this is how it looks like under My Course on the dashboard for this user, one is

![image](https://github.com/mitodl/mitxonline/assets/3138890/707ad037-4b7a-4a4b-8d75-9ad3c024180d)



For [mitodl/hq/issues/1549](https://github.com/mitodl/hq/issues/1549), this user is currently enrolled in `course-v1:MITxT+14.100x+2T2022`, but there is only passing grade for `course-v1:MITx+14.100x+1T2021`, so it's not displaying passed tag for course 14.310x because no passing grade for the enrolled run `course-v1:MITxT+14.100x+2T2022`

<img width="791" alt="image" src="https://github.com/mitodl/mitxonline/assets/3138890/21c63dd7-d176-4166-8e0f-8bb6b20294c0">

Their enrollments/grades are from MicroMasters, I don't know why sometimes users have grades (finalgrade) for runs they didn't enroll (dashboard_cachedenrollment) on MM. 
 
On the dashboard, each card is corresponding to that specific run (enrollment/grade needs to be matched for the run). On the program record, it's pulling the grade/certificate based on the course. I think this is why we see the difference between the two.

[...]

PR #1687, opened 2023-06-23
fix bug on renderCourseInfoCard and add migration to backfill certificate enrollments

rachellougee:

https://github.com/mitodl/mitxonline/issues/1643

 - Fixing bug on renderCourseInfoCard on program drawer. 
      - Current behavior: If a user has enrolled multiple enrollments for the course and has earned a certificate, it displays the most recent enrollment for the course, not necessarily the one with a certificate. This caused inconsistency between the dashboard and program record because user has no way to view the certificate on the program dashboard but then it says "certificate earned" on the program record page
      -  Expected behavior: It should always display the course card with a certificate link and course passed tag if user has earned a certificate for any run. If none of the enrollments has a certificate, it defaults to the most recent enrollment which is the current behavior.
 -  Adding a migration to backfill the missing enrollments for certificates. 
    -  There are 98 in production https://bi.odl.mit.edu/queries/1290, these are all old DEDP certificates

# How can this be tested?

- Setup course, and program if not exist
- Add courserun enrollment from admin/courses/courserunenrollment/ for program course
- Add passing grade for the above run from admin/courses/courserungrade/
- Add courserun certificate for the above run from admin/courses/courseruncertificate/. 
   - Note that for this step, you will need to set up certificates in CMS following https://github.com/mitodl/mitxonline#certificates
-  Repeat step 2 to add another course run enrollment for the same course, but this time don't create a grade or certificate for it
-  Go to your dashboard to view your program course, you should see something like this
    
image

This should be consistent with the program record 
image


annagav (approved):
Works for me 👍

PR #1685, opened 2023-06-22
Remove program field from course model

collinpreston:
# What are the relevant tickets?
https://github.com/mitodl/mitxonline/issues/1683
https://github.com/mitodl/hq/issues/1663
https://github.com/mitodl/mitxonline/issues/1464
https://github.com/mitodl/mitxonline/pull/1480
https://github.com/mitodl/mitxonline/issues/1220
https://github.com/mitodl/mitxonline/issues/1681

# Description
Removes the program FK field from the Course model.  The Program/Course relationship is derived from the requirements and electives defined on the Program instance.

There is a new program relationship table created which is a link table for associating 2 programs together.  By associating 2 programs together in this link table, financial aid approvals in one course will be applied to the linked program's courses.

# How can this be tested?
Non-Related Program

 Can create program, tiers, courseware page, and financial assistance request form
 Can create a course as a requirement in the program, with products and an about page
 Can create a course as an elective in the program, with products and an about page
 "Financial assistance" CtAs appear in the appropriate spots - in the program dashboard cards, in the upsell cards, etc.
 Can submit and approve a financial assistance request for the program
 Personalized pricing displays appropriately on the course about pages
 Can enroll in the course programs
 After enrolling, the program appears successfully in the dashboard and displays all associated courses within the program
 Upgrading course enrollment results in the proper financial assistance tier being applied as discount
Related Programs, No Overlapping Courses

 Can create program, tiers, courseware page, and financial assistance request form
 Can create a course as a requirement in the program, with products and an about page
 Can create a course as an elective in the program, with products and an about page
 "Financial assistance" CtAs appear in the appropriate spots - in the program dashboard cards, in the upsell cards, etc.
 Can submit and approve a financial assistance request for the program
 Personalized pricing displays appropriately on the course about pages
 Can enroll in the course programs
 After enrolling, the program appears successfully in the dashboard and displays all associated courses within the program
 Upgrading course enrollment results in the proper financial assistance tier being applied as discount
 Can add a second program to the system, with a relation to the first, and complete the first three steps for it
 Personalized pricing appears in the courses assigned to the second program
 Can create a secondary financial assistance request page for the second program
 Attempting to apply for financial assistance should be disallowed
This delivered me to the approved page for the proper program (test program 2 rather than 1)
 The dashboard should automatically see both related programs
 Enrolling in a course in the second program should result in a program enrollment for the second program; one should not exist prior to this
[...]

PR #1690, opened 2023-06-23
Fixes the course count on cards in My Programs

jkachel:

Part of #1685 

The course count in the program cards in the My Programs tab was incorrect - it was looking through the program requirements, but we have a flat list of courses in the data structure that can be used instead.

# How can this be tested?

Add a program with a nested operator and course requirements within that operator. (Use DEDP as an example of this - you can take one elective, or one of a set of two electives, to satisfy the electives requirement.) The course count should display the correct number.


pdpinch:
Where is the course count displayed?

collinpreston:
> Where is the course count displayed?

Dashboard, "My programs" tab, program card.

collinpreston (approved):
Nice and cleaner solution.   Tested locally.

PR #1701, opened 2023-06-26
Remove call to `transaction.rollback` if the payment gateway does something unexpected

### LINKED ISSUE #1686: Transaction rollback fails during refund processing if the app gets a response it's not expecting
When processing a refund for order mitxonline-production-18147, CyberSource sent back a successful result of "Created". This was unexpected - the code looks for "Accept" or "Pending", so it attempted to roll back what it was doing and failed to do so. The code is wrapped in `transaction.atomic()` so the explicit rollback is the problem. Simply raising the exception should cause the rollback to happen. 

As part of this, we should _not_ at this point adjust the code to accept a "Created" response. (As I am writing the ticket, I'm unsure if the transaction was refunded or not - Created is ambiguous.)

### Steps to Reproduce

1. Create a transaction in the system. This should be using the normal credit card flow, not PayPal, which will fail for other reasons.
2. Attempt to refund it through the Google Sheets workflow.

### Expected Behavior

The refund should complete successfully.

### Actual Behavior

The refund fails with an exception: `This is forbidden when an 'atomic' block is active.`. 

### Stacktrace

https://mit-office-of-digital-learning.sentry.io/issues/4268494759/?project=5864687&query=is%3Aunresolved&referrer=issue-stream&stream_index=15

### Related Issues

https://github.com/mitodl/hq/issues/1757

jkachel:
It's in a `with` block so explicit rollbacks aren't necessary or supported.

Closes #1686 

During a refund, if the PaymentGateway gets an unexpected response from the processor, it gives up processing the refund. It was explicitly rolling back the transaction in this case, though, which was unnecessary.

# How can this be tested?

This appears to happen when CyberSource returns "Created" rather than something we expect. I'm not sure how to fake that specific case, but re-running a processed transaction should also generate an error, I think, which should hit the same code branch.

1. Complete an order.
2. Attempt to refund the order using the Google Sheets system. 
3. The refund should fail successfully.

### DIFF
--- a/ecommerce/api.py
+++ b/ecommerce/api.py
@@ -8,7 +8,6 @@ from django.core.exceptions import ObjectDoesNotExist, PermissionDenied, Validat
 from django.db import transaction
 from django.db.models import Q
 from django.urls import reverse
-from hubspot_sync.task_helpers import sync_hubspot_deal
 from ipware import get_client_ip
 from mitol.common.utils.datetime import now_in_utc
 from mitol.payment_gateway.api import CartItem as GatewayCartItem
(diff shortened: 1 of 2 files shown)


jkachel:
Closing, Anna also ended up doing something similar in https://github.com/mitodl/mitxonline/pull/1706

Issue #1702, opened 2023-06-26
Line serializer cannot handle Orders with no current enrollment


collinpreston:
Using the command introduced in this PR: https://github.com/mitodl/mitxonline/pull/1674
### Steps to Reproduce

1. Create a user, course, course run, product.
2. Create an order for your user referencing the items from step 1.
3. Run the following management command `sync_db_to_hubspot --deals create`.

### Expected Behavior

Line items are synced with HubSpot even if the user does not have a CourseRunEnrollment record for an existing Order.

### Actual Behavior

https://mit-office-of-digital-learning.sentry.io/issues/4276444301/?project=5864687

PR #1703, opened 2023-06-26
1702: line serializer cannot handle orders with no current enrollment

collinpreston:
https://github.com/mitodl/mitxonline/issues/1702

Until recently, the Line serializer did not rely on CourseRunEnrollment.  There is the possibility that old data exists for a learner who unenrolls or has no enrollment for an Order and Line.

# How can this be tested?

1. Create a user, course, course run, product.
2. Create an order for your user referencing the items from step 1.
3. Run the following management command `sync_db_to_hubspot --deals create`.


annagav (approved):
LGTM

PR #1651, opened 2023-06-05
Convert grade into letter for program records

annagav:
#### Pre-Flight checklist

- [ ] Screenshots and design review for any changes that affect layout or styling
  - [x] Desktop screenshots
  - [ ] Mobile width screenshots

#### What are the relevant tickets?
Fix https://github.com/mitodl/mitxonline/issues/1647

#### What's this PR do?
Populates all passing grades missing letter grade before Fall 2022 semester by a fixed formula.

#### How should this be manually tested?
Create a grade for your user. Go to the program record view to see how it is displayed.
Make sure there is a passing grade for a course run that ended before Sept 1st 2022, where letter_grade is set to 'Pass'
Then run migrations and make sure that the letter_grade was assigned a letter.
Screen Shot 2023-06-05 at 2 00 35 PM

Issue #1647, opened 2023-06-02
Program Record: Show letter grades instead of pass 


annagav:
Few users have complained that they would like to see letter grades instead of "PASS" in their program record in MITxonline dashboard.
Related tickets: https://github.com/mitodl/hq/issues/1577, https://github.com/mitodl/hq/issues/1574
In MM:
<img width="1341" alt="Screen Shot 2023-05-15 at 8 27 48 AM" src="https://github.com/mitodl/mitxonline/assets/7574259/c596effc-6d52-4dc3-b269-4e6ed149c7cb">

In MITxOnline:
<img width="340" alt="Screen Shot 2023-06-02 at 5 04 09 PM" src="https://github.com/mitodl/mitxonline/assets/7574259/d5eef493-7076-48c7-af1c-e08b30a95329">

pdpinch:
@annagav please add some detail about the root cause and the plan to fix it.

annagav:
I think the root cause of this is that we use the letter grade that is in the model CourseRunGrade.letter_grade to display in the Program Record. This is very inconsistent, some grades are showing letter, some are showing 'pass'.

We should compute the letter grade based on the score, if `letter_grade` is 'pass'.

pdpinch:
I think the problem here is that we never stored the letter grades in MicroMasters, and left it to a runtime calculation. When we migrated the data, there were no letter grades to migrate. @rachellougee can you confirm?

Let's not perpertuate the same mistake here. Instead of hard-coding the grade bands in the serializer as #1651 does, we should migrate the data. 

At the end of the day, the letter grades should follow the grading policy as defined by the course team. Since 1T2023 they can do this in open edX Studio.

pdpinch:
I'd like to know which course runs we are missing letter grades for. 

- For most of the legacy course DEDP runs, we should follow the grading scheme that was in use in MicroMasters, for consistency. 
- I'm uncertain what to do with the DEDP grades from 3T2021, 1T2022, 2T2022 and 3T2022 grades, so I'd like to be able to look at those first. In particular, we have to be careful about passing status. 
- We shouldn't change any grades that aren't from DEDP migrated data. I don't know offhand what the grading configurations are for non-DEDP courses, and we should defer to the open edX configuration for those.

rachellougee:
@annagav https://bi.odl.mit.edu/queries/1289 Here are DEDP course IDs that have a "Pass" letter grade

annagav:
[...]

PR #1704, opened 2023-06-26
fix courses migration conflict

rachellougee:

https://github.com/mitodl/mitxonline/pull/1651
https://github.com/mitodl/mitxonline/pull/1687

Fixes course migrations conflict on the main branch

`
Static file directory was missing: /home/runner/work/mitxonline/mitxonline/frontend/public/build Static file directory was missing: /home/runner/work/mitxonline/mitxonline/frontend/staff-dashboard/build CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0031_backfill_enrollments_from_certificates, 0031_populate_letter_grades, 0032_add_related_programs_table in courses). To fix them run 'python manage.py makemigrations --merge'
`

# How can this be tested?

Run migrations from 0030
```
manage.py migrate --fake courses 0029 (run this if you have already applied 0030_remove_course_position_in_program.py)
manage.py migrate courses 0030
 ```

### DIFF
--- a/courses/migrations/0033_populate_letter_grades.py
+++ b/courses/migrations/0033_populate_letter_grades.py
@@ -5,7 +5,21 @@ import pytz
 from django.db import migrations
 
 from courses.models import Program
-from courses.utils import convert_to_letter
+
+
+def convert_to_letter(grade):
+    """Convert a decimal number to letter grade"""
+    grade = round(grade, 1)
+    if grade >= 0.825:
+        return "A"
+    elif grade >= 0.65:
+        return "B"
+    elif grade >= 0.55:
+        return "C"
+    elif grade >= 0.50:
+        return "D"
+    else:
+        return "F"
 
 
 def populate_letter_grade(apps, schema_editor):
@@ -27,11 +41,11 @@ def populate_letter_grade(apps, schema_editor):
     )
     for grade in grades:
         grade.letter_grade = convert_to_letter(grade.grade)
+        grade.set_by_admin = True
         grade.save()
 
 
 class Migration(migrations.Migration):
-
     dependencies = [
         ("courses", "0032_add_related_programs_table"),
     ]
--- a/courses/migrations/0033_populate_letter_grades.py
+++ b/courses/migrations/0033_populate_letter_grades.py
@@ -9,12 +9,15 @@ from courses.utils import convert_to_letter
 
 
 def populate_letter_grade(apps, schema_editor):
-    """Populate the certificate_available_date from course run's end_date"""
+    """Populate the letter grade with A/B/C/D for the DEDP courses prior to 3T2022"""
     CourseRunGrade = apps.get_model("courses", "CourseRunGrade")
 
-    dedp_program = Program.objects.filter(readable_id="program-v1:MITx+DEDP").get()
+    dedp_program = Program.objects.filter(readable_id="program-v1:MITx+DEDP").first()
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: courses/migrations/0033_populate_letter_grades.py [hunk 1]
```diff
@@ -4,16 +4,22 @@
...
 
+from courses.models import Program
 from courses.utils import convert_to_letter
 
 
 def populate_letter_grade(apps, schema_editor):
     """Populate the certificate_available_date from course run's end_date"""
     CourseRunGrade = apps.get_model("courses", "CourseRunGrade")
+
+    dedp_program = Program.objects.filter(readable_id="program-v1:MITx+DEDP").get()
```
collinpreston, line 15: Will this break locally if a developer does not have this program configured?  If so, this might be difficult for someone to recognize when first setting up their environment.

courses/migrations/0033_populate_letter_grades.py
```diff
@@ -4,16 +4,22 @@
 import pytz
 from django.db import migrations
 
+from courses.models import Program
 from courses.utils import convert_to_letter
 
 
 def populate_letter_grade(apps, schema_editor):
     """Populate the certificate_available_date from course run's end_date"""
```
collinpreston, line 12: I know it wasn't part of this PR, but could you update this docstring to match the function?

collinpreston (changes requested):

courses/migrations/0033_populate_letter_grades.py
rachellougee: good call. I will update to use `.first()` so it will not throw error if no such program locally

collinpreston:
Does the conversion algorithm in `convert_to_letter()` apply to all courses or just courses in this specific DEDP program?

rachellougee:
> Does the conversion algorithm in convert_to_letter() apply to all courses or just courses in this specific DEDP program?

It should only apply to courses (before 3T2022) in the existing DEDP program. The original PR https://github.com/mitodl/mitxonline/pull/1651 doesn't specify that in the description, but its indicated in https://github.com/mitodl/mitxonline/issues/1647#issuecomment-1582499637 and we don't change letter grade for any other courses

collinpreston (commented):
Should we be setting `grade.set_by_admin` to True after updating the letter grade?

collinpreston:
> > Does the conversion algorithm in convert_to_letter() apply to all courses or just courses in this specific DEDP program?
> 
> It should only apply to courses (before 3T2022) in the existing DEDP program. The original PR #1651 doesn't specify that in the description, but its indicated in [#1647 (comment)](https://github.com/mitodl/mitxonline/issues/1647#issuecomment-1582499637) and we don't change letter grade for any other courses

I would consider whether that function will ever be reused and therefore added as a utility.  If it's likely to be reused, I think we should very clearly indicate that it will only apply to courses in that DEDP program which were before 3T2022.

rachellougee:
`convert_to_letter` is added to utils.py @annagav should have more context on this since I am not the one who created that PR. This is to address the merge conflict for migration, but I am happy to address other feedback you might have for https://github.com/mitodl/mitxonline/pull/1651

rachellougee:
addressed above comments in https://github.com/mitodl/mitxonline/pull/1704/commits/3f4de4b159008add2abb47997397cd76d07fe799

collinpreston (approved):
Looks good to me.  Thanks for the additional stuff

Issue #1705, opened 2023-06-26
Refund Order: refund as an atomic transaction doesn't work.


annagav:
### Steps to Reproduce

In google sheets two transactions showed error for trying to rollback a transaction refund that was not successful.

### Expected Behavior

If call to Cybersource for refund is not successful, throw an error with the precise error message. There are no database transaction to roll back.

### Actual Behavior
If call to Cybersource for refund is not successful, it reverts the database transactions, and shows a error message about trying to rollback inside of an atomic block.

Reverting at this point does not help, because cybersource complains if you try to refund again: `Error: Declined - The transaction identifier sent with this request has already been used in the last 60 days`

PR #1706, opened 2023-06-26
Refund Order: remove atomicity and rollback

annagav:

Fixes # https://github.com/mitodl/mitxonline/issues/1705

remove the atomic block and rollback

# How can this be tested?
tests should pass

# Additional Context
The atomic block is not needed since we are not updating the database. After the api call to Cybersource was made usually a second attempt doesn't work: `Error: Declined - The transaction identifier sent with this request has already been used in the last 60 days`


rachellougee (approved):
tested refund order, it looks good now

Issue #1700, opened 2023-06-26
Increase efficiency for program enrollments API


jkachel:
The program enrollments API call (`/api/program_enrollments`) is pretty slow - testing locally, it generates a bunch of n+1 errors and it takes many seconds to resolve (tested at ~10s with debug turned on and ~3 with it turned off).

### Steps to Reproduce

Observe the program enrollments API.

### Expected Behavior

It returns enrollments in a reasonable timeframe.

### Actual Behavior

It takes several seconds to return data.

pdpinch:
Seems like this is still very slow, at least on rc.mitxonline.mit.edu. What more can we do?

jkachel:
#1708 wasn't intended to be the last word in this; there's still a bunch of n+1s that need investigation and we can/should also look at how the frontend uses the APIs, and either optimize that usage, create new/refactor existing APIs that provide the data, or do more processing on the client side. (Or maybe all of those.)

PR #1708, opened 2023-06-27
Optimize program enrollments API

jkachel:

#1700 

Refactors some queries in the `ProgramSerializer` and `UserProgramEnrollmentDetailsSerializer` and `UserProgramEnrollmentsViewSet` so they should be more efficient. This also removes the implicit program enrollments from the view set - if you're not explicitly enrolled in a program, you won't see it anymore (we should have explicit enrollments currently, though). 

There are more things that could be optimized; I believe the amount of data getting pulled in through CourseSerializer is somewhat excessive too but this should result in a bit of an improvement for now. 

# How can this be tested?

As a user with enrollments in programs, load `/api/program_enrollments`. It should return reasonable data and should do so faster than the same call on the `main` branch. For best results, test each a few times sequentially. 

Additionally, the Dashboard should work as expected. Programs should appear normally; their courses should appear normally; and certificates should display as usual.


pdpinch:
I don't want to slow this down, but how/can we monitor performance after the new program code is released?

courses/serializers.py
```diff
@@ -390,6 +388,56 @@ class Meta:
         ]
 
 
+class SimpleProgramSerializer(serializers.ModelSerializer):
```
collinpreston, line 391: Does it make sense to update the existing ProgramSerializer?

collinpreston (commented):
Looks good, just one comment I'm curious about.

courses/serializers.py
jkachel: https://github.com/mitodl/mitxonline/pull/1708/commits/5e18fc739fb27ba420f76e75160854d56af5591b does that; there were other fields that the current frontend doesn't use that got pulled into `FullProgramSerializer` that inherits from `ProgramSerializer`.

collinpreston (approved):
Nice, looks good and is speedy quick!

PR #1709, opened 2023-06-28
Merging migrations from #1708 and #1704

jkachel:

#1708 and #1704

Both these PRs had migrations at the same time, so there needed to be a merge; this just does that.

# How can this be tested?

Everything should work properly - tests should pass, etc.


rachellougee (approved):
👍

PR #1716, opened 2023-06-29
remove non-standard issue templates

pdpinch:

n/a

Removes the issue templates custom to this repo. Instead, it will use the standard mitodl issue templates. 

# How can this be tested?

AFAIK, there is no way to test it. We have to merge it and see what happens.


JenniWhitman:
Sounds good to me - worse comes to worse we put them back - version control helps :)

JenniWhitman (approved):
Removing them seems like a good course to me/what I found in my 5 minutes of research

PR #1719, opened 2023-06-30
Update MM grade import, add missing enrollment import script

rachellougee:

https://github.com/mitodl/hq/issues/1561

This PR is to modify `letter_grade` in the grade script for MicroMasters import. It also adds a new script to populate the missing enrollments after certificates migration as enrollments might be deleted from MM

# How can this be tested?

Run `docker-compose run --rm web ./manage.py import_micromasters_data --num 013` for the new script
If you need to run other scripts, please follow [this](https://github.com/mitodl/mitxonline/tree/main/micromasters_import#local-setup) instruction 

# Additional Context
There are some MM users who register on MITx Online after we ran the initial migration in Oct 2022, those users data are still in MicroMasters so we need to run micromasters_import script to bring over their enrollment/grades/certificates


annagav (approved):
I was able to populate missing enrollment from a certificate for a course. 👍

PR #1721, opened 2023-06-30
Clears discounts when reusing an existing PendingOrder

### LINKED ISSUE #1715: Order have more than one discount redemption
### Steps to Reproduce

There are 22 orders with more than one discount redemption. See https://bi.odl.mit.edu/queries/1293. Some orders have the same coupon applied to the order multiple times (they look like duplicates)

e.g.
https://mitxonline.mit.edu/admin/ecommerce/order/17428/change/ 
https://mitxonline.mit.edu/admin/ecommerce/order/19464/change/

### Expected Behavior

Order shouldn’t have more than one discount redemption

### Actual Behavior
22 orders have more than one discount redemption, especially the same coupon applied to the order multiple times

### Related Issues

https://github.com/mitodl/hq/issues/1765

jkachel:

Closes #1715 

When a learner attempts to check out, their basket is converted to a pending Order. If there's already a pending Order for the learner with the same items in it, the system will reuse that existing Order. This is fine, but if the pending Order had discounts applied to it, those discount redemptions will stick around in the new order - they don't have an effect on the amount charged but they do count as redemptions for the discount, which could potentially result in a discount code becoming invalid when it shouldn't. (And, it definitely causes issues for reporting as we only support a single discount per order.) 

This adds some code to delete any DiscountRedemptions that exist for the order if it's being reused. Any discounts applied are applied to the user's basket first, so they will just be copied back into the order when the basket is converted into an order (which includes automatically applied discounts and financial assistance). 

# How can this be tested?

1. Set up a cart with a product in it.
2. Apply a discount (either manually or via financial assistance). Do not apply a discount that fully discounts the cart - doing this will cause it to automatically process, so your order won't stay in the Pending state. 
3. Proceed to the payment processor, and stop.
4. In a separate window/session/etc., pull up the resulting Order record in Django Admin. The order should have one discount redemption on it. 
5. In the first window, go back to the MITx Online system and set up the same cart again. Do not cancel out from CyberSource - either use the Back button or type the URL in.
6. Apply the same discount.
7. Proceed to the payment processor, and then stop.
8. In the Django Admin window, check the order again. You should see only the new discount redemption.

### DIFF
--- a/ecommerce/models.py
+++ b/ecommerce/models.py
@@ -640,12 +640,16 @@ class PendingOrder(FulfillableOrder, Order):
             product_content_types.append(product.content_type_id)
 
         # Get or create a PendingOrder
-        orders = Order.objects.select_for_update().filter(
-            lines__purchased_object_id__in=product_object_ids,
-            lines__purchased_content_type_id__in=product_content_types,
-            lines__product_version__in=product_versions,
-            state=Order.STATE.PENDING,
-            purchaser=user,
+        orders = (
+            Order.objects.select_for_update()
+            .prefetch_related("discounts")
+            .filter(
+                lines__purchased_object_id__in=product_object_ids,
+                lines__purchased_content_type_id__in=product_content_types,
+                lines__product_version__in=product_versions,
+                state=Order.STATE.PENDING,
+                purchaser=user,
+            )
         )
         # Previously, multiple PendingOrders could be created for a single user
         # for the same product, if multiple exist, grab the first.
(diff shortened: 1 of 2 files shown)


collinpreston (commented):
Looks good, just a couple questions:
- Does it make sense to add a `prefetch_related("discounts")` when filtering the Orders?

jkachel:
> Looks good, just one question.
> 
>     * Does it make sense to add a `prefetch_related("discounts")` when filtering the Orders?

Yup - got this in https://github.com/mitodl/mitxonline/pull/1721/commits/6ffb1981ddba63936f11b3f4759cd24f9bb1232b

collinpreston (approved):
Looks good to me

PR #1722, opened 2023-06-30
add migration to fix letter grade set by 0033

rachellougee:

https://github.com/mitodl/hq/issues/1691

The letter_grade set by [0033_populate_letter_grades](https://github.com/mitodl/mitxonline/blob/main/courses/migrations/0033_populate_letter_grades.py#L12C1-L12C1) is not correct. It rounded the grade between 0.8 to 0.85 to 0.8, so 0.834 would round to 0.8 and have B which should be A. The grades don't need to be rounded. This PR is to re-run the migration.

# How can this be tested?

Create a program (program-v1:MITx+DEDP) and program requirements
Create program courses, runs and make sure one of the runs start date is before 2022-09-01
Create a passing grade record for that run and make sure set_by_admin is checked
Run migration 
Check grade
image
You can also test grades in other range and make sure A/B/C/D is updated correctly


annagav (approved):
Works for me 👍 
Grade was correctly updated
Screen Shot 2023-06-30 at 5 54 46 PM

Issue #1671, opened 2023-06-09
In the google sheets refunds some rows are being ignored when they shouldn't


annagav:
### Expected Behavior

If the row has no 'completed date', no 'Ignore', and no errors, then it should should be processed.

### Actual Behavior

In the google sheets refunds few rows are being ignored even though now error, not ignore setting, no completed date is there.
<img width="816" alt="Screen Shot 2023-06-09 at 11 18 53 AM" src="https://github.com/mitodl/mitxonline/assets/7574259/35b36bb5-80ed-40c6-9d33-c5a6a1eaa612">

PR #1726, opened 2023-07-05
Refund Order should return propper error message

annagav:

Fixes https://github.com/mitodl/mitxonline/issues/1671

Updates the refund_order to return error messages along with the success value.

# How can this be tested?
Have google sheets refunds setup and running locally, being processed by your local mitxonline instance.
Submit a request that has order reference number missing, or missing transaction for the order, or order not being in a "Fulfilled" state.
In the Refunds tab you should be able to see these errors in the error column

### DIFF
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -335,7 +335,7 @@ class AdminRefundOrderView(LoginRequiredMixin, PermissionRequiredMixin, Template
                 refund_amount = refund_form.cleaned_data.get("refund_amount")
 
                 # Call the refund CyberSource API with provided reason and amount
-                refund_api_success = refund_order(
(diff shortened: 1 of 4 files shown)


rachellougee (commented):
I tested the refunds without order numbers, but I don't see all errors are written into Errors columns in google sheets. See screenshot below, some rows are processed with errors `Ignored rows in Enrollment Change Request sheet (Refunds): [4, 5, 9]`, but the other 3 rows don't have anything, for these 3 I do see errors are logged in celery with `Either order_id or reference_number is required to fetch the Order`

image

image

@annagav Can you take a look and let me know if I am missing anything

rachellougee (approved):
It works for me now. I thought I did the same thing yesterday. Sorry for the confusion.

PR #1732, opened 2023-07-07
add migration to delete duplicate discount redemption

rachellougee:

https://github.com/mitodl/hq/issues/1765

Adds a migration to delete the duplicate discount redemptions on production https://bi.odl.mit.edu/queries/1293.
It keeps the last DiscountRedemption per order.

https://github.com/mitodl/mitxonline/pull/1721 has fixed the logic, but we also need to fix the data. 

# How can this be tested?

- Create orders
- Create duplicate DiscountRedemption for same orders either from shell or from admin. 
   - To add from admin, you will need to comment out `readonly_fields` in `DiscountRedemptionAdmin`
- Run migration
- Verify from http://mitxonline.odl.local:8013/admin/ecommerce/discountredemption/ . There should be only one redemption per order. For duplicates, it should keep the most recent one.

Before Migration
image

After migration
image


collinpreston (approved):
This looks good to me.  I'm curious, what's the reason for having this as a migration instead of running it via Django shell?

rachellougee:
> what's the reason for having this as a migration instead of running it via Django shell?
In this case, it makes sense to do a migration because it's a one-time fix for existing data. This is a bug that should not happen again since we have fixed the logic in the code at https://github.com/mitodl/mitxonline/pull/1721. My understanding of management command is if we plan to use it in the future as part of maintenance.

PR #1736, opened 2023-07-11
Refactors serializers/views for speed in enrollment APIs, adds flag to designate requirements as electives or not

### LINKED ISSUE #1695: The program requirement operation titles should not be used in logic
As a program administrator, when adding requirements to a program, defining the title of the operators in the requirements tree should be hard coded and not left up to the administrator.  This will ensure that requirements operator titles such as "Elective courses", "Required courses", and "One of" are reliably referenced throughout the code of our application.

#### Acceptance Criteria:

- [ ] Remove dependency on program requirement and elective operator titles for logic in code.
- [ ] Documentation is updated to reflect this.

### LINKED ISSUE #1700: Increase efficiency for program enrollments API

jkachel:

Closes #1695 
Closes #1700 

For #1695 -
- Adds `electives_flag` - if `depth=2` in a ProgramRequirement node, this means the child course nodes are electives (and thus eliminates the need to depend on the title for this distinction). 

For #1700 - 
- Optimizes queries in a few places to make them pull data with less nested queries
- Culls some of the recursion that was being used; we were walking the requirements tree to find things and there were more efficient ways to do this (e.g. using `path` in combination with the new `electives_flag`). 

Testing this locally dropped the response time for `/api/program_enrollments` from ~11-12s to ~1.2s (with debug set to True) or ~3-5s to ~1.2s (with debug set to False). This was running thousands of queries before; it should be much less now.

# How can this be tested?

* Automated tests should work.
* All functionality should work as expected - the main thing that uses the program enrollments API is the My Programs tab and program drawer in the dashboard. In addition, certificate generation should work as normal.


collinpreston (commented):
Just started reviewing this, but would you be able to also update this section of documentation to include the elective flag?  https://docs.google.com/document/d/1kw-ChTQK2YZZr8QjexhQX7P-IH5wwQeXNr4xxG0i2r8/edit#heading=h.iuzfhhiijaeo

collinpreston (commented):
I need to check if this is occurring in the main branch:
- Unenrolling from all courses belonging to a program results in the program tab on the dashboard displaying a blank page.
- Enrolling into a course run associated with a program, unenrolling from the program, unenrolling from the course run, and then enrolling in the course run again does not result in the learner being enrolled in the program again.

courses/models.py
```diff
@@ -301,7 +302,33 @@ def courses(self):
...
+                .select_related("course", "course__page")
+                .all()
+            )
+
+            heap.extend(
+                [
+                    (
+                        req.course,
+                        "Required Courses"
+                        if not op.elective_flag
```
collinpreston, line 324: Can the requirement operator be used instead of elective_flag to determine whether the course is a requirement or elective? `ProgramRequirement.Operator.MIN_NUMBER_OF.value` vs `ProgramRequirement.Operator.ALL_OF.value`?

courses/models.py
```diff
@@ -301,7 +302,33 @@ def courses(self):
         - list of tuple (Course, string): courses that are either requirements or electives, plus the requirement type
         """
 
-        return self._req_course_walk(self.requirements_root, [])
```
collinpreston, line 304: Can we remove `_req_course_walk` since it's not used anywhere?

courses/forms.py
```diff
@@ -145,6 +145,12 @@ def program_requirements_schema():
                                     },
                                 },
                             },
+                            "elective_flag": {
+                                "type": "boolean",
+                                "title": "Contains Electives",
+                                "default": False,
+                                "options": {},
```
collinpreston, line 152: Will this field only be applicable for operators with `MIN_NUMBER_OF`?  If so, it might be nice to have something like:
```
"elective_flag": {
                                "type": "boolean",
                                "title": "Contains Electives",
                                "default": False,
                                "options": {
                                    "dependencies": {
                                        "node_type": ProgramRequirementNodeType.OPERATOR.value,
                                        "operator": ProgramRequirement.Operator.MIN_NUMBER_OF.value,
                                    }
                                },
                            },
```
So the field doesn't exist on every item in the program requirements tree.

collinpreston (changes requested):

courses/migrations/0037_update_programrequirements_add_elective_flag.py
```diff
@@ -0,0 +1,43 @@
...
+
+from django.db import migrations, models
+
+
+def forwards_func(apps, schema_editor):
+    ProgramRequirement = apps.get_model("courses", "ProgramRequirement")
+    db_alias = schema_editor.connection.alias
+    update_reqs = (
+        ProgramRequirement.objects.using(db_alias)
+        .filter(depth=2, title="Elective Courses")
```
collinpreston, line 11: Should `elective_flag = True` also be set for nested elective operators?

courses/models.py
jkachel: It depends. Having the flag here allows for electives to also be "all of" at its root, and I can think of a scenario where that would be needed: the electives are grouped further into two or more "min number of" groups. 
```
Elective ---------> Group 1       ---------> Course 1
ALL_OF     \        MIN_NUMBER_OF   \ -----> Course 2
            \
             \----> Group 2       ---------> Course 3
                    MIN_NUMBER_OF   \ -----> Course 4
```

In this case, both second-level nodes would have the same operator, so telling them apart programmatically isn't possible (without the flag).

collinpreston: Got it, good thinking and nice diagram!

courses/migrations/0037_update_programrequirements_add_elective_flag.py
jkachel: No, it's really only meant to count if `depth=2` (1 being the root node for the tree). It doesn't have any meaning past that.

courses/forms.py
jkachel: Did this in https://github.com/mitodl/mitxonline/pull/1736/commits/a483bcf205e7993f10b15f8df0930fc005a269ea but only if `node_type` is operator (see previous note about electives potentially starting with an "all of").

collinpreston (approved):
Looks good to me, up to you if you want to remove the prefetch import that's not being used.  This definitely makes things faster.

Issue #1675, opened 2023-06-13
Only courses in future are getting processed for certificate generation and grade sync


annagav:
There is a user whose grade for 3T2022 recently got updated(Early June 2023) on open edx, combining the exam and course together. The users grade was not getting updated on mitxonline keeping the original grade from the course, and no certificate generated.
Looking into this closer I discovered that the grades get updated only for courses whos `certificate_available_date__gt=now-24hours`
### Expected Behavior

Updated grades should get updated in mitxonline and certificates generated accordingly.

### Actual Behavior

Grade changes after `certificate_available_date` won't get synced automatically.

### Related Issues
Discovered from this issue https://github.com/mitodl/hq/issues/1639

annagav:
So it looks like setting `certificate_available_date` to today reprocesses the certificates and updates the grades. But the course team are unaware that when they update grades after `certificate_available_date` that grades don't get synced and they need to notify us.
Would be nice to automate this.

pdpinch:
Can we just tell the course teams to update the certificate generation date when they update a grade?

annagav:
@pdpinch do they have access to mitxonline django admin?

pdpinch:
They don’t, but I thought we synced the certificate generation dates from Studio.

annagav:
Okay, I need to check how often do we sync course run info.

annagav:
yes, course run info syncs every hour so it should work, if they update the date.

annagav:
@pdpinch is there a document where I can add this step as a requirement for the procedure of updating grades after certificates had been generated. So that it is documented somewhere...

pdpinch:
Thanks for asking. I'll have to bring it up with the MITx team, since they do the training and docs for course teams. 

Can you write out the process as you see it here, so I can make sure I understand?

annagav:
I added a signal to sync grades and generate certificates on `CourseRun` update. https://github.com/mitodl/mitxonline/pull/1745

annagav:
After a discussion with @collinpreston, we thought it would be most efficient to allow grades to sync and certificates to be generated for another 6 months after the course has ended to avoid wrong grades and missing certificates.

pdpinch:
> After a discussion with @collinpreston, we thought it would be most efficient to allow grades to sync and certificates to be generated for another 6 months after the course has ended to avoid wrong grades and missing certificates.

[...]

PR #1745, opened 2023-07-12
When course run is updated sync grades and generate certificates

annagav:
Related to https://github.com/mitodl/mitxonline/issues/1675

Adds a signal on updating a course run to sync in current grades and certificates.

The problem is that certificates get generated with 24 hour widow of 'certificate_available_date'. If any grades are updated on edX after that, the grades don't get sync and no certificates generated.

This PR allows for the generated of certificates to take place if in the django admin you update the course run.

# How can this be tested?
1) Make sure you have a course run with a past certificate_available_date. 
2) Create user enrollment for this course run. 
3) Go to django admin and set the certificate_available_date to now.
4) Check the logs that an api call was made to edx to sync grades.

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -785,9 +785,7 @@ def generate_course_run_certificates_for_course(run):
     For given course run syncs grades and
     if certificate_available_date is in the past, also generates certificates.
     """
-    edx_grade_user_iter = exception_logging_generator(
-        get_edx_grades_with_users(run)
-    )
+    edx_grade_user_iter = exception_logging_generator(get_edx_grades_with_users(run))
     created_grades_count, updated_grades_count, generated_certificates_count = (
         0,
         0,
--- a/courses/api.py
+++ b/courses/api.py
@@ -783,7 +783,7 @@ def generate_course_run_certificates():
 def generate_course_run_certificates_for_course(run):
     """
     For given course run syncs grades and
-    if certificate_available_date is in the past, also generates certificates.
+    if certificate_available_date is in the past or course id self paced, also generates certificates.
 
     Args:
         run (models.CourseRun): a course run instance.
--- a/courses/api.py
+++ b/courses/api.py
@@ -784,6 +784,9 @@ def generate_course_run_certificates_for_course(run):
     """
     For given course run syncs grades and
     if certificate_available_date is in the past, also generates certificates.
+
+    Args:
+        run (models.CourseRun): a course run instance.
     """
     edx_grade_user_iter = exception_logging_generator(get_edx_grades_with_users(run))
     created_grades_count, updated_grades_count, generated_certificates_count = (
--- a/courses/api.py
+++ b/courses/api.py
@@ -766,7 +766,8 @@ def get_certificate_grade_eligible_runs(now):
 
 def generate_course_run_certificates():
     """
-    Hits the edX grades API for eligible course runs and generates the certificates and grades for users for course runs
+    Finds all course runs that are eligible for certificate generation.
+    Hits the edX grades API for eligible course runs, syncs grades and generates certificates.
     """
     now = now_in_utc()
     course_runs = get_certificate_grade_eligible_runs(now)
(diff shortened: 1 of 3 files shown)


courses/api.py
```diff
@@ -776,48 +777,56 @@ def generate_course_run_certificates():
...
-            get_edx_grades_with_users(run)
-        )
-        created_grades_count, updated_grades_count, generated_certificates_count = (
-            0,
-            0,
-            0,
+        generate_course_run_certificates_for_course(run)
+
+
+def generate_course_run_certificates_for_course(run):
```
collinpreston, line 783: It might be helpful to rename both `generate_course_run_certificates()` and `generate_course_run_certificates_for_course()` to more clearly distinguish what they do.  Can you also add in type hints for the parameter?

courses/signals.py
```diff
@@ -33,3 +41,25 @@ def handle_create_course_run_certificate(
...
+def handle_update_course_run(
+    sender, instance, created, **kwargs
+):  # pylint: disable=unused-argument
+    """
+    When a CourseRun model is updated, sync course run grades and generate missing certificates
+    """
+    if not created:
+        now = now_in_utc()
+        if (
+            instance.certificate_available_date
```
collinpreston, line 60: Can you create a unit test for this logic?

collinpreston (changes requested):
Couple questions I'm curious about:

1. Is there a need to have the certificate generated right after the course run is updated?
2. Is there a use case that a CourseRun will have `certificate_available_date=None` and `self_paced=True` but will need to still generate certificates?  This is currently being handled in `generate_course_run_certificates_for_course()`

annagav:
@collinpreston answering some questions:

> 1.  Is there a need to have the certificate generated right after the course run is updated?

When the course team have updated a learners grade it is useful to have a quick way to sync grades with edx and generate missing certificates.

> 2.Is there a use case that a CourseRun will have certificate_available_date=None and self_paced=True but will need to still generate certificates? This is currently being handled in generate_course_run_certificates_for_course()

The self paced courses get processed daily by a scheduled celery task.

courses/api.py
annagav: I think the names are pretty accurate. What would be your suggestion?

annagav:
@collinpreston I think I addressed all your comments. Thanks for your review!

courses/api.py
collinpreston: Yeah I'm not sure why I was originally thinking both should be renamed.  It might be helpful to rename `generate_course_run_certificates_for_course` to `generate_course_run_certificates_for_course_run` but that's nit-picking.  I would still add type hints for the `run` parameter.

courses/api.py
```diff
@@ -776,48 +777,59 @@ def generate_course_run_certificates():
...
+            updated_grades_count += 1
+
+        # Check certificate generation eligibility
+        #   1. For self_paced course runs we generate certificates right away irrespective
+        #   of certificate_available_date
+        #   2. For others course runs we generate the certificates if the certificate_available_date of course run
+        #   has passed
+        now = now_in_utc()
+        if run.is_self_paced or (
+            run.certificate_available_date and run.certificate_available_date <= now
```
collinpreston, line 814: I was writing some [documentation](https://docs.google.com/document/d/1kw-ChTQK2YZZr8QjexhQX7P-IH5wwQeXNr4xxG0i2r8/edit#heading=h.j7mv17u378xp) last week about how certificates are generated and it took me way too long to understand all the different clauses we have scattered around regarding `certificate_available_date`.  Would it make sense to remove this `if` statement completely, and add the check of `is_self_paced` within `get_certificate_grade_eligible_runs`?

Right now we check certificate_available_date in the following places:
handle_update_course_run: certificate_available_date exists and is in the past.
generate_course_run_certificates_for_course: certificate_available_date exists and is in the past, or the run is self paced.
get_certificate_grade_eligible_runs: certificate_available_date does not exist or is in the past (default 24 hours).

I've tried to list out all of the scenarios in which course run certificates should be created in the documentation I linked.  I think there is a lot of overlap being handled by each one of these functions that we could consolidate.

annagav: Yes, I think it could be refactored.

annagav:
@collinpreston I removed the extra check for certificate date being in the past. I think it is helpful to sync the current grades whenever a course run is updated independently of certificates.

The question I have in general: why do we not want to sync grades for courses after certificates were generated? But this can be a separate issue.

courses/signals.py
```diff
@@ -33,3 +39,20 @@ def handle_create_course_run_certificate(
...
+)
+def handle_update_course_run(
+    sender, instance, created, **kwargs
+):  # pylint: disable=unused-argument
+    """
+    When a CourseRun model is updated, sync course run grades and generate missing certificates
+    """
+    if not created:
+        transaction.on_commit(
+            lambda: generate_course_run_certificates_for_course(instance)
```
collinpreston, line 57: If a CourseRun has `self_paced = True`, we will generate certificates for eligible learners whenever the CourseRun is saved, irregardless of whether `certificate_available_date` is defined.  Is that expected?

courses/api.py
```diff
@@ -776,48 +777,59 @@ def generate_course_run_certificates():
...
-        created_grades_count, updated_grades_count, generated_certificates_count = (
-            0,
-            0,
-            0,
+        generate_course_run_certificates_for_course(run)
+
+
+def generate_course_run_certificates_for_course(run):
+    """
+    For given course run syncs grades and
```
collinpreston, line 785: I would update this docstring to specify that the course certificate is generated if the `certificate_available_date` is in the past, OR the CourseRun has `self_paced = True`

courses/signals.py
annagav: I don't see anything wrong with having self paced courses also available for this signal.

collinpreston: Okay, that is different from how course certificates are created for self paced courses today.  I just wanted to make sure that won't surprise anyone.

annagav: What we could do is make it very specific and say only if `certificate_availble_date` was updated then sync grades. I don't imagine course runs being updated often after their creation, but we could be more specific on what kind of update is eligible to sync grades. 
What do you think?

collinpreston: So the current nightly task does what is described here: https://docs.google.com/document/d/1kw-ChTQK2YZZr8QjexhQX7P-IH5wwQeXNr4xxG0i2r8/edit#heading=h.j7mv17u378xp

You could remove the `certificate_availble_date` and `self_paced` logic from within generate_course_run_certificates_for_course() and add that into your signal.  You would then need to add a check for `self_paced = True` in `get_certificate_grade_eligible_runs()`.  But then you're not trying to make the the if statement at https://github.com/mitodl/mitxonline/blob/18c5a9fd13ec67f7e45666996647c4eaa7b6cd99/courses/api.py#L814 work for two different requirements.

annagav:
After a discussion with @collinpreston, we thought it would be most efficient to allow grades to sync and certificates to be generated for another 6 months after the course has ended to avoid wrong grades and missing certificates.

Issue #1742, opened 2023-07-12
Program records are showing a grade for audited course run


annagav:
### Steps to Reproduce

<img width="721" alt="Screen Shot 2023-07-11 at 11 03 25 PM" src="https://github.com/mitodl/mitxonline/assets/7574259/ebf9471a-c1aa-4c01-8476-7448c85d0994">

### Expected Behavior

Grade should be empty if there is no verified enrollment  and grade for course

### Actual Behavior

The user enrolled in audit track, a grade of 0 was created, and is showing up on the record.

pdpinch:
@annagav how did you come to the conclusion that audits should not appear on the program record? Is that what we do on MicroMasters?

PR #1747, opened 2023-07-13
No audit grades in program records

annagav:
Fix https://github.com/mitodl/mitxonline/issues/1742

This PR filter only grades earned in a verified more of enrollment for display on the Learner Records page.

# How can this be tested?
1. create a course run grade for your user
2. create a verified enrollment for that course run
3. view Program records, the grade should show
4. then set the enrollment to audit
5. Check the Program Records again, now the grade should not be there

### DIFF
--- a/courses/serializers.py
+++ b/courses/serializers.py
@@ -719,10 +719,18 @@ class LearnerRecordSerializer(serializers.BaseSerializer):
                 "grade": None,
                 "certificate": None,
             }
+            verified_enrollment_course_runs_list = (
+                models.CourseRunEnrollment.objects.filter(
+                    user=user,
+                    run__course=course,
+                    enrollment_mode=EDX_ENROLLMENT_VERIFIED_MODE,
+                    change_status=None,
+                ).values_list("run__id", flat=True)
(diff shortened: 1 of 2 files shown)


jkachel:
I tested this with a set of enrollments that resulted in certificates being generated. When I set one of them to `audit`, it still shows that I have a certificate for the course. Should this be the case or should the certificate flag be hidden if there's not a verified enrollment for the user? 

Functionality regarding the program completion badge and the grade display works as expected.

pdpinch:
> should the certificate flag be hidden if there's not a verified enrollment for the user?

We already have a mechanism for revoking certificates. Additional logic to hide a certificate in a degenerate state like this (and a test for that logic) adds unneeded complexity.

annagav:
@pdpinch I think what James was asking is - if a learner is missing doesn't have a verified enrollment should the certificate be not shown on their dashboard or revoked?

My answer would be we leave the certificate as is even if the enrollment mode changes for the user.

jkachel (approved):
LGTM 👍

Issue #1738, opened 2023-07-11
Program dashboard tab does not load if program does not have a feature image


collinpreston:
### Steps to Reproduce

Remove the feature image from a program's CMS page.
Visit the program tab in the MITx Online dashboard.

### Expected Behavior

Program dashboard tab is loaded with a default image for any programs without a feature image.

### Actual Behavior

A blank page is shown.

### Stacktrace
```
instrument.js:109 TypeError: Cannot read properties of null (reading 'feature_image_src')
    at hy.renderProgramEnrollment (EnrolledItemCard.js:590:49)
    at hy.render (EnrolledItemCard.js:689:14)
    at Ra (react-dom.production.min.js:182:192)
    at Ca (react-dom.production.min.js:181:224)
    at bi (react-dom.production.min.js:263:490)
    at mu (react-dom.production.min.js:246:265)
    at du (react-dom.production.min.js:246:194)
    at ru (react-dom.production.min.js:239:172)
    at react-dom.production.min.js:123:115
    at t.unstable_runWithPriority (scheduler.production.min.js:19:467)
    at Io (react-dom.production.min.js:122:325)
    at Ko (react-dom.production.min.js:123:61)
    at Vo (react-dom.production.min.js:122:496)
    at Y (react-dom.production.min.js:287:95)
    at Vt (react-dom.production.min.js:68:371)
    at HTMLDocument.o (helpers.js:73:23)
```

### Related Issues

https://github.com/mitodl/mitxonline/pull/1724/

### Screenshot or Screencast
![Screenshot 2023-07-11 at 15 30 43](https://github.com/mitodl/mitxonline/assets/8311573/b6c1bb11-7706-4053-9ed9-bc38486f2114)

jkachel:
Can't reproduce - on RC, I cleared the featured image from the Something Else Test Program and it pulled up OK. There is a bug where it'll fail if there's no page at all for the program, though.

pdpinch:
Should we close this, or re-write the issue?

collinpreston:
We can close this based on the work James did.

PR #1750, opened 2023-07-14
ProgramSerializer should now return default image if there's no program page

jkachel:

Related to #1738

If the program doesn't have a page, the frontend will crash because it expects a featured image. This adds a default featured image if there's no page, so it doesn't. 

# How can this be tested?

1. Create a new program. (Manually using the Django Admin is the recommended method so you don't inadvertently get a page created or anything.) 
2. Create a program enrollment for one of your accounts.
3. Log in and navigate to the dashboard.
4. Click the My Programs page. You should see the new program.


collinpreston (approved):
Nice

PR #1760, opened 2023-07-19
Updates import_course a bit for program requirements

### LINKED ISSUE #1680: import_courserun command should rely on Program requirements
### Steps to Reproduce

1. Create a Course which is not associated with a Program.
2. Create a Program.
3. Create a Course which is associated with the Program created.
4. Add both Courses created as requirements to the Program created.
5. Create 1 course instance for each course in Edx using the same run tag (ex, `3T2023`)
6. Run the `import_courserun` using the `--program` and `--run-tag` flags.

### Expected Behavior

2 CourseRuns should be created since both courses are defined as requirements for the Program.

### Actual Behavior

Only 1 CourseRun is created for the Course associated with the Program.  The Course which is only defined as a requirement of the Program does not have a CourseRun generated for it.

### Additional

Courses should be able to associated with multiple Programs.  This allows us to create multiple Programs with requirements that specify the same Course(s), or to create new Programs with required Courses from Programs that have since ended.

jkachel:

Closes #1680

Fixes a minor error when creating a standalone course using import_course. (This came about after testing the command for #1680.) The command would try to add a program to the new course (even if the program was None), which would cause an error because there's no FK to Program anymore.

# How can this be tested?

Full test:
1. In Studio, create three courses. 
2. In MITx Online, create two matching courses (but no courseruns) and assign them to a program. (These can be all electives, all requirements, or a mix.)
3. Run `manage.py import_courseware --program <your program> --run-tag <run tag>`. It should pick up the course runs from edX successfully. 
4. Run `manage.py import_courseware --courserun <the remaining one>` (add whatever other flags you like). It should pick up the courserun and create the course as well.

### DIFF
--- a/courses/management/commands/import_courserun.py
+++ b/courses/management/commands/import_courserun.py
@@ -181,21 +181,6 @@ class Command(BaseCommand):
                         )
                     )
 
-                    if kwargs["make_elective"]:
-                        program.add_elective(course)
-                        self.stdout.write(
-                            self.style.SUCCESS(
-                                f"Added course {course_readable_id} to program {program.readable_id} as an elective"
-                            )
-                        )
-                    else:
-                        program.add_requirement(course)
-                        self.stdout.write(
-                            self.style.SUCCESS(
-                                f"Added course {course_readable_id} to program {program.readable_id} as required course"
-                            )
-                        )
-
                 new_run = CourseRun.objects.create(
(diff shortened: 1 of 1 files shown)


collinpreston (approved):
Looks good to me.  Just need to remove the --make-elective kwarg which we spoke about.

Issue #1764, opened 2023-07-20
Two flaky tests are flaky (hubspot_sync.tasks & courses.views)


collinpreston:
### Steps to Reproduce

Run the following two tests:
hubspot_sync.tasks_tests.test_sync_failed_contacts()
courses.views_test.test_get_courses()

### Expected Behavior

Tests always pass.

### Actual Behavior

Tests sometimes pass due depending on the randomness of ordering.

PR #1765, opened 2023-07-20
Ignore ordering in tests where it doesn't matter

collinpreston:
https://github.com/mitodl/mitxonline/issues/1764

Improves the integrity of some tests which expected results in a certain order.  Ordering was not integral to the tested functionality.

# How can this be tested?
Tests should pass.

### DIFF
--- a/hubspot_sync/tasks_test.py
+++ b/hubspot_sync/tasks_test.py
@@ -426,7 +426,9 @@ def test_sync_failed_contacts(mocker):
     )
     result = tasks.sync_failed_contacts(user_ids)
     assert mock_sync.call_count == 4
-    assert result == [user_ids[1], user_ids[3]]
+
+    # Expect two failed contacts that correspond with the ApiExceptions above.
+    assert len(result) == 2
 
 
 @pytest.mark.parametrize("for_contacts", [True, False])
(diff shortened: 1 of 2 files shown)


hubspot_sync/tasks_test.py
```diff
@@ -426,7 +426,9 @@ def test_sync_failed_contacts(mocker):
     )
     result = tasks.sync_failed_contacts(user_ids)
     assert mock_sync.call_count == 4
-    assert result == [user_ids[1], user_ids[3]]
+
+    # Expect two failed contacts that correspond with the ApiExceptions above.
+    assert len(result) == 2
```
jkachel, line 431: Would this be better if it checked for the IDs in the result (`user_ids[1] in result`)?

jkachel (commented):
Had a question about one of the tests.

hubspot_sync/tasks_test.py
collinpreston: I think the issue that causes this test to be flaky is the `filter()` in `sync_failed_contacts()` which doesn't guarantee the order of the users.  So `filter()` might return [user_ids[1], user_ids[2], user_ids[3], user_ids[4] which would mean that user_ids[1] and user_ids[3] are going to experience an ApiException, `filter()` could also return [user_ids[2], user_ids[4], user_ids[3], user_ids[1] which would mean that user_ids[2] and user_ids[3] are going to experience an ApiException and be returned by `sync_failed_contacts()`.

Issue #1762, opened 2023-07-20
Letter grade is set to None when grade is populated through a management command


annagav:
### Steps to Reproduce
Example course run grade id 43640.

I noticed that some grades are missing `letter_grade` value. We think that happens through the management command `manage_certificates`.  If `is_grade_override` is true then we set letter grade to None: [code here]
(https://github.com/mitodl/mitxonline/blob/2f951479e328c2a9f9d21d79f84fc3e04837ff43/courses/api.py#L871-L897)
### Expected Behavior

If the grade is passing and enrollment verified we need an actual letter grade.

### Actual Behavior

The learner is missing a letter grade.

<img width="688" alt="Screen Shot 2023-07-20 at 8 51 09 AM" src="https://github.com/mitodl/mitxonline/assets/7574259/c69a7458-9b0c-4886-9ce9-730d3228e86f">

pdpinch:
When we pass a grade to the management command, we should require a letter grade (string) as well.

pdpinch:
> Example course run grade id 43640.

Solving this issue also requires doing a search for other missing letter grades and fixing them up. If the courses are DEDP courses, it's straightforward since they always use the same grading policy. If there are missing letter grades from other courses, we will have to look at their grading policy in Studio.

arslanashraf7:
@pdpinch FYI.

>Solving this issue also requires doing a search for other missing letter grades and fixing them up. If the courses are DEDP courses, it's straightforward since they always use the same grading policy. If there are missing letter grades from other courses, we will have to look at their grading policy in Studio.


This might be difficult in the sense that, a `None` letter grade in Django would not mean that it was only set by the override command. So, To reduce the scope, We can check those grades that are `set_by_admin=True` to back fill the data.

Also, Just to add more thoughts to what's already defined above:

1. I believe that the Letter grade from edX can have possible values (`A`, `B`, `C`, `D`, `E`, `F`, `Pass`, `Fail`, `None`)
2. So, along with the override management command, The grade can come as `None` from edX as well.
3. I think we'll need to update validation on letter grade as defined [here in the PR](https://github.com/mitodl/mitxonline/pull/1767/files#diff-1891631991481dca5fef7d88592f36f41fe1422ddc3f13e172d8973bc77c8da6R30-R32)
4. Usually, The `Pass`/`Fail` comes when there is a 50/50 grading policy set which is also the default in edX.


See some screenshots below:
[...]

PR #1767, opened 2023-07-21
Manage certificates command: require letter grades when overriding grade

annagav:
Related to https://github.com/mitodl/mitxonline/issues/1762

require a letter grade when overriding learners grade.

# How can this be tested?
Run the management command 
`./manage.py manage_certificates --user <username> --run <run_id>  --grade <score> --create`
This should show a validation error asking you to provide a letter grade
The run:
`./manage.py manage_certificates --user <username> --run <run_id>  --grade <score> --letter-grade <letter_grade> --create`

### DIFF
--- a/courses/management/commands/manage_certificates.py
+++ b/courses/management/commands/manage_certificates.py
@@ -73,6 +73,12 @@ class Command(BaseCommand):
             required=False,
         )
         parser.add_argument(
+            "--letter_grade",
+            type=float,
+            help="Override a grade with a corresponding letter grade. Range: A-F",
+            required=False,
+        )
+        parser.add_argument(
             "--revoke", dest="revoke", action="store_true", required=False
         )
         parser.add_argument(
@@ -92,6 +98,7 @@ class Command(BaseCommand):
         create = options.get("create")
         run = options.get("run")
         override_grade = options.get("grade")
+        letter_grade = options.get("letter_grade")
 
         if not (revoke or unrevoke) and not create:
             raise CommandError(
@@ -142,6 +149,11 @@ class Command(BaseCommand):
 
             if override_grade and not is_grade_valid(override_grade):
                 raise CommandError("Invalid value for grade. Allowed range: 0.0 - 1.0.")
+
+            if override_grade and not letter_grade:
+                raise CommandError(
+                    "Override grade needs a letter grade, allowed range: A-F"
+                )
 
             if override_grade and not user:
                 raise CommandError(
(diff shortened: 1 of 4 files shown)


courses/utils.py
```diff
@@ -26,6 +27,10 @@ def is_grade_valid(override_grade: float):
     return 0.0 <= override_grade <= 1.0
 
 
+def is_letter_grade_valid(letter_grade: str):
+    return False if re.match("^[A-F]$", letter_grade) is None else True
```
asadali145, line 31: Could be simplified as `return re.match("^[A-F]$", "a") is not None`

courses/management/commands/manage_certificates.py
```diff
@@ -72,6 +72,12 @@ def add_arguments(self, parser):
                 (>0.0) allows certificate creation. Range: 0.0 - 1.0",
             required=False,
         )
+        parser.add_argument(
+            "--letter_grade",
+            type=float,
```
asadali145, line 77: Shouldn't it be a string?

asadali145: It raised an error that `A` cannot be converted to float.

courses/management/commands/manage_certificates.py
```diff
@@ -143,6 +150,11 @@ def handle(self, *args, **options):  # pylint: disable=too-many-locals
             if override_grade and not is_grade_valid(override_grade):
                 raise CommandError("Invalid value for grade. Allowed range: 0.0 - 1.0.")
 
+            if override_grade and not letter_grade:
```
asadali145, line 153: We needed to pass the letter grade to `override_user_grade` after line 186.
Screen Shot 2023-07-24 at 5 38 06 PM

asadali145 (changes requested):
Command throughs a couple of exceptions as commented. Also, added a minor suggestion.

courses/utils.py
```diff
@@ -26,6 +27,10 @@ def is_grade_valid(override_grade: float):
     return 0.0 <= override_grade <= 1.0
 
 
+def is_letter_grade_valid(letter_grade: str):
+    return re.match("^[A-F]$", letter_grade) is not None
```
asadali145, line 31: Could you please have a look at this [comment](https://github.com/mitodl/mitxonline/issues/1762#issuecomment-1649697708) by @arslanashraf7 and make the relevant changes?

annagav: This PR doesn't try to solve the inconsistency of the letter grades in general. This PR just makes letter grade required when running this management command.

asadali145 (approved):
Looks good to me. Just documenting one of the things that we discussed is that we haven't allowed `Pass` and `Fail` as letter grades as those might never get used.

PR #1775, opened 2023-07-24
Program records: show verified grades or grades with certificate

annagav:
Related to https://github.com/mitodl/hq/issues/1525

Some users have a passing grade, payment, and certificate but no verified enrollment for a course run.
This PR attempts to show grades only if the user enrolled in verified track or received a certificate.

# How can this be tested?
Set up your data to test that the following grades are showing up on the Program Records page:
1) The user has a grade and a certificate for course
2) The user has a grade and a verified enrollment for a course run (the grade may be failing)

### DIFF
--- a/courses/serializers.py
+++ b/courses/serializers.py
@@ -16,6 +16,7 @@ from cms.serializers import CoursePageSerializer, ProgramPageSerializer
 from courses import models
 from courses.api import create_run_enrollments
 from courses.constants import CONTENT_TYPE_MODEL_COURSE, CONTENT_TYPE_MODEL_PROGRAM
+from courses.models import CourseRunCertificate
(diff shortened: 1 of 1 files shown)


courses/serializers.py
```diff
@@ -16,6 +16,7 @@
 from courses import models
 from courses.api import create_run_enrollments
 from courses.constants import CONTENT_TYPE_MODEL_COURSE, CONTENT_TYPE_MODEL_PROGRAM
+from courses.models import CourseRunCertificate
```
rachellougee, line 19: there is already `from courses import models` on line 16

rachellougee (approved):
Looks good, just one small comment.

PR #1776, opened 2023-07-25
Deferrals better message

annagav:
N/A

Makes sure the a failed status is returned if enrollment to a course run was not successful.

Removes the atomic block and allows for repeating reruns of the deferral procedure. This is helpful if this deferral request was already processed and partially succeeded.
If an api call was made to to edx to uneroll a user but they already have been unenrolled edx would return 400 and the deferral procedure would throw an error.

# How can this be tested?
There are few things to check:
1) There have been cases where the deferral would not succeed and no exception would get raised, and the row would get marked as completed. An example here https://github.com/mitodl/hq/issues/2009
I was not able to fully replicate that scenario locally, but I would like to pass a message to deferral result for better messaging.
2) If the user's deferment already happened the google sheet procedure should be able to reflect that on the spreadsheet, instead of trying to process it again and throwing an error.
3) if the enrollment into course did not happen the error row should state that.
4) if the user has unenrolled from course, but was not enrolled into a repeated processing of the row (by deleting the error message) should try to complete the enrollment.

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -512,19 +512,29 @@ def defer_enrollment(
                 from_enrollment.run.course.title, to_run.course.title
             )
         )
-    with transaction.atomic():
-        to_enrollments, enroll_success = create_run_enrollments(
-            user,
-            [to_run],
-            keep_failed_enrollments=keep_failed_enrollments,
-            mode=from_enrollment.enrollment_mode,
-        )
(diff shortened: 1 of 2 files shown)


rachellougee (approved):
Code looks good, tested the deferrals from one course to another, it worked as expected.

PR #1777, opened 2023-07-25
Adds a "type_label" to the Program model

jkachel:

mitodl/hq#1982

This provides some grouping for programs. It defaults to Series. This isn't exposed anywhere yet - it's for the new course and program admin stuff. 

# How can this be tested?

After running the migration, you should see the field in the Django Admin.


jkachel:
@collinpreston requested review because this field'll need to be there for the catalog pages (and I need it also, I think)

jkachel:
There are some migrations in the past that reference the Program model object and thus cause issues (since at step 0033 there's no new field, but it's there in the model object, so it gets confused). Working on that..

PR #1778, opened 2023-07-25
Add program_type field to database

jkachel:

Closes mitodl/hq#1982

Adds a `program_type` field to the Program model and defaults it to `Series`. This is part of the new lite redesign - this field is needed elsewhere and having it in the current design won't hurt anything, so going forward with adding it now.

This also changes a couple migrations to use the versioned app models for `Program`. The affected migrations were directly importing the model, which ends up having a field that hasn't been applied yet and thus causes issues. This is really only an issue if you're spinning up a new instance or for CI; existing databases should have the migrations applied already and there's no functional change to them. 

# How can this be tested?

Run the migration. Afterwards, you should see a "Program type" field in the Django admin.


collinpreston (commented):
Looks good.  Just a couple questions:

1. Does it make sense to make the field an enum? (ref: https://docs.djangoproject.com/en/4.2/ref/models/fields/#enumeration-types)
2. When would a program not have a `program_type`?

jkachel:
I opted for a simple text field because the list is going to change at some point - we'll start with two but there will be a new one later and we're not using it for logic so it seemed like more of a future hassle to make it an enum. 

Not sure a program would ever have a blank type. It defaults to Series in that case.

collinpreston (approved):

PR #1782, opened 2023-07-26
Fixing amounts for the default tiers for the configure_tiers command

### LINKED ISSUE #1781: configure_tiers command incorrectly creates default discount tiers
Not sure when this happened, but it appears the `configure_tiers` management command is improperly setting the discount amount when it uses the default (DEDP-esque) tier set. The discounts are being created with decimal percentages rather than whole numbers (0.75 vs 75) so the discounts are off by a bit. Affected tiers include course-v1:MITxT+3.012Sx, course-v1:MITxT+8.S50.2x, and course-v1:MITxT+3.034.1x. (The rest seemed to be OK and this is pretty easy to identify in the Django Admin.) 

### Steps to Reproduce

1. Use the `configure_tiers` command to create a set of financial assistance tiers for a courseware object.

### Expected Behavior

Tiers are created for 25%, 50%, and 75% discounts (as well as 0, which isn't a problem here).

### Actual Behavior

The tiers are 0.25%, 0.50%, and 0.75%. 

### Related Issues

mitodl/hq#2025

jkachel:

Closes #1781 
mitodl/hq#2025

Just changes the default tiers to be whole-number percentages. They were decimal but that isn't correct now.

# How can this be tested?

Run `configure_tiers` for a course. You should get the proper discount amounts (0%, 25%, 50%, and 75%) and not ones that are < 1.

### DIFF
--- a/flexiblepricing/management/commands/configure_tiers.py
+++ b/flexiblepricing/management/commands/configure_tiers.py
@@ -107,7 +107,7 @@ class Command(BaseCommand):
                 "discount": {
                     "discount_code": f"{discount_abbrev}-fa-tier1-{current_year}",
                     "discount_type": "dollars-off" if not course else "percent-off",
-                    "amount": 750 if not course else 0.75,
+                    "amount": 750 if not course else 75,
                 },
             },
             {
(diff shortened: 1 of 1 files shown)


collinpreston (approved):
Looks good to me

Issue #1075, opened 2022-09-29
Improve frontend asset loading times


ChristopherChudzicki:
**Routes and Bundles:** Currently `/create-account`, `/signin`, `/dashboard` and I believe most other routes serve all bundles chunks. The homepage and `/courses` serve all chunks except root.

<img width="800" alt="Screen Shot 2022-09-29 at 11 10 51 AM" src="https://user-images.githubusercontent.com/9010790/193074682-e4c98087-9fb0-46db-935e-47e38bf53596.png">

<table>
    <tr>
        <td>Chunk</td>
        <th>gzipped / parsed</th>
        <td colspan="2">totals</td>
    </tr>
    <tr>
        <th>common:</th>
        <td>283 KB / 1.24 MB</td>
        <td rowspan="3">
            514 KB / 2.1MB
            <br>
            homepage, /courses
        </td>
        <td rowspan="4">
            635 KB / 2.6MB
            <br>
            most other routes
        </td>
    </tr>
    <tr>
        <th>django</th>
        <td>214 Kb / 765 KB</td>
    </tr>
    <tr>
        <th>style</th>
        <td>17 KB / 77 KB</td>
    </tr>
    <tr>
        <th>root</th>
        <td>121 KB / 502 KB</td>
        <td></td>
    </tr>
</table>

Since all routes are loading common+django, many pages are probably loading things they don't need. Some ideas to decrease the amount of network transfer:

1. Load things more dynamically. (I'm guessing the homepage does not need to load videojs, for example.)
2. import less of `moment` (and possibly replace it with `luxon`)

    Our bundle currently includes all the localizations of momentjs (german, french, japanese, ...) but I am fairly certain we are not actually using them. So we should not include them.

    Additionally, we might consider using something else. [Momentjs is a legacy project](https://momentjs.com/docs/#/-project-status/); [dayjs](https://www.npmjs.com/package/dayjs), [luxon](https://www.npmjs.com/package/luxon), or [date-fns](https://www.npmjs.com/package/date-fns) might be reasonable alternatives.
3. Put large external dependencies (e.g., videojs) in a separate chunk even if they are only loaded once; this means when we cut a new release, their cache would not be broken.
4. Don't accidentally include dev dependencies. (Like parts of `chai`, which made it into the common chunk).

  *Ideally, we'd enforce that this does not happen through linting. But js code tends not to distinguish between development dependencies and runtime dependencies. It might be useful to do that; I will start a discussion.*

[...]

PR #1784, opened 2023-07-26
Remove unnecessary MomentJS locales

ChristopherChudzicki:
#1075 

This removes unused locales from `moment`, the datetime library we're using:

- 'en' is the default locale; we never alter it.
- Adds 'moment-locales-webpack-plugin' to remove the unnecessary locales, as recommended in https://momentjs.com/docs/#/use-it/webpack/ *MomentJS is bit of a legacy project. In Modern packages, you'd import just the ones you need. Moment imports everything, so we need to strip out the unused ones instead.*

### How can this be tested?

*Going to ask about this in standup.*


frontend/public/src/lib/util_test.js
```diff
@@ -216,7 +216,6 @@ describe("utility functions", () => {
   })
 
   it("formatPrettyDate should return a formatted moment date", () => {
-    moment.locale("en")
```
ChristopherChudzicki, line 219: `en` is the default locale. There are now no calls to `locale()` in the codebase.

ChristopherChudzicki:
> Do we ever display start dates and times in the user's timezone? Does that rely on this package?

We do (on enrollment cards and order history), but timezones do **not** rely on the extra locales. (They rely on `moment-timezone`, which you can still see in the bundle chart.)

Locales determine how dates are displayed:
```
now.locale('fr').format('LLL') // '1 août 2023 10:49'
now.locale('en').format('LLL') // 'August 1, 2023 10:49 AM'
now.format('LLL') // 'August 1, 2023 10:49 AM', English is the default
```

We never change the locale, so all our dates are always displayed in the `'en'` locale.

collinpreston (approved):
Looks good to me.  Comment on the plugin will be helpful for the future

Issue #697, opened 2022-07-14
retire_users management command doesn't unenroll learner from the course in edX


briangrossman:
### Steps to Reproduce

- Find a learner who is enrolled in at least one course
- Run `retire_users` command to retire the learner

### Expected Behavior

- The learner should be unenrolled from the courses they're enrolled in in edX

### Actual Behavior

- The learner's enrollment is removed from the MITx Online Django admin
- ... but the learner is still actively enrolled in the course on the edX side

### Screenshot or Screencast

Enrollments before retiring learner (Django admin):
![Before](https://user-images.githubusercontent.com/21342568/179052516-05687791-faa2-4236-ac06-d34900126eef.png)

Retire Learner:
![image](https://user-images.githubusercontent.com/21342568/179052570-3cdb69da-2e2a-417c-b7d0-3d27ae1ff178.png)

Enrollments after retiring learner (Django admin): 
![After](https://user-images.githubusercontent.com/21342568/179052602-6dc124d9-6668-42ec-b035-c667e33e00f3.png)

Enrollment on the edX side:
![image](https://user-images.githubusercontent.com/21342568/179052824-870636dd-edee-4876-abd9-0a552c2c1fbd.png)

### related issue:

https://github.com/mitodl/ol-infrastructure/issues/1007

jkachel:
Related Sentry log message?
https://sentry.io/organizations/mit-office-of-digital-learning/issues/3426657144/?project=5864687&referrer=slack

briangrossman:
Kinda, but not really... I accidentally tried to remove my test user from production first--they were on RC. This error was generated when I accidentally ran the retire command on production. So, it's a red herring. 

Here's the code for the management command, BTW: https://github.com/mitodl/mitxonline/blob/main/users/management/commands/retire_users.py

pdpinch:
While we're investigating this, can we also revisit whether there is a proper API in open edX for retiring users that we can use?

pdpinch:
Take a look at this API: https://github.com/openedx/edx-platform/blob/7686014144930bb9fda4ae8beb66f0392855e55a/lms/djangoapps/bulk_user_retirement/urls.py#L11-L15

pdpinch:
That API may not work without some further configuration of open edX. When Tobiase tried to retire a user in production, we got the error: 

https://sentry.io/organizations/mit-office-of-digital-learning/issues/3513101534/?project=5939801&query=is%3Aunresolved

[...]

PR #1785, opened 2023-07-27
feat: add ability to retire users on edX

asadali145:

Fixes https://github.com/mitodl/mitxonline/issues/697

This PR updates the retire_users management command to initiate user retirement on edX. Currently, the command retires a user in MITxOnline and we need to run an extra command in edX to initiate the retirement.

# How can this be tested?

- Checkout this branch in MITxOnline locally.
- Clone and Checkout [edx-api-client](https://github.com/mitodl/edx-api-client) branch `asad/bulk-retirement-api`
- Install the local clone of edx-api-client in MITxOnline by using `pip install -e path/to/edx-api-client`
- Start edX locally, go to the lms shell and run `./manage.py lms populate_retirement_states` to create retirement states.
- Now run the following in lms shell
  - `app_name=retirement`
  - `user_name=retirement_service_worker`
  - `./manage.py lms manage_user $user_name $user_name@example.com --staff --superuser`
  - `./manage.py lms --settings=<your-settings> create_dot_application $app_name $user_name`
  - Copy the client id and secret for later.
- Update setting in edX, which can be done in private.py or lms.yml `RETIREMENT_SERVICE_WORKER_USERNAME = 'retirement_service_worker'`
- Clone https://github.com/mitodl/mitxpro-openedx-extensions and update line 22 in `mitxpro_core/settings/common.py` to `r"^/(admin|auth|login|logout|register|api|oauth2|user_api)", r"^/v1/accounts/bulk_retire_users"`
  - now build a pip-installable package using the guidelines provided in the [readme](https://github.com/mitodl/mitxpro-openedx-extensions).
  - copy and paste the .tar.gz file into edx-platform and install the package in the LMS shell by using `pip install mitxpro-openedx-extensions-1.0.0.tar.gz`.
- Now update the MITxOnline .env file and add the Client ID and Secret copied from the LMS shell.
  - `OPENEDX_RETIREMENT_SERVICE_WORKER_CLIENT_ID=<client_id>`
  - `OPENEDX_RETIREMENT_SERVICE_WORKER_CLIENT_SECRET=<client_secret>`
- We are done with all of the configurations.
- Create a user that we want to retire.
- Go to the MITxOnline web container and run `./manage.py retire_users --user <username>`. Output should be similar to the SS
  Screen Shot 2023-08-08 at 1 05 33 PM
- Go to the LMS Django Admin at `/admin/user_api/userretirementstatus/`. Verify that a retirement status exists with Pending state for the user.

## Checklist:
- [x] Publish new version of edx-api-client
- [x] Merge https://github.com/mitodl/salt-ops/pull/1574 and https://github.com/mitodl/ol-infrastructure/pull/1667
- [x] Create retirement service worker and OAUTH2 application in edX
- [x] Add OPENEDX_RETIREMENT_SERVICE_WORKER_CLIENT_ID and OPENEDX_RETIREMENT_SERVICE_WORKER_CLIENT_SECRET to the vault so that they can be used at https://github.com/mitodl/salt-ops/pull/1574/files#diff-70bbf14162dbce0218569098f3d487edbd7944ac86ad4fbeb9d5af8429364c1cR134-R135


openedx/api.py
```diff
@@ -921,3 +955,19 @@ def validate_username_with_edx(username):
         raise EdxApiRegistrationValidationException(username, resp)
     result = resp.json()
     return result["validation_decisions"]["username"]
+
+
+def bulk_retire_edx_users(usernames):
+    """
+    Bulk retires edX users.
+
+    This method calls edX bulk enrollment API to initiate the user retirement on edX.
```
arslanashraf7, line 964:
```suggestion
    This method calls edX bulk retirement API to initiate the user retirement on edX.
```

arslanashraf7: Can we also mention the possible states for retirement in edX? Or would that be confusing information?

openedx/api.py
```diff
@@ -515,6 +515,40 @@ def get_edx_api_service_client():
     return edx_client
 
 
+def get_edx_retirement_service_client():
```
arslanashraf7, line 518: Can we write a test for this? Something similar to `test_get_edx_api_client` in `openedx/api_test.py`.

users/management/commands/retire_users.py
```diff
@@ -106,6 +107,25 @@ def handle(self, *args, **kwargs):
...
+            resp = bulk_retire_edx_users(user.username)
+            if user.username not in resp["successful_user_retirements"]:
+                self.stderr.write(
+                    self.style.ERROR(
+                        "Could not initiate retirement request on edX for user {user}".format(
+                            user=user
+                        )
+                    )
+                )
+                sys.exit(1)
```
arslanashraf7, line 119: Use case: Multiple Users

This would halt the process if edX fails to initiate the retirement process for a single user. The existing implementation continues the process if retirement fails for a single user in MITxOnline. We should continue the process in case of failure from edX as well.

users/management/tests/retire_users_test.py
arslanashraf7, line 1: nit: We should add a simple assertion test to check that the retire_command also calls the functionality(``bulk_retire_edx_users``) from openedx/api.py .

arslanashraf7 (changes requested):

openedx/api.py
asadali145: > Can we also mention the possible states for retirement in edX? Or would that be confusing information?

Retirement states are dynamic and configured by a setting in edX `RETIREMENT_STATES`. Pending is the initial state that we wanted to create for a user and its mentioned on the next line.

arslanashraf7 (approved):

Issue #1773, opened 2023-07-24
Epic: Redesign course and program about pages


jkachel:
As a learner, I'd like to:
* know that I can enroll for free.
* see that I can pay for a certificate and how much it will cost, and when the upgrade deadline is
* know how I can apply for financial assistance
* be able to learn more about certificates
* be able to enroll in a course that is closed and understand that I can't get a certificate
* know when a course is part of a MicroMasters program or a Learning Path

#### Designs and Mockups

* Figma design: https://www.figma.com/proto/Gs4zIhOFv5gVvafawwl6PF/MITx-Online?page-id=37%3A4307&type=design&node-id=37-4311
* Updated design (whole thing): https://www.figma.com/proto/Gs4zIhOFv5gVvafawwl6PF/MITx-Online?page-id=0%3A1&type=design&node-id=558-2708&viewport=-1130%2C359%2C0.1&t=gwT4VDKtWVtv71cH-1&scaling=scale-down&mode=design

#### Acceptance Criteria:

These are high level requirements and split into functional areas. Items will link to individual tickets where they can.

- **CMS Changes**
    - [ ] Add "Show More" functionality to rich text content blocks
    - [ ] Make sure these context blocks are rich text: Course Intro, About (new), What You'll Learn
    - [ ] [Add selectable Meet Your Instructors block](https://github.com/mitodl/hq/issues/1543)
- **Main Content**
    - [ ] Add new CMS block for Course Intro
    - [ ] Implement "Show More" functionality within all content blocks
    - [ ] Add anchor tab
    - [ ] Update styling to match the design
- **Infobox**
    - [ ] Add container and styling for courseware object information
    - [ ] Add timing information 
    - [ ] Add certificate track/pricing information
    - [ ] Add program path information 
- **Modals**
    - [ ] Reformat upsell message for individual courses
    - [ ] Add upsell message for programs (this will contain an enrollment selector)
    - [ ] Any other modals
- **Miscellaneous**
    - [ ] Add feature flag to enable/disable the new design
    - [ ] Add feature flag to enable/disable program about pages 

#### Out of Scope

- Feature flag implementation: this should not be done in the old manner and instead use the new infrastructure coming in https://github.com/mitodl/hq/issues/1981.

#### Related issues

Course About Page - https://github.com/mitodl/hq/issues/1542
Reusable Faculty Members - https://github.com/mitodl/hq/issues/1543
Program About Page - https://github.com/mitodl/hq/issues/1544
[...]

PR #1786, opened 2023-07-27
Migrate instructor data to individual selectable pages

jkachel:

Closes mitodl/hq#1543
#1773 

Converts the current method of storing resource page instructor data (a stream block within the page marked Faculty Members) to a new method that uses individual pages for the instructors that can be linked to the resource page instead. Also contains some helper stuff to convert the data, so the existing data is not lost. 

#### Instructor pages

These are new and are designed to be located underneath an Instructor index page in Wagtail. The `configure_wagtail` command has been updated to create this index page if it doesn't exist.

Otherwise, instructor pages are normal Wagtail CMS pages. Some notes on fields:
* Title and Name are separate - in the case of duplicates, titles can be different to allow instructor pages to be more easily found in the CMS while the instructor's name remains the same. (This is especially important for the `convert_faculty_to_pages` command - see below.) 
* Name is what's used in the card on the page.
* There is a new Title field, this is displayed as a `<h4>` if present in the card.
* The short bio is analogous to the old description field.
* The long bio is not used currently.

#### `convert_faculty_to_pages`

This is a helper command that will load the data that's in the Faculty Members field in existing course pages, creates new Instructor Page instances for them, and then links them to the course page. It does _not_ operate on Program pages as we're not using them now. 

An Instructor Page is created for the faculty member listings _regardless of whether or not the instructor already exists in the system_ so we preserve whatever distinct imagery and copy is there from that particular page. The output of the command should tell you the title of the page that was created for the particular course page so you can find it later. These can be coalesced into a single instructor page at your leisure later. 

As part of the conversion process, the Faculty Members section is cleared for the page it's processing so duplicate cards aren't shown.

If there isn't an index page, the command will make one. 

#### Template changes

The `product_page.html` template is now set up to display the `Meet your instructors` section if there's linked Instructor Pages or old-style Faculty Members. Instructor pages will be displayed first. 

Instructor Pages have an additional Title field that wasn't in the old Faculty Members section - this is displayed as an `h4` under the name if there is one. 

No other changes for the cards - they should look the same as they did before. 

# How can this be tested?

1. Create a course page with some faculty members in it. 
2. Run `convert_faculty_to_pages`. This should convert the faculty member listings to pages, and apply them back to the course page.
3. View the course page. It should not be visibly different.
4. View the course page in the CMS. You should see the linked Instructor Pages and Faculty Members should be empty.
5. Explore the CMS - you should see an Instructors section with the new pages underneath. They should be Drafts.


pdpinch:
Does this work with the existing design, the upcoming redesign, or both?

How will we transition from the current system to the new one? What should I be telling the MITx folks who manage the about pages?

jkachel:
> Does this work with the existing design, the upcoming redesign, or both?

Both! 

> How will we transition from the current system to the new one? What should I be telling the MITx folks who manage the about pages?

There's a command to facilitate the migration process that will need to be run manually. The command scrapes the instructor data off the existing course pages, puts them into new instructor pages, links them to the course page it was working on, and then clears the Faculty Members section on the page; the end result is that the faculty members all get moved to instructor pages and there's no duplication of cards on the course pages. We end up with three states for this:

- Until the command is run: do nothing. The template still displays the people in the Faculty Members section so there's no change, and they can ignore the new one. 
- When we perform the migration (immediately after running the command): All the data in each Faculty Members section will be translated into an Instructor Page. The command doesn't make any effort to dedup this so there will likely be a bunch of similar pages (we can use the output from the command to determine which one goes to which page); it'd be best to clean that up some. But, immediately after running the migration command, the course about pages should look exactly as they did. 
- After the migration: I'll need to put up a new PR to eliminate the old Faculty Members section (and move the title for the section; it's out of order now), and they'll just link in instructor pages as necessary.

cms/models.py
```diff
@@ -472,6 +473,112 @@ def create_country_field(self, field, options):
         return ChoiceField(**options)
 
 
+class InstructorPage(Page):
+    """
+    Detail page for instructors.
```
collinpreston, line 478: I think you can add in something like: https://github.com/mitodl/mitxonline/blob/e09ce6f99c0ed119e7a7aa31880a3cd8a02ff996/cms/models.py#L1525
Which will not allow child pages to be created for an InstructorPage.

collinpreston (changes requested):
Looks good and works for me.

I noticed that a faculty page that is in "draft" status will still display if added to a Course about page.

I am also able to create an Instructor Page in the Courses "folder" in the CMS.

cms/management/commands/convert_faculty_to_pages.py
```diff
@@ -0,0 +1,138 @@
...
+    def add_arguments(self, parser) -> None:
+        parser.add_argument(
+            "--only",
+            type=str,
+            help="Only process the specified courseware object (a Course or Program). Specify the readable ID.",
+            dest="courseware",
+        )
+
+        parser.add_argument(
+            "--live", action="store_true", help="Make the page live. (Defaults to not.)"
```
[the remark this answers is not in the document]
jkachel: Made the flag and text more understandable (Wagtail calls it Draft/Published in the UI but the DB table calls it "live"). df1ea622 also adds some code to disallow serving of InstructorPages specifically - we don't want them exposed right now (but we might later and I repurposed other code for this so it's in there from that).

cms/management/commands/convert_faculty_to_pages.py
```diff
@@ -0,0 +1,140 @@
...
+from django.utils.text import slugify
+from wagtail.images.models import Image
+
+from cms.models import (
+    CoursePage,
+    HomePage,
+    InstructorIndexPage,
+    InstructorPage,
+    InstructorPageLink,
+    ProgramPage,
```
collinpreston, line 19: This import can be removed.

collinpreston (approved):
Looks good to me!

PR #1787, opened 2023-07-27
Populate passing DEDP grades that were set to None by Admin

annagav:
Related to https://github.com/mitodl/mitxonline/issues/1762

Populate passing DEDP grades that were set to None by Admin
Question: Do we need to make sure these grades are verified?

# How can this be tested?
Have a user with a passing grade with letter_grade set to None. Then run migrations.
The grade should get populated with a letter grade.


rachellougee (approved):
Looks good.  There are 7 grades will be converted by this migration https://bi.odl.mit.edu/queries/1304
and those 7 grades are all verified so we should be good.

We did 4 migrations to set letter grade for DEDP. If we need do a similar migration in the future, it would make sense to make a management command for that. But hopefully we don't need it again.

PR #1790, opened 2023-07-31
When deferring enrollment set to verified

annagav:
None

When an enrollment is deferred we always want the enrollment to be verified instead of copying the mode of the enrollment of the "from course".
There has been a case where the user was deferred to a course successfully but set to audit. I don't know why the "from course" was in audit mode, but I think it is safe to assume that when the deferral procedure is run we want to enroll the user in a verified mode.
This would allow for a deferral of cases that were partially complete. 

# How can this be tested?
Try to run the deferral management command while having current 'verified' enrollment for a "From course". Make sure that the user get's successfully unenrolled from the course, and enrolled into the verified mode of the "To course".

Then try the same with an 'audit' enrollment. The same result should proceed.

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -530,7 +530,7 @@ def defer_enrollment(
         user,
         [to_run],
         keep_failed_enrollments=keep_failed_enrollments,
-        mode=from_enrollment.enrollment_mode,
+        mode=EDX_ENROLLMENT_VERIFIED_MODE,
     )
     if not enroll_success and not keep_failed_enrollments:
         raise Exception(
--- a/courses/api_test.py
+++ b/courses/api_test.py
@@ -823,7 +823,7 @@ def test_defer_enrollment(
             existing_enrollment.user,
             [target_run],
             keep_failed_enrollments=keep_failed_enrollments,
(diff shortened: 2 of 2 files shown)


collinpreston (commented):
Where are deferrals initiated from?  I'm thinking it might be nice to offer some way of deciding what enrollment mode a user should be deferred into a course course with.

annagav:
We initiate deferals from the google sheet, I am not aware if anyone else is using deferrals in other ways. 

The reason I opened this PR is to limit the enrollment mode which I think should not be allowed to be anything else other than 'verified'. The point of the deferral operation is to transfer a verified enrollment from one course run to a future course run. This is only relevant for verified enrollments. For audit enrollments the user can just register by themselves in any number of course runs.

annagav:
@collinpreston you don't need to set up google sheets for this, you can just run the management command:
`./manage.py defer_enrollment --user <username> --from-run <from_run_id> --to_run <to_run_id>`

collinpreston (approved):
Makes sense to me. Thanks for the extra detail!

PR #1792, opened 2023-07-31
Make hashed webpack output filenames more stable

ChristopherChudzicki:
#1075 

This PR changes how filenames are generated by webpack: Now, filenames only change when content changes. If you change a CSS file, the JS filename will not change.

# How can this be tested?
1. Enter the `watch` container: `docker compose exec watch bash`
2. Run production build: `yarn workspace mitx-online-public build --clean`, `ls frontends/public/build`.
    - note the filenames for `requirementsAdmin` chunk
3. Make a temporary change to `requirements.scss` 
4. Run production build: `yarn workspace mitx-online-public build --clean`, `ls frontends/public/build`.
    - note the filenames for `requirementsAdmin` chunk
    - The JS file should have the same name, the CSS file should have a different name.

### DIFF
--- a/frontend/public/webpack.config.js
+++ b/frontend/public/webpack.config.js
@@ -21,8 +21,8 @@ module.exports = function (env, argv) {
     output: {
       path: path.resolve(__dirname, "build"),
       ...(isProduction ? {
-        filename: "[name]-[chunkhash].js",
-        chunkFilename: "[id]-[chunkhash].js",
+        filename: "[name]-[contenthash].js",
+        chunkFilename: "[id]-[contenthash].js",
         crossOriginLoading: "anonymous",
         hashFunction: "xxhash64"
       } : {


frontend/public/webpack.config.js
```diff
@@ -21,8 +21,8 @@ module.exports = function (env, argv) {
     output: {
       path: path.resolve(__dirname, "build"),
       ...(isProduction ? {
-        filename: "[name]-[chunkhash].js",
-        chunkFilename: "[id]-[chunkhash].js",
+        filename: "[name]-[contenthash].js",
+        chunkFilename: "[id]-[contenthash].js",
```
ChristopherChudzicki, line 25: A "chunk" is all the files associated with a particular entrypoint (our entrypoints are `root`, `header`, `style`, `django`, `requirementsAdmin`). Typically this is css+js. 
-`chunkhash` is the hash of the chunk
- `contenthash` is a per-file hash

Previously: css change --> chunk hash changes --> js filename changes, which was unnecessary.

This doesn't really make much of a difference for mitxonline, since almost all CSS is in its own entrypoint. But this is a simple change, and is the recommended approach based on https://webpack.js.org/guides/caching/

rachellougee (approved):
LGTM. Tested by modifying layout.scss and set NODE_ENV as production in .env, css file got unique hash name and js file is unchanged

PR #1807, opened 2023-08-08
Fixes per-course instructor page migrations

jkachel:

n/a

Noticed that trying to convert instructors to pages for a single course failed - this fixes that.

# How can this be tested?

Run `convert_faculty_to_pages --only <course>` for a course (ideally, one with old-style faculty members in it). The command should run successfully.


collinpreston (approved):
Looks good to me

PR #1812, opened 2023-08-10
Updates to prefer long bio over short for existing design

### LINKED ISSUE #1811: Currently displayed instructor bios should be the long ones
When we made [separate instructor bio pages](https://github.com/mitodl/mitxonline/pull/1786), we added a field for a short bio and a long bio. The short bio is really intended to be used in the new product page redesign (which shows a shorter bio on the page, and a longer one in a modal) and the current design was set up to display the short bio as well. The current design should really display the longer bio, though, since we'll need to set up specific short bios before the new design goes live. 

### Steps to Reproduce

1. Go to an instructor page and fill out the long bio. 

### Expected Behavior

On the course pages that link that instructor page, you should see the long bio.

### Actual Behavior

You see the short bio.

### Related Issues

https://github.com/mitodl/mitxonline/pull/1786
https://github.com/mitodl/mitxonline/issues/1773

jkachel:

Closes #1811 

* Adds a migration to migrate instructor short bios into the long bio field (this will overwrite all long bios (but there aren't any right now))
* Updates the template to display the long bio rather than the short one
* Updates the `convert_faculty_to_pages` command to populate the short bio with the instructor name rather than the bio (to prevent confusion)

# How can this be tested?

1. Change one of the short bios so you can tell that the migration worked.
2. Run the migration and check to make sure instructor info shows up as expected (esp. with the page that was edited in step 1).
3. Add an old-style faculty member to a course.
4. Run `convert_faculty_to_pages` - the new page should have the bio in the long field.


collinpreston (approved):
Looks good to me

PR #1816, opened 2023-08-11
New design: Course page layout changes

jkachel:

Part of #1773 

This does a lot but is focused on the course page redesign:

- Adds feature flagging for all new designs - this will expect a PostHog flag called `jkachel-new-design` if you want to see the changes
- Incorporates the new design for course pages, including:
   - Updates to the global style sheet (site has a white background now)
   - Adds info box for the relevant course details on the right side
   - Displays related program information (with links, though they don't go anywhere right now - program pages aren't in this PR)
   - Adds a tab bar to the course info
   - Adds expandable About This Course section - text after the first paragraph is hidden (and it supports splitting on a tag with a specific class name, which isn't exposed in Wagtail right now)
   - Adds instructor cards with more info modals according to the redesign
   - Adds redesigned upsell message dialog
- Adds some new data to the APIs to allow for these changes
   - Course API now provides for pulling a single course
   - Doing this pulls in some additional data (including program info) that it didn't before
   - Adds API for pulling instructor page details
- Refactors the feature flag code _in the Django app_ to look at Posthog as well as the existing `.env` based feature flags
- Removes the faculty members block from the resource page Wagtail model - this will result in any faculty member data in there being _lost_ so make sure you've converted them to instructor pages first. 

Note that this doesn't include mobile layouts - that will be in a separate PR. This is just the desktop view and the feature flag stuff. 

Top half:
image

Bottom half: 
image

Instructor modal:
image

New upsell modal:
image

# How can this be tested?

To test, you'll need a Posthog account. (There's a free tier if you're creating one.) Once you have that set up, you will need to put your API token and the host in your `.env` file:

```
POSTHOG_API_HOST=https://app.posthog.com
POSTHOG_API_TOKEN=get your own
```

Then, create a new feature flag called `mitxonline-new-product-page`. The payload can be anything (I set it to True) and make sure to set the Release Condition to roll out to 100% of users:

image

_(I changed the feature flag but have not updated this screenshot so the flag name won't match.)_ 

Finally, before testing, make sure you have some courses set up with a variety of settings - some should be in programs, some should be in more than one program, some shouldn't be in a program, some should have products, some should have financial assistance, etc. Also, _make sure your ad blocker is disabled for your local instance_ - it will likely not be happy about Posthog being there and will then block API calls to it. (This was very much the case for uBlock Origin for me.) 

To test: You can toggle between the old and new designs by *enabling* or *disabling* the flag. (You just need to enable/disable, not adjust the payload.) 

With the flag enabled, you should see the new design. With it disabled, the old design should appear. The new design should match the Figma layout for the course about page (as much as possible - some elements may not appear on a given course depending on what settings they have). The old design should match RC or production. 

Note that switching the flag on or off may not take full effect until you reload a time or two - the React app itself looks for the flag and sometimes browser caching flummoxes that.


jkachel:
FYI changed the feature flag name to `mitxonline-new-product-page` which is what it's (currently) called in the more real Posthog.

annagav (commented):
Code looks good! Comparing the style with the design now.

cms/migrations/0031_merge_20230817_1538.py
```diff
@@ -0,0 +1,12 @@
+# Generated by Django 3.2.20 on 2023-08-17 15:38
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("cms", "0030_add_faq_remove_faculty_reorder_things"),
```
annagav, line 8: Could you recreate this migration with number 0031 so that the merge migration is not needed?

annagav (changes requested):

cms/migrations/0031_merge_20230817_1538.py
jkachel: Done in https://github.com/mitodl/mitxonline/pull/1816/commits/6775405dcd4bdf9dc8dd8962dd718b11540718d7

annagav:
I managed to add meet your instructors twice to the course page:
Screen Shot 2023-08-18 at 3 04 50 PM
Screen Shot 2023-08-18 at 3 05 05 PM

frontend/public/src/entry/style.js
```diff
@@ -1 +1,7 @@
-import "../../scss/layout.scss"
+import { checkFeatureFlag } from "../lib/util"
+
+if (checkFeatureFlag("mitxonline-new-product-page")) {
```
JenniWhitman, line 3: So I just thought (way too long) about this - and I don't think we can do this AND maintain the ability to flip flags on and off - we're back to doing a deploy to see UI changes and that's no so good. I think we might have to v2 or otherwise newly name all new scss files as well as variables.

JenniWhitman (changes requested):
I left a comment about building styles - down to hash it out and be wrong, but I can't think of a way to have both.

frontend/public/src/entry/style.js
jkachel: Not sure - it works for me in dev. I'd sorta hope(d) it'd notice the import calls and load in both sets of stylesheets but maybe not.. I can test with a local production build, though.

jkachel:
Doubling up of the instructors section is fixed in https://github.com/mitodl/mitxonline/pull/1816/commits/4d1499fbe668736a93b335c78e58ba42d07039f3 (this was more than that; a rebase error pulled in a bunch of old code into the template so a bunch of stuff was getting duplicated).

frontend/public/src/entry/style.js
jkachel: So, I did some testing with a production-style build locally (basically setting environment = prod in the `.env` everywhere that it made sense to) and tested it out, and it actually did seem to work. Would you be able to test as well to verify? (I have my own Posthog account so I'll just swap back over to that locally for now so changes there won't affect the stuff I'm still working on.) 

FWIW, my criteria for success were:
- I can set the feature flag up to apply when `environment = prod` for 100% of users
- I can see the call to Posthog happen in the network pane in devtools, and see that it's sending me back a 1 (there should only be a couple of places checking the feature flag in the code - once in main app init thing and once in the product detail app)
- The Watch container remains dormant - no logs emitted during testing
- The CSS and JS bundles are coming from port *8013* rather than *8012* signifying that I'm getting a static build rather than one coming out of webpack
- (Obvs) the flag works and the design changes :-) 

This all seemed to pass (with the requisite couple of reloads when the flag status changes) but if you can give it a shot too I'd appreciate it. (Or, if there's something else I should be looking for here to make sure I'm getting a production-level build.)

frontend/public/src/components/CourseInfoBox.js
```diff
@@ -0,0 +1,143 @@
...
+        </div>
+        {course && course.programs && course.programs.length > 0 ? (
+          <div className="program-info-box">
+            <strong>
+              Part of the following program
+              {course.programs.length === 1 ? null : "s"}
+            </strong>
+
+            <ul>
+              {course.programs.map(elem => (
```
annagav, line 129: This seems to list one program as two:
Screen Shot 2023-08-21 at 3 47 21 PM

annagav (changes requested):

frontend/public/src/components/CourseInfoBox.js
jkachel: I can't reproduce this - for a course that's just in one program, or is in two, it lists the proper programs out. (Do you perhaps have a program named "program" in your copy somewhere?)

frontend/public/src/entry/style.js
JenniWhitman: OK, now what if there are 3 feature flags - course page, home page, and... IDK potato page. Course page is set to go, because we love it and it's great, so we turn that on in prod. Now, Home page - not ready yet. We don't want that on in prod. Herein lies the issue. Yes, this works from feature to feature swapping around the imports, but can we dynamically build said styles in each of those possible T T T, TFT, TFF, FFF scenarios for those 3 pages, components, or whatever level they may be at?

We weren't using webpack when I used flags last so I'm not sure where the correct answer is outside of just... versioning class names or something simple because everything else I came up with seemed needlessly complex.

jkachel: We discussed via slack and i think the next step here is:
- Pulling out the imports from the main `loader.scss` that apply to this code
- Importing those scss files specifically within the module itself. (In this case, that'd be the ProductDetailEnrollApp component, since it should get loaded on any product page.) 
- For this specifically, there's a style that I put into the `loader.scss` (for the page background) that will move to one of the module imports

This will keep the specific styling bits that need to change separate from the main CSS. If we need to flag more stuff, we can do that and keep the flags atomic by pulling the CSS imports out of the main loader and into the components they affect. This will probably temporarily make CSS loading less efficient but since it'll be temporary it shouldn't make too much difference.

annagav (approved):
LGTM 👍

PR #1820, opened 2023-08-16
Featured Products Carousel

JenniWhitman:
Updates #https://github.com/mitodl/hq/issues/1677

Adds programs to featured products in Wagtail
Adds feature flagging to be able to turn these changes off and on depending on environment
New Styling & carousel for Featured Products

Desktop

Mobile

# How can this be tested?
Make sure that you have a number of courses and programs set up
Go into the wagtail admin and add some of each to your Featured products on the home page
If in dev, this should automatically be on.
Look at both desktop and mobile and manipulate both.

# Additional Context
The flag is called mitxonline-new-featured-carousel in post-hog. Let me know if you need access.
[Figma Link](https://www.figma.com/file/Gs4zIhOFv5gVvafawwl6PF/MITx-Online?type=design&node-id=558%3A2708&mode=design&t=REGKyMeHdusZuvU1-1)


cms/migrations/0031_alter_homeproductlink_course_product_page.py
```diff
@@ -0,0 +1,26 @@
...
+        migrations.AlterField(
+            model_name="homeproductlink",
+            name="course_product_page",
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=django.db.models.deletion.SET_NULL,
+                related_name="+",
+                to="wagtailcore.page",
+                verbose_name="Featured Product Page",
```
JenniWhitman, line 23: All this actually does is change the verbose name - it was Course-centric prior.

cms/models.py
```diff
@@ -684,11 +690,18 @@ def products(self):
         return page_data
 
     def get_context(self, request, *args, **kwargs):
+        posthog = Posthog(settings.POSTHOG_API_TOKEN, host=settings.POSTHOG_API_HOST)
+        show_new_featured_carousel = posthog.feature_enabled(
+            "mitxonline-new-featured-carousel",
+            "randomID",
+            person_properties={"environment": settings.ENVIRONMENT},
+        )
```
JenniWhitman, line 698: The only time we really care about the flag is on the front end - but the django template and feature flags don't get along all the way, so I'm passing from front to back end rather than loading on the html page with vanilla js or changing post-page load

cms/models.py
```diff
@@ -707,10 +720,13 @@ class HomeProductLink(models.Model):
         blank=True,
         on_delete=models.SET_NULL,
         related_name="+",
+        verbose_name="Featured Product Page",
```
JenniWhitman, line 723: Again - this was the only change to the model.

frontend/public/scss/layout-v2.scss
```diff
@@ -0,0 +1,63 @@
...
+  .std-page-body {
+    @include media-breakpoint-down(md) {
+      padding-top: 32px;
+      padding-bottom: 32px;
+    }
+
+    padding-top: 50px;
+    padding-bottom: 50px;
+  }
+}
```
JenniWhitman, line 63: Will merge conflict wtih James' work, will handle.

JenniWhitman: So I actually want to rethink this because this defeats the purpose of the feature flags - we'd still haev to redeploy to see the changes.

frontend/public/scss/v2/featured-product-cards.scss
```diff
@@ -0,0 +1,241 @@
...
+        .carousel-item {
+          gap: 24px;
+
+          .featured-product-card {
+            width: calc(25%);
+            border-radius: 5px;
+            background: white;
+            box-shadow: 0 2px 3px 0 rgba(18, 38, 49, 0.15);
+            @include media-breakpoint-down(sm) {
+              width: 282px;
```
rachellougee, line 115: This might need to adjust a little bit to match with design in mobile UI

cms/templates/home_page.html
```diff
@@ -18,35 +18,110 @@ <h3>{{ page.hero_subtitle }}</h3>
...
+                {% endif %}
+              </div>
+              <div class="featured-product-info card-body">
+                {% if product.start_date and not product.is_program and show_new_featured_carousel %}
+                  <p class="date">
+                    Starts {{ product.start_date | date:"F j, Y" }}
+                  </p>
+                {% elif not product.is_program %}
+                  <p class="date">
+                    Start Anytime
```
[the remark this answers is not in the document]
rachellougee: Nevermind. I guess it makes sense to say Start Anytime if all existing runs are expired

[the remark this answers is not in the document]
rachellougee: I would add `not product.start_date` to be clear.

cms/models.py
JenniWhitman: Wagtail does this for us in the CMS :) "Only the following page types may be chosen for this field: Course page, Program page. Search results will exclude pages of other types."

cms/templates/home_page.html
```diff
@@ -18,35 +18,110 @@ <h3>{{ page.hero_subtitle }}</h3>
...
+      </div>
+      <div class="featured-product-listing">
+        <div class="carousel-inner" role="listbox">
+          {% for product in products %}
+          <div class="carousel-item{% if forloop.counter == 1 %} active{% endif %} ">
+            <div class="card featured-product-card">
+              <a href="{{ product.url_path }}" class="featured-product-card-link">
+              <div class="featured-product-thumb">
+                <img src="{% feature_img_src product.feature_image %}" alt="" />
+                {% if product.is_program and product.program_type %}
```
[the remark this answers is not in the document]
JenniWhitman: True. Will update.

cms/templates/home_page.html
```diff
@@ -18,35 +18,110 @@ <h3>{{ page.hero_subtitle }}</h3>
...
+                  <span class="badge badge-program-type">{{ product.program_type }}</span>
+                {% endif %}
+              </div>
+              <div class="featured-product-info card-body">
+                {% if product.start_date and not product.is_program and show_new_featured_carousel %}
```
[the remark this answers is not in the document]
JenniWhitman: That's... odd. Not sure why that's there actually. I think from when it wasn't just one big if statement. Can remove.

cms/templates/home_page.html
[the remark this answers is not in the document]
JenniWhitman: so not is_program and not  start_date? without new_featured carousel the prior if statement is just "if start date and is not program" - this is "else if" so the not start_date is implied, we just need to make sure it's not a program. Not sure what we gain.

cms/models.py
rachellougee: Yes, there is a message in CMS. I mainly mentioned it here for developer who look at model because field name is course_product_page so my instinct would be its only for course. Not big deal, just for clarification.

JenniWhitman: I can add in a comment about hte historical context/name. Help text shows to user and that page is already busy as heck.

cms/templates/home_page.html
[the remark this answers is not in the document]
rachellougee: I think its because the previous if `product.start_date and not product.is_program and show_new_featured_carousel`  made this unclear, but if you remove `show_new_featured_carousel`, then yes no need to update here. It took me a sec to realize what this implies when I first look at it.

cms/models.py
JenniWhitman: Added.

cms/templates/home_page.html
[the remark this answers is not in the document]
JenniWhitman: updated

cms/templates/home_page.html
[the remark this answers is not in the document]
JenniWhitman: Modified

frontend/public/scss/v2/featured-product-cards.scss
JenniWhitman: Done so. The measurements used match the figma, I've swapped some hard measurements out for % though and I think I fixed it. Note that this is only the featured component so it will still shift left on moving the page. Once the full page is done, will get center-responsive.

rachellougee:
Ipad view looks better now. Since this PR only touches carousel, page isn't full centered until other components are done per notes. So this PR looks good to me

rachellougee (approved):

PR #1825, opened 2023-08-21
update full name on user profile

rachellougee:

https://github.com/mitodl/hq/issues/2182

This PR allows full name get updated on user profile from dashboard. Currently full name isn't updated on MITx Online but it gets updated on edX because `sync_hubspot_user` is called before the update on`CurrentUserRetrieveUpdateViewSet` runs so User.name gets overwritten.

# How can this be tested?

- Go to http://mitxonline.odl.local:8013/profile/
- Edit `Full Name`, `First Name`, `Last Name`, updates should go through locally
image

- Check `Name` field on your edX instance, it should be updated to match with `Full Name` from MITx Online
image

- Check  `First Name`, ` Last Name` field of contact in Hubspot `rl-mitxonline-dev`
   Full Name on MITx Online doesn't map to any field in Hubspot  
image

### DIFF
--- a/hubspot_sync/api.py
+++ b/hubspot_sync/api.py
@@ -662,7 +662,7 @@ def sync_contact_with_hubspot(user: User):
         body=body,
     )
     user.hubspot_sync_datetime = now_in_utc()
-    user.save()
+    user.save(update_fields=["hubspot_sync_datetime"])
 
     return result
 
--- a/users/serializers.py
+++ b/users/serializers.py
@@ -338,7 +338,6 @@ class UserSerializer(serializers.ModelSerializer):
 
             user = super().update(instance, validated_data)
 
-        sync_hubspot_user(user)
         return user
 
     class Meta:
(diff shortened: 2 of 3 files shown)


mbertrand (commented):
Seems to work fine.  I have a small concern that it might still be possible for the hubspot sync task to run before the change_edx_user_name_async task, for instance maybe if they end up being run by different celery workers.  One idea I just thought of would be to update the `hubspot_sync.api.sync_contact_with_hubspot` function to save only the one updated field instead of everything:

```python
user.save(update_fields=["hubspot_sync_datetime"])
```

instead of just:

```python
user.save()
```

What do you think of that idea?  If you prefer to keep it as is then I'll approve it.

rachellougee:
Thank you @mbertrand. I added `user.save(update_fields=["hubspot_sync_datetime"])` for your suggestion. It's a good idea to add that.

mbertrand (approved):
👍

PR #1830, opened 2023-08-24
Fixes text alignment for the Enrolled button, hide the upsell card in new design

jkachel:

Part of #1773

The text on the Enrolled button wasn't aligned properly, and the upsell card was appearing erroneously when the new design is turned on. This fixes both of those issues.

Additionally, after a discussion with @JenniWhitman , I've reworked what gets imported by default in here to just pull in the framework stuff and global variables into the feature flagged CSS loader. This should keep different flagged features from stepping on each other. This should be functionally equivalent to the previous code - it should all work the same.

# How can this be tested?

Toggle the new design on and go to a course that you have an audit enrollment in. The text for the Enrolled button should be centered in the button, and you should not see the upsell card.

Toggle the new design off and reload the page a couple of times (to allow the changes to take effect). The upsell card should appear.

### DIFF
--- a/frontend/public/scss/meta-product-page.scss
+++ b/frontend/public/scss/meta-product-page.scss
@@ -1,4 +1,12 @@
-@import "layout";
+@import "variables";
+@import "~@material/layout-grid/dist/mdc.layout-grid";
+@import "~@material/top-app-bar/dist/mdc.top-app-bar";
+@import "~react-picky/dist/picky";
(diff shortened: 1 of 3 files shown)


rachellougee (approved):
Looks good to me. Tested that I don't see 'Get Certificate' when mitxonline-new-product-page feature is on, and I can see it when feature is off (by modifying code)

Issue #1837, opened 2023-08-28
CSS regression: forms are full width instead of narrow


pdpinch:
### Steps to Reproduce

1. View the profile form on RC: https://rc.mitxonline.mit.edu/profile/ (this may require a feature flag to see the effect)
2. Compare to the profile form on production: https://mitxonline.mit.edu/profile/ 

### Expected Behavior

The profile form should not be full-width. The production version is correct. 

![Screenshot 2023-08-28 at 4 05 17 PM](https://github.com/mitodl/mitxonline/assets/430126/becbd7fe-338f-4227-8051-6eb06100c4c7)

### Actual Behavior

Currently, on RC, the profile form (and other forms) extends to the full width of the container. 

![Screenshot 2023-08-28 at 4 01 47 PM](https://github.com/mitodl/mitxonline/assets/430126/107687bc-e9eb-42c7-b4e1-614dfb2c296d)

PR #1838, opened 2023-08-29
Changes to feature-flag related scss files and imports

JenniWhitman:
Fixes https://github.com/mitodl/mitxonline/issues/1837

This PR moves the home-page related files to their own directory, removes unnecessary imports from the product-details page to stop the writing over of the container css, and cuts down on the extra css imports. It would be helpful to also revisit the naming of the product page and use new class names in order to cut back on this more, but James is on vacation and this solution works with what we have.

# How can this be tested?
Check that the homepage has not been altered (some of this is on RC, not all)
Check that product pages look as they did before this (RC is current)
Check that your local `/profile` looks as it does at mitxonline.mit.edu/profile/

# Additional Context
You can see the CSS in its incorrect state at https://rc.mitxonline.mit.edu/profile/

### DIFF
--- a/frontend/public/scss/meta-product-page.scss
+++ b/frontend/public/scss/meta-product-page.scss
@@ -1,11 +1,5 @@
 @import "variables";
-@import "~@material/layout-grid/dist/mdc.layout-grid";
-@import "~@material/top-app-bar/dist/mdc.top-app-bar";
-@import "~react-picky/dist/picky";
-@import "~react-day-picker/lib/style";
-@import "~bootstrap/scss/bootstrap";
-@import "~video.js/dist/video-js";
-
+@import "~bootstrap/scss/bootstrap-utilities";
 
 @import "product-page/user-menu";
 @import "product-page/product-details";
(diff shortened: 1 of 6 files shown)


frontend/public/scss/meta-product-page.scss
```diff
@@ -1,11 +1,5 @@
 @import "variables";
-@import "~@material/layout-grid/dist/mdc.layout-grid";
-@import "~@material/top-app-bar/dist/mdc.top-app-bar";
-@import "~react-picky/dist/picky";
-@import "~react-day-picker/lib/style";
-@import "~bootstrap/scss/bootstrap";
-@import "~video.js/dist/video-js";
-
+@import "~bootstrap/scss/bootstrap-utilities";
```
JenniWhitman, line 2: Variables and the mixins available from bootstrap-utilities are the only pieces needed and neither contains css that will overwrite things when reimported because of the styles.js conditional import

rachellougee (approved):
Nice! Tested home/profile/dashboard/course product page with feature flags on/off, they look as expected, I didn't notice anything being overwritten

JenniWhitman:
Thank you @rachellougee !

PR #1841, opened 2023-08-30
Anonymous user handling for posthog

JenniWhitman:
Fixes https://mit-office-of-digital-learning.sentry.io/issues/4435442344/?environment=rc&project=5864687&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

Added a conditional for un-authenticated users to generate and store a session uuid to track each user's session in the feature flag history 

# How can this be tested?
Load up the application, make sure you are not logged in. You should be able to navigate around and in the [posthog event history](https://app.posthog.com/events), see a random UUID on your requests - but it should be the same UUID for your requests.

The home page should no longer 500 whether you are logged in or not.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -2,6 +2,7 @@
 import json
 import logging
 import re
+import uuid
 from datetime import datetime, timedelta
 from json import dumps
 from urllib.parse import quote_plus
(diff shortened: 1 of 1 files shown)


rachellougee (approved):
Code and test look good to me

Issue #1842, opened 2023-08-30
Duplicate ProductPage.get_context in cms/models.py


rachellougee:
### Steps to Reproduce

Duplicated get_context in https://github.com/mitodl/mitxonline/blob/main/cms/models.py#L1049-L1071. It might be duplicated during merge conflict fix.

### Expected Behavior

Only one `get_context` in `ProductPage`

PR #1846, opened 2023-08-31
1842: duplicate ProductPage get_context method in model

collinpreston:
https://github.com/mitodl/mitxonline/issues/1842

Removes a duplicate method which was introduced in this PR: https://github.com/mitodl/mitxonline/pull/1816/

# How can this be tested?
All tests should pass and the ProductPage model should function as expected.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -27,7 +27,6 @@ from wagtail.contrib.forms.models import (
     FORM_FIELD_CHOICES,
     AbstractForm,
     AbstractFormField,
-    AbstractFormSubmission,
 )
 from wagtail.contrib.routable_page.models import RoutablePageMixin, route
 from wagtail.coreutils import WAGTAIL_APPEND_SLASH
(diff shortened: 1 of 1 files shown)


cms/models.py
```diff
@@ -27,20 +27,18 @@
...
 from wagtail.embeds.exceptions import EmbedException
 from wagtail.fields import RichTextField, StreamField
 from wagtail.images.models import Image
-from wagtail.models import Orderable, Page, Site
+from wagtail.models import Page
 from wagtail.search import index
 
 from cms.blocks import (
     CourseRunCertificateOverrides,
-    FacultyBlock,
```
rachellougee, line 43: Looks like FacultyBlock is no longer used anywhere, probably can be removed as well. But up to you

rachellougee (approved):
LGTM

cms/models.py
collinpreston: Nice catch.  I created an issue for it here: https://github.com/mitodl/mitxonline/issues/1855

Issue #1850, opened 2023-08-31
Homepage 500s due to using the python sdk rather than initializing at the app level


JenniWhitman:
### Steps to Reproduce

When installing initially, we had used the python SDK rather than the Django implementation. This is causing 500 errors on RC now.

[more context](https://mitodl.slack.com/archives/C02649X2P1V/p1693507094521869)

pdpinch:
I think this is the relevant sentry error: https://mit-office-of-digital-learning.sentry.io/issues/4443267622/?referrer=github_integration

PR #1851, opened 2023-08-31
Change posthog initialization

JenniWhitman:
Fixes https://github.com/mitodl/mitxonline/issues/1850

Removed the initialization of the posthog client from features.py and models.py and created the client in settings to be imported as used.

# How can this be tested?
Ensure feature flags still work and can be toggled while active. Trying to come up with a meaningful load test, but it might need to happen on RC (anything I've thought up is more cumbersome than I think we should be doing for this)

# Additional Context
I ended up using more of this: https://posthog.com/tutorials/django-analytics#installing-and-using-the-library than the documentation as documentation's method looks like it's importing the client, but really importing the library without a client connected and does not yet work.


main/features.py
```diff
@@ -1,29 +1,27 @@
...
 
-def is_enabled(name, default=None):
+def is_enabled(name, default=None, unique_id=settings.HOSTNAME):
     """
     Returns True if the feature flag is enabled
 
     Args:
         name (str): feature flag name
         default (bool): default value if not set in settings
+        unique_id (str): person identifier passed back to posthog
```
collinpreston, line 24: How should a developer decide whether to set this value to `settings.HOSTNAME` or `User.id`, or maybe something else?

collinpreston (changes requested):

main/features.py
JenniWhitman: That's a good question - what I went with for the home page, considering the output in the posthog application itself, was either the user's readable ID (in this case, email) OR, in the case of an anonymous user, a uuid generated per session - so we can track if/when something has gone awry. This would also allow, in the future, to filter by user if need be (if we're passing user as email, then we know, for example, to add a rule for jenniw@mit.edu rather than hostname which can vary).

JenniWhitman: I would prefer we change all of them to use that, but for the time being, went in this direction. If you think it'd be ok to swap them over, I can, but I wasn't sure all places would have access to request/session/user.

collinpreston: Makes sense.  I would add a description of how this value is used in PostHog to the argument description you currently have.  Maybe even suggest the policy you defined for using user ID or UUID so that PostHog isn't filled with a bunch of values derived from different developer's functionality.

collinpreston: You could also add a description of your policy within here: https://docs.google.com/document/d/1OQhnF1t00s0yhHWSZTWSoduHlF9aHo7YWZILF5LlX1M/edit?pli=1#heading=h.2wm6l084bpzh

JenniWhitman: Yes - Going to do that now. Sorry, a lot of this didn't become super clear until we actually started to USE the platform and some of these uses are future use, not now. I will add a descrip in the argument description and after can help flesh out the doc more.  I have been meaning to get in there, but keep finding things that need fixing :(

JenniWhitman:
The tests seemingly pass (they pass on local and on ci they look like they pass then just hang), but there's no real output to let me know what and I've cleared the cache to no avail. It's not happening elsewhere so likely related to the settings addition, but trying to pinpoint and fix.

main/features.py
```diff
@@ -1,39 +1,42 @@
...
-def is_enabled(name, default=None):
+def is_enabled(name, default=None, unique_id=settings.HOSTNAME):
     """
     Returns True if the feature flag is enabled
 
     Args:
         name (str): feature flag name
         default (bool): default value if not set in settings
+        unique_id (str): person identifier passed back to posthog which is the display value for person. I recommend
+                         this be a readable id for logged-in users to allow for user flags as well as troubleshooting.
```
collinpreston, line 24: I'd explicitly say User.id for logged in users.

JenniWhitman: gotcha

JenniWhitman:
A rogue merge while trying to figure out the other stuff did some strange things here. Going to close this PR and move to https://github.com/mitodl/mitxonline/pull/1857 since it's JUST the changes I want

PR #1857, opened 2023-09-01
Posthog refactor 

JenniWhitman:
Fixes https://github.com/mitodl/mitxonline/issues/1850

Removed the initialization of the posthog client from features.py and models.py and created the client in settings to be imported as used.

# How can this be tested?
Ensure feature flags still work and can be toggled while active. Trying to come up with a meaningful load test, but it might need to happen on RC (anything I've thought up is more cumbersome than I think we should be doing for this)

# Additional Context
I ended up using more of this: https://posthog.com/tutorials/django-analytics#installing-and-using-the-library than the documentation as documentation's method looks like it's importing the client, but really importing the library without a client connected and does not yet work.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -19,7 +19,6 @@ from django.urls import reverse
 from django.utils.text import slugify
 from mitol.common.utils.datetime import now_in_utc
 from modelcluster.fields import ParentalKey
-from posthog import Posthog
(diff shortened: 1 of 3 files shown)


collinpreston (approved):
Looks good to me.
Nice work troubleshooting that silent issue!

PR #1861, opened 2023-09-05
fix expandable Read More 500 error

rachellougee:

https://github.com/mitodl/mitxonline/issues/1856

Fixing 500 error on About section if text is less than 2 paragraphs

When there are more than 2 paragraphs, and Show More is displayed
image
When there is only 1 sentence, display the original text
image

# How can this be tested?

Edit About section from a course cms page
   - add one sentence there, make sure no 500 error
   - add more than one paragraphs, 'Show More' should be displayed. It can be toggled which implemented in https://github.com/mitodl/mitxonline/pull/1816

### DIFF
--- a/cms/templatetags/expand.py
+++ b/cms/templatetags/expand.py
@@ -1,12 +1,14 @@
 """
-Adds an expandable Read More section to a text block. 
+Adds an expandable Read More section to a text block.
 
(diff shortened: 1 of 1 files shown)


jkachel (approved):
LGTM 👍

Issue #1862, opened 2023-09-05
Apply new classnames to the course about page


jkachel:
This is mostly for me from the discussion in the planning meeting this (9/5) morning.

### Actual Behavior

The course about page reuses some class names between the new design and the old. These should be separated out and redone so that they don't step on each other anymore, allowing us to include both CSS files in the built stylesheet. 

### Related Issues

#1838 
#1840 
#1773 
https://github.com/mitodl/hq/issues/1542

pdpinch:
Is this still an issue?

jkachel:
Closing this as it's not a thing anymore.

PR #1865, opened 2023-09-06
Updates the course about page CSS

jkachel:

https://github.com/mitodl/mitxonline/issues/1862

This wraps the feature-flagged class changes (and just those changes) in a selector that should only apply if the feature flag is tagged on.

# How can this be tested?

Set the feature flag to Enabled and check - it should look the same as it did in https://github.com/mitodl/mitxonline/pull/1816 . 

Set the feature flag to Disabled and check - it should successfully revert to the old design.


JenniWhitman (approved):
Looks good - matches with current RC, doesn't modify other pages, and pulls us out of the business of loading extra stuffs.

Issue #1870, opened 2023-09-10
Remove duplicate catalog URL path


collinpreston:
There is a duplicate catalog URL in the main URL paths: 
https://github.com/mitodl/mitxonline/blob/366b675f0f852160835e64563c390beeeb837570/main/urls.py#L90

PR #1871, opened 2023-09-10
1870: Remove duplicate catalog URL path

collinpreston:
https://github.com/mitodl/mitxonline/issues/1870

Removes a duplicate URL path.

# How can this be tested?

- http://mitxonline.odl.local:8013/catalog/ should still be accessible when `ENABLE_NEW_DESIGN = True`
- http://mitxonline.odl.local:8013/catalog/ should display a 404 page when `ENABLE_NEW_DESIGN = False`


JenniWhitman (approved):
LGTM :)

PR #1873, opened 2023-09-11
Add new header design

annagav:
Fix https://github.com/mitodl/hq/issues/2313

Add new design of the header

Screen Shot 2023-09-08 at 7 44 29 AM
Screen Shot 2023-09-11 at 8 09 19 AM

# How can this be tested?
Make sure the feature flag is on. Go to the home page and all other pages and make sure it looks good.

### DIFF
--- a/frontend/public/src/components/Header.js
+++ b/frontend/public/src/components/Header.js
@@ -6,6 +6,8 @@ import TopAppBar from "./TopAppBar"
 
 import type { CurrentUser } from "../flow/authTypes"
 import type { Location } from "react-router"
+import {checkFeatureFlag} from "../lib/util";
+import TopBar from "./TopBar"
 
 type Props = {
   currentUser: CurrentUser,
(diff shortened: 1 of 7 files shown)


frontend/public/src/components/Header.js
```diff
@@ -27,11 +29,20 @@ const Header = ({ currentUser, location }: Props) => {
       scope.setUser(null)
     })
   }
-  return (
-    <React.Fragment>
-      <TopAppBar currentUser={currentUser} location={location} />
-    </React.Fragment>
-  )
+  const showNewDesign = checkFeatureFlag("mitxonline-new-header")
```
rachellougee, line 32: Spent some time debugging this with Jenni, as I noticed this still returns true even when local posthog call is disabled after testing the feature flag is on. We think it's cache issue with JS call since other feature flags use python call. It should be fine but let's see on RC

rachellougee (approved):
LGTM. Let's test it on RC to flip the flag.

Issue #1876, opened 2023-09-11
Not all departments are displaying when the catalog page loads


collinpreston:
When the catalog page is first loaded, only the departments for the courses retrieved via the first API call are shown.  At most, the departments for only 12 courses are displayed.

## Expected Behavior
All departments which have at least one course or program associated to them, depending on the current catalog tab, should be displayed.

## Current Behavior
Only departments associated with the courses or programs loaded by the catalog are displayed.

## Steps to Reproduce

1. Create 13 courses and course runs which are expected to be displayed in the catalog.
2. Add a unique department to the 13th course.
3. Load the catalog page.  Verify that the unique department is not shown until you scroll to the bottom of the catalog page.

PR #1877, opened 2023-09-11
Department REST API

collinpreston:
https://github.com/mitodl/mitxonline/issues/1876

Adds a REST API endpoint for retrieving departments in MITx Online.  The endpoint is http://mitxonline.odl.local:8013/api/departments.  The data returned from the endpoint is:
```
[
    {
        "name": "Science",
        "courses": 2,
        "programs": 0
    },
    {
        "name": "Math",
        "courses": 1,
        "programs": 0
    },
    {
        "name": "test",
        "courses": 1,
        "programs": 1
    }
]
```

Calls to this endpoint are added from the catalog page in order to display all departments for courses or programs upon initial load of the page.

Departments are only displayed based on the following:
1. If the current tab is "courses", the department must have 1 or more courses associated with it.
2. If the current tab is "programs, the department must have 1 or more programs associated with it.

# How can this be tested?

1. Create 13 courses and course runs which are expected to be displayed in the catalog.
2. Add a unique department to the 13th course.
3. Load the catalog page. Verify that the unique department associated with the 13th course is shown even if the 13th course has not yet been displayed on the catalog page.

# Additional Context
There is the possibility that a department is associated with 1 or more courses or programs that will not be visible on the catalog page due to the catalogs business logic for filtering courses and programs.  The department will still be shown.


Ferdi:
Why just the nr of course and not the list of courses ?     Is there any api/departments/[departmentid]/courses endpoint ?

collinpreston:
> Why just the nr of course and not the list of courses ? Is there any api/departments/[departmentid]/courses endpoint ?

I was thinking about including the list of course ID's since you could then use those to make a second API call to retrieve the courses themselves.  I didn't want this endpoint to become a substitute for the courses endpoint.

jkachel (changes requested):
Seems to work OK but there's some weirdness with the DRF frontend noted below.

Ferdi:
@collinpreston thanks, is there any endpoint do you get a list of courses in a given department  (api/departments/[departmentid]/courses ) ?

collinpreston:
> @collinpreston thanks, is there any endpoint do you get a list of courses in a given department (api/departments/[departmentid]/courses ) ?

@Ferdi , there is not.

PR #1878, opened 2023-09-11
Course passed if has course run certificate

annagav:
Fix https://github.com/mitodl/hq/issues/2319

Don't show green label "Course Passed" if there is no certificate.

This green label "Course Passed" shouldn't show, if there is no certificate.
Screen Shot 2023-09-11 at 3 33 30 PM

# How can this be tested?
create a passing grade for user and a revoked coure run certificate. Check the dashboard course card enrollment info. It should not show the certificate and not show the "Course Passed" tag.

### DIFF
--- a/frontend/public/src/components/EnrolledItemCard_test.js
+++ b/frontend/public/src/components/EnrolledItemCard_test.js
@@ -147,7 +147,7 @@ describe("EnrolledItemCard", () => {
     })
   })
 
-  it("doesn't renders the course passed label if course is not passed", async () => {
+  it("doesn't renders the course passed label if course is passed but no certificate", async () => {
     // If there are no grades at all
     let inner = await renderedCard()
     let coursePassed = inner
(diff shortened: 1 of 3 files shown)


frontend/public/src/components/EnrolledItemCard_test.js
```diff
@@ -147,41 +147,39 @@ describe("EnrolledItemCard", () => {
...
-    grade.passed = false
+    grade.passed = true
     enrollmentCardProps.enrollment.grades = [grade]
     inner = await renderedCard()
     coursePassed = inner.find(".enrolled-item").find(".badge-enrolled-passed")
     assert.isFalse(coursePassed.exists())
   })
   ;[true, false].forEach(isSelfPaced => {
-    it("renders the course passed label based on self_paced and course passed", async () => {
+    it("renders the course passed label based on course certificate independent of course pacing", async () => {
```
collinpreston, line 167: What is this specifically testing for?

frontend/public/src/components/EnrolledItemCard_test.js
```diff
@@ -147,41 +147,39 @@ describe("EnrolledItemCard", () => {
...
-        assert.isTrue(coursePassed.exists())
-      } else {
-        assert.isFalse(coursePassed.exists())
-      }
+      assert.isTrue(coursePassed.exists())
     })
   })
   ;[true, false].forEach(past => {
-    it("renders the course passed label based on course end and certificate available dates in past", async () => {
+    it("doesn't render the course passed label based on course end and certificate available dates in past", async () => {
```
collinpreston, line 182: Is this test needed in order to enforce a requirement?

collinpreston (changes requested):
Testing worked for me.  Couple comments and a suggestion.

Should we rename https://github.com/mitodl/mitxonline/blob/c2dc6f58fbd12248e47db4715c8ce88ad0ce90bd/frontend/public/src/lib/courseApi.js#L120 ?

frontend/public/src/components/EnrolledItemCard_test.js
annagav: It checks that the "course passed" label is shown independently of whether the course is self paced or not. But I don't know the history of why that matters. Do you have any ideas?

collinpreston: I don't think this should exist if the label is independent of whether the course is self-paced or not.

frontend/public/src/components/EnrolledItemCard_test.js
collinpreston: I agree that this should be removed.

collinpreston (approved):
Nice, looks good to me!

PR #1887, opened 2023-09-12
Updates create_courseware to allow you to specify departments (optionally creating them)

jkachel:

n/a

This adds support for assigning a courseware object created via `create_courseware` to a department or a series of departments. Optionally, you can also specify that the departments should be created if they don't already exist.

Departments are specified with the `-d/--dept/--department` flag. Creating can be toggled on by specifying `--create-depts`. 

# How can this be tested?

With just assignment: run `create_courseware <necessary bits> -d <department name>`. If the department exists, the new object will be assigned to it. If the department does not exist, it won't.

With creation toggled: run `create_courseware <necessary bits> -d <dept name> --create-depts`. The courseware will be created with the department, which will also be created if necessary.

You can specify any number of departments by specifying more than one `-d` flag. Any that don't exist will be created if that's turned on.

### DIFF
--- a/courses/management/commands/create_courseware.py
+++ b/courses/management/commands/create_courseware.py
@@ -4,7 +4,7 @@ a course run).
 """
 from django.core.management import BaseCommand
 
-from courses.models import Course, CourseRun, Program
+from courses.models import Course, CourseRun, Department, Program
 from main.utils import parse_supplied_date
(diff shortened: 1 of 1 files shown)


collinpreston (approved):
Looks good to me

PR #1891, opened 2023-09-15
Catalog page css updates - v1

collinpreston:
https://github.com/mitodl/hq/issues/1676

# Description:
This PR cleans up some of the extraneous CSS on the catalog page.  
The only visible changes this PR makes are:
1. Setting the font family.
2. Slightly decreasing the width of the department filter sidebar and catalog items.

# How can this be tested?

1. Visit http://mitxonline.odl.local:8013/catalog/.
2. Ensure that the page renders correctly for all screen sizes.
3. Ensure that the page is accessible via screen readers and can be navigated with a keyboard.

### DIFF
--- a/frontend/public/scss/catalog.scss
+++ b/frontend/public/scss/catalog.scss
@@ -1,6 +1,5 @@
 #catalog-page {
     background: var(--grey-lite, #F0F5F7);
-    font-family: 'Poppins', 'san-serif';
 
     button {
         border: none;
@@ -95,6 +94,7 @@
             color: #FFF;
             text-align: center;
             text-overflow: ellipsis;
+            font-family: 'Poppins', 'san-serif';
             /* H3 */
             font-size: 28px;
             font-style: normal;
--- a/frontend/public/scss/catalog.scss
+++ b/frontend/public/scss/catalog.scss
@@ -1,9 +1,6 @@
-#catalog-mobile-filter {
-    background: var(--blue-dark, #03152D);
-}
-
 #catalog-page {
     background: var(--grey-lite, #F0F5F7);
+    font-family: 'Poppins', 'san-serif';
 
     button {
         border: none;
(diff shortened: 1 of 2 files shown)


frontend/public/scss/catalog.scss
```diff
@@ -79,379 +78,359 @@
...
+        
+                .catalog-item {
+                    display: flex;
+                    flex-direction: column;
+                    align-items: center;
+                    border-radius: 5px;
+                    border: 1px solid var(--background-grey, #F0F5F7);
+                    background: #FFF;
+                    padding: 0;
+                    flex: none;
```
JenniWhitman, line 353: The parent a is display: block so this has no effect here. Can remove.

frontend/public/scss/catalog.scss
```diff
@@ -79,379 +78,359 @@
...
+            overflow: hidden;
+            color: #FFF;
+            text-align: center;
+            text-overflow: ellipsis;
+            /* H3 */
+            font-size: 28px;
+            font-style: normal;
+            font-weight: 700;
+            line-height: normal;
+            display: flex;
```
JenniWhitman, line 106: SHould this have `font-family: Poppins, san-serif`?

frontend/public/scss/catalog.scss
```diff
@@ -79,379 +78,359 @@
...
-                width: inherit;
-                position: absolute;
-                margin-right: 64px;
-                opacity: 0;
-            }
+    
+        #course-catalog-navigation {
+            display: flex;
+            margin: 0 auto;
+            font-family: 'Inter';
```
JenniWhitman, line 133: I add the font fallbacks in just in case - if google fonts is down, they have fonts turned off, etc - just to make sure we get the san serif default at least. This also helps with folks with special fonts turned on for accessibility reasons. In my browser, it shows the full font-family prior crossed out, and only "inter" for this which leads me to believe it might not have the fallback family in place.

JenniWhitman (changes requested):
Just a small note about one unnecessary flex: none and a bit about font family fallbacks. Otherwise looks good.

PR #1892, opened 2023-09-15
Add courses api filters to catalog page

rachellougee:

https://github.com/mitodl/hq/issues/2329

Adds additional `page__live` and `courserun_is_enrollable` filters to `/api/courses` to satisfy the course display logic in https://github.com/mitodl/hq/issues/1676#issuecomment-1655795103
- `page__live` covers 1  for Course About page 
     - this is default filter name, but can be changed if needed
-  `courserun_is_enrollable` covers 2-6 to narrow down the `course runs` that meet the following criteria
      - run is live
      - start date is not null
      - enrollment_start is in the pass
      - enrollment_end is either null or in the future

Updates the catalog page to call `page=${page}&live=true&page__live=true&courserun_is_enrollable=true`. It doesn't remove the JS logics as I am not sure if we want to do it in this PR, but API call should give the course result we need for frontend, the JS logic will not filter anything

Adds index to courserun.live 

# How can this be tested?

Adds more than 12 courses, make fewer than 12 of them meet the criteria indicated in https://github.com/mitodl/hq/issues/1676#issuecomment-1655795103
Go to http://mitxonline.odl.local:8013/catalog/, notice that page should display these courses that meet the criterial when initially loaded


courses/views/v1/__init__.py
```diff
@@ -102,20 +104,54 @@ def paginate_queryset(self, queryset):
...
     filter_backends = [DjangoFilterBackend]
     serializer_class = CourseWithCourseRunsSerializer
-    filterset_fields = ["id", "live", "readable_id"]
+    filterset_class = CourseFilterSet
 
     def get_queryset(self):
         return (
             Course.objects.filter()
             .select_related("page")
-            .prefetch_related("courseruns", "departments")
```
collinpreston, line 118: Should prefetching `courseruns` remain?  If we do decide to remove it, I noticed that a few n+1 warnings pop-up when calling the courses API.  
I was able to improve the query performance with the following changes:
1. Add `@cached_property` to `first_unexpired_run` on the `Course` model.
2. Remove `unexpired_runs` and `available_runs` from the `Course` model - these are not being used anywhere and could be replaced by filters in the future.
3. Change https://github.com/mitodl/mitxonline/blob/c09207714b4a728d8c7e2e613ed5c59b6c5b6d1c/courses/models.py#L491 to be `course_runs = self.courseruns.all().iterator()`.

Even after these changes I still notice one n+1 warning when performing GET: http://mitxonline.odl.local:8013/api/courses?id=1&live=true&page__live=true.

collinpreston (changes requested):
This is great!  Just one question and a comment.

pdpinch:
> page__live covers 1 for Course About page
> this is default filter name, but can be changed if needed

I think the filter name is ok, esp. since it is a default. 

> courserun_is_enrollable

This name kind of makes sense to me, but can you check with someone working on Open to see if we have any similar filters there? Maybe @mbertrand or @abeglova or @ChristopherChudzicki 

Thinking forward a bit, how do these filters get documented? I know we don't have drf-spectacular installed here, but is there something you can do to prepare for, with doctstrings or something? @JenniWhitman might be able to guide you if this question doesn't make sense.

courses/views/v1/__init__.py
rachellougee: I tested it with `courseruns` in prefetch_related here, there were two issues:
- It can't filter out any child `courserun` that already cached, it would just filters out results at the parent course level. For example, if we have a course with multiples runs, but only one of them meet criteria 3-6. With prefetch_related courseruns here, it would list all the runs for that course.
- To avoid the above issue, I added the prefetch_related `courserun` to `filter_courserun_is_enrollable` so we can filter out courseruns at the time of getting the object using Prefetch. But we can't have two prefetch_related courseruns, so I remove it from here.

I am wondering if it can be conditional based on the filter `courserun_is_enrollable` is applied or not. When `courserun_is_enrollable` isn't applied, add prefetch_related("courseruns"). I can check it out

courses/views/v1/__init__.py
```diff
@@ -102,20 +104,54 @@ def paginate_queryset(self, queryset):
...
 
 
+class CourseFilterSet(django_filters.FilterSet):
+
+    courserun_is_enrollable = django_filters.BooleanFilter(
+        field_name="courserun_is_enrollable",
+        method="filter_courserun_is_enrollable",
+    )
+
+    def filter_courserun_is_enrollable(self, queryset, _, value):
```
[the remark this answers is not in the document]
rachellougee: yes, I can add it

rachellougee:
> Thinking forward a bit, how do these filters get documented? I know we don't have drf-spectacular installed here, but is there something you can do to prepare for, with doctstrings or something? @JenniWhitman might be able to guide you if this question doesn't make sense.

Jenni and Matt both mentioned drf-spectacular for MIT open. I think we can add `drf-spectacular` in MITx Online. I use  `FilterSet` class here, my understanding is that it should automatically be documented when we install drf-spectacular

courses/views/v1/__init__.py
rachellougee: Ah,`prefetch_related("courseruns")` can be conditional based on `courserun_is_enrollable`. I put it back so when we call http://mitxonline.odl.local:8013/api/courses?id=1&live=true&page__live=true, it would use the prefetched courseruns object.

For these changes you mentioned, I can open a issue for that as I am hesitant to make these changes in this PR. These are good improvements to make, but I rather do some tests to make sure no side effects.

courses/views/v1/__init__.py
```diff
@@ -112,6 +112,10 @@ class CourseFilterSet(django_filters.FilterSet):
     )
 
     def filter_courserun_is_enrollable(self, queryset, _, value):
+        """
+        courserun_is_enrollable filter to narrow down runs that are open for
```
collinpreston, line 116: Can you just add a definition for what determines if a course run is open for enrollment, and what `value` is doing in regards to narrowing down runs?

courses/views/v1/__init__.py
collinpreston: Nice!

courses/views/v1/__init__.py
collinpreston: Something like:

```
Filters courses based on associated courseruns for which: live is True, start_date is not None, enrollment_start is in the past, and enrollment_end is either None or in the future.
@param {boolean} value description about how value is used.
```

JenniWhitman:
@rachellougee - we discussed drf-spectacular - is super quick to add, a little longer to make it work correctly at times. Either way, worth it, in my opinion.

Issue #1896, opened 2023-09-18
Dropdown menu does not open


annagav:
This issue is on https://rc.mitxonline.mit.edu/. This issue seems to be independent of the `mitxonline-new-header
` feature flag.
## Expected Behavior
When I click the arrow dropdown menu should open
<img width="205" alt="Screen Shot 2023-09-18 at 9 27 36 AM" src="https://github.com/mitodl/mitxonline/assets/7574259/3c28bb10-8ca9-4584-aed0-cfd87e268558">

## Current Behavior
The arrow flips but no menu is opened.
<img width="204" alt="Screen Shot 2023-09-18 at 9 28 40 AM" src="https://github.com/mitodl/mitxonline/assets/7574259/efbaa07a-667e-4c8d-b07d-cf40584397b9">

collinpreston:
Appears the cause is: https://github.com/mitodl/mitxonline/blob/c3fbddebfe11bfdb6f4ad018d23ecc5b41d4ca13/frontend/public/scss/home-page/home-page-hero.scss#L9

JenniWhitman:
@collinpreston  This makes more sense than what I was thinking - can remove that and see if it still sits above the featured carousel, if not, drop that number back... more than a bit.

JenniWhitman:
@collinpreston you're correct. I have a literal one line PR from me testing that locally

PR #1897, opened 2023-09-18
Hero sits over dropdown because of carousel z index change

JenniWhitman:
Fixes https://github.com/mitodl/mitxonline/issues/1896

Remove the z-index which is no longer necessary

# How can this be tested?
Open the home page, click the drop down on the top right while logged in - you should see it, it should not be covered by the hero.


collinpreston (approved):
Looks good.

Issue #1898, opened 2023-09-18
Styling updates to the new header design


annagav:
Here is the video with Bilbal's updates: https://www.loom.com/share/Code-Changes-and-Navigation-Improvements-70812078ae124440b1534de39b33c1cb?sid=6f96a373-e2ea-4c01-a92b-8612404a9035

PR #1899, opened 2023-09-18
User dropdown menu updates

annagav:
Fix https://github.com/mitodl/mitxonline/issues/1898

Adding styling update:
1. The dropdown menu arrow has now transition through rotation, better centering, and less thick
2. There is a light grey bar to the left of the learners name
3. The dropdown menu items are differently positioned when the menu is open

Before:
Screen Shot 2023-09-18 at 11 43 00 AM

After:

# How can this be tested?
Log in as a users and play around the the top right menu.


rachellougee (approved):
Looks good

PR #1872, opened 2023-09-10
1869: Course program api performance improvements

collinpreston:
# What are the relevant tickets?
https://github.com/mitodl/mitxonline/issues/1869

# Description (What does it do?)
Improves the performance of courses REST APIs with the following changes:
1. Remove the course page from the CourseRunSerializer since the Course serializer already contains the course page and is also defined in the CourseRunSerializer.  PR updates the front end to utilize this change.
2. Select_related and prefetch_related where possible on the Programs, Courses, and CourseRun viewsets.
3. Use iterator() when a filter query exists within a loop.
4. Set the `live` attribute as an index field on the Program and Course models.
5. Add REST API filtering to the Course and Program viewsets for id, live, and readable_id.
6. Create a DepartmentSerializer to more easily handle departments related to courses and programs.  Department serialization was previously being handled the same way across many serializers using serializer methods.
7. Cache properties where possible.

# How can this be tested?

1. Attempt making API calls to the REST endpoints listed in the Testing Results section.  Verify that you achieve a lower response time than the current `main` branch, and that all data expected is provided in the results.
2. Ensure that front-end pages function as they previously did.

# Additional Context
## Testing results
Testing was performed via GET requests using Postman.  The database records used for testing may not be extensive and do not contain all possible related records, such as a course, product, flexible price record, and flexible price request.

Recorded times indicate the lowest response time achieved for each set of GET requests for REST endpoint.  The set of GET requests was roughly 5-10 requests performed immediately after each other.

### `/courses` (32 records)
-- Before -------
9.24s

-- Now -------
516ms

### `/courses?id=4&live=true` (32 records)
-- Before ------- (filtering by readable_id)
78ms
```
mitxonline-web-1      | [2023-09-10 14:13:41] DEBUG 2339 [django.db.backends] utils.py:123 - [c18a6235ffb0] - (0.002) SELECT "courses_course"."id", "courses_course"."created_on", "courses_course"."updated_on", "courses_course"."title", "courses_course"."readable_id", "courses_course"."live" FROM "courses_course" WHERE ("courses_course"."live" AND "courses_course"."readable_id" = 'course1') ORDER BY "courses_course"."id" ASC; args=('course1',)
mitxonline-web-1      | [2023-09-10 14:13:41] ERROR 2339 [nplusone] notifiers.py:40 - [c18a6235ffb0] - Potential n+1 query detected on `Course.courseruns`
[...]

PR #1900, opened 2023-09-18
Updates the code behind the More Dates selector to work without the course page

jkachel:
The course page was removed from the courserun API response in [an earlier PR](https://github.com/mitodl/mitxonline/pull/1872), but the More Dates selector depended on it being there to determine which course run to show you. This PR changes that logic to something more simple - either you've selected a course run or you haven't. 

Fixes mitol/hq#2396 partially

If a course run has been selected, use that instead of the default (which is just the first courserun that appears in the list when we go to fetch them). 

# How can this be tested?

Set up a course with multiple concurrent runs. (For my testing, I made two with run tags ending in A and B so it'd be obvious which is which.) Then, load the course page. You should see the More Dates selector. 

Clicking Enroll Now without choosing a course in the More Dates selector should enroll you in the first course on the list. 

Choosing a course from the More Dates selector and then clicking Enroll Now should enroll you in that specific course run. 

# Additional Context

Your Enroll Now button may not change to Enrolled depending on whether or not you have a working edX instance and a corresponding course there for your test course runs. For example, I am enrolled in both these courses. However, Tutor isn't aware of either - I set one manually to be EdX Enrolled = True. (These screenshots also demonstrate that the selector works.)

Screenshot 2023-09-18 at 1 14 18 PM
Screenshot 2023-09-18 at 1 14 26 PM

### DIFF
--- a/frontend/public/src/containers/ProductDetailEnrollApp.js
+++ b/frontend/public/src/containers/ProductDetailEnrollApp.js
@@ -156,6 +156,10 @@ export class ProductDetailEnrollApp extends React.Component<
   }
 
   setCurrentCourseRun = (courseRun: EnrollmentFlaggedCourseRun) => {
+    console.log(
+      "==> we're setting currentCourseRun but is the context right??",
+      courseRun
+    )
     this.setState({
       currentCourseRun: courseRun
     })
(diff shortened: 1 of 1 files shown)


collinpreston (commented):
Removed comment

jkachel:
> Should the "more dates" section only show CourseRuns which have `live = True`?

Don't think so - I updated the template to filter out runs that aren't live in https://github.com/mitodl/mitxonline/pull/1900/commits/8081f984912e4b003ab3e89ebca8e4f9145e75c2

jkachel:
Pulled the live filtering - we can address that in a separate PR (per Slack convo) since the better way to do this potentially changes other things.

collinpreston (approved):
Nice find and fix!

PR #1904, opened 2023-09-19
Swap posthog to use js sdk with minimal settings

JenniWhitman:
Updates https://github.com/mitodl/hq/issues/2353
Updates https://github.com/mitodl/hq/issues/2388

This takes the posthog setup into one place, out of the util function, and removes most of the config options.  I also turned on posthog debug when environment is dev to assist in debugging.

# How can this be tested?
Load pages - the flags should still work and you should see the ph cookie is not very small (126 bytes give or take, instead of over 1500 as on RC presently). when using locally, should see posthog request info in console. You should also see user data for posthog in the session storage. There should not be events generated for everything you do on the page any more in the posthog application.

# Additional Info
Will have a follow up PR to swap to the React SDK instead. This one will help alleviate short term 400 issues.


jkachel (approved):
LGTM 👍 cookie size was 126 after a bunch of fiddling with it

PR #1909, opened 2023-09-20
Re-add the fix from PR 1900 to the reworked product page components

jkachel:

https://github.com/mitodl/mitxonline/pull/1900

https://github.com/mitodl/mitxonline/pull/1900 fixed an issue with the course selector in the More Dates section, which then didn't get migrated into the Program Page code (which split out the upsell stuff into separate components for course and program pages). This re-adds it to the new course page component.

# How can this be tested?

See testing parameters in https://github.com/mitodl/mitxonline/pull/1900 - it should be exactly the same.


collinpreston (approved):
Looks good to me

PR #1911, opened 2023-09-21
Self paced dates -> start any time

JenniWhitman:
Fixes https://github.com/mitodl/hq/issues/2375

adds self paced on the relevant run to the context & logic to make it visible on the page.

# How can this be tested?
Having a CMS page for a self paced and a non self paced course. Non self paced should show the relevant start date (next in the future, or recently passed with enrollment still open), self paced should show "start anytime".


annagav (approved):
Works great 👍

PR #1914, opened 2023-09-22
Hide programs tab when department has none

annagav:
Fix https://github.com/mitodl/hq/issues/2413

Hides the program tab when the department has no programs.

Coming

# How can this be tested?
Make sure to create a course for a department with no programs in the department. Go to the catalog page. When you select the department the programs tab should disappear.

### DIFF
--- a/frontend/public/scss/common.scss
+++ b/frontend/public/scss/common.scss
@@ -424,3 +424,7 @@ button.btn-secondary.unstyled {
 .form-control {
   padding: .375rem 0.75rem;
 }
+
+.display-none {
+  display: none;
+}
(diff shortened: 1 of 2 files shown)


rachellougee (commented):
The hide program tab works great. But I noticed another issue, I am not sure if you want to fix it, just let me know.

e.g. I have a program for Economic, so it displays Programs tab which is expected.
image

When I click on Programs tab, the sidebar highlights "All department" instead of "Economics" with one program
image

Then I click on "All Departments", it changes to two programs (one program doesn't have any department)
image

It's a bit confusing, but I noticed the same issue on RC.

annagav:
@rachellougee thanks for noticing that. I also thought it was strange behavior, that needs to be fixed. Since this is how it was before my changes and is not caused by this PR let's create a separate issue for this.

Issue #1919, opened 2023-09-25
Catalog page - department filter doesn't stay when clicking on "Programs" tab


rachellougee:
Noticed this bug when reviewing https://github.com/mitodl/mitxonline/pull/1914

## Expected Behavior

When clicking on "Programs" tab with a specific department, I would expect the department filter stays. 
If it doesn't stay, then the result should be accurate on new catalog page

## Current Behavior

When clicking on "Programs" tab with a specific department, the filter got reset to "All departments" but result is still for that specific department.

## Steps to Reproduce

1. Go to https://rc.mitxonline.mit.edu/catalog/
2. Click on `Economics` department, you should see
<img width="1243" alt="image" src="https://github.com/mitodl/mitxonline/assets/3138890/0951db52-3f14-4949-bf96-aa45dd6bae21">

3. Click on Programs tab, it displays two programs with "All Departments" instead of "Economics" selected
<img width="1245" alt="image" src="https://github.com/mitodl/mitxonline/assets/3138890/e06e6b93-30ed-4047-8ec4-2240be66f412">

4. Click on "All Departments", it now shows 11 programs (conflicts with above result)
<img width="1261" alt="image" src="https://github.com/mitodl/mitxonline/assets/3138890/eebf83d1-9809-4f22-adbe-c64c740f08fc">

PR #1921, opened 2023-09-26
Course Calatog: Stay on selected department

annagav:
Fix https://github.com/mitodl/mitxonline/issues/1919

Don't switch departments when courses or programs tab is selected.

# How can this be tested?
Go to catalog page. Select a department, switch to the programs tab. The same department should be selected.

### DIFF
--- a/frontend/public/src/containers/pages/CatalogPage_test.js
+++ b/frontend/public/src/containers/pages/CatalogPage_test.js
@@ -722,7 +722,7 @@ describe("CatalogPage", function() {
 
     // Change to the programs tab.
     inner.instance().changeSelectedTab("programs")
-    // Confirm that the selected department resets to "All Departments".
+    // Confirm that the selected department is the same as before.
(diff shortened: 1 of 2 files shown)


frontend/public/src/containers/pages/CatalogPage_test.js
```diff
@@ -722,14 +722,14 @@ describe("CatalogPage", function() {
 
     // Change to the programs tab.
     inner.instance().changeSelectedTab("programs")
-    // Confirm that the selected departement resets to "All Departments".
-    expect(inner.state().selectedDepartment).equals("All Departments")
+    // Confirm that the selected department resets to "All Departments".
```
rachellougee, line 725: Minor: This comment doesn't match with code now. Is this test still needed?

rachellougee (approved):
LGMT, just a small comment. Tested it that filter doesn't reset to All department when switching to programs tab now. It works as expected.

PR #1923, opened 2023-09-26
Update course info bar to respond to archived courses

JenniWhitman:
Fixes https://github.com/mitodl/hq/issues/2472

This PR conditionally displays "Course content available anytime", the callout, and "No certificate available" on courses which fall under "isArchived" which is:
- end date is earlier than now AND
- enrollment_end is empty or later than now

Screenshot from 2023-09-26 12-58-09

# How can this be tested?
To test this properly, before pulling, set up 2 courses (amongst others) with course runs that match the following:
For both: start_date and _end date are both in the past, enrollment start is in the past.
1. enrollment_end is in the future
2. enrollment_end is empty

Open these prior and should see the old behavior (look like other pages)
pull the PR
reload page
Should see the new callout & words.

### DIFF
--- a/frontend/public/scss/product-page/product-details.scss
+++ b/frontend/public/scss/product-page/product-details.scss
@@ -178,14 +178,30 @@ body.new-design {
           }
 
           .callout {
-
+            padding: 0.9375rem;
+            gap: 0.625rem;
+            flex-wrap: nowrap;
+            i.warning {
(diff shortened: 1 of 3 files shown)


rachellougee (approved):
This looks good on unenrolled archived courses. but I am also seeing this message on enrolled archived course. Not sure if that's ok.  
image

PR #1931, opened 2023-09-28
don't leak grades in program record before course completed

rachellougee:

https://github.com/mitodl/hq/issues/2303

Updates `LearnerRecordSerializer` to only return grades if one of these conditions are met:
- A certificate is generated (this has already covered in existing code)
- The course's certificate available date has passed
- If there is no course certificate available date, then the course has ended

# How can this be tested?

- Setup a program and related courses
- Setup a run 
     - start date is in the pass and end date is in the future
     - certificate available date in the future
- Enroll in program
- Create passing grade for run created above
- Go to your program record, make sure grade is not displayed for that run
- Test any condition mentioned in the description, grade should be displayed for that run

### DIFF
--- a/courses/serializers.py
+++ b/courses/serializers.py
@@ -9,7 +9,7 @@ from django.contrib.auth.models import AnonymousUser
 from django.templatetags.static import static
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
-from django.db.models import CharField
+from django.db.models import CharField, Q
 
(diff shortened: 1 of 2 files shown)


annagav:
This PR works well for the solution that is outlined in the description. 

However, I am thinking we might need a little different solution here.  
@pdpinch Do we want to show a grade that has not earned a certificate on Program Records page? 

I am thinking we can tie the grade to a certificate, and only show those grade in the grade records page.

annagav (approved):
LGTM 👍

PR #1937, opened 2023-10-02
Revert "Assorted fixes for More Dates selector in existing design"

jkachel:
Reverts mitodl/mitxonline#1903

mitodl/hq#2573

Reverts most of #1903. Specficially:

* In the CoursePage model, the relevant runs should now filter out runs with enrollment end dates in the past, so past course runs should not display
* The ProductDetailEnrollApp component should now favor the first apparent unenrolled courserun rather than just grabbing the first if one isn't explicitly selected.
* The relevant course runs queryset function now only includes live courses

(I kept the code to stop you from being able to choose a date that wasn't enrollable.) 

# How can this be tested?

Testing should be pretty close to the scenario described in mitodl/hq#2573

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -1129,15 +1129,11 @@ class CoursePage(ProductPage):
         return f"{self.course.readable_id} | {self.title}"
 
     def get_context(self, request, *args, **kwargs):
-        now = now_in_utc()
-
         relevant_run = get_user_relevant_course_run(
             course=self.product, user=request.user
         )
         relevant_runs = list(
-            get_user_relevant_course_run_qset(
-                course=self.product, user=request.user
-            ).filter(models.Q(enrollment_end=None) | models.Q(enrollment_end__gt=now))
+            get_user_relevant_course_run_qset(course=self.product, user=request.user)
         )
         is_enrolled = (
             False
--- a/cms/templates/product_page.html
+++ b/cms/templates/product_page.html
@@ -222,15 +222,13 @@
                         <p>Click below to enroll in one of these dates</p>
                       </div>
                       {% for course_run in course_runs %}
-                        <div>
-                          {% if course_run.is_not_beyond_enrollment %}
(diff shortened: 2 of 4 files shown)


JenniWhitman:
So, when I'm enrolled in a previous run, but not the open one, I still see "enrolled". I do see the correct dates (not the old one, only the future), but the button isn't changing.

jkachel:
This is what I have for testing:

I set up two new courses in a program (in a fresh MITx Online instance as it turned out), one that ran from 1-Jan-2023 to 29-May-2023 (1T2023) and one that ran from 1-Jun-2023 to 31-Dec-2023 (2T2023) (with enrollment start stop inclusive there). I had to fabricate an enrollment for my user in the first course. Once I did that I got the Enrolled button when going to the course, and clicking More Dates got me the option to select the 2T2023 run, which gave me the option to enroll. Fabricating an enrollment for 2T2023 gave me an Enrolled button - my local copy doesn't have an edX at the moment so enrolling thru the UI doesn't work - and switching between the dates in More Dates gives me the right course runs. (They have different courseware URLs.) 

Looking through the video sent, the user is jjmchale@mit.edu and that users is enrolled in 14.750x 1T2023 and 3T2023. However, in the video supplied, they don't get the More Dates at all, so that's a problem. It still favors the earlier course and that seems to work (you can see that they end up at 1T2023) but they don't have the option to select 3T2023. 

This is where I am on testing - with the additional caveat that this is with the new design stuff turned off (using my own Posthog account). I haven't tested with the new design. Let me know if I'm missing a test param here or anything like that.

JenniWhitman:
Yup - so same - but if I'm enrolled in the past run, not the present, I shouldn't see "enrolled" I should see "enroll" per the second part of that - I see the correct available date (future/present), but the button should be "enroll"

JenniWhitman:
Yes!

JenniWhitman:
Awesome. Does the thing now.

JenniWhitman (approved):
works as issue asks

PR #1940, opened 2023-10-03
Updated get_financial_assistance_form_url to also consider related programs

jkachel:

Closes mitodl/hq#2391

Updates the course page serializer to include related programs in its search for a financial assistance form URL. 

# How can this be tested?

Create two programs. One should have a full financial assistance buildout (published form, tiers, etc.); one should not. The two programs should have a RelatedProgram record and should have independent courses with products.

Attempt to view a course in the program that does not have financial assistance. You should see links to the other program's financial assistance forms, and you should be able to apply for and be approved for financial assistance as per usual. 

Attempt to purchase the verified track for a non-FA program's course after approving your financial aid request. You should get the personalized pricing for the course. (This has not changed - you should have gotten this before - but it should still work.) 

# Additional Context

We already consider related programs to determine whether or not to grant financial assistance; this just shores up the courses API for that. 

The serializer unrolls the course programs so combining the directly related course programs and the related programs for the course programs don't end up overwriting the course programs.

### DIFF
--- a/cms/serializers_test.py
+++ b/cms/serializers_test.py
@@ -246,10 +246,22 @@ def test_serialized_course_finaid_form_url(
     own_program_has_form, related_program, related_program_has_form
 ):
     """
-    Tests the financial assistance form URL returned for a course that's in a
-    program that has no financial assistance form and is related to a program
-    that does have a financial assistance form. (The course should get the
-    form URL for the program that does have the financial assistance form.)
+    Tests a few scenarios for financial assistance form URL retrieval to ensure
+    that the proper form is displayed. The FA form URL should prefer the form
+    that belongs to the program that the course is a member of; if that program
+    lacks a FA form, it should use the FA form of a related program.
+
+    - own_program_has_form flags whether or not the course under test's program
+    has its own financial assistance form. If this is set, then the course
+    should return back the form for the program it's in directly.
+
+    - related_program flags whether or not the course under test's program is
+    related to another program at all. (We also want to make sure the financial
+    assistance form for the other program isn't used by the course under test.)
+
+    - related_program_has_form flags whether or not the secondary related
+    program has a financial assistance form. This form will then be used only if
+    own_program_has_form is False.
     """
 
     program1 = ProgramFactory.create()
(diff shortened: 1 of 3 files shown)


collinpreston (commented):
When running GET `http://mitxonline.odl.local:8013/api/courses/?readable_id=course30&live=true` I saw two N+1 warnings:
```
web_1       | [2023-10-11 10:39:46] ERROR 26 [nplusone] notifiers.py:40 - [2bc4de06c073] - Potential n+1 query detected on `RelatedProgram.first_program`
web_1       | [2023-10-11 10:39:46] ERROR 26 [nplusone] notifiers.py:40 - [2bc4de06c073] - Potential n+1 query detected on `RelatedProgram.first_program`
```
I was able to get rid of these warning by appending `.iterator()` to `self.related_programs_qs.all()` at https://github.com/mitodl/mitxonline/blob/1e2f0b7ea5fd5b6e7e366c4dab3aa3b9659ac055/courses/models.py#L188

collinpreston (changes requested):
Functional testing looks good.  I think we should have a unit test for this use case (flexible price retrieved for a course based on course -> program -> related program -> flexible price).

cms/serializers_test.py
```diff
@@ -225,3 +230,65 @@ def test_serialize_course_page_with_flex_price_form_as_child_no_program(
...
+    ],
+)
+def test_serialized_course_finaid_form_url(
+    own_program_has_form, related_program, related_program_has_form
+):
+    """
+    Tests the financial assistance form URL returned for a course that's in a
+    program that has no financial assistance form and is related to a program
+    that does have a financial assistance form. (The course should get the
+    form URL for the program that does have the financial assistance form.)
```
collinpreston, line 252: I would just update this docstring to better describe the different cases defined by your parameter values (`own_program_has_form,related_program,related_program_has_form`).  This unit test is testing more situations than the current docstring describes.

collinpreston (approved):
Everything looks good.  Comment just to add more descriptive docstring to the unit test.

cms/serializers_test.py
jkachel: Got this in https://github.com/mitodl/mitxonline/pull/1940/commits/0531caa2c5967acc0cdafbcf593ac8f7882125c0 - since this is just a comment change, going ahead with merging.

PR #1941, opened 2023-10-04
Catalog Page: Add content loader 

annagav:
Fix https://github.com/mitodl/hq/issues/2328

Installs [react-content-loader](https://github.com/danilowoz/react-content-loader)
adds content loader for courses and programs in the catalog page

Screen Shot 2023-10-04 at 10 22 55 AM

# How can this be tested?
Go to catalog page while the courses are loading it should show the loader.

### DIFF
--- a/frontend/public/src/components/CourseLoader.js
+++ b/frontend/public/src/components/CourseLoader.js
@@ -1,7 +1,7 @@
 import React from "react"
 import ContentLoader from "react-content-loader"
 
-const SkeletonLoader = props => (
+const CourseLoader = props => (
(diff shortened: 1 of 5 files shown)


rachellougee (approved):
Looks great. Not sure if we need 12 ContentLoader to match with page_size? Maybe 3 is fine as we don't have design for it

annagav:
@rachellougee thanks for the review! I made few more changes to show course loader at the bottom of the page. Could you take a look?

rachellougee (approved):
Change looks great. 👍

PR #1942, opened 2023-10-04
Course About Redesign - Accessibility issues on instructor bio

rachellougee:

https://github.com/mitodl/hq/issues/2601

Address accessibility issues on instructor section of course about page (new design)
 - make instructor name keyboard focusable
 - return focus on instructor name when modal closes

# How can this be tested?

- Setup course/program about page for new design
- Add Faulty Members on course/program about page from CMS
- Go to course/program page
- Use screen reader to tab through the above page, you should now be able to tab to each instructor and open modal. 
- When instructor modal closes, the screen reader should now focus on instructor name.


annagav (approved):
LGTM 👍

Issue #1682, opened 2023-06-20
Tune uWSGI settings based on application profile


rhysyngsun:
See my writeup on how to approach this: https://github.com/mitodl/hq/discussions/393

The current settings on xPro (workers = 3, threads = 25) is probably a good starting point for values.

NOTE: MicroMasters is an older application, it's probably going to be more difficult to do this work, so I would recommend not spending a lot of time fine-tuning the settings and instead apply some reasonable settings, monitor on newrelic, and tune further if needed.

pdpinch:
> Creating the locustfile.py for this - Ibrahim did not add an example file to xpro that I saw, but I plan to both to aid in quicker testing and as a breadcrumb for future tests

@ibrahimjaved12 was working on ocw-studio. If he hasn't shared the file in that repo, he should. 

Otherwise, this looks like the right plan. Let's get started!

JenniWhitman:
I meant OCW :facepalm: The locust file is in a comment on the PR, so I have his structure (thankfully! Hence why I expect this to be wrapped up so quickly), but i think the better breadcrumb moving forward would be to include things in a doc or example file. We have a docs folder on mitxonline, so I'm planning to make sure it's noted so we can "re-tune" quickly and easily in the future :)

JenniWhitman:
This work helped identify performance issues on /catalog of the redesign.  That is being tracked here: https://github.com/mitodl/hq/issues/2781. The completion of this work is pending changes to the catalog page and completion of that ticket. At that point I will rerun tests and we'll work with what we have then.

pdpinch:
@JenniWhitman are you actively working on this? If not, we should move it back to in triage.

JenniWhitman:
@pdpinch 
No, I'm not actively working on it right now, I was waiting on completion of the catalog work, however, seeing as we're not making any more major API changes we COULD go back and rerun this/do some tuning.

JenniWhitman:
(so, ready column over triage potentially)

PR #1947, opened 2023-10-06
wsgi tuning setup config

JenniWhitman:
Updates https://github.com/mitodl/mitxonline/issues/1682

Installs uwsgitop which we'll need on rc to run this
has instructions and an example file to set up locust locally which will be used to generate both local and RC traffic

# How can this be tested?
Using the information provided, get locust running and sending requests to your local application and rc - if you can use this to set it up, I've done my job.

# Additional Context
This is essentially this: https://github.com/mitodl/ocw-studio/pull/1898 with a little more information around the locust setup

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,3 +126,4 @@ celerybeat-schedule
 docker-compose.*.yml
 
 .python-version
+/locustfile.py
--- a/locustfile.py.example
+++ b/locustfile.py.example
@@ -1,0 +1,92 @@
+import logging
+from locust import HttpUser, task, between, events
+
+
+def _get_product_json(response_json, page_type):
+    logging.info(f'User was displayed {response_json["count"]} {page_type}(s)')
+    pages = response_json["results"][:5]
+    return [page["readable_id"] for page in pages]
+
+
+class DjangoAdminUser(HttpUser):
+    wait_time = between(1, 5)
+
+    def on_start(self):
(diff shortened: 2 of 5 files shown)


JenniWhitman:
I will say the only reason the locustfile.py is here right now is so locust starts (you need a valid user). Don't worry about the test passing, not, etc - just make sure you can load it and stuff happens when you hit the button :) We'll critique the test itself in the last PR

.gitignore
```diff
@@ -126,3 +126,4 @@ celerybeat-schedule
 docker-compose.*.yml
 
 .python-version
+/locustfile.py
```
JenniWhitman, line 129: This is just so that when we DO add locustfiles, they don't accidentally go to prod - locust won't run without one, so we won't accidentally... ddos production somehow?

locustfile.py.example
```diff
@@ -0,0 +1,92 @@
+import logging
```
JenniWhitman, line 1: Again - we're not looking at this with heavy scrutiny yet (or ever really - adjust this to your needs - this is just what I'm using this time around). But feel free to take a look and check out the output as things run (yes, 2 endpoints are angry and I'm still deciding what to do about it)

annagav (approved):
I was able to follow the instruction in the RST file and set it up successfully. 👍

Issue #1934, opened 2023-09-29
Learners that completed DEDP before the split need to see their certificate


annagav:
14.009x used to be one of the electives for DEDP. Now that the Program was split into 2 programs 14.009x belongs only to one of the programs. So now the user has one elective passed in one program, and one in the other.
 
Based on the rules from before the separation the learner would have earned the program certificate.

Related issues:
https://github.com/mitodl/hq/issues/1861
https://github.com/mitodl/hq/issues/2501
https://github.com/mitodl/hq/issues/2499
https://github.com/mitodl/hq/issues/2685
https://github.com/mitodl/hq/issues/2681
https://github.com/mitodl/hq/issues/2680
https://github.com/mitodl/hq/issues/2678
https://github.com/mitodl/hq/issues/2699
https://github.com/mitodl/hq/issues/2696

## Expected Behavior
Learner that completed requirements for old program should receive their certificate.

## Current Behavior
The learner does not receive a program certificate.

annagav:
What needs to be done:

1. Create a new hidden program.
2. Enroll select users in the program. I think the title would need to change not to confuse it with current DEDP program.
3. Make sure that program shows on their dashboard.
4. Generate certificates for this program.
5. Make sure they can see the certificate on their dashboard.

pdpinch:
Can we try this on RC first?

pdpinch:
We can give the "legacy" program an id of `program-v1:MITx+DEDP+2T2023` because these were folks who passed 14.009x in 2T2023

annagav:
I tested this on RC, everything seems to have worked 👍 
I am ready to do this in production.

pdpinch:
Great, thanks. What's the process for generating a certificate for a learner? 

Can you give me some links to take a quick look at? I'd like to see what the program looks like in django admin, and also what a certificate looks like.

annagav:
I just created a certificate manually in the django admin for my user https://rc.mitxonline.mit.edu/admin/courses/programcertificate/. You can then view the certificate on the user's dashboard. You can view my users certificate `annagav2`.

pdpinch:
Ok, so in production, we would create ~7 certificates manually and we're done?

annagav:
Here is what the dashboard looks like for these users:
No certificate yet:
[...]

PR #1954, opened 2023-10-16
Filter out not live programs

annagav:
Fix https://github.com/mitodl/mitxonline/issues/1934

When enrolling user in a course, enroll them also in associated LIVE programs only.

# How can this be tested?
Enroll you user in a course run that is a requirement in a program. If the program is not "live" the should be no program enrollment.
If you set  `program.live = True` then when you enroll in a course a program enrollment should be created.


rachellougee (approved):
LGTM

PR #1957, opened 2023-10-20
Course About Page: update font and styles

annagav:
Fix https://github.com/mitodl/hq/issues/2618
Fix https://github.com/mitodl/hq/issues/2647

Following the designs here https://www.figma.com/file/Gs4zIhOFv5gVvafawwl6PF/MITx-Online?type=design&node-id=0%3A1&mode=dev.

Updated the font family for new-design styles.

Before:
Screen Shot 2023-10-20 at 9 12 52 AM

After:
Screen Shot 2023-10-20 at 10 20 35 AM

### DIFF
--- a/frontend/public/scss/product-page/product-details.scss
+++ b/frontend/public/scss/product-page/product-details.scss
@@ -1,14 +1,23 @@
 // sass-lint:disable mixins-before-declarations
 
 body.new-design {
+  font-family: Inter;
+
+  h1, h2 {
+      font-family: Poppins;
+
+  }
+
   .product-page {
     margin-top: 1.5rem;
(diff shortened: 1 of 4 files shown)


rachellougee (approved):
The fonts are Inter now. The line-height also matched with design, which 38px, but it doesn't look the same as in the screenshot reported in https://github.com/mitodl/hq/issues/2647. That's design issue or my eyes
image

Anyway, LGTM if we match to design

annagav:
@rachellougee I was also confused about the screenshots in the issue. But those were taken from production old design. And for this fix we are implementing the figma design, which is different.

Thanks for the review!

PR #1960, opened 2023-10-23
Course About Page: tablet and mobile view adjustments

annagav:
Fix https://github.com/mitodl/hq/issues/2570
Adds padding for before the Course Info Box.
Aligns the course info box to be in the middle for tablet and mobile view.

Screen Shot 2023-10-23 at 10 31 37 AM
Screen Shot 2023-10-23 at 10 30 59 AM

### DIFF
--- a/cms/templates/product_page.html
+++ b/cms/templates/product_page.html
@@ -13,7 +13,7 @@
 {% if new_design %}
     <div id="main" class="product-page">
       <div class="container">
-        <div class="row d-flex flex-row align-items-center">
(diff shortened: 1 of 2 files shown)


rachellougee:
Should we make course info container and product info container the same size? It's now centered but the issue says "The containers would be the same size". I am not sure though. @annagav

annagav:
We can't really make them same size bc the the course info box was designed to be like a card on the right side of the page. If we were to make it flexible it would distort the design.

rachellougee (approved):
Sounds good to me.

PR #1962, opened 2023-10-26
Course API test updates

JenniWhitman:
Closes https://github.com/mitodl/hq/issues/2795

Adds in:
- upper limit query counting, based on current, bad state
- nplus one fixture (auto use)
- utility to look for duplicate queries
- more robust/true to "life" fixture for API data 

# How can this be tested?
Pull the branch and run tests via pytest. You can run with --v to see more output (including information about duplicated queries.

One helpful bit might be my math on the number of queries is a little fuzzy as there's a weird edge case. I'm hoping this is solved once we go in and fix these, but calling this out.

checklist, per issue:
- make sure we're looking for nplusone situations
- make sure there are test assertions around query count with baselines based on current state
- Ensure tests are run against sufficiently sized datasets

### DIFF
--- a/courses/views_test.py
+++ b/courses/views_test.py
@@ -9,7 +9,6 @@ import random
 import reversion
 from django.db.models import Count, Q
 from django.urls import reverse
-from nplusone.core.profiler import Profiler
 from pytest_django.fixtures import django_assert_max_num_queries
 from requests import ConnectionError as RequestsConnectionError
 from requests import HTTPError
@@ -41,6 +40,7 @@ from courses.serializers import (
 from courses.views.v1 import UserEnrollmentsApiViewSet
 from ecommerce.factories import LineFactory, OrderFactory, ProductFactory
 from ecommerce.models import Order
+from fixtures.common import raise_nplusone
 from main import features
 from main.constants import (
     USER_MSG_COOKIE_NAME,
@@ -52,19 +52,11 @@ from main.test_utils import assert_drf_json_equal, duplicate_queries_check
 from main.utils import encode_json_cookie_value
 from openedx.exceptions import NoEdxApiAuthError
 
-pytestmark = [pytest.mark.django_db]
+
+pytestmark = [pytest.mark.django_db, pytest.mark.usefixtures("raise_nplusone")]
 
 
 EXAMPLE_URL = "http://example.com"
-
-
-@pytest.fixture(autouse=True)
-def _raise_nplusone(request):
-    if request.node.get_closest_marker("skip_nplusone"):
-        yield
-    else:
-        with Profiler():
-            yield
 
 
 def _num_queries_from_course(course):
@@ -108,8 +100,13 @@ def courses():
     return CourseFactory.create_batch(3)
 
 
-@pytest.fixture
-def course_catalog_api():
+@pytest.fixture()
+def course_runs():
+    """Fixture for a set of CourseRuns in the database"""
+    return CourseRunFactory.create_batch(3)
+
+
+def populate_course_catalog_data(num_courses, num_programs):
     """
     Current production data is around 85 courses and 150 course runs. I opted to create 3 of each to allow
     the best course run logic to play out as well as to push the endpoint a little harder in testing.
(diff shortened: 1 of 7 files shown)


courses/views_test.py
```diff
@@ -50,6 +58,44 @@
 EXAMPLE_URL = "http://example.com"
 
 
+@pytest.fixture(autouse=True)
+def _raise_nplusone(request):
+    if request.node.get_closest_marker("skip_nplusone"):
+        yield
+    else:
+        with Profiler():
+            yield
```
rhysyngsun, line 67: So we can reuse this in other test files, define this at the root of the project, but without `autouse-True` and instead in this module and any others that need it use:
```python
pytestmark = pytest.mark.usefixture("raise_nplusone")
```

courses/views_test.py
```diff
@@ -62,6 +108,79 @@ def courses():
     return CourseFactory.create_batch(3)
 
 
+@pytest.fixture
```
rhysyngsun, line 111: This shouldn't be a fixture the way it's currently written because it's just a plain function.

If you want to go with a fixture, you'd want to set it up with [indirect parameterization](https://docs.pytest.org/en/7.3.x/example/parametrize.html#indirect-parametrization) with `num_progams` and `num_courses` fixtures that use `request.param` or a default so that you'd be able configure it like this:

```python
@pytest.mark.parametrize("num_progams", [100], indirect=True)
@pytest.mark.parametrize("num_courses", [3], indirect=True)
def test_func(course_catalog): ...
```

main/test_utils.py
```diff
@@ -25,19 +28,23 @@ def assert_not_raises():
...
+        ignore_order (bool): Boolean to ignore the order in the result
     """
     json_renderer = JSONRenderer()
     converted1 = json.loads(json_renderer.render(obj1))
     converted2 = json.loads(json_renderer.render(obj2))
-    assert converted1 == converted2
+    if ignore_order:
+        assert DeepDiff(converted1, converted2, ignore_order=ignore_order) == {}
+    else:
+        assert converted1 == converted2
```
rhysyngsun, line 47: Could you explain why this change was needed/useful?

rhysyngsun (changes requested):

main/test_utils.py
JenniWhitman: Yup, I thought I addressed it, but I did not - 

The serializer (`CourseWithCourseRunsSerializer`) we use presently in `/courses` doesn't order the course runs, in theory you get them back in the same order, but I was getting mixed results.  The endpoints themselves are all ordered (I think, I might be wrong, I have not checked all), but the related objects are not. The order does not matter that I am aware of, so I added in this option to ignore order. 

This can be removed with changes to the API - the reason this worked prior and does not now is that we did not ever look at the course runs on the course endpoint - we would only add the top level model an endpoint returned, which meant the lack of ordering was not a problem prior as the course runs were not included on `/courses`.

courses/views_test.py
JenniWhitman: Yes, fair. Thank you! On it.

courses/views_test.py
JenniWhitman: Yes yes.
Should have to start.

courses/views_test.py
```diff
@@ -115,62 +222,132 @@ def test_delete_program(user_drf_client, programs):
     assert resp.status_code == status.HTTP_405_METHOD_NOT_ALLOWED
 
 
-def test_get_courses(user_drf_client, courses, mock_context):
+@pytest.mark.parametrize("num_courses", [100])
+@pytest.mark.parametrize("num_programs", [15])
```
JenniWhitman, line 226: @rhysyngsun since it really is a function and not a fixture, I opted instead to just parameterize two variables for the test

courses/views_test.py
JenniWhitman: I opted instead to just parameterize the values a la https://github.com/mitodl/mitxonline/pull/1962/files#r1373812611 - it really is a function, so I also renamed `course_catalog_api` to `populate_course_catalog_data`.

courses/views_test.py
JenniWhitman: https://github.com/mitodl/mitxonline/pull/1962/files#diff-4ff824f0186901e292c8550b4150c3c05f94653f487c498a3dc384a4ff998da5 moved it.

rhysyngsun (approved):

PR #1963, opened 2023-10-26
Program API test updates

JenniWhitman:
Closes https://github.com/mitodl/hq/issues/2801

Adds in:
- upper limit query counting, based on current, bad state
- nplus one fixture (auto use)
- utility to look for duplicate queries
- more robust/true to "life" fixture for API data 

# How can this be tested?
Pull the branch and run tests via pytest. You can run with --v to see more output (including information about duplicated queries.

One helpful bit might be my math on the number of queries is a little fuzzy as there's a weird edge case. I'm hoping this is solved once we go in and fix these, but calling this out.

checklist, per issue:
- make sure we're looking for nplusone situations
- make sure there are test assertions around query count with baselines based on current state
- Ensure tests are run against sufficiently sized datasets

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -1,6 +1,11 @@
 Release Notes
 =============
 
+Version 0.77.1 (Released October 23, 2023)
+--------------
+
+- Catalog Page: Mobile view changes (#1956)
+
 Version 0.77.0 (Released October 18, 2023)
 --------------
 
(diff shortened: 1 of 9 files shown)


rhysyngsun (approved):
LGTM

jkachel (approved):
LGTM too 👍

PR #1965, opened 2023-10-27
Enrollment dialog WCAG fixes

jkachel:

Closes mitodl/hq#2616

Makes a bunch of changes to the enrollment upsell modal and the more dates selector:

**More Dates**
- No longer clickable - you can view the dates but it's just informational now.
- Enrolled course runs should now appear green with the "Enrolled" text if you're enrolled in the course. (This was there, but wasn't working - it now considers your actual enrollments rather than rely on the courserun API.)

**Upsell Modal**
- If there's an upgrade available, clicking Enroll Now won't immediately enroll you in the course (at all)
- If there are multiple course runs, you should be able to select what course run you want to enroll in
- Selecting a different course run should update the modal text appropriately (upgrade deadlines, prices, etc.) 
- Selecting the "Enroll and Pay" button should drop you into the cart
- Selecting "Enroll for Free" should just enroll you in the course and should bring you to the dashboard as per usual
- If financial assistance is available, you should see "Need financial assistance?"
- Clicking the "Need financial assistance?" link should bring you to the appropriate financial assistance form
- Closing the upsell modal should leave you on the page - you should _not_ be enrolled in the course and you should not go to the dashboard, as you did before
- If the course has multiple course runs and you're enrolled in one of them, the one you're enrolled in should be selected

Date selector version:
image
Date selector version with a payment deadline:
image
No date selector:
image

# How can this be tested?

Automated tests should pass.

Selecting More Dates in the Course Info Box on a course with multiple course runs should display the dates. Selecting individual dates should have no effect, and runs that you're enrolled in should have "<date> - Enrolled" text and be presented in green.

Selecting Enroll Now on a course with a single active course run and a product should pop the upsell modal (see screenshot). You should not see the date selector. An enrollment should not be created when you see the modal. Clicking Enroll and Pay should allow you to pay for the course. Clicking Enroll for Free should enroll you in the course and deliver you to the dashboard. Clicking Need financial assistance? (if available) should bring you to the financial assistance form. Clicking the close button or outside the modal should close the modal, and you should remain on the course detail page; again, no enrollment should be created.

Selecting Enroll Now on a course with multiple active course runs (with products) should pop the upsell modal. You should see the date selector, and choosing different dates should update the modal as appropriate. (For testing, it would be a good idea to have at least one with a payment deadline.) The modal should otherwise work like the version for a single active course run.

# Additional Context

Some styling changes were made to select controls and to modal headers that are global - the design indicates that there were changes here so I applied them globally. 

This needs some more tests to cover some scenarios but will do that in a separate PR so we can get this out the door.


jkachel:
Per conversation following standup on 10/30, I'm waiting on new design for this. The gist of it is this:
- Eliminating the program membership section entirely
- For courses with multiple available course runs: 
   - Hiding the upsell messages if there's not a selected course run
   - No automatic enrollment of the learner - they must make a choice of date first
   - Updates to the button styling for the date selector bar
   - Making the toast message appear only after the enrollment has been completed when a date is selected
- For courses that are eligible for upgrade but don't have multiple available course runs:
   - The toast should appear once the enrollment is completed
   - Continue to hide the date selector

These may be separate modals or a combined modal with individual pieces componentized out; that's an implementation detail for me, though. 

Potential questions (mostly for @pdpinch and @mbilalmughal): 
- If the learner's already enrolled in an active course run, do we just show them the simple upsell dialog for the course that they're already in? This seems like the easiest option.
- When the learner selects a course run, do we then lock out the date selector (and make them cancel and restart if they've chosen the wrong date)? We're now going to fire the enrollment when the date selector changes (we want them to enroll in audit mode first, then upgrade if the upsell messaging works on them) so if we allow them to change the date, we now have to unenroll them. The code is already there for this, and we can continue to do it this way, but it does potentially mean some additional enrollments if someone's particularly indecisive.

pdpinch:
> If the learner's already enrolled in an active course run, do we just show them the simple upsell dialog for the course that they're already in? This seems like the easiest option.

We have a user story that users need to be able to re-enroll in a course, even if they are already enrolled. So I think it would be best to follow the same flow, even if the user is already enrolled. I guess this means that ideally we would disable the button for the already-enrolled dates and somehow indicate that the user is already enrolled.

pdpinch:
> We're now going to fire the enrollment when the date selector changes

@mbilalmughal suggested have a button on the 1st modal to proceed or continue. I think we should hold off on the making the enrolment until the user has clicked that.

frontend/public/scss/product-page/product-details.scss
```diff
@@ -325,7 +321,7 @@ body.new-design {
 
     .acheiving-text {
```
pdpinch, line 322: I know it doesn't matter how this is spelled, but after we make the text changes, it will prob confuse folks to have this as the class name.

frontend/public/src/components/CourseProductDetailEnroll.js
```diff
@@ -199,28 +330,24 @@ export class CourseProductDetailEnroll extends React.Component<
...
+                </div>
               </div>
-            </div>
+            ) : null}
 
-            <div className="row">
-              <div className="col-12">
-                <p className="acheiving-text">
-                  Acheiving a certificate has its advantages:
-                </p>
```
pdpinch, line 223: We need a clearer call to action here. 

"Do you want you want to earn a certificate?"

frontend/public/src/components/CourseProductDetailEnroll.js
```diff
@@ -168,12 +223,88 @@ export class CourseProductDetailEnroll extends React.Component<
...
+
+  getEnrollmentForm(run: EnrollmentFlaggedCourseRun, showNewDesign: boolean) {
+    const csrfToken = getCookie("csrftoken")
+
+    return showNewDesign ? (
+      <form action="/enrollments/" method="post">
+        <input type="hidden" name="csrfmiddlewaretoken" value={csrfToken} />
+        <input type="hidden" name="run" value={run ? run.id : ""} />
+        <button type="submit" className="btn enroll-now enroll-now-free">
+          Or, Enroll for Free without a certificate
```
pdpinch, line 268: Remove "Or, " -- it just reads strange. 

Can we bold Enroll for Free? 

```suggestion
          <em>Enroll for Free</em> without a certificate
```

frontend/public/src/components/CourseProductDetailEnroll.js
```diff
@@ -248,7 +375,17 @@ export class CourseProductDetailEnroll extends React.Component<
...
+                      fill="#6F7175"
+                    />
+                  </svg>
+                </div>
                 <p>
                   Certitficate track:{" "}
```
pdpinch, line 390: typo 
```suggestion
                  Certificate track:{" "}
```

pdpinch (changes requested):
Some text changes

frontend/public/src/components/CourseProductDetailEnroll.js
rhysyngsun: There's a typo in ^that, just in case you were going to copy/paste.

jkachel:
> Some questions:
> 
>     * When in the flow does the user get enrolled? I'm expecting it to be after they click "Enroll and Pay button" or when they click the "Enroll for free" link.

When they click either of those two links/buttons.

>     * How/does this handle re-enrollment?

If you're already enrolled in the audit track for the course, the selector should default to that course run. 

>     * Instead of hiding the drop-down when there is only one date, is it possible to keep the drop-down with only one choice and change the display text to "There is one session available:" ? This could be an enhancement in a future PR.

I rolled this in with the other changes - it's more uniform looking I think (but maybe the dropdown should be set to disabled).

frontend/public/src/components/CourseProductDetailEnroll.js
```diff
@@ -119,16 +140,34 @@ export class CourseProductDetailEnroll extends React.Component<
...
+    // Find an existing enrollment - the default should be the audit enrollment
+    // already have, so you can just upgrade in place. If you don't, you get the
+    // current run (which should be the first available one).
+    const { enrollments } = this.props
+
+    if (enrollments) {
+      const firstAuditEnrollment = enrollments.find(
+        (elem: RunEnrollment) =>
+          elem.run.course.id === run.course.id &&
+          elem.enrollment_mode === "audit"
```
rhysyngsun, line 153: Nit: `elem` as a var name I wouldn've expected to be a dom element.

frontend/public/src/components/CourseProductDetailEnroll.js
```diff
@@ -248,9 +381,19 @@ export class CourseProductDetailEnroll extends React.Component<
...
+                      fill="#6F7175"
+                    />
```
rhysyngsun, line 392: Nit: this should probably be extracted into a file-level `const` definition since it clutters the function a tad.

rhysyngsun (commented):
Some very nitpicky comments.

jkachel:
> Some very nitpicky comments.

These should be resolved in https://github.com/mitodl/mitxonline/pull/1965/commits/ff6d4be16597fb65109db59c396c4773b08a53de .

annagav:
The styles for smaller resolution and phone view feels a bit off, but everything seems to work as expected:

Screen Shot 2023-11-01 at 5 18 59 PM
Screen Shot 2023-11-01 at 5 19 30 PM

annagav (approved):
The dialog works as described in the issue. I was able to select a particular run in the dropdown selection and it took me to checkout page with the correct price. 👍

PR #1966, opened 2023-10-30
Programs API v2

JenniWhitman:
Closes https://github.com/mitodl/hq/issues/2802

- Namespaces the v1 and v2 APIs
- refactors the API code into v2 and v2
- Removes courses from the programs endpoint

# How can this be tested?
Check that tests run.  You'll notice that we're doing far less queries than we were on the current main branch.
Hit the endpoint (http://mitxonline.odl.local:8013/api/v2/programs/) and verify the structure is the same as main branch but with flat list of IDs for courses rather than full serialized objects.


courses/serializers/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+        if page
+        and page.thumbnail_image
+        and page.thumbnail_image.file
+        and page.thumbnail_image.file.url
+        else static("images/mit-dome.png")
+    )
+    return urljoin(settings.SITE_BASE_URL, relative_url)
+
+
+class BaseProgramRequirementTreeSerializer(serializers.ListSerializer):
```
jkachel, line 33: We talked about this (very) briefly in Slack but I'm not sure this goes here now? Since we're not using it in v2 and we're leaving the v1 stuff as-is (so v2 is just deltas) I think this is best kept alongside the v1 serializers that use it. My thinking is that if we need a v2 API for this then we'll probably also have a use case for refactoring the `ProgramRequirementTreeSerializer` and/or this serializer as well. 

(The only thing that uses this, I think, is the `ProgramAdminForm` - it's the only place right now that needs a fully realized requirements tree. If we expose this in the frontend later, we will have some additional requirements for that at that point in time.)

What do you think?

jkachel (commented):
I had a comment about the location of the `BaseProgramRequirementTreeSerializer` but otherwise this functionally looks good and the query count is much better.

courses/serializers/__init__.py
JenniWhitman: So, it is subclassed by the `[ProgramRequirementTreeSerializer](https://github.com/mitodl/mitxonline/pull/1966/files#diff-551e72a56450d28fbe49c33b7992eccf1b5b990308a7198a77cec2bc1726b2c3R94)` - while there's no real change, I think that was the reason I pulled it up. I can drop it if you don't think it's useful to have at the root, I just didn't want to import from v1 here.

jkachel: Ok, yeah.. I think my brain went splat on this again. I think this is OK as is in this case.

jkachel (approved):
LGTM 👍

rhysyngsun (changes requested):
Some general feedback before I get further into the review:

The module structure is a bit odd, I assume it's meant to mirror how the views modules are setup but I've not really been a fan of having a bunch of code live in an `my/module/name/__init__.py` file when there's no sibling files as opposed to `my/module/name.py` - typically we don't put code in `__init__.py` files, although an exception was made previously for the views - so there's some extra cognitive overhead that when I'm searching for view code that I need to remember that for the courses app, it lives in an `__init__.py` file. The existence of `my/module/name/views_test.py` that tests this init kind of illustrates the weirdness here too.

So I'd propose doing one of the following package structures instead:

```bash
courses/serializers/base.py
courses/serializers/base_test.py
courses/serializers/v1.py
courses/serializers/v1_test.py
courses/serializers/v2.py
courses/serializers/v2_test.py
```
Or:
```bash
courses/serializers/base.py
courses/serializers/base_test.py
courses/serializers/v1/programs.py
courses/serializers/v1/programs_test.py
courses/serializers/v1/courses.py
courses/serializers/v1/courses_test.py
courses/serializers/v2/programs.py
courses/serializers/v2/programs_test.py
courses/serializers/v2/courses.py
courses/serializers/v2/courses_test.py
```

The latter is probably more beneficial because it also chops up the large file, but at the cost of trickier imports probably. I'd optimize this for time, so if the latter option is more time consuming to implement, do the former.

courses/views/v1/__init__.py
```diff
@@ -38,16 +38,18 @@
     Program,
     ProgramEnrollment,
 )
-from courses.serializers import (
+from courses.serializers.v1 import (
```
jkachel, line 41: Prob should be `courses.serializers.v1.courses` and `.programs`

flexiblepricing/serializers.py
```diff
@@ -2,7 +2,8 @@
 from rest_framework import serializers
 
 from courses.models import Course, Program
-from courses.serializers import BaseCourseSerializer, BaseProgramSerializer
+from courses.serializers.v1 import BaseCourseSerializer
```
jkachel, line 5: should also be `.base` I think

jkachel (changes requested):
Couple imports need to be reworked a bit more for the new layout

courses/views/v2/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+class ProgramViewSet(viewsets.ReadOnlyModelViewSet):
+    """API viewset for Programs"""
+
+    permission_classes = []
+
+    serializer_class = ProgramSerializer
+    pagination_class = Pagination
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = ["id", "live", "readable_id"]
+    queryset = Program.objects.filter().prefetch_related("departments")
```
rhysyngsun, line 30: There's probably some other prefetches we could be doing but it looks like that's probably a taller order because a lot of it is inlined into the model class itself.

rhysyngsun (changes requested):

cms/serializers.py
```diff
@@ -164,11 +165,41 @@ def get_page_url(self, instance):
     def get_price(self, instance):
         return instance.price[0].value["text"] if len(instance.price) > 0 else None
 
+    def get_financial_assistance_form_url(self, instance):
```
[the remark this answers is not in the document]
JenniWhitman: This is less bad, I think - still not sure if there's a better way to grab the children forms - 
I COULD grab a `values_list` off the first query to determine the `selected_program` conditionals - there's also apparently a `specific()` method from wagtail that will return the most subclassed page object rather than the generic page, through fewer queries, but I'm not sure whether or not that will help here.I can easily do the first bit (grab a list of tuples and work with that rather than call `selected_program_id` twice), but I'm not well versed in the ways of wagtail and worry I'll make the `get_children()` call worse.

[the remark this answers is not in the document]
JenniWhitman: Something like 
```
        fin_form_related_program_ids = [program.id for program in instance.program.related_programs]
        fin_form_program_ids = fin_form_related_program_ids + [instance.program.id]
        flex_pricing_forms = (FlexiblePricingRequestForm.objects.filter(live=True).values_list("id", "selected_program_id", "slug", ))
        fin_assistance_pages_by_program_ids = (FlexiblePricingRequestForm.objects.filter(
            selected_program_id__in=fin_form_program_ids
        )).live().values_list("id", "selected_program_id", "slug", "get_parent_product_page__id").order_by("-id")
        selected_fin_assistance_page = next(
            (x for n, x in enumerate(fin_assistance_pages_by_program_ids) if x[1] == instance.program.id), None
        )
        if selected_fin_assistance_page is None:
            selected_fin_assistance_page = instance.get_children().type(FlexiblePricingRequestForm).live().first()

        if selected_fin_assistance_page is None and (len(instance.program.related_programs) > 0):
            selected_fin_assistance_page = next(
                (x for n, x in enumerate(fin_assistance_pages_by_program_ids) if x[1] in fin_form_related_program_ids), None
            )
        return(
            f"{instance.get_url()}{selected_fin_assistance_page[2]}/"
            if selected_fin_assistance_page is not None
            else ""
        )


```

courses/views/v2/__init__.py
JenniWhitman: Yeah - the relationships run fairly deep and I ended up adding more queries than saving. I can look at adding some more to the requirements_tree ... tree, which might help some of this, but overall we're still saving a lot with the removal of the courses model from here.  I added 1 per course by adding flexpricingform in, too, but that was to be expected as it wasn't called prior.

cms/serializers.py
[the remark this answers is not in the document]
rhysyngsun: I'd say just leave it for now cause we want to get this out today. I was more making a note about it.

rhysyngsun (commented):
The code itself looks good to me, although I might not be able to provide a timely functional test, @jkachel are you able to?

jkachel (approved):
Focused on functional testing and that looked OK to me - the data elements are where I would expect for programs. 👍

rhysyngsun (approved):

PR #1971, opened 2023-11-01
Change Catalog Page to use the new v2 programs endpoint

JenniWhitman:
Updates https://github.com/mitodl/hq/issues/2804

This PR adds the v2 update to teh ProgramQuery in queries/programs. This updates the catalog page to correctly display programs in a fraction of the time.  In order to do this I had to make a small change to teh endpoint swapping the IDs for Strings. 

I did NOT make the change on course or program pages as the current endpoints would require some backbending to get the data up front and I'd rather be able to directly get the data we want rather than either all of it or hit the endpoint one by one.

# How can this be tested?
Load up with some programs and courses, make sure you set departments!

load the catalog, make sure you see both courses and programs, as you expect. Click filters, make sure they still work. Toggle departments and make sure the correct courses/programs show.

Go to Course and Program pages and make sure they load with data.


rhysyngsun (commented):
Code looks good will do a functional test shortly.

jkachel (approved):
I think functionality looks good to me. 👍 Looks and works like it does on main.

PR #1973, opened 2023-11-01
Update how we call the posthog API to include another person identifier to allow for easier targetting

JenniWhitman:
Through the redesign, I've discovered that it's impossible to filter by user in our current setup. This adds user.id to all calls so we can use the django admin user id for a user to target them for feature flags in posthog

Closes https://github.com/mitodl/hq/issues/2896

Updates both check feature flag functions to accept a unique identifier.
Adds user id to react components/js places where it was missing.
Adds to python calls where it was missing as well.

# How can this be tested?

- Load the app locally.
- Go to your inspect window and remove the ph_ cookie for the app OR use a new browser/private session.
- Navigate around the app as usual. I suggest hitting home and catalog at the very least.
- Go to [Posthog's event viewer](https://app.posthog.com/events) - I have modified this for all users to automatically show person properties of `environment` and `user_id`. If you don't have access OR don't see those columns, I can help.
- Make sure you see events for yourself. `user_id` should match your django user's ID (likely 12, or 3 unless you generated a lot of users in your instance). The should now include `user_id` and `environment` for both python and javascript("web") SDKs.
- Now, go to one of those flags ([mitxonline-new-product-page](https://app.posthog.com/feature_flags/15708) is a good candidate as it is used in a number of places - both js and python as well. Turn the dev flag to 0% of users. This simulates the current state of production.
- Check that you can not see the changes expected (try going to http://mitxonline.odl.local:8013/catalog/ - should get a 404). To see as prod presently is, do these changes on ALL flags.
- Now add a new condition (I already have one in there that I'll leave set to 0%, but this is so you know what I did) - user_id = your_django_user_id - mine is 1, yours might vary, be sure to change this to match yours. environment=dev. Turn this to 100% - This will be how we will turn this on for a specific set of users.
- Check that the state changes and you can see redesign changes under the flag(s) you changed.


cms/models.py
```diff
@@ -750,25 +750,28 @@ def products(self):
         return page_data
 
     def get_context(self, request, *args, **kwargs):
-        if request.user.is_authenticated:
-            user = request.user.email
-        else:
-            if "anonymous_session_id" not in request.session:
-                request.session["anonymous_session_id"] = str(uuid.uuid4())
-            user = request.session["anonymous_session_id"]
```
JenniWhitman, line 758: So I was adding this UUID - but we're not using it. And I don't forsee an actual reason for it. What I did do in this PR is make it so ALL anonymous users will have the same tag - meaning we can very easily target not-logged-in users.

frontend/public/src/lib/util.js
```diff
@@ -262,12 +262,18 @@ export const getFlexiblePriceForProduct = (product: Product) => {
 
 export const intCheckFeatureFlag = (
   flag: string,
+  uniqueID: any,
```
jkachel, line 265: Nit: could this be `string|number` rather than `any`? (Also on line 285.)

jkachel (approved):
Works well - I was able to turn off random bits for my 2nd user locally without issue and the event viewer is more useful now too. I had a question about typing but it's not critical. 

One other question though - could we use a username instead of the ID? Getting the user ID out of the Django Admin isn't particularly hard but it'd probably be easier for nontechnical people to be able to work with usernames in there, as well as if we're ever to use the data in Posthog to track someone's path through the system. (That said I don't know whether or not that might pose more problems, especially with usernames that have Unicode in them. I have no real horse in this race - the ID is fine with me.)

frontend/public/src/lib/util.js
JenniWhitman: fair point - I think it was nullable in my mind, and that was my way of allowing all 3 rather than string|number. Bad plan. I tweaked it to specify both and also to not have the extraneous if statement - we are declaring there is a uniqueID so it will blow up before that point if there is not... and in some cases could evaluate to false there when it should not (it's fine to send an empty user_id... not ideal, but won't break posthog).

PR #1977, opened 2023-11-02
Removing final vestiges of clickable more dates selectors

jkachel:

Closes https://github.com/mitodl/hq/issues/2902

There were parts of the More Dates selector that were dependent on enrollments being there, which doesn't happen sometimes (either you're not logged in or you haven't got any), which caused the info box to not display. In addition, if you weren't  enrolled in the course, the additional course dates were still clickable, which we have gone away from. 

# How can this be tested?

Load a course while logged out of the app. The info box should display successfully. 

If the course has multiple course runs, you should be able to expand the list and the additional start dates should not be clickable. This should be the case if you're logged out or if you're logged in and viewing a course you're not enrolled in.


JenniWhitman (approved):
looks good

PR #1979, opened 2023-11-02
Fixing date to display if you're logged in

jkachel:

https://github.com/mitodl/hq/issues/2902

If you were logged in but not enrolled in a course, the More Dates link would work but not show anything. This fixes that - we were looking for you to be an anonymous user to display the link. 

# How can this be tested?

Log in and view a course with multiple course runs that you're not enrolled in. Clicking More Dates should show you the additional dates.

### DIFF
--- a/frontend/public/src/components/CourseInfoBox.js
+++ b/frontend/public/src/components/CourseInfoBox.js
@@ -30,12 +30,6 @@ const getStartDateText = (run: BaseCourseRun, isArchived: boolean = false) => {
     : "Start Anytime"
 }
 
-const getStartDateForRun = (run: BaseCourseRun) => {
-  return run && !emptyOrNil(run.start_date) && !run.is_self_paced
-    ? formatPrettyDate(moment(new Date(run.start_date)))
-    : "Start Anytime"
-}
-
 export default class CourseInfoBox extends React.PureComponent<CourseInfoBoxProps> {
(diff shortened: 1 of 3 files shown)


frontend/public/src/components/CourseInfoBox.js
```diff
@@ -51,11 +51,7 @@ export default class CourseInfoBox extends React.PureComponent<CourseInfoBoxProp
   }
 
   renderUnenrolledDateLink(run: EnrollmentFlaggedCourseRun) {
-    const { currentUser } = this.props
-
-    return !currentUser || !currentUser.id ? (
-      <>{getStartDateForRun(run)}</>
-    ) : null
+    return <>{getStartDateForRun(run)}</>
```
annagav, line 54: since this is so simple now, do we still need this function?

annagav (changes requested):
Maybe we could just eliminate `renderUnenrolledDateLink` function all together.  And have it like this:
```
? this.renderEnrolledDateLink(courseRun)
 : getStartDateForRun(run)}
```

annagav (approved):
👍

jkachel:
@annagav can you give this a quick re-look? functionality should be the same, but have now added tests for the functionality that broke.

annagav:
LGTM 👍

PR #1982, opened 2023-11-07
Fix Program Share URLs

JenniWhitman:
fixes https://github.com/mitodl/hq/issues/2933

Changes the URLs from re_path to path

# How can this be tested?
be enrolled in a local program, go to dashboard and click to go into the record. Record should load and you should be able to enable sharing (if everything else is correctly set up)

# Additional Context
I broke this during the API refactor.  Here is urls.py prior to that refactor: https://github.com/mitodl/mitxonline/blob/709b0e3a8d7c5dedb674009314ee4cad800154e9/courses/urls.py

### DIFF
--- a/courses/urls/__init__.py
+++ b/courses/urls/__init__.py
@@ -17,10 +17,10 @@ urlpatterns = [
 ]
 
 urlpatterns += [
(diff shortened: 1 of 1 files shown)


jkachel (approved):
LGTM 👍

PR #1983, opened 2023-11-07
Adds v2 of the courses API for the program page

jkachel:

mitodl/hq#2798

Adds a refactored courses API to the app. This API:
* Allows for filtering on specific course IDs (numeric IDs)
* By default excludes the approved financial assistance check
* Uses a cut down version of the CoursePageSerializer (named aptly `BaseCoursePageSerializer`)

With these changes, the `/api/v2/courses` page runs ~600 queries (using the normal test case) versus v1's ~2000 queries. It also contains less data. 

The v2 API is intended to be used by the catalog; the course product pages should continue to use the v1 API so that financial assistance, instructors, and pricing data is available.

# How can this be tested?

Compare loading a course (or courses) in the v1 api versus the v2 one: `/api/courses` vs `/api/v2/courses` optionally with a PK at the end. They should be largely the same but the v2 should not include financial assistance page info, instructors, and pricing data in the page, and the approved financial assistance key should always be False. 

Load multiple courses by appending `?id=<IDs separated by commas>` - i.e. `/api/v2/courses/?id=1,3,5,7`. You should just get back those courses.

### DIFF
--- a/cms/serializers.py
+++ b/cms/serializers.py
@@ -11,15 +11,12 @@ from courses.constants import DEFAULT_COURSE_IMG_PATH
 from ecommerce.models import Product
 
 
-class CoursePageSerializer(serializers.ModelSerializer):
+class BaseCoursePageSerializer(serializers.ModelSerializer):
     """Course page model serializer"""
 
     feature_image_src = serializers.SerializerMethodField()
     page_url = serializers.SerializerMethodField()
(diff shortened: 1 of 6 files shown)


JenniWhitman (approved):
Looks good. Functions as expected. I was able to see faster load time and get data in a few ways that's super helpful.

PR #1985, opened 2023-11-07
Adding a test for archived courses

jkachel:

mitodl/hq#2912

After some investigation into the ticket noted, the archived message was in fact working properly - this just adds an automated test for that case. 

This also adds a few classes to the parts of the course infobox so it's easier to write tests for them.

# How can this be tested?

Run the automated test, which should pass.

You should additionally see new class names in the course infobox for the various sections of the box:
```
course-archived-message (if visible)
course-timing-message
course-effort-message
course-pricing-message
course-certificate-message
```

# Additional Context

The rules for this are:

A course is archived (and displays the yellow archived callout) if:
- The `enrollment_end` date is unset, or set in the future
- The `end_date` is set


JenniWhitman (approved):
Looks good. I appreciate the additional classes for ease of testing. :D

PR #1987, opened 2023-11-08
Change Catalog to use v2 API endpoint for courses

JenniWhitman:
closes https://github.com/mitodl/hq/issues/2798

Since we don't want to change other pages, I've added this to a new query under catalogCourses (open to changing this name, but it made the most sense to me).

# How can this be tested?
Load up and load /catalog... careful, you might experience whiplash it's so fast now (kidding... but it's a major improvement.)

Screenshot_2023-11-08_11-23-06
image


jkachel (approved):
LGTM 👍 seems a good bit faster now

PR #1989, opened 2023-11-08
Updated v1 course API to only pull live programs; added tests

jkachel:

Closes mitodl/hq#2916

Updates the courses API to filter programs by Live - if the program isn't live, then the course's programs list shouldn't have it. 

Also added a test for the course API by readable id - there didn't seem to be one, and the new tests also checks for program data to be included (or not) appropriately. 

# How can this be tested?

Automated tests should pass.

Load a course page for a course that's in a program. You should see the program in the list in the course infobox. 

Turn off the program by going to Django Admin and unchecking the Live box, then reload the course page. You should no longer see the program in the list.

### DIFF
--- a/courses/serializers/v1/courses.py
+++ b/courses/serializers/v1/courses.py
@@ -33,7 +33,9 @@ class CourseSerializer(BaseCourseSerializer):
         if self.context.get("all_runs", False):
             from courses.serializers.v1.base import BaseProgramSerializer
 
-            return BaseProgramSerializer(instance.programs, many=True).data
+            return BaseProgramSerializer(
+                [program for program in instance.programs if program.live], many=True
+            ).data
 
         return None
 
--- a/courses/serializers/v1/courses.py
+++ b/courses/serializers/v1/courses.py
@@ -33,9 +33,15 @@ class CourseSerializer(BaseCourseSerializer):
         if self.context.get("all_runs", False):
             from courses.serializers.v1.base import BaseProgramSerializer
 
-            return BaseProgramSerializer(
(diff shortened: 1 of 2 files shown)


courses/serializers/v1/courses.py
```diff
@@ -33,7 +33,9 @@ def get_programs(self, instance):
         if self.context.get("all_runs", False):
             from courses.serializers.v1.base import BaseProgramSerializer
 
-            return BaseProgramSerializer(instance.programs, many=True).data
+            return BaseProgramSerializer(
+                [program for program in instance.programs if program.live], many=True
+            ).data
```
JenniWhitman, line 38: So, I think this also needs to check page__live - I'm adding an endpoint call for that now actually (similar behavior on catalog), but it should also check if program.page.live

JenniWhitman: Otherwise a program which is live but does not yet have a page will be linked (and said link will be broken)

JenniWhitman (changes requested):

courses/serializers/v1/courses.py
jkachel: Should be fixed in https://github.com/mitodl/mitxonline/pull/1989/commits/eb53287296b28a682c923bea12f58e689e603042

JenniWhitman: yeah :)  I don't see the program I tested with any more!

There's a test failing now, but it's passing locally so I'm not sure. The data on that test makes sense on first pass, but I didn't think too long on it.

JenniWhitman (approved):
Looks good once that test passes.

PR #1990, opened 2023-11-08
Add page__live filter to programs API and update url in catalog page query

JenniWhitman:
closes https://github.com/mitodl/hq/issues/2907

updates programs to have a  page__live query

# How can this be tested?
Have a live program with no live associated page, should not see it on catalog page.


jkachel (approved):
LGTM 👍

PR #1995, opened 2023-11-15
Headings for required courses start date

annagav:
Fix https://github.com/mitodl/hq/issues/2644

Change the start time label to div instead of h4.

Screen Shot 2023-11-15 at 10 24 59 AM

# How can this be tested?
Make sure that the design looks the same as before and nothing is broken.


JenniWhitman (approved):
After much fighting with my setup, I was able to verify that not only do things stay where they should, but the date and course name are now appropriately referenced.

PR #1996, opened 2023-11-15
Product Page: Display none for hidden text

annagav:
Fix https://github.com/mitodl/hq/issues/2596

Sets display none for hidden text on the page

Screen Shot 2023-11-15 at 1 56 27 PM
Screen Shot 2023-11-15 at 1 56 36 PM

# How can this be tested?
Use voiceover to make sure the it doesn't read text that is not visible on the page.


JenniWhitman (approved):
Screen reader complied :)

PR #1997, opened 2023-11-16
Departments v2 endpoint - now with more course and program counts

JenniWhitman:
Updates
https://github.com/mitodl/hq/issues/2992
https://github.com/mitodl/hq/issues/2911
https://github.com/mitodl/hq/issues/2910
https://github.com/mitodl/hq/issues/2823

This update to teh Departments endpoint:
- Makes the course and department counts accurate
- Moves the logic to the serializer rather than the view
- adds serializer tests
- adds view tests

# How can this be tested?
Run tests
Add departments to your local. The count in the catalog page should match the API output.
Also, double check my "math" on that logic.

### DIFF
--- a/courses/serializers/v2/departments.py
+++ b/courses/serializers/v2/departments.py
@@ -27,12 +27,19 @@ class DepartmentWithCountSerializer(DepartmentSerializer):
             & Q(live=True)
             & Q(start_date__isnull=False)
             & Q(enrollment_start__lt=now)
-            & (Q(enrollment_end=None) | Q(enrollment_end__gt=now))).values_list("id", flat=True)
+            & (Q(enrollment_end=None) | Q(enrollment_end__gt=now))
+        ).values_list("id", flat=True)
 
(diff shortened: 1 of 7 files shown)


courses/serializers/v2/departments.py
```diff
@@ -0,0 +1,49 @@
...
+class DepartmentSerializer(serializers.ModelSerializer):
+    """Department model serializer"""
+
+    class Meta:
+        model = Department
+        fields = ["id", "name"]
+
+
+class DepartmentWithCountSerializer(DepartmentSerializer):
+    """CourseRun model serializer that includes the number of courses and programs associated with each departments"""
```
jkachel, line 17: nit: this isn't a CourseRun serializer

jkachel (approved):
I have a nit but otherwise this looks good to me. 👍

courses/serializers/v2/departments.py
JenniWhitman: copy pasta. oops. I'll fix before I merge.

PR #1999, opened 2023-11-17
Reoder product info for mobile view

annagav:
Fix https://github.com/mitodl/hq/issues/2567

Reoder product info for mobile view
Screen Shot 2023-11-17 at 8 46 26 AM
Screen Shot 2023-11-17 at 8 46 45 AM

# How can this be tested?
Go to product page make sure that Desktop and tablet view look the same as before. Go to mobile view and make sure that the Product Info Box goes right after product description.

### DIFF
--- a/cms/templates/product_page.html
+++ b/cms/templates/product_page.html
@@ -14,8 +14,9 @@
     <div id="main" class="product-page">
       <div class="container">
         <div class="row d-flex flex-row align-center">
-          <div class="col flex-grow-1" id="product-details">
-            <div class="text">
+        <div class="col d-block d-md-none">
+        {# Mobile only #}
+          <div class="text">
             <nav aria-label="breadcrumb">
               <ol class="breadcrumb">
                 <li class="breadcrumb-item">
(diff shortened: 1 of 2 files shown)


cms/templates/product_page.html
```diff
@@ -33,10 +34,38 @@
...
               <h1>{{ page.title }}</h1>
               {# Description field contents are already rendered wrapped in a <p> tag #}
-              
+
               <section class="course-description">
                 {{ page.description | richtext }}
               </section>
+        </div>
+        </div>
+          <div class="col order-md-1 order-2 flex-grow-1" id="product-details">
```
jkachel, line 43: This seems to throw the info box to the left on my system (Firefox 119 on macOS Sonoma in the screenshot but also in Chrome 119):

image

(FWIW, I was able to fix this in the inspector by adding `order-sm-1 order-md-2` to the infobox column but that broke in the mobile view.)

jkachel (changes requested):
Looks good in iOS Simulator but the desktop version seems to do something weird. (I did explicitly restart the watch container to make sure it completely rebuilt the frontend.)

jkachel (approved):
LGTM :+1:

PR #2004, opened 2023-11-21
Accessibility: Make label visible to screen reader

annagav:
Fix https://github.com/mitodl/hq/issues/2486
Make Email input label visible to screen reader

Screen Shot 2023-11-21 at 11 03 21 AM

# How can this be tested?
Use voiceover to to read the Signup section at the bottom of the home page. It should read the "Email*" label.

### DIFF
--- a/frontend/public/scss/home-page/home-page-contact-row.scss
+++ b/frontend/public/scss/home-page/home-page-contact-row.scss
@@ -121,6 +121,10 @@ body.new-design {
               width:1px;
               height:1px;
               overflow:hidden;
+
+              .hs-form-required {
+                display: none;
+              }
             }
             .input {
               .hs-input {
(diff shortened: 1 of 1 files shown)


frontend/public/scss/home-page/home-page-contact-row.scss
```diff
@@ -117,7 +117,10 @@ body.new-design {
               max-width: 100%;
             }
             label {
```
jkachel, line 119: Can we hide elements with class `hs-form-required` from the screen reader? This works but it's a little awkward sounding since it reads the label and the required `*` as "email star".

jkachel (changes requested):
One small change, otherwise looks good to me.

jkachel (approved):
LGTM 👍

PR #2005, opened 2023-11-27
Catalog Page: updating header and list elements

annagav:
Fix https://github.com/mitodl/hq/issues/2491

1. Update main header to h1
2. The items count should be h2
3. departments list should be wrapped in nav element
4. the courses grid should be a list

# How can this be tested?
Everything should look the same as before. Use voice over to navigate the page by heading, navigate the page by landmark.


JenniWhitman (approved):
Looks good on multiple sizes. As far as I can tell, things look in the right order.

PR #2006, opened 2023-11-27
Announce course count change

annagav:
Fix https://github.com/mitodl/hq/issues/2494

Adds `role="status" aria-atomic="true"` to the course count div.

# How can this be tested?
Use voiceover to navigate the page click the department filters, when the course count updates it should announce the update.


jkachel (approved):
This looks OK and tests OK in Safari and Chrome. (I had some issues with it in Firefox but not sure if that's because of bugs in Firefox or not.)

PR #2007, opened 2023-11-27
Updates CSS to add hover states on the program course cards

jkachel:

Closes mitodl/hq#2986

The program course cards should default to a "dimmed" mode, and then undim on hover. This adds the CSS to support that. See the [relevant Figma link](https://www.figma.com/proto/Gs4zIhOFv5gVvafawwl6PF/MITx-Online?page-id=0%3A1&type=design&node-id=1373-5892&viewport=-17610%2C515%2C0.97&t=8L3q92hSlGEYTV7H-1&scaling=min-zoom&mode=design). 

https://github.com/mitodl/mitxonline/assets/945611/96cf69ae-6bb1-49ed-957f-25481cb3b098

# How can this be tested?

Load the program page for a program with courses. The course cards should be dimmed, and hovering over them should bring the opacity of the image and the text to 1.

# Additional Context

We were using the red button background (ala the Enroll Now button) for the course title text - in Figma it isn't like this.

### DIFF
--- a/frontend/public/scss/product-page/program-courses.scss
+++ b/frontend/public/scss/product-page/program-courses.scss
@@ -87,8 +87,9 @@ body.new-design {
                     .start-date {
                         font-weight: 400;
                         font-size: 12px;
-                        color: #6f7175;
-                        margin-bottom: 0.5rem
+                        color: rgba(112, 114, 117, 0.5);
+                        margin-bottom: 0.5rem;
+                        transition: $product-page-hover-time;
                     }
 
                     h3 {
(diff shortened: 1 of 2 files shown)


annagav (changes requested):
Looks Good! Just a couple of suggestions:
1. Could we include the "start time" text in the hover also? For now it stays the same.
2. Could we make the  transition for the opacity just a bit slower?

jkachel:
> Looks Good! Just a couple of suggestions:
> 
>     1. Could we include the "start time" text in the hover also? For now it stays the same.
> 
>     2. Could we make the  transition for the opacity just a bit slower?

Got these done in https://github.com/mitodl/mitxonline/pull/2007/commits/cab46a04421400a70dc3eb52fd6da59a236ac823

annagav (approved):
LGTM 👍

PR #2008, opened 2023-11-28
Fix just the counts from Catalog

JenniWhitman:
Fixes https://github.com/mitodl/hq/issues/2823

I pulled back on the loading changes to get a PR out to JUST the count changes. This means an accurate count will always show for all courses, programs, or a subsection of either, even if they're not yet loaded.

# How can this be tested?
Have data. Lots of it. Load the page and check that the counts for programs and courses are accurate and greater than 12 where they should be. 

There will be a follow-on to make sure the loading is clearer, but rather than refactor how we were infinitely loading, this seemed to be a simpler and shorter term solution to me refactoring the whole loader... which I may have about half done.


annagav (approved):
LGTM 👍

PR #2010, opened 2023-11-29
Accessibility: make CourseInfoBox discoverable

annagav:
Fix https://github.com/mitodl/hq/issues/2608

Make changes to the course info box to make it more discoverable with screen reader.

# How can this be tested?
Visually nothing should change.
Use voiceover to navigate through the headers, the course info box should be read.


jkachel (approved):
I think this looks OK - I could navigate through the infobox with the rotor and that worked pretty well. 👍

PR #2015, opened 2023-12-04
Make department sidebar sticky on mobile

annagav:
Fix https://github.com/mitodl/hq/issues/2531

Make department sidebar stay on the page when scrolling down on mobile

# How can this be tested?
Go to Catalog on mobile, scroll down to the bottom of the page, click the department filter button, a nav menu will open up. You should be able to see the list of the departments.

### DIFF
--- a/frontend/public/scss/layout.scss
+++ b/frontend/public/scss/layout.scss
@@ -47,11 +47,12 @@ body {
   display: flex;
   min-height: 100vh;
   flex-direction: column;
+  position: relative;
 
   .main-panel {
     flex: 1;
     min-height: calc(100vh - 351px);
-    position: relative;
+    position: static;
 
     // TODO: temporary fix for home page iamge size, please remove it after home page design
     p {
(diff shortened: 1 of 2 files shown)


jkachel (changes requested):
Functionality is good but I think this should cover the remaining part of the viewport - the footer is visible for me if I have it set to a phone-sized screen (iPhone 11):

image

The [design in Figma ](https://www.figma.com/file/Gs4zIhOFv5gVvafawwl6PF/MITx-Online?type=design&node-id=1592-7157&mode=dev) has this covering the footer. Otherwise looks/works well.

mbilalmughal:
@annagav Here's what you need to do: The 'position: relative' is applied to the main-panel, which does not include the page footer. That's why the absolute element is not covering the entire page.

Here is the proposed solution (See the first attachment).

 
remove-position-add-static


You have to add 'position: relative' to the body so that the absolute overlay container covers the entire page. (See 2nd attachment)


add-body-position

Result: The overlay will cover the entire body now.

annagav:
@jkachel The overlay is now covering the whole page.

PR #2017, opened 2023-12-04
Fixes course run selection code and standardizes it

jkachel:

Closes mitodl/hq#3083

The issue in 3083 is that the course run selection was selecting two different courses - the Enroll button code used one method and the infobox uses a separate one, which ended up selecting a non-archived course where it should have grabbed the archived one. This reworks that code to ensure it selects the most recent course run if they're all expired (and thus fixes the issue) and moves the selection into a utility function so it's only in one place. 

In addition, I [fixed an issue](https://github.com/mitodl/mitxonline/pull/2013) this (12/4) morning and forgot to add `key`s to the `li` tags - this fixes that too, for both course cards and program cards (neither had that).

# How can this be tested?

Set up two courses:
- One entirely in the past, with an enrollment end date and an end date in the past.
- One in the past, but with open-ended enrollment (i.e. end date in the past but no enrollment end date)

You should see the "archived course" message in the Info Box for the course. 

All other course run date scenarios should work as before - if you have one that's currently open or in the future, it should select it appropriately, etc.

# Additional Context

I found some weirdness with products and unupgradeable courseruns - that has been captured here: https://github.com/mitodl/hq/issues/3109

### DIFF
--- a/frontend/public/src/components/CourseInfoBox.js
+++ b/frontend/public/src/components/CourseInfoBox.js
@@ -6,6 +6,7 @@ import {
   formatLocalePrice,
   getStartDateText
 } from "../lib/util"
+import { getFirstRelevantRun } from "../lib/courseApi"
 import moment from "moment-timezone"
 
 import type { BaseCourseRun } from "../flow/courseTypes"
@@ -52,10 +53,20 @@ export default class CourseInfoBox extends React.PureComponent<CourseInfoBoxProp
 
     const course = courses[0]
 
+    const testRun = getFirstRelevantRun(courses, courseRuns)
+    console.log(
+      "qq infobox btw new thingy says this is the relevant run",
+      testRun
+    )
+
+    console.log("qq infobox thinks it has these courseruns", courseRuns)
+
     const run = course.next_run_id
       ? course.courseruns.find(elem => elem.id === course.next_run_id)
       : course.courseruns[0]
     const product = run && run.products.length > 0 && run.products[0]
+
+    console.log("qq the infobox thinks this one is the right courserun", run)
 
     const isArchived =
       moment().isAfter(run.end_date) &&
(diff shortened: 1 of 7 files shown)


annagav (commented):
I have couple of questions.

1. Do we want/need the Enroll Now button and the CourseInfoBox to use different methods for identifying the most relevant run? The Enroll button is using `getFirstUnexpiredRun`.
2. Would it be better to compute the most relevant run in the backend?

jkachel:
> I have couple of questions.
> 
>     1. Do we want/need the Enroll Now button and the CourseInfoBox to use different methods for identifying the most relevant run? The Enroll button is using `getFirstUnexpiredRun`.

Missed this, should be fixed in https://github.com/mitodl/mitxonline/pull/2017/commits/b782f7c2661074a48cec2b020f0fc254417eafee

>     2. Would it be better to compute the most relevant run in the backend?

I don't see a whole lot of benefit to that - we would still need to deliver the remaining course runs so that the learner can potentially select between them. (And I'm generally pretty hesitant to want to add to these APIs since they're a performance choke point.)

annagav (commented):
I am enrolled in two currently open for enrollment runs. I am enrolled the one that starts a bit later than the other. `getFirstRelevantRun` returns the run that I am not enrolled in, so it show the "enroll now" button.

Should we consider enrollment data when choosing the most relevant run?

jkachel:
> I am enrolled in two currently open for enrollment runs. I am enrolled the one that starts a bit later than the other. `getFirstRelevantRun` returns the run that I am not enrolled in, so it show the "enroll now" button.
> 
> Should we consider enrollment data when choosing the most relevant run?

I tested this in a few ways:
- Two course runs that overlap, no products in either, enrollment in the later one = Enroll Now button
- Two course runs that overlap, no products in either, enrollment in the first one = Enrolled button
- Two course runs that overlap, no products in either, enrollment in both = Enrolled button

This was consistent on main and this branch so it's working as expected, I think.. I'm not sure whether or not we _should_ take this into consideration, though, since there's some opportunity to upsell too.

PR #2019, opened 2023-12-05
Fixes the keyboard trap issue with instructor modals

jkachel:

Closes mitodl/hq#3077

- Moves the instructor modal code into the ProductDetailEnrollApp (eliminating inline JS calls)
- Adds taborder to the instructor image so it can be focused upon
- When the modal is popped, set focus to the close button
- When the modal closes, set focus to the instructor's featured image

This sets the focus to the featured image because otherwise there ends up being a stray keyup that re-toggles the modal to open again if focus is set to the h3.

# How can this be tested?

Navigate to a course page that has instructors on it. You should be able to tab down to the instructor name, hit Enter, and pop the modal. The focus should be on the Close button so hitting Enter again should close, or you can tab around. The modal should close but not reopen.

Similarly, navigation should work as expected with a screen reader turned on.

### DIFF
--- a/cms/templates/product_page.html
+++ b/cms/templates/product_page.html
@@ -115,11 +115,11 @@
                   <ul class="faculty-members card-listing">
                     {% if instructors %}
                     {% for member in instructors %}
-                      <li>
-                        <div class="member-card highlight-card" onClick="$('#instructor-modal-{{ member.id }}').modal('show');">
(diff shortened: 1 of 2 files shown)


annagav (approved):
I was able to tab through the instructor card, open the instructor modal and close it, and instructor img was selected. 👍

PR #2023, opened 2023-12-07
Filter out courseruns that can't be upgraded when displaying the upsell dialog

jkachel:

Partial fix for mitodl/hq#3109

If there are two active course runs, both with products but one of which is ineligible for upgrade, the upsell dialog will allow you to upgrade that particular course run where you shouldn't be able to. This adds some filtering to keep that from happening. 

# How can this be tested?

Set up two course runs that overlap. Both should have products and allow for upgrading, but the one that starts the latest should have an upgrade deadline in the past. (So, if you have ones that has both start dates as 2023-11-01 and one that has both start dates as 2023-11-15, the 2023-11-15 one should be the one with the upgrade deadline set in the past.) 

- On main, you should be able to upgrade, and you should be able to select the second course run and upgrade it.
- In this branch, you should be able to upgrade but you won't see the selector at all. 

If you add a _third_ course run that overlaps and is eligible for upgrade, you should see two course runs but not the ineligible one.

### DIFF
--- a/frontend/public/src/components/CourseProductDetailEnroll.js
+++ b/frontend/public/src/components/CourseProductDetailEnroll.js
@@ -240,16 +240,18 @@ export class CourseProductDetailEnroll extends React.Component<
           className="form-control"
         >
           {courseRuns &&
-            courseRuns.map((elem: EnrollmentFlaggedCourseRun) => (
-              <option
-                selected={run.id === elem.id}
-                value={elem.id}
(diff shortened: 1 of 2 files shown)


jkachel:
> @jkachel flow is complaining.

fixed in https://github.com/mitodl/mitxonline/pull/2023/commits/492c7b0656c10b68096a059ef374376120330501

collinpreston:
Thanks for reviewing this @annagav

PR #2027, opened 2023-12-15
Prevent department selection when loading courses

annagav:
Fix https://github.com/mitodl/hq/issues/2909

Prevent department selection when loading courses

# How can this be tested?
Make sure you can't select a tab while courses are loading.

### DIFF
--- a/frontend/public/src/containers/pages/CatalogPage.js
+++ b/frontend/public/src/containers/pages/CatalogPage.js
@@ -287,6 +287,9 @@ export class CatalogPage extends React.Component<Props> {
    * @param {string} selectedDepartment The department name to set selectedDepartment to and filter courses by.
    */
   changeSelectedDepartment = (selectedDepartment: string) => {
+    if (this.props.coursesIsLoading) {
+      return
+    }
     this.setState({ selectedDepartment: selectedDepartment })
     this.setState({
       filteredCourses: this.filteredCoursesBasedOnCourseRunCriteria(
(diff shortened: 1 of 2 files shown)


frontend/public/src/containers/pages/CatalogPage.js
```diff
@@ -287,6 +287,9 @@ export class CatalogPage extends React.Component<Props> {
    * @param {string} selectedDepartment The department name to set selectedDepartment to and filter courses by.
    */
   changeSelectedDepartment = (selectedDepartment: string) => {
+    if (this.props.coursesIsLoading) {
```
JenniWhitman, line 290: Because this value is set to true once the FIRST set loads, this doesn't have any effect on further queries (since it returns as done because of how we're running this query). the next queries use isLoadingMoreItems and even that resets with each load. What you COULD do here, and I've thought of doing before, but it seemed messy when we can just use useRequests from a future version, is to check that the length of the filtered departments matches. The problem here, though, is that the rest will not load unless the user scrolls.

JenniWhitman (changes requested):
Doesn't look like this has any effect.

frontend/public/src/containers/pages/CatalogPage.js
annagav: Could you write a scenario for when this bug occurs in the issue? I thought in the issue you were describing the bug happening on initial load.

annagav:
It looks like whatever Jenni is working on is going to fix the issue.

PR #2030, opened 2023-12-19
Update react-redux and related packages to facilitate catalog changes

JenniWhitman:
Supports https://github.com/mitodl/hq/issues/2992

Upgrades:
react-redux: 5.1.2->7.1.0
redux: 3.7.2->4.0.0
redux-query: 2.3.1->3.3.1 at which point it split some things out into also:
redux-query-interface-superagent (literally the same, just in a new spot in their v3)
redux-query-react (same as the superagent)
Added some test dependencies top-level

Keeps shallow rendering for many tests - thank you to Chris for the help!
Moved a few over to use full mount(), but because it's using a different renderer, it's not behaving on requests. I'm going to attempt to mock out the responses one more time, but I'm posting this to see if anyone else has another/ a better idea.

I've added a few comments as you go along to help explain a little of this - but there's a lot.

# How can this be tested?
First - run the tests and everything - be sure they pass.
Check the tests that were changed to be sure the changes make sense.
Lastly, Check some of the processes/workflows:
- registration especially - this is impacted by the changes and pretty important
- catalog bits
- anything that pops up a notification
- look for errors, weird behavior, etc.


frontend/public/.eslintrc
```diff
@@ -8,7 +8,8 @@
   "rules": {
     "no-unused-vars": [
       "error", {
-        "argsIgnorePattern": "^__"
+        "argsIgnorePattern": "^__",
+        "ignoreRestSiblings": true
```
JenniWhitman, line 12: Update from Chris C to help make flow happy after react-redux things changed

frontend/public/src/components/CourseProductDetailEnroll_test.js
```diff
@@ -266,18 +270,171 @@ describe("CourseProductDetailEnroll", () => {
...
   it(`shows form based enrollment button when upgrade deadline has passed but course is within enrollment period`, async () => {
     isWithinEnrollmentPeriodStub.returns(true)
     courseRun.is_upgradable = false
-    const { inner } = await renderPage()
+    course.next_run_id = courseRun.id
 
-    sinon.assert.calledWith(
-      helper.handleRequestStub,
-      "/api/course_runs/?relevant_to=",
-      "GET"
```
JenniWhitman, line 278: A bunch of these asserts are removed as this is only checking that the connect Request is working - this is making sure Redux knows to ask for this data which the shallow renderer can not see. I tried mocking this out with Mount, but it's so many layers down due to changes in the Redux-Query library and lack of changes to Enzyme. The long term solution here is to swap to Jest which uses full react rendering. Enzyme does not and as a result, you have to jump through hoops to get a full integration test.  here, I felt it was more than adequate to test the component using what is in the store.

frontend/public/src/containers/HeaderApp_test.js
```diff
@@ -18,16 +18,46 @@ describe("Top-level HeaderApp", () => {
...
 
   afterEach(() => {
     helper.cleanup()
   })
 
-  it("fetches user data on load and initially renders an empty element", async () => {
+  /* Because mount will automatically call componentDidMount, I'm simulating a null response from the server
+   * to test the loading state.  Until the component gets a positive response, it will continue to not render the header.
+   * This includes it receiving no parseable response.  I do have a new test after to show what a positive response does.
+   */
```
JenniWhitman, line 36: This, like hte above comment, is largely because Enzyme is not using React as the renderer, but another third party renderer which has slightly different behavior

frontend/public/src/components/NotificationContainer_test.js
```diff
@@ -26,7 +25,7 @@ describe("NotificationContainer component", () => {
     getAlertPropsStub = helper.sandbox
       .stub(notificationsApi, "getNotificationAlertProps")
       .returns({})
-    render = helper.configureHOCRenderer(
+    render = helper.configureMountRenderer(
       NotificationContainer,
       InnerNotificationContainer,
       {
```
jkachel, line 31: Not here but on line 71 there's an empty comment that should be pruned.

frontend/public/src/components/NotificationContainer_test.js
```diff
@@ -127,10 +126,8 @@ describe("NotificationContainer component", () => {
 
     await timeoutPromise
     wrapper.update()
-    assert.deepEqual(
-      wrapper.prop("userNotifications"),
-      omit(["message1"], messages)
-    )
+    // Due to changes in rendering due to enzyme, this now returns as undefined. Future test updates should
```
jkachel, line 129: think missing the rest of this sentence

jkachel (changes requested):
I have some small things that need a bit of attention. Outside that,
- the tests passed ok
- the changes made to the tests look OK, except where they don't (and those have comments)
- functionality looked pretty OK to me when I tested it on a clean setup
- flow hasn't complained

So, this looks pretty good to me.

frontend/public/src/containers/pages/CatalogPageOld.js
```diff
@@ -0,0 +1,716 @@
+import React from "react"
```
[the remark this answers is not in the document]
JenniWhitman: Good catch - no, it's a remnant from the other branch. I'll look through again

frontend/public/src/components/CourseProductDetailEnroll_test.js
```diff
@@ -664,55 +776,4 @@ describe("CourseProductDetailEnroll", () => {
 
     assert.isTrue(enrolledItem.exists())
   })
-
-  it("CourseInfoBox renders the archived message if the course is archived", async () => {
```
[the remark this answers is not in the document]
JenniWhitman: It didn't, it just moved - there was a bunch of shuffling in this file because of the different renderers, figuring out how much I could convince enzyme to do. https://github.com/mitodl/mitxonline/pull/2030/files/9e4f6eccc567f15e6caa19c31837857555f462b0#diff-04610edacb3613ea80cd276232f83b4b3e771a18e7a25c3d2156e3e4d90ae9f8R363

frontend/public/src/components/NotificationContainer_test.js
JenniWhitman: Good lord - I'll update that with the other comment. Train of thought derailed mid-comment

PR #2044, opened 2024-01-03
Check to make sure the end date and upgrade deadline haven't passed before grabbing your audit enrollment

jkachel:

Closes mitodl/hq#3295

If you have an existing enrollment in a course that is upgradeable, the enroll now button component will try to funnel you into upgrading that enrollment. However, it did this funneling even if the run's deadline had passed. This fixes that by adding some checks to make sure the course run is actually eligible for upgrade. 

# How can this be tested?

Set up two course runs (and related course, if necessary). One should be fully in the past and one should be eligible for enrollment and upgrade now. The course runs must have products and the course must have a CMS page. Create an audit enrollment in the past course run, then load the course page. You should see the normal messaging - current upgrade deadline and price, Enroll Now button - and clicking Enroll Now should prompt you to upgrade into the new course.

### DIFF
--- a/frontend/public/src/components/CourseProductDetailEnroll.js
+++ b/frontend/public/src/components/CourseProductDetailEnroll.js
@@ -145,13 +145,18 @@ export class CourseProductDetailEnroll extends React.Component<
     // Find an existing enrollment - the default should be the audit enrollment
     // already have, so you can just upgrade in place. If you don't, you get the
     // current run (which should be the first available one).
+    // This was changed to also make sure the run you're enrolled in is upgradeable.
     const { enrollments } = this.props
 
     if (enrollments) {
       const firstAuditEnrollment = enrollments.find(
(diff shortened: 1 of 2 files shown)


pdpinch:
What happens if you have an existing audit enrollment but you want to re-enroll in an upcoming course run. Is there a way to do that? (sorry if this is off-topic)

jkachel:
I think I noted this in https://github.com/mitodl/hq/issues/3109 .. if there are multiple course runs and at least some of them don't have products you're pretty well stuck with the one the app chooses for you. (This is a side effect of pulling the selector from the info box and moving it into the upsell dialog.) If all the course runs in question here have products, though, you should get the dropdown in the upsell dialog and you should be able to enroll in the other course from there.

JenniWhitman (approved):
Works. Barring Peter's question being a blocker it passed all scenarios I tried to break it in :)

PR #2046, opened 2024-01-04
When resolving the current run, make sure it's also enrollable

jkachel:

mitodl/hq#3295

If the app has to resolve which course run it should be working with, sometimes it will fail back to just the first one in the list that comes back from the API. This can result in a course run that's not enrollable being chosen. This PR adds a check to favor the first enrollable course run over just the first one in the list, and then fall back to the first one in the list if there's not one that's eligible. 

# How can this be tested?

I tested by copying the courserun info for `course-v1:MITxT+14.100x+3T2023` and `1T2024` from production, making sure both had products, and then forcing an audit enrollment in 3T2023. Once done, you should be able to load the course page for 14.100x, and clicking Enroll Now should offer to upgrade you to 1T2024. Clicking upgrade _or_ hitting Enroll for Free should enroll you in the correct (1T2024) run.

### DIFF
--- a/frontend/public/src/components/CourseProductDetailEnroll.js
+++ b/frontend/public/src/components/CourseProductDetailEnroll.js
@@ -86,7 +86,7 @@ export class CourseProductDetailEnroll extends React.Component<
   resolveFirstEnrollableRun() {
     const { courseRuns } = this.props
 
-    const enrollableRun = courseRuns
+    const enrollableRun = courseRuns && courseRuns
       .sort((a: EnrollmentFlaggedCourseRun, b: EnrollmentFlaggedCourseRun) => {
         if (moment(a.enrollment_start).isBefore(moment(b.enrollment_start))) {
           return -1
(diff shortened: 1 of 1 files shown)


pdpinch:
I thought we had a design for when there is no enrollable run. @mbilalmughal do you know?

jkachel:
If there's really nothing for you to enroll in, you just don't get a button at all.

JenniWhitman (approved):
After setting up a similar test, I was enrolled in the more current course run despite my old enrollment.

collinpreston (commented):
I think `getFirstUnenrolledCourseRun` can be removed from frontend/public/src/components/CourseProductDetailEnroll.js since it's not used anywhere.

jkachel:
saw @collinpreston 's note as I was merging.. I agree with that but in the interest of getting this out sooner I went ahead and merged. I'll throw a separate PR up with that change later.

PR #2048, opened 2024-01-04
display course number on program enroll modal

rachellougee:

https://github.com/mitodl/hq/issues/2623#issuecomment-1875972496

display the course number instead of course ID on program enroll modal

image

# How can this be tested?

Go to program enrollment page (It requires setting up courses, programs locally)
Click on "Enroll now"
A upgrade enrollment dialog should popup, it should list all the courses in that program.
Note that the display name should be course title + course number
The screen reader should read the course title and course number as well


JenniWhitman (approved):
ezpz looks good.

PR #2050, opened 2024-01-05
add next parameter to course and program enrollment pages for unauthenticated users

rachellougee:

https://github.com/mitodl/hq/issues/714

Adding `?next=` to course and program enrollment pages to remember where user was going after login 

# How can this be tested?

As anonymous users, go to course and program page
Click on enroll now
You are redirected to login page with `?next=` on url
Log in with your credential
You should be taken to the course or program page where you were before login

# Additional Context


jkachel (approved):
LGTM 👍

PR #2051, opened 2024-01-08
Fix flow errors in redux-query updates

ChristopherChudzicki:

- [ ] Desktop screenshots

### DIFF
--- a/frontend/public/.eslintrc
+++ b/frontend/public/.eslintrc
@@ -8,7 +8,8 @@
   "rules": {
     "no-unused-vars": [
       "error", {
-        "argsIgnorePattern": "^__"
+        "argsIgnorePattern": "^__",
+        "ignoreRestSiblings": true
         }
       ],
     // module importing pattern have huge impact over performance, especially when it comes to lodash
--- a/frontend/public/src/containers/pages/profile/EditProfilePage.js
+++ b/frontend/public/src/containers/pages/profile/EditProfilePage.js
@@ -13,7 +13,7 @@ import users, { currentUserSelector } from "../../../lib/queries/users"
 import queries from "../../../lib/queries"
 import EditProfileForm from "../../../components/forms/EditProfileForm"
 
-import type { Response } from "redux-query"
+import type { HttpResponse } from "../../../flow/httpTypes"
 import type { Country, User } from "../../../flow/authTypes"
 import type { RouterHistory } from "react-router"
 
@@ -23,8 +23,8 @@ type StateProps = {|
 |}
 
 type DispatchProps = {|
-  editProfile: (userProfileData: User) => Promise<Response<User>>,
-  getCurrentUser: () => Promise<Response<User>>
+  editProfile: (userProfileData: User) => Promise<HttpResponse<User>>,
+  getCurrentUser: () => Promise<HttpResponse<User>>
 |}
 
 type ProfileProps = {|
--- a/frontend/public/src/containers/pages/register/RegisterConfirmPage.js
+++ b/frontend/public/src/containers/pages/register/RegisterConfirmPage.js
@@ -113,8 +113,12 @@ const mapStateToProps = createStructuredSelector({
   qsParams: qsSelector
 })
 
-const mapPropsToConfig = ({ qsParams }) =>
-  mutateAsync(queries.auth.registerConfirmEmailMutation(qsParams))
+const mapPropsToConfig = ({ qsParams }) => {
+  const { type, ...config } = mutateAsync(
+    queries.auth.registerConfirmEmailMutation(qsParams)
+  )
+  return config
+}
 
 const mapDispatchToProps = {
   addUserNotification
--- a/frontend/public/src/lib/queries/auth.js
+++ b/frontend/public/src/lib/queries/auth.js
@@ -19,7 +19,7 @@ export const updateEmailSelector = pathOr(null, ["entities", "updateEmail"])
 const nextState = nthArg(1)
 
 const DEFAULT_OPTIONS = {
-  transform: (auth: AuthResponse) => ({ auth }),
+  transform: (auth: ?AuthResponse) => ({ auth }),
   update:    {
     auth: nextState
   },
@@ -154,7 +154,7 @@ export default {
   confirmEmailMutation: (code: string) => ({
     queryKey:  "updateEmail",
     url:       `/api/change-emails/${code}/`,
-    transform: (json: updateEmailResponse) => ({
+    transform: (json: ?updateEmailResponse) => ({
       updateEmail: json
     }),
     update: {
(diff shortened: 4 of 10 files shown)


frontend/public/src/containers/pages/register/RegisterConfirmPage.js
```diff
@@ -113,8 +113,12 @@ const mapStateToProps = createStructuredSelector({
...
 })
 
-const mapPropsToConfig = ({ qsParams }) =>
-  mutateAsync(queries.auth.registerConfirmEmailMutation(qsParams))
+const mapPropsToConfig = ({ qsParams }) => {
+  const { type, ...config } = mutateAsync(
+    queries.auth.registerConfirmEmailMutation(qsParams)
+  )
+  return config
+}
```
ChristopherChudzicki, line 121: This was giving an error:
```
Cannot call connectRequest with mapPropsToConfig bound to mapPropsToConfigs
because property type is missing in QueryConfig [1] but exists in
MutateAsyncAction [2] in the return value.
```

Again, Flow did not complain on `main` because redux-query is untyped on main, so anything is allowed.

The error occurs because:
1. Now `redux-query` has Flow types...
2. `redux-query` expects `mapPropsToConfig` to return a QueryConfig, which [their type definition](https://github.com/amplitude/redux-query/blob/562599ee1c482098879ae30e52898862c4639bec/packages/redux-query/src/types.js#L25C13-L25C24) defines as an [exact type](https://flow.org/en/docs/types/objects/#exact-and-inexact-object-types), so **it is not allowed to have extra keys**. But the return value of `mutateAsync` does have an extra key, namely `"type"`.
    - Javascript doesn't care about the extra properties, which is why everything seems to work fine. 
    - We use Flow 0.95, in which exact types are denoted with pipes `type Exact = {| ... |}`. Exact types are now the default in Flow.
    - In contrast, Typescript does not have the concept of exact types. So in Typescript, a declaration like
     ```ts
    type Dog = { name: string } 
    const sayWoof = (dog: Dog) => console.log(`${dog.name): woof woof`)
    const leo = { name: "Leo", owner: "Chris" }
    sayWoof(leo) // TS won't complain about extra "owner" property, but Flow will. 
     ``` 
    is perfectly fine. (TS does actually demand no extra properties when assigning an object literal to a type, a pragmatic, extra-check they call [excess property checking](https://www.typescriptlang.org/docs/handbook/2/objects.html#excess-property-checks). 


IMO Flow's enthusiasm for exact types is paranoid and unhelpful. TS's behavior is more natural, IMO. For reference: https://github.com/microsoft/TypeScript/issues/12936#issuecomment-284590083 

### Resolution
I resolved the issue by bending to Flow's demand :shrug:. There's probably also a way to typecast it to QueryConfig, or suppress the error with a `$FlowFixMe`.

frontend/public/src/lib/queries/auth.js
```diff
@@ -19,7 +19,7 @@ export const updateEmailSelector = pathOr(null, ["entities", "updateEmail"])
 const nextState = nthArg(1)
 
 const DEFAULT_OPTIONS = {
-  transform: (auth: AuthResponse) => ({ auth }),
+  transform: (auth: ?AuthResponse) => ({ auth }),
```
ChristopherChudzicki, line 22: There was a Flow error:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/containers/pages/register/RegisterDetailsPage.js:184:5

Cannot call mutateAsync with auth.registerDetailsMutation(...) bound to the
first parameter because AuthResponse [1] is incompatible with null or
undefined [2] in the first argument of property transform.
```
This error is occurring because:
1. Now `redux-query` has flow types...
2. And their flow type [definition for `Transform` is](https://github.com/amplitude/redux-query/blob/562599ee1c482098879ae30e52898862c4639bec/packages/redux-query/src/types.js#L72C8-L72C94)
    ```js
    type Transform = (body: ?ResponseBody, text: ?ResponseText) => { [key: string]: any };
    ```
    which says `body` might be undefined or null.
3. ...whereas we were saying it's not allowed to be.

### Resolution
If `body` ever actually is undefined/null, I suspect we will have issues at runtime. Currently, the transformed object would be `{ auth: undefined }` in that situation. (Maybe that's fine actually. Might just be treated as unauthenticated?) 

Anyway, this doesn't change the runtime behavior, just makes the signature match what redux-query's flow definitions are expecting.

frontend/public/.eslintrc
```diff
@@ -8,7 +8,8 @@
   "rules": {
     "no-unused-vars": [
       "error", {
-        "argsIgnorePattern": "^__"
+        "argsIgnorePattern": "^__",
+        "ignoreRestSiblings": true
```
ChristopherChudzicki, line 12: This means that in `const { a, ...others } = thing`, `a` won't be treated as unused, even if it is never used again. This allows object spread `...` to be used for omitting properties from an object without triggering errors. https://eslint.org/docs/latest/rules/no-unused-vars#ignorerestsiblings

ChristopherChudzicki (commented):
I left comments explaining and resolving the Flow errors.

I will say, overall, I found these issues annoying not helpful, and in contrast, Typescript handles al these issues better. *Admittedly, our version of Flow is 5 years out of date. But I believe these are philosophical differences between TS and Flow, not things that have improved.*

Another option, rather than using the resolutions I've provided, would be to just add `redux-query` to the [`[untyped]`](https://flow.org/en/docs/config/untyped/) section of our Flowconfig, which would make `redux-query` be treated as untyped. 

untyped doesn't make me feel good but:
- That's the current status on `main`
- Many of these warnings seem annoying

So if Flow isn't being helpful in regards to this library, it worth considering, IMO.

frontend/public/src/containers/pages/profile/EditProfilePage.js
```diff
@@ -13,7 +13,7 @@ import users, { currentUserSelector } from "../../../lib/queries/users"
 import queries from "../../../lib/queries"
 import EditProfileForm from "../../../components/forms/EditProfileForm"
 
-import type { Response } from "redux-query"
+import type { HttpResponse } from "../../../flow/httpTypes"
```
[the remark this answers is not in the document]
JenniWhitman: After much futzing with my own installation, was able to confirm. (registration/login isn't something I usually care about since i just work in the front end app, but being able to tell that the registration works requires the connection to edx to actually... work.

frontend/public/src/containers/pages/register/RegisterConfirmPage.js
JenniWhitman: Basically all i got from this one was TS>Flow

JenniWhitman: (kidding - makes sense, although it doesn't make sense. If that... makes sense)

frontend/public/src/lib/queries/auth.js
JenniWhitman: Yes, to confirm - that's no logged-in user which is mutated to a not-logged in user, but auth stays undefined because they're not authed

JenniWhitman (approved):

JenniWhitman:
@ChristopherChudzicki THANK YOU. I'm going to merge this in and double check things before I fix those tests that are hanging (if I haven't already, TBH at this point IDK any more)

PR #2053, opened 2024-01-09
Add css in for visual focus on modal buttons

JenniWhitman:
Closes https://github.com/mitodl/hq/issues/2750

Adds visual focus to two elements mentioned above - the enroll now button and enroll free "button" (that looks like a link)

Screenshot_2024-01-09_10-03-03
Screenshot_2024-01-09_10-03-22

### How can this be tested?
Make sure your course has pricing set so it launches the modal. You should be able to tab through and it should highlight the various elements. These should be:

- The X (previously worked and should still)
- The Enroll Now button (did not have a border prior)
- The Financial aid link (previously worked and should still - you will have to add a form to make this work OR - as I did - make a small change to the logic in the component to display it regardless)
- The No thanks, I'll take it free "button"/link at the end (did not have a border prior)

### DIFF
--- a/frontend/public/scss/product-page/product-details.scss
+++ b/frontend/public/scss/product-page/product-details.scss
@@ -397,6 +397,10 @@ body.new-design {
       font-size: 16px;
       font-weight: 400;
     }
+    .enroll-now-free:focus-visible {
+      border-radius: 0;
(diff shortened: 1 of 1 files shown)


rachellougee (approved):
LGTM. The focus works on both course and program page.

PR #2057, opened 2024-01-12
create_courseware command no department defined error

collinpreston:
NA

When running the quickstart management command (main/management/commands/configure_instance.py) an error occurs when the quickstart command executes the create_courseware command (courses/management/commands/create_courseware.py).  The create_courseware command is executed without specifying a department argument.  The create_courseware throws an error when it checks the value of the local variable `add_depts` which, when no department argument is provided, is undefined.

This PR resolves the issue by checking whether `add_depts` is defined as a local variable.  `add_depts` is defined when the department argument is defined, otherwise the local variable is not defined.

The `add_depts` variable is checked when creating a program or course.

### How can this be tested?
Run the configure_instance management command using instructions here (docs/source/configuration/quickstart.rst).

### DIFF
--- a/courses/management/commands/create_courseware.py
+++ b/courses/management/commands/create_courseware.py
@@ -227,7 +227,7 @@ class Command(BaseCommand):
                 live=kwargs["live"],
             )
 
-            if add_depts:
+            if "add_depts" in locals():
                 new_program.departments.add(*add_depts)
                 new_program.save()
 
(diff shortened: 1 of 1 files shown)


jkachel (approved):
LGTM 👍  - I was able to run the configure_instance command and was also able to just create a course with create_courseware without listing a department successfully.

PR #2059, opened 2024-01-16
Delay alert when rendering the page in order to engage screen reader alert

collinpreston:

After paying for a course, a user is redirected to the user's Dashboard in MITx Online where they are presented with an alert indicating whether the payment process was successful or not.  For screen readers, the alert is rendered when the page is loaded which does not trigger the use of `role="alert` on the alert.  This PR adds a 3 second delay, after the page has loaded, before displaying the alert.  This delay means that the alert is rendered 3 seconds after the page has loaded and results in the alert being read by the screen reader in a fashion expected of `role="alert`.  

The functionality that`role="alert` provides to screen readers can also be observed when unenrolling from a course from the user dashboard.

### How can this be tested?

1. Enroll and pay for a course using the instructions here: docs/source/configuration/ecommerce.rst
2. After completing payment for the course and being redirected back to the MITx Online user Dashboard, verify that the alert is read by a screen reader as soon as it's rendered.

### DIFF
--- a/frontend/public/src/components/NotificationContainer.js
+++ b/frontend/public/src/components/NotificationContainer.js
@@ -11,7 +11,7 @@ import {
   firstNotNil,
   newSetWith,
   newSetWithout,
-  timeoutPromise,
+  timeoutPromise
 } from "../lib/util"
 import { getNotificationAlertProps } from "../lib/notificationsApi"
 import { notificationTypeMap, TextNotification } from "./notifications"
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki:
Commenting just to say... That's an interesting issue I hadn't heard of before. I found this:
- https://github.com/w3c/aria/issues/1360

3s seems like a long time. Seems like some people in that issue are using 500ms.

Fundamentally seems like a screenreader issue, but we certainly should mitigate =/

collinpreston:
> Commenting just to say... That's an interesting issue I hadn't heard of before. I found this:
> 
> * [Consider a new aria-live property that reads on page load w3c/aria#1360](https://github.com/w3c/aria/issues/1360)
> 
> 3s seems like a long time. Seems like some people in that issue are using 500ms.
> 
> Fundamentally seems like a [screenreader issue,](https://github.com/w3c/aria/issues/1360#issuecomment-739319614) but we certainly should mitigate =/

@ChristopherChudzicki Yeah I stumbled upon that Github issue while working on this.  I can update to 500ms.  I agree that 3s is probably too long, but it was difficult to gauge while running this on my local machine.

annagav:
@collinpreston is this issue happening only for after payment? My screen reader is reading other alerts okay. Just checking that I am testing the right thing.

collinpreston:
> @collinpreston is this issue happening only for after payment? My screen reader is reading other alerts okay. Just checking that I am testing the right thing.

@annagav Yes this only happens after the payment process redirects you back to MITx Online.  The reason it only happens then is because the alert is loaded with the page, so the screen reader is not notified of any alert on page load.  Screen readers are only notified of alerts when they appear - after the page has already loaded.

annagav (approved):
LGTM 👍

PR #2063, opened 2024-01-18
Show More link on product about page accessibility

annagav:
Fix https://github.com/mitodl/hq/issues/2596

Fixes two things:

- the hide class should be on the hidden element from the very begining
- the Show More link is now in between the visible text and the hidden text.

Screenshot 2024-01-18 at 1 54 00 PM

### How can this be tested?
Use a screen reader to tab through a product page that has hidden "about" section text. Make sure the the links in the hidden text are not read by the reader.
Play with the Show More button and make sure that the hidden text appears after the button.


rachellougee (approved):
👍 Tested with screen reader, it doesn't read the link inside hidden text anymore. Also, the show more link appears before the new content.

PR #2064, opened 2024-01-18
Removes instances where we've feature flagged the new (current) design

### LINKED ISSUE #1864: Remove New Design Feature Flag
Checklist of things to do when de-flagging the new design:

- [x] Remove calls to `is_enabled`
- [x] Specifically remove flagged block from `cms/templates/product_page.html`
- [x] Remove catalog view https://github.com/mitodl/mitxonline/blob/0b6e49d7aed2426951e3021864a6521863aa36d0/main/views.py#L43 and update url path to reference index view.

Tabled:
- [ ] Clean up flagged CSS - remove CSS classes that are for the old design
- [ ] Audit static images included in the repository.  I'm guessing that some of these are no longer used in the new design.

jkachel:

Closes #1864

This removes remnants of the old MITx Online visual design and eliminates checks for the feature flag that toggled between the two designs. It keeps the code that allows for checking for feature flags in the front end.

#1864 also specifies removing CSS and images from the old design but [some pages still use elements of the old design](https://github.com/mitodl/mitxonline/issues/1864#issuecomment-1899188983), so this isn't done here. 

### How can this be tested?

Test against RC or production. The look and feel should not change; all UI elements should be there; there should be no changes for accessibility. Turning the new design off by flipping the feature flags, or blocking Posthog in some method (e.g. via an ad blocker), should not revert you back to the old design.


collinpreston (changes requested):
I believe that the `TopAppBar` component can be removed.

collinpreston (approved):
Looks good to me

Issue #2, opened 2021-06-22
Account creation


rhysyngsun:
As a user I would like to create an account on mitx online.
- Bring over the best of our implementation from xpro , etc. 
- Consolidate `courseware` app into the `users` app:
> `courseware` as a name is a holdover from when we envisioned xPro supporting alternative backends other than `openedx` in the future, so it served as an abstraction layer between users and arbitrary courseware backends. It's highly unlikely we'd ever do it for `mitx-online`, so we should just remove the abstraction (and any confusion caused by the name)

#### Designs and Mockups

N/A - should be barebones / wireframe style.

#### Acceptance Criteria:

- [ ] Learner can create an account
- [ ] Learner can login

#### Out of Scope

- Export compliance checks
- Entering profile data

PR #2065, opened 2024-01-18
3400 remove enrolled state and modal default

rachellougee:

part of https://github.com/mitodl/hq/issues/3404 https://github.com/mitodl/hq/issues/3400

- Remove the "Enrolled" box, so it always display Enroll Now button.
- When there are multiple courseruns open for enrollment, the enrollment modal should default to the courserun with the earliest start date, even if it is in the past.

### How can this be tested?

- Create two runs that both open for enrollments. e.g.
image

- Create CMS course page and products for these two runs

- Go to course about page and click on Enroll now, it defaults to Jan 16 run 
image

-  Enroll and Pay for this run
- Attempt to enroll it again (Either Enroll for free or Enroll and Pay), you should get message indicated that you have already paid for
image

In case of refunded, it should be ok since this PR does't change the flow/logic for "Enroll and Pay"


collinpreston:
If there is a course with 2 course runs:
1. start date in 2023 and enrollment start date in 2023, end date and enrollment end date in 2024, instructor paced.
2. start date in 2022 and enrollment start date in 2022, no end date or enrollment end date defined, self paced.

Will the #2/self paced course run always be the default selected date?

rachellougee:
> Will the https://github.com/mitodl/mitxonline/issues/2/self paced course run always be the default selected date?

Yes, it defaults to the earliest start date #2 since these two runs are both open

collinpreston:
> > Will the #2 paced course run always be the default selected date?
> 
> Yes, it defaults to the earliest start date #2 since these two runs are both open

I could see that being an issue if a user intends on enrolling into an upcoming instructor led course since the user will need to recognize that the course run selected needs to be updated.  I brought this up during the call after standup this morning where that situation was agreed to be okay functionally.

jkachel (approved):
I think this looks ok - I was able to test OK with two courseruns that had some overlap, and at least one of them as self-paced.

rachellougee:
@jkachel Can you take another look? I've added some validation for "Enroll for free" button in case of user has already paid for, per https://github.com/mitodl/hq/issues/3404#issuecomment-1900854907. Thank you.

jkachel (approved):
I think this works OK - I tried re-enrolling for free in a course that I was already in and got the "you're already enrolled" message, rather than it downgrading my enrollment.

PR #2069, opened 2024-01-23
Updated fixed price test to be less flaky

jkachel:

Closes mitodl/hq#2979

After some examination, it was determined that the `test_discount_product_calculation` test was failing because, when testing fixed price discounts, the price specified in the generated discount was greater than the price in the generated product. Since we won't _raise_ a price to make the discount code work, the app correctly just returns back the original price in the `discount_product` call; this was then making the test fail, since that (maybe) wasn't always the case and the test wasn't updated for this. 

There wasn't an issue with the algorithm so this just adjusts the test to assert that the discounted amount is the unchanged product price if the discount amount is greater than the product price for fixed price discounts. This also adds in an explicit test for the case where the price is less than the discount amount. I've also added a couple comments to better describe what the `less_than_zero_or_discount` argument means. 

The test for dollars-off discounts and percent-off discounts was already OK: the dollars-off one already checked for a discount that would result in a less-than-zero purchase price, and the percent-off discount is a percentage of the purchase price and has a hard limit at 100%. 

### How can this be tested?

You may want to try adjusting the factories temporarily so that a fixed-price discount is always above the product price (maybe make the product price always be < 1 and discount amount be > 10?). 

To emulate what was going on with the runners, I just made it run the test until it failed: 
```shell
while [ 1 ]; do docker compose exec web pytest "ecommerce/models_test.py::test_discount_product_calculation[fixed-price-False]" --pdb ; done
```

This took a while to get to a failure state because the amounts are randomized. (But maybe you'll get lucky and it'll do it sooner rather than later.)

### DIFF
--- a/ecommerce/models_test.py
+++ b/ecommerce/models_test.py
@@ -371,22 +371,31 @@ def test_create_transaction_with_no_transaction_id():
 
 
 @pytest.mark.parametrize(
-    "discount_type, less_than_zero",
+    "discount_type, less_than_zero_or_discount",
     [
         [DISCOUNT_TYPE_DOLLARS_OFF, False],
         [DISCOUNT_TYPE_DOLLARS_OFF, True],
         [DISCOUNT_TYPE_FIXED_PRICE, False],
+        [DISCOUNT_TYPE_FIXED_PRICE, True],
         [DISCOUNT_TYPE_PERCENT_OFF, False],
     ],
 )
 def test_discount_product_calculation(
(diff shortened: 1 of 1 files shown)


JenniWhitman (approved):
Behaved locally for me. I was able to break it on first run without changing things, but I ran it a few times both with factories changed to make it err and without to see - haven't seen anymore wild behavior.  I like the additional assertion as well.

PR #2070, opened 2024-01-24
Update Publish Documentation to newer versions of stuff so it runs

jkachel:

n/a

For version 0.85.0, the production deploy got stuck because the Publish Documentation action failed. It looks like this was mostly due  to Pants being too old, so this bumps that to 2.17.1 and makes some minor changes to the `docs` plugin to make it work. 

### How can this be tested?

Running `./pants docs ::` should work as expected.

### Additional Context

The latest Pants is 2.18.2 but that didn't seem to work; [there's a warning here](https://github.com/mitodl/mitxonline/actions/runs/7641957862/job/20820461026#step:5:100) that is likely the issue. (We're not using SCIE pants here yet so moving to that is probably the right move but there's also [specifically an init-pants action](https://github.com/pantsbuild/actions/tree/main/init-pants) that we could also look at. Or, both, though bumping to 2.18.2 locally worked OK without also switching to SCIE pants.)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.11
       - name: Install ghp-import
         run: pip install ghp-import
       - name: Build documentation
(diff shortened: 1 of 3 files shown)


JenniWhitman (approved):
Ran locally. Seems to be fine. I'm not opposed to upgrading to a newer version, but this works for now.

PR #2073, opened 2024-01-25
Wrapping product creation in a create_version block so versions happen like they should

jkachel:

Closes mitodl/hq#3463

`create_product` wasn't creating the initial version of the product it created - it just made the row in the `ecommerce_product` table, so trying to buy the resulting product would fail. This fixes that.

### How can this be tested?

Run `create_product` - the new product it creates should have a version. (Easy way to check is just to grab the last row of the `reversion_version` table, or doing a `like` on `object_repr` in that table with the product name you created.) 

Checkout should work with the newly created product.

### DIFF
--- a/courses/views/v2/views_test.py
+++ b/courses/views/v2/views_test.py
@@ -1,24 +1,25 @@
 """
 Tests for courses api views v2
 """
+import logging
 import operator as op
 import random
 
 import pytest
+from django.db import connection
 from django.urls import reverse
 from rest_framework import status
-from django.db import connection
-import logging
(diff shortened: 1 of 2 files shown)


collinpreston (approved):
Looks good to me

jkachel:
Re-requesting review because this ended up also including some fixes for tests that were failing.

collinpreston (approved):
Looks good.  There are some unused imports in the views_test.py file.  Only remove them if you want, they are unrelated to your changes.

PR #2077, opened 2024-01-31
Fixes use of "redemption_type" when generating discount codes in bulk

jkachel:

n/a

I noticed when processing a bulk code creation form in the staff dashboard that codes that I'd requested be created as one-time discounts were being created as unlimited ones instead. As it turns out, the API call for this was not looking at the `POST`ed redemption type - it expected flags instead. This fixes the API to use _either_ the flags or the `redemption_type` kwarg (which is how it's sent via POST), so the redemption type is saved correctly. 

This also fixes the `test_get_courses` test - this had similar problems to `test_get_programs`, which was fixed in https://github.com/mitodl/mitxonline/pull/2073 , which was also rolled into that PR because the tests were failing. 

### How can this be tested?

Create a set of discount codes from the staff dashboard that are of a redemption type that's not unlimited. The type should be correctly set.

### Additional Context

I noticed this processing https://github.com/mitodl/hq/issues/3515 . Upon further examination, this is a fairly longstanding bug - I went ahead and fixed all the discounts in production that had the `in_bulk` flag set to be `one-time` redemption instead. This was around 1170 discount codes. (I did look to make sure the codes were all ones that should have been one-time discounts, and they were.) 

I know multiple redemptions has come up once but I don't know at this point if any codes have been redeemed more than once. Note that the redemption status is calculated by looking at actual redemptions so any unlimited discount that had any redemptions in the past will now not be redeemable.

### DIFF
--- a/courses/views/v2/views_test.py
+++ b/courses/views/v2/views_test.py
@@ -10,7 +10,7 @@ from django.urls import reverse
 from rest_framework import status
 
 from courses.factories import DepartmentFactory
-from courses.models import Program
+from courses.models import Course, Program
 from courses.serializers.v2.courses import CourseWithCourseRunsSerializer
 from courses.serializers.v2.departments import DepartmentWithCountSerializer
 from courses.serializers.v2.programs import ProgramSerializer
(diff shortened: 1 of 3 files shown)


JenniWhitman (approved):
Works as intended (once I got my dashboard fixed)

Issue #2075, opened 2024-01-30
Some dialogs ("Email Settings", "Unenroll") not keyboard navigable. 


ChristopherChudzicki:
### Expected Behavior
Dialog buttons should be actionable via keyboard.

### Current Behavior
Dialog buttons not keyboard navigable

### Steps to Reproduce
This issue seems to affect both the "Email Settings" and "Unenroll" dialogs on dashboard page.

1. Visit https://mitxonline.mit.edu/dashboard/
2. Click (or navigate via keyboard) to the `...` icon on one of the course cards.
3. From the dropdown menu, click "Unenroll" (or "Email Settings")
4. Navigate the dialog with keyboard controls. With any of the buttons focused, press "Enter". The button should activate but it does not.

### Possible Solution
This should be double checked, but:
- The button elements all seem to be real buttons (using a `<div role="button">` is one source of similar issues)
- The event handlers seem to be attached with `onClick`, not with something like `onPointerUp`, which would also cause this.

Button element click handlers **should** be firing on "click" events, so I'm confused. Perhaps there is some other keyboard-only event handler that is interfering with the events?

ChristopherChudzicki:
<img width="1546" alt="Screenshot 2024-01-29 at 8 04 31 PM" src="https://github.com/mitodl/mitxonline/assets/9010790/b021fe04-d16b-4fe7-a20d-9773694aec5e">

collinpreston:
Could be related to https://github.com/jaredpalmer/formik/issues/2468.

collinpreston:
@ChristopherChudzicki, I've created a branch with changes which do the following:
1. Pressing escape closes the modal (worked as before).
2. Pressing enter submits the modal, regardless of which button is focused via keyboard navigation.

I think the functionality of #2 above is correct based on what I read here: https://accessibility.asu.edu/articles/forms.  The accessibility requirements for a modal containing a form are difficult to narrow down.

PR #2079, opened 2024-02-01
2075: some dialogs email settings unenroll not keyboard navigable

collinpreston:
https://github.com/mitodl/mitxonline/issues/2075

Resolves in issue with using the keyboard to navigate and control the browser when the either the unenroll modal or email settings modal is open for a course enrollment.

### How can this be tested?
You should have a course created and be enrolled.
1. Visit http://mitxonline.odl.local:8013/dashboard/.
2. Using only your keyboard, open the unenroll modal or email settings modal from the course enrollment card.
3. Verify that you can move focus to one of the buttons on the modal.  When you press enter, the focused button should perform it's expected action (close the modal, or submit).

### DIFF
--- a/frontend/public/src/components/EnrolledItemCard.js
+++ b/frontend/public/src/components/EnrolledItemCard.js
@@ -255,8 +255,10 @@ export class EnrolledItemCard extends React.Component<
               enrollmentId:    enrollment.id,
               courseNumber:    enrollment.run.course_number
             }}
-            render={({ values }) => (
-              <Form className="text-center">
+            render={({ values, handleSubmit }) => (
+              <Form className="text-center"
+                onSubmit={handleSubmit}
+              >
                 <section>
                   <Field
                     type="hidden"
@@ -276,7 +278,6 @@ export class EnrolledItemCard extends React.Component<
                   />{" "}
                   <label
                     id={`verified-unenrollment-${values.enrollmentId}-email-checkbox`}
-                    check
                   >
                     Receive course emails
                   </label>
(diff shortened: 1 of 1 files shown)


frontend/public/src/components/EnrolledItemCard.js
```diff
@@ -543,10 +541,11 @@ export class EnrolledItemCard extends React.Component<
                     >
                       Email Settings
                     </DropdownItem>
-                    {this.renderEmailSettingsDialog(enrollment)}
                   </span>
                 </DropdownMenu>
               </Dropdown>
+              {this.renderRunUnenrollmentModal(enrollment)}
+              {this.renderEmailSettingsDialog(enrollment)}
```
ChristopherChudzicki, line 548: **Quick explanation:** Reactstrap Dropdowns attach event handlers to all children that call `preventDefault` on keydown events, see https://github.com/reactstrap/reactstrap/blob/6085ac0bb340f3e2cdae1aec4e6abec38c37ebad/src/Dropdown.js#L354. Normally, `onClick` handlers will fire on a keydown event bubbling from a button element. But if you call preventDefault on the keydown event, that behavior is prevented.

There's no reason, afaik, that the modal needs to be a child of the dropdown, so moving it outside the dropdown fixes the issue.

ChristopherChudzicki (approved):
👍 Working as expected now.

ChristopherChudzicki:
@collinpreston Actually, looking at the code, I suspect that `renderProgramUnenrollmentModal` has the same issue. Might be good to change that here, too.

PR #2081, opened 2024-02-02
Fix for program unenrollment modal

collinpreston:
Same issue is fixed with this PR as in https://github.com/mitodl/mitxonline/pull/2079.

Resolves keyboard accessibility issue with the program un-enroll modal.


ChristopherChudzicki (approved):
👍

PR #2103, opened 2024-02-14
Admin improvements to CourseRunEnrollmentAudit model

annagav:
Related to https://github.com/mitodl/hq/issues/3580

adds search fields to CourseRunEnrollmentsAuditAdmin model and inlines to `CourseRunEnrollmentAdmin`

Make sure you have an enrollment instance. View it in the admin.
Then modify the course enrollment. A new audit model should get created. Try searching for the audit by username or email.
View the enrollment in the admin and see the audit instances related to the enrollment.

### DIFF
--- a/courses/admin.py
+++ b/courses/admin.py
@@ -191,6 +191,13 @@ class ProgramEnrollmentAuditAdmin(TimestampedModelAdmin):
     def has_delete_permission(self, request, obj=None):
         return False
 
+class CourseRunEnrollmentAuditInline(admin.TabularInline):
+    """Inline editor for CourseRunEnrollmentAudit"""
+
+    model = CourseRunEnrollmentAudit
+    readonly_fields = ["enrollment_id", "created_on", ]
+    min_num = 0
+    extra = 0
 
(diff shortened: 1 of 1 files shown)


courses/admin.py
```diff
@@ -192,6 +192,18 @@ def has_delete_permission(self, request, obj=None):
...
+class CourseRunEnrollmentAuditInline(admin.TabularInline):
+    """Inline editor for CourseRunEnrollmentAudit"""
+
+    model = CourseRunEnrollmentAudit
+    readonly_fields = [
+        "enrollment_id",
+        "created_on",
+    ]
+    min_num = 0
+    extra = 0
```
rachellougee, line 204: Can we add also add code to prevent add or delete? This table is for audit, so I don't think it should be editable for admin unless there is a use case?

```
    can_delete = False
    can_add = False

    def has_add_permission(self, request, obj=None):
        return False
```

rachellougee (commented):
Just comments about disabling edit or delete feature.

rachellougee (approved):
👍

PR #2105, opened 2024-02-15
fix course card not keyboard accessible issue

rachellougee:

https://github.com/mitodl/hq/issues/3361

Fixing the keyboard accessible issue for the course card on program page. 

Course card is now focusable
image

### How can this be tested?

- You will need to setup courses and programs to test it
- Turn on voiceover and go to program page
- you should be able to tab to the course card and then use keyboard to open it

### DIFF
--- a/cms/templates/course_list_card.html
+++ b/cms/templates/course_list_card.html
@@ -1,20 +1,20 @@
 <ul>
     {% for card in cards %}
     <li>
+      <a href="{{card.course.page.url}}" class="program-course-card-link">
         <div class="program-course-card">
-            <a href="{{card.course.page.url}}" class="program-course-card-link">
-                <div class="image">
-                    <img src="{{card.featured_image}}" alt="">
-                    <div class="overlay"></div>
+            <div class="image">
+                <img src="{{card.featured_image}}" alt="">
+                <div class="overlay"></div>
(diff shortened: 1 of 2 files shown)


cms/templates/course_list_card.html
```diff
@@ -1,17 +1,19 @@
...
+    <li>
         <div class="program-course-card">
-            <div class="image">
-                <img src="{{card.featured_image}}" alt="">
-                <div class="overlay"></div>
-            </div>
-            <div class="program-course-card-info">
-                <div class="start-date">
-                {{card.start_descriptor}}
+            <a href="{{card.course.page.url}}" class="program-course-card-link">
```
annagav, line 5: I think it be better to put the link around `program-course-card` div.

annagav (changes requested):

cms/templates/course_list_card.html
annagav: and when move the <a element outside you need to add this to the styles:
```
program-course-card-link {
    display: block;
    text-decoration: none;
    width: fit-content;
}```

rachellougee:
@annagav Thank you for the feedback. Can you take another look?

annagav (approved):
👍

PR #2108, opened 2024-02-16
Create certs only for live programs

annagav:
None

Updates the signal to check whether the program is live before generating the cert.
### How can this be tested?
Take a look at the tests. 

Create a program with a required course. When you create a cert for that course, a program cert should get generated. 
If you set the program `live=False` and then create a course cert then no program cert should get generated.


rachellougee (approved):
Looks good to me. Manually tested that program certificate is not created for inactive program

PR #2114, opened 2024-02-26
Fix mobile counts on catalog

JenniWhitman:
Fixes https://github.com/mitodl/hq/issues/3664
blocks https://github.com/mitodl/hq/issues/3663

Splits the render number of items function into two - one for courses and one for programs, and uses those in the desktop and mobile versions alike. This will also be used to determine how many loading divs are needed in a follow-on PR

### How can this be tested?
Make sure the counts match on desktop and mobile and are correct for each item type and filter


jkachel (approved):
LGTM 👍

PR #2117, opened 2024-02-28
adding enrollment change tracking

annagav:
Fix https://github.com/mitodl/hq/issues/3580

Track enrollment change data by adding call stack info to a field "modified_by" to the audit model.

### How can this be tested?
Create an enrollment for user then try modifying it through running a task or calling some of the methods that modify an enrollment from django shell:
`deactivate_run_enrollment(enrollment, "unenrolled")`

Then go to the admin and view the "CourseEunEnrollmentAudit" it should look something like this:

Screenshot 2024-02-29 at 10 48 12 AM


courses/models.py
```diff
@@ -1117,6 +1119,34 @@ def update_mode_and_save(self, mode, no_user=False):
...
+        self.refresh_from_db()
+        before_dict = None
+        if before_obj is not None:
+            before_dict = before_obj.to_dict()
+
+        call_stack = "".join(traceback.format_stack()[-6:-2])
+
+        audit_kwargs = dict(
+            acting_user=acting_user,
+            modified_by=call_stack,
```
rachellougee, line 1142: Not sure if any side effect but why can't it be handled in https://github.com/mitodl/mitxonline/blob/main/main/models.py#L72 since we are adding `modified_by` to AuditModel base model anyway?

main/models.py
```diff
@@ -18,6 +19,7 @@ class AuditModel(TimestampedModel):
     """An abstract base class for audit models"""
 
     acting_user = ForeignKey(settings.AUTH_USER_MODEL, null=True, on_delete=CASCADE)
+    modified_by = CharField(default="", max_length=750, null=True, blank=True)
```
rachellougee, line 22: Can it be renamed to traceback or call_back? It's a little confusing when I first saw it. Not a big deal, just a thought.

rachellougee (commented):
The functionality works fine. Just comment and question below.

rachellougee (approved):
👍

PR #2118, opened 2024-03-01
3689: add department filter to the api endpoints for courses and programs

collinpreston:
https://github.com/mitodl/hq/issues/3689

### Description
This PR modifies the `/api/v2/departments/` endpoint.  The endpoint previously provided the number of courses and programs associated with each department.  This PR removes the number of courses and programs, and instead provides a list of course and program IDs associated with each department.

**Example response**
```
{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 1,
            "name": "dept1",
            "course_ids": [
                1,
                2
            ],
            "program_ids": []
        },
        {
            "id": 2,
            "name": "dept2",
            "course_ids": [
                1
            ],
            "program_ids": [
                1
            ]
        }
    ]
}
```

The intention of this change is to better support filtering by department on the catalog page.  When a user filters by department on the catalog page, future development will utilize the course and program IDs provided by the department API in this PR.  Using those course and program IDs, the front-end can make an API call to retrieve those course or programs.

### How can this be tested?
1. You should have a few courses, a program, and departments created.  The courses and program should be associated with one or more departments.
2. You can make a GET request to the departments API http://mitxonline.odl.local:8013/api/v2/departments/ and verify the response.
3. You can then verify that the course catalog continues to display as it currently does.

### DIFF
--- a/frontend/public/src/containers/pages/CatalogPage.js
+++ b/frontend/public/src/containers/pages/CatalogPage.js
@@ -313,7 +313,7 @@ export class CatalogPage extends React.Component<Props> {
   /**
    * Returns a filtered array of Course Runs which are live and:
    * - Have a start_date before the current date and time
-   * - Have an enrollment_start_date that is before the current date and time 
+   * - Have an enrollment_start_date that is before the current date and time
    * - Has an enrollment_end_date that is not defined or is after the current date and time.
    * @param {Array<BaseCourseRun>} courseRuns The array of Course Runs apply the filter to.
    */
--- a/frontend/public/src/containers/pages/CatalogPage.js
+++ b/frontend/public/src/containers/pages/CatalogPage.js
@@ -224,7 +224,7 @@ export class CatalogPage extends React.Component<Props> {
           ...new Set([
             ALL_DEPARTMENTS,
             ...departments.flatMap(department =>
-              department.courses > 0 ? department.name : []
+              department.course_ids.length > 0 ? department.name : []
             )
           ])
         ]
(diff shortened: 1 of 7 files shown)


frontend/public/src/containers/pages/CatalogPage_test.js
```diff
@@ -223,14 +223,14 @@ describe("CatalogPage", function() {
           },
           departments: [
             {
-              name:     "History",
-              courses:  0,
-              programs: 5
+              name:        "History",
+              course_ids:  [],
+              program_ids: [2, 2, 2, 2, 2]
```
JenniWhitman, line 228: nit: should the program be different for each? technically this is 1 program, not 5.

frontend/public/src/containers/pages/CatalogPage.js
```diff
@@ -311,8 +311,8 @@ export class CatalogPage extends React.Component<Props> {
   }
 
   /**
-   * Returns a filtered array of Course Runs which are live, define a start date,
-   * enrollment start date is before the current date and time, and
+   * Returns a filtered array of Course Runs which are live, define a start date
+   * and enrollment start date that is before the current date and time, and
```
JenniWhitman, line 315: This still isn't fully English, if we're going to clean it up, maybe bullets?
```
Returns a filtered array of Course Runs which are live and:
- Have a start_date before the current date and time
- Have an enrollment_start_date that is before the current date and time 
- Has an enrollment_end_date that is not defined or is after the current date and time.
```

JenniWhitman (approved):
Two small comments, neither stop the actual working, both are more for future clarity than anything else.  Approving, you can choose whether or not to update them

PR #2120, opened 2024-03-05
Fetch by department when selected

JenniWhitman:
Closes https://github.com/mitodl/hq/issues/3748

This PR adds in a `queryIDListString` to the `getNextCoursePage` function which takes in a string of course IDs and appends that as a parameter for` id=` to the end of the API url.  This is set when a department other than `ALL_DEPARTMENTS` is selected and `getNextCoursePage` is called.

This should make it so that on a page with less than the expected numebr of courses already loaded, the missing courses are loaded.

### How can this be tested?
Load up with more than 12 courses. You should have a few in a couple of departments. This will allow you to see if, when you swap to one, it loads after swapping.

### DIFF
--- a/frontend/public/src/containers/pages/CatalogPage.js
+++ b/frontend/public/src/containers/pages/CatalogPage.js
@@ -71,7 +71,8 @@ export class CatalogPage extends React.Component<Props> {
     items_per_row:              3,
     courseQueryPage:            1,
     programQueryPage:           1,
-    isLoadingMoreItems:         false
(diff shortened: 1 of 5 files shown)


annagav (changes requested):
While I was testing I noticed that when I selected a department tab with many course in it and then scrolled down to load more, I saw the same course repeated twice. Also, it seems that the sorting gets messed up also.

While under "All departments" the repetition did not occur.

annagav (approved):
LGTM 👍

PR #2121, opened 2024-03-07
Change program end date to certificate creation date

annagav:
Fix https://github.com/mitodl/hq/issues/3622

The end date for the program on the certificate is the date the certificate was created.

To test this:
Create a program certificate, go to view the certificate, make sure the the end date on the certificate is showing the date it was created.
Screenshot 2024-03-07 at 8 37 26 AM

### DIFF
--- a/courses/models.py
+++ b/courses/models.py
@@ -992,7 +992,7 @@ class ProgramCertificate(TimestampedModel, BaseCertificate):
     def start_end_dates(self):
         """
         Start date: earliest course run start date
-        End date: latest course run end date
+        End date: program certificate creation date
         """
         course_ids = [course[0].id for course in self.program.courses]
         dates = CourseRunCertificate.objects.filter(
--- a/courses/models.py
+++ b/courses/models.py
@@ -997,9 +997,7 @@ class ProgramCertificate(TimestampedModel, BaseCertificate):
         course_ids = [course[0].id for course in self.program.courses]
         dates = CourseRunCertificate.objects.filter(
             user_id=self.user_id, course_run__course_id__in=course_ids
-        ).aggregate(
-            start_date=models.Min("course_run__start_date")
-        )
+        ).aggregate(start_date=models.Min("course_run__start_date"))
         return dates["start_date"], self.created_on
 
     def __str__(self):
--- a/courses/models.py
+++ b/courses/models.py
@@ -998,10 +998,9 @@ class ProgramCertificate(TimestampedModel, BaseCertificate):
         dates = CourseRunCertificate.objects.filter(
             user_id=self.user_id, course_run__course_id__in=course_ids
         ).aggregate(
-            start_date=models.Min("course_run__start_date"),
-            end_date=models.Max("course_run__end_date"),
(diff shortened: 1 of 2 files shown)


courses/models.py
```diff
@@ -997,11 +997,8 @@ def start_end_dates(self):
         course_ids = [course[0].id for course in self.program.courses]
         dates = CourseRunCertificate.objects.filter(
             user_id=self.user_id, course_run__course_id__in=course_ids
-        ).aggregate(
-            start_date=models.Min("course_run__start_date"),
-            end_date=models.Max("course_run__end_date"),
-        )
-        return dates["start_date"], dates["end_date"]
+        ).aggregate(start_date=models.Min("course_run__start_date"))
```
collinpreston, line 1000: It looks like the docstring for this method should be updated.

collinpreston (changes requested):

courses/models_test.py
```diff
@@ -523,7 +523,7 @@ def test_program_certificate_start_end_dates_and_page_revision(user):
     certificate = ProgramCertificateFactory.create(program=program, user=user)
     program_start_date, program_end_date = certificate.start_end_dates
     assert program_start_date == early_course_run.start_date
-    assert program_end_date == later_course_run.end_date
+    assert program_end_date == certificate.created_on
```
collinpreston, line 526: The variable `program_end_date` should probably be renamed since it's not actually equal to the program's end_date.  I would also update the docstring for this unit test - I would describe how the test ensures `start_end_dates` is working as expected.

collinpreston:
Functional testing passed.  Just some documentation updates requested then this is good to go.

courses/models_test.py
annagav: I think ` program_end_date` sounds correct in its meaning. Because the program end date is unique for each learner dependent on when they finished all the courses.

collinpreston: Does the program model have an `end_date` attribute?

pdpinch:
> Does the program model have an end_date attribute?
> 

No. It's going to be unique to each learner, based on when they completed their last required course. 

Let me know if that makes sense or needs more background.

annagav:
The `get_end_date` in the FullProgramSerializer has no value since the program never ends, it is ongoing.

annagav:
Should we just remove `get_end_date`?

collinpreston:
> Why do you think the doc string for `get_end_date` is incorrect?

I guess I'm confused why `end_date` is in the FullProgramSerializer if it never has a value.  Looking at the code for it, it seems like it could have a value as long as there are course_runs, with end_dates, that are associated with the Program.

annagav:
The `get_end_date` in the `FullProgramSerializer` has no relation to certificates. Can we merge the PR?

collinpreston:
> The `get_end_date` in the `FullProgramSerializer` has no relation to certificates. Can we merge the PR?

Yeah sure.  Sorry to bother.

collinpreston (approved):
Approving to quickly resolve the problem.  Not clear how we're defining program end_date across the application.

PR #2124, opened 2024-03-08
Upgrade dialog choose any course run

annagav:
Fix https://github.com/mitodl/hq/issues/3671 and partially https://github.com/mitodl/hq/issues/3109

Changes the enroll now button to open the dialog whenever there is more then one course run available for enrollment.
In the dialog the dropdown select course run shows all enrollable course runs. The runs that can't be upgraded have a label  `(no certificate available)`

### How can this be tested?
There are few scenarios to test:
**1. All course runs for a given course don't have products but are enrollable**
Screenshot 2024-03-20 at 2 19 36 PM

**2. One course run has product and other not, but both are enrollable**
Screenshot 2024-03-20 at 2 33 10 PM

**3. two course runs both have products both are enrollable, but only one is upgradable** (meaning payment_deadline is in the future) - in this scenario the dialog looks the same as case 2.
**4. Has one enrollable course run with no product** - in this case when you click enroll button it just enrolls you, no dialog is shown.
**5. Has one enrollable and upgradable run**   
Screenshot 2024-03-20 at 2 35 41 PM

**6. has two runs both self paced and enrollable** ( and upgradable?)
If at least one of the course runs is upgradable the dialog would look like case 2. Otherwise, the dialog would show up as case 1.

### DIFF
--- a/frontend/public/src/components/CourseProductDetailEnroll.js
+++ b/frontend/public/src/components/CourseProductDetailEnroll.js
@@ -44,7 +44,6 @@ import AddlProfileFieldsForm from "./forms/AddlProfileFieldsForm"
 import CourseInfoBox from "./CourseInfoBox"
 
 import type { User } from "../flow/authTypes"
-import type { Product } from "../flow/cartTypes"
 
 type Props = {
   courseId: ?string,
@@ -537,14 +536,12 @@ export class CourseProductDetailEnroll extends React.Component<
       enrollments,
       enrollmentsIsLoading
     } = this.props
-    let run,
-      product = null
+    let run = null
 
     if (courses && courseRuns) {
       run = getFirstRelevantRun(courses[0], courseRuns)
 
       if (run) {
-        product = run && run.products ? run.products[0] : null
         this.updateDate(run)
       }
     }
--- a/frontend/public/src/components/CourseProductDetailEnroll.js
+++ b/frontend/public/src/components/CourseProductDetailEnroll.js
@@ -509,19 +509,33 @@ export class CourseProductDetailEnroll extends React.Component<
   }
 
   renderEnrollNowButton(run: EnrollmentFlaggedCourseRun) {
-    const { currentUser } = this.props
+    const { currentUser, courseRuns } = this.props
+    const csrfToken = getCookie("csrftoken")
     return currentUser &&
       currentUser.id &&
       run &&
       isWithinEnrollmentPeriod(run) ? (
         <h2>
-          <button
-            id="upgradeEnrollBtn"
-            className="btn btn-primary btn-enrollment-button btn-lg btn-gradient-red highlight enroll-now"
-            onClick={() => this.toggleUpgradeDialogVisibility()}
-          >
-          Enroll now
-          </button>
+          {courseRuns && courseRuns.length > 1 ? (
+            <button
+              id="upgradeEnrollBtn"
+              className="btn btn-primary btn-enrollment-button btn-lg btn-gradient-red highlight enroll-now"
+              onClick={() => this.toggleUpgradeDialogVisibility()}
+            >
+            Enroll now
+            </button>
+          ) : (
+            <form action="/enrollments/" method="post">
+              <input type="hidden" name="csrfmiddlewaretoken" value={csrfToken} />
+              <input type="hidden" name="run" value={run ? run.id : ""} />
+              <button
+                type="submit"
+                className="btn btn-primary btn-enrollment-button btn-gradient-red highlight enroll-now"
+              >
+              Enroll now
+              </button>
+            </form>
+          )}
         </h2>
       ) : null
   }
--- a/frontend/public/src/components/CourseProductDetailEnroll.js
+++ b/frontend/public/src/components/CourseProductDetailEnroll.js
@@ -183,7 +183,6 @@ export class CourseProductDetailEnroll extends React.Component<
 
   hndSetCourseRun = (event: any) => {
     const { courseRuns } = this.props
-
     if (event.target.value === "") {
       this.setCurrentCourseRun(null)
       return
@@ -194,7 +193,6 @@ export class CourseProductDetailEnroll extends React.Component<
         (elem: EnrollmentFlaggedCourseRun) =>
           elem.id === parseInt(event.target.value)
       )
-
     if (matchingCourseRun) {
       this.setCurrentCourseRun(matchingCourseRun)
     }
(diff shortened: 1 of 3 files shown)


frontend/public/src/components/CourseProductDetailEnroll.js
```diff
@@ -274,18 +273,17 @@ export class CourseProductDetailEnroll extends React.Component<
...
-                  value={elem.id}
-                  key={`courserun-selection-${elem.id}`}
-                >
-                  {formatPrettyDate(moment(new Date(elem.start_date)))} -{" "}
-                  {formatPrettyDate(moment(new Date(elem.end_date)))}
-                </option>
-              ))}
+            courseRuns.map((elem: EnrollmentFlaggedCourseRun) => (
+              <option
+                selected={run.id === elem.id}
```
pdpinch, line 278: Can we remove the default selection & force the user to make a choice?

annagav:
If we want to have the select, not having something selected by default it would look something like this:

Screenshot 2024-03-12 at 9 13 42 AM

annagav:
We could try to add a default "nothing selected" option. And then we would need to disalbe all the buttons in the dialog:

Screenshot 2024-03-12 at 2 28 50 PM

pdpinch:
@annagav this works for me, but I'd like to tweak the "nothing selected" text to be something based on best practices. 

@mbilalmughal can you weigh in on this? 

Some context: customer support has brought up that we get a number of tickets from learners who enroll without realizing what course dates they are signing up for. CS would like to cut down on these tickets and they have suggested that we **not** provide a default course run in situations when there are multiple runs like this. Learners would be required to make an active choice between the runs. 

In the past, I had been reluctant to make this change, since it adds friction to the enrollment process, but now that I've seen the frustration it is causing customer support, I'm willing to try something like this.

pdpinch:
I think this will address https://github.com/mitodl/hq/issues/3400 as well.

annagav:
@mbilalmughal Thanks for the feedback. The only thing is that when the run is not selected we can't show the price and certificate availability info since it depends on the run selected. So it would look a bit different on my screenshots.

Ferdi:
Perhaps just make the date field required ?

I don't think users will understand "can upgrade". What is that supposed to do ?

annagav:
> Perhaps just make the date field required ?
> 
> I don't think users will understand "can upgrade". What is that supposed to do ?

We are trying to mark courses that could be upgraded for a certificate track. What wording do you think would be better?

pdpinch:
Instead of marking the ones that can be upgraded with `(can upgrade)`, we can mark the ones that can't be upgraded with `(no certificate available)` -- which matches the text in the info box when there is no certificate.

pdpinch:
> For the same course, you can upgrade for some runs but other runs ? I did't know that was the case
> 
This shouldn't happen, but we've had edge cases where the course team has left enrollment open in one of the runs after the course upgrade deadline. This is what happened with https://github.com/mitodl/hq/issues/3671

annagav:
Latest screenshots:

Screenshot 2024-03-13 at 9 24 34 AM
Screenshot 2024-03-13 at 9 24 56 AM

annagav:
@pdpinch @mbilalmughal could you review the latest screenshots?

frontend/public/src/components/CourseProductDetailEnroll.js
```diff
@@ -273,19 +240,17 @@ export class CourseProductDetailEnroll extends React.Component<
           onChange={this.hndSetCourseRun.bind(this)}
           className="form-control"
         >
+          <option value="" key="default-select">
+            Nothing Selected
```
pdpinch, line 244:
```suggestion
            Please Select
```

annagav:
@collinpreston this is ready for review now.

collinpreston (changes requested):
Can you write what is expected for each of the test cases listed in the PR description?

frontend/public/src/components/CourseProductDetailEnroll_test.js
```diff
@@ -532,7 +534,8 @@ describe("CourseProductDetailEnrollShallowRender", () => {
       if (multiples) {
         assert.isTrue(selectorControl.exists())
         const selectorControlItems = selectorControl.find("option")
-        assert.isTrue(selectorControlItems.length === 2)
+        assert.isTrue(selectorControlItems.length === 3)
+        assert.isTrue(selectorControlItems.at(0).text() === "Nothing Selected")
```
collinpreston, line 538: This should be updated.

collinpreston:
@pdpinch just my opinion, but feels a bit confusing to have the disabled "Enroll and pay" button.  I don't think it's natural to identify that the "Enroll for free" link is what should be clicked on.

pdpinch:
> I don't think it's natural to identify that the "Enroll for free" link is what should be clicked on when a course does not offer an upgraded enrollment.

What are our options?

@mbilalmughal what do you think?

mbilalmughal:
> > I don't think it's natural to identify that the "Enroll for free" link is what should be clicked on when a course does not offer an upgraded enrollment.
> 
> What are our options?
> 
> @mbilalmughal what do you think?

I agree with @collinpreston. It doesn't make sense to disable it.
@pdpinch if you remember, we added the disabled button state just to ensure the user selects the date first. Nothing else is required. If you are okay with it, let's remove the disabled button state.

annagav:
@collinpreston If there is no upgradable run the dialog should look like this: 
Screenshot 2024-03-20 at 2 19 36 PM

We could remove the "without certificate" part.

annagav:
@collinpreston I updated the description with test scenarios and corresponding pictures.

frontend/public/src/components/CourseProductDetailEnroll_test.js
```diff
@@ -532,12 +534,95 @@ describe("CourseProductDetailEnrollShallowRender", () => {
...
+          upgradeForm
+            .find("button.btn-upgrade")
+            .at(0)
+            .prop("disabled")
+        )
       }
     })
   })
+
+  it(`renders enrollment dialog for a course with multiple non-upgradable active runs`, async () => {
```
collinpreston, line 585: Can you update the description of this test, and the one above that you modified, to indicate what the test is ensuring happens?

collinpreston (changes requested):
Looks good.  Just a documentation request.

collinpreston (approved):
Looks good.  I wouldn't merge until we get a decision from Peter/Bilal regarding the UX/UI of the disabled button.

pdpinch (commented):
Let's merge this and get in onto RC so I can try it out with (semi-)real courses.

PR #2128, opened 2024-03-15
1257: updating meta tags on mitx online site

collinpreston:
https://github.com/mitodl/hq/issues/1257

This PR adds the Wagtail metadata plugin.  This same plugin is used in xPro.  The plugin ensures that our product pages are best suited for SEO, in addition to adding metadata, entered into the CMS for each course and program page, into the page.

### How can this be tested?

1. Using this branch with some test courses or programs.  This PR adds a new migration.  Ensure the migration is run prior to the following steps.
2. Update the course's or program's  metadata via the CMS as shown below.

Screenshot 2024-03-15 at 1 48 06 PM

3. Publish the CMS changes and view the live program or course page.
4. View the source code for the course or program page.
5. Verify that the source code contains a metadata tag in the header, as well as other SEO related tags added via the plugin, similar to the example below.

Screenshot 2024-03-15 at 1 47 07 PM


JenniWhitman (approved):
Works after I was able to migrate (personal issue, nothing wrong with it once I fixed my data)

PR #2132, opened 2024-03-19
3649 all mitxonline courses should have a department

collinpreston:
https://github.com/mitodl/hq/issues/3649

### Description
Makes "department" a required field for courses and programs.  Updates the create_courseware management command to require department.

This PR removes functionality that created courses or programs in some management commands where the main functionality was not intended to create courses or programs.  This should simplify those management commands and keep them specific in their intended use cases.

### How can this be tested?
You must have edX (Tutor) up and running locally in order to fully test these changes.  You only need to configure Tutor so that the service user is setup in MITx Online.  This PR includes changes to update the Tutor instructions.

1. Login to Django Admin in MITx Online.
2. http://mitxonline.odl.local:8013/admin/courses/course/ Attempt to create and save a course without a department selected.  Verify an error message is shown and the course is not created.
3. http://mitxonline.odl.local:8013/admin/courses/program/ Attempt to create and save a program without a department selected.  Verify an error message is shown and the program is not created.
4. Run the `create_courseware` management command to create a course without specifying the `--department` argument.  Verify that no course is created and the command outputs an error message.
5. Run the `create_courseware` management command to create a program without specifying the `--department` argument.  Verify that no program is created and the command outputs an error message.
6. Run the `configure_instance` command and verify, via Django Admin, that courses and a program are created in MITx Online.
7. Run the `import_courserun` command.  I used the following command which references a course I created in my local instance of edX: `docker-compose run --rm web ./manage.py import_courserun --courserun course-v1:MIT+101+2022`.  Verify that you receive an error message indicating that a department must be defined.
8. Append `--dept Science` to the command in step 7 and run the command again.  Verify, via Django Admin in MITx Online, that the course and corresponding course_run were successfully created.


pdpinch:
I was examining the data in production recently, and I see there are a handful of "test" courses that don't have a department. Will those need to be assigned before you deploy this?

collinpreston:
> I was examining the data in production recently, and I see there are a handful of "test" courses that don't have a department. Will those need to be assigned before you deploy this?

They should have a department.  I'm not sure what would happen if they don't, but I don't think defining a department for them is much to ask.

pdpinch:
Are we doing anything in this PR that would conflict with making the setting of departments self-service some time in the future (see https://github.mit.edu/mitxonline/mitxonline-issues/issues/598)?

pdpinch:
We may have made too strong an assumption here about every course needing a department. Please don't merge this until I've resolved this question: https://github.com/mitodl/hq/issues/3601#issuecomment-2018647709

collinpreston:
> Are we doing anything in this PR that would conflict with making the setting of departments self-service some time in the future (see https://github.mit.edu/mitxonline/mitxonline-issues/issues/598)?

Can you please explain what "making the setting of departments self-service" is?

JenniWhitman (commented):
Worked as expected when testing. Comments made clear what is going on and as written, this follows the requirements, but after discussing with Peter I think we're holding on this per outcome of discussion re: the model change (`blank=false`) as that might be too hard of a requirement for some uses. My only other comment/thought while testing was some weirdness in course/program about pages when I didn't have a department set, also a decision around "test" courses/programs and their department. Obviously we can have a "bucket" in which to put those but I'm not sure what that should look like or if this is the place to solve that problem. That's also a one-off problem on implementation and not a code issue.  So, I'm approving, but this is pending the decision/response from Peter

JenniWhitman (approved):

pdpinch:
We can go ahead and merge this. I had some concern that the MITx/OWLS team needed there to be some courses with no department, but https://github.com/mitodl/hq/issues/3601#issuecomment-2018680231 clarified for me that we do indeed need a minimum of one department for each course:

> every course needs a department or it won’t show if people look by department.

Regarding @collinpreston's question:

> Can you please explain what "making the setting of departments self-service" is?

At some future point, it would be good to allow CMS users (i.e. the MITx/OWLS team) set the departments via Wagtail. So instead of directing Bon or me to add or change a department, they could do it themselves.

PR #2139, opened 2024-03-25
Showing Course rather than Courses (or program/s) when there is only 1 to display

JenniWhitman:
Closes https://github.com/mitodl/hq/issues/2488

Adds a function to render the correct display text on the desktop view for 1 course or program as opposed to multiple (without an s to make the noun plural).

### How can this be tested?
Create a department that only has one course and/or one program assigned. You should see the text change from 1 courses (in current main branch) to 1 course when you select that department on the catalog page.


frontend/public/src/containers/pages/CatalogPage.js
```diff
@@ -561,6 +561,20 @@ export class CatalogPage extends React.Component<Props> {
     } else return 0
   }
 
+  renderCatalogCount() {
```
collinpreston, line 564: I would just add a description above this method describing what it does.

frontend/public/src/containers/pages/CatalogPage_test.js
```diff
@@ -882,4 +882,36 @@ describe("CatalogPage", function() {
       JSON.stringify([displayedProgram, newProgram])
     )
   })
+
+  it("renderCatalogCount is plural for more than one course", async () => {
```
collinpreston, line 886: Nice tests!

collinpreston (approved):

PR #2140, opened 2024-03-25
Add URL routing for catalog for courses/programs and department

JenniWhitman:
Closes https://github.com/mitodl/hq/issues/3690 and https://github.com/mitodl/hq/issues/2516

Removes the possibility of courses being shown that are not in the count from the API. We were checking this on the front end, but the API call we're using checks for this now, so I removed that bit from the filter function as it COULD have, in a race condition, created an infinite loading loop.

When you go to catalog/programs or catalog/courses it will pick the correct tab. When you add a department slug, that department will be selected. This PR also introduces a slug to the department field & migrates in order to standardize those slugs and ensure there is nothing wild in the future. It also adds to the management command.

### How can this be tested?
Create a department via management command or admin, should be able to. Should have a slug.
Go to /catalog/programs, should be programs tab. 
Go to /catalog/courses/department-name and should get there, same with programs.
Go to /catalog/fake and should get all departments, same with programs.
Should then be able to click around and not get jumped back. Should have correct counts and everything still.


annagav (changes requested):
When I reload the page it doesn't show the 1 program that I have:
Screenshot 2024-03-29 at 6 34 36 PM

JenniWhitman:
> When I reload the page it doesn't show the 1 program that I have: Screenshot 2024-03-29 at 6 34 36 PM

I've tried to recreate what you're seeing with the program and I can't - have you rerun migrations? I got all tests passing and my local looks good. If you want to paste the API output I can see if there's something missing? If so, we should make sure I add in a check for whatever it is. But I'm not sure why yours is doing that.

annagav (approved):
Works for me 👍

PR #2141, opened 2024-03-25
Upgrade enrollment dialog fixes

annagav:
none

This PR corrects some behaviors when `is_upgrade` and `products` could be true or false independent of each other.

fixes couple of things:
1. When the dialog is closed and opened again the current run is reset to null
2. if a run has a product but the upgrade deadline has passed it should say "certificate not available", and disable upgrade buttons accordingly
3. removes unnecessary code
### How to test

to test this :
1. create a product for your run and set the upgrade deadline to the future.
2. create another run with a product and set upgrade deadline to the past
3. open the upgrade dialog as you select course the messaging should be consistent.

For example:
Here it says in the select element (no certificate available), but underneath shows the cost

Screenshot 2024-03-25 at 3 19 34 PM

### DIFF
--- a/frontend/public/src/components/CourseProductDetailEnroll.js
+++ b/frontend/public/src/components/CourseProductDetailEnroll.js
@@ -132,44 +132,9 @@ export class CourseProductDetailEnroll extends React.Component<
       this.toggleAddlProfileFieldsModal()
     }
   }
-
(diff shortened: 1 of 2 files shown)


JenniWhitman (approved):
Works well. Nice to get a lot gone.  Glad to have a test for this as well!

PR #2142, opened 2024-03-26
Distinguish change forms in account settings

annagav:
Fix https://github.com/mitodl/hq/issues/3830

add aria-label to the change password and email forms for voiceover to read.

### How can this be tested?
Go account-settings and use voiceover to read the forms. Make use it is clear when one form ends and where the other starts.


jkachel (approved):
LGTM - it's a lot clearer now that there are two separate forms when using VoiceOver.

PR #2146, opened 2024-04-02
Deferrals: set to audit instead of unenrolling

annagav:
Fix https://github.com/mitodl/hq/issues/3781

Update the deferral procedure to downgrade enrollment to audit instead of unenrolling.

Testing:
Have a user with existing enrollment and fullfilled order for that run.
Create another run that is open for enrollment
Then run the management command:
`./manage.py defer_enrollment --user <username> --from-run <readable_id> --to-run <readable_id> -k`
Then check in the admin that the `from_run` enrollment is now set to audit, and the `to_run` is set to verified.

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -34,7 +34,7 @@ from courses.models import (
     Program,
     ProgramCertificate,
     ProgramEnrollment,
-    ProgramRequirement,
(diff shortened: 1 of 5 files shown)


collinpreston (approved):
looks good.

Issue #2116, opened 2024-02-28
Improve resilience of posthog integration


rhysyngsun:
### Description/Context
Posthog had [an outage](https://status.posthog.com/incidents/p9fhh5gnd5ky) which caused calls to their API to timeout.

We worked around this by setting `IN_TEST_SUITE=True` and `FEATURES_DEFAULT=True` because we were fortunate enough to have all features enabled. It took some time to evaluate if this was the correct way to address the issue because `IN_TEST_SUITE` sounds like it could have other undesirable effects.

### Plan/Design

- Rename `IN_TEST_SUITE` to `POSTHOG_ENABLED`
- Investigate options to tune the timeout/retry  settings in posthog
  - The default appears to be 3 retries w/ 15 second timeouts, pushing us well over the 30 second Heroku request timeout.
  - The client itself appears to have settings for these: https://github.com/PostHog/posthog-python/blob/master/posthog/client.py#L32
  - These aren't all exposed through module-level properties, but it looks like we can set [`default_client`](https://github.com/PostHog/posthog-python/blob/master/posthog/__init__.py#L22)
  - See also how the library itself creates a default `default_client`: https://github.com/PostHog/posthog-python/blob/master/posthog/__init__.py#L441-L455
  - Handle errors from the call to Posthog
- Document procedures for Posthog outages

pdpinch:
How similar is this to the work @jkachel is doing on https://github.com/mitodl/mit-open/pull/693

collinpreston:
@rhysyngsun what does "pushing us well over the 30 second Heroku request timeout." mean?

rhysyngsun:
Heroku has a 30 second request timeout: https://devcenter.heroku.com/articles/request-timeout

collinpreston:
This should not be automatically closed.  This can be closed once the changes added in https://github.com/mitodl/ol-django/pull/152 are integrated with MITx Online.

collinpreston:
Blocked by https://github.com/mitodl/ol-django/pull/156

pdpinch:
Even though we made this change, we had another outage. Unlike the first outage, instead of PostHog timing out, it returned an error. So we need to add better error handling.

PR #2148, opened 2024-04-03
2116 improve resilience of posthog integration

collinpreston:
https://github.com/mitodl/mitxonline/issues/2116

1. Adds `POSTHOG_FEATURE_FLAG_REQUEST_TIMEOUT_MS` as an environment variable. Default is 3000 milliseconds.
2. Adds `POSTHOG_MAX_RETRIES` as an environment variable.  Default is 3.
3. Defines `feature_flags_request_timeout_seconds` as an option when calling PostHog (https://posthog.com/docs/feature-flags/adding-feature-flag-code#request-timeout).
4. Rename IN_TEST_SUITE to POSTHOG_ENABLED.

### Testing

1. PostHog functionality should not be impacted when running MITx Online locally.  POSTHOG_ENABLED should be set to True locally.
2. If you define `POSTHOG_FEATURE_FLAG_REQUEST_TIMEOUT_MS=1` in your .env file and restart MITx Online.  When you load the homepage you should not see any features that require the feature flags to be active.
3. If you define `POSTHOG_FEATURE_FLAG_REQUEST_TIMEOUT_MS=3000` in your .env file and restart MITx Online.  When you load the homepage you should see features that require the feature flags to be active.


jkachel:
I was looking through this and it seems pretty close to the implementation [that's in Open](https://github.com/mitodl/mit-open/pull/682). Rather than squaring these two implementations up, should they be merged and put instead into ol-django's common app? That way MITx Online and Open will have consistent implementations (and MITx Online would gain a local feature flag cache). Thoughts?

PR #2149, opened 2024-04-03
Enrollment request make sure not enrolled

annagav:
Fix https://github.com/mitodl/hq/issues/3013

If you are already enrolled in this course run, and you go to product page and try to enroll again, it will not call `create_run_enrollments`

### How can this be tested?
Enroll in a course run in certificate track. Make sure that PaidCourseRun is deleted.
Go to course product page try to enroll in the same run again. The enrollment post request should return duplicate error.
Repeat the same with the program product page. The program page enrolls you in the run it find convenient so make sure you are already enrolled for that run.
### Question
I am not sure if we should should restrict re-enrollment only if the user is enrolled in verified track, or any type.

### DIFF
--- a/courses/views/v1/__init__.py
+++ b/courses/views/v1/__init__.py
@@ -274,7 +274,7 @@ def _validate_enrollment_post_request(
     if (
         PaidCourseRun.fulfilled_paid_course_run_exists(user, run)
         or CourseRunEnrollment.objects.filter(
-            user, run=run, change_status=None
+            user=user, run=run, change_status=None
         ).exists()
     ):
         resp = redirect_with_user_message(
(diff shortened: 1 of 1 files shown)


rachellougee (approved):
LGTM

This was mentioned in https://github.com/mitodl/hq/issues/3404 that we restrict re-enrollment if user is enrolled in verified track, assuming no harm enrolling twice, but if that's the cause of bug, then it makes sense to restrict it. But I am wondering why PaidCourseRun was deleted if user enrolled in certificate track?

annagav:
I think PaidCourseRun was deleted due to deferral procedure. But that is being corrected in this PR https://github.com/mitodl/mitxonline/pull/2146

PR #2152, opened 2024-04-08
Fix catalog page for empty app

collinpreston:
### What are the relevant tickets?
NA

### Description

1. Removes a bunch of unused or duplicate code.
2. Refines logic and methods to be more single-purpose/functional.
3. Allows for Program + department URLs such as http://mitxonline.odl.local:8013/catalog/programs/science
4. Resolves and issue where https://rc.mitxonline.mit.edu/catalog/programs loads an empty catalog page.

### How can this be tested?

1. Verify that you can navigate the catalog page for both courses and programs, switching between departments, switching between tabs, and loading more courses and programs as you scroll.
2. Verify that you can navigate to the course and program tab using a URL such as http://mitxonline.odl.local:8013/catalog/programs/
3. Verify that you can navigate to the course and program tab, and filter by department, using a URL such as http://mitxonline.odl.local:8013/catalog/programs/science
4. Accessibility of the catalog page should not change.

PR #2153, opened 2024-04-08
Ensure filteredPrograms is set earlier and change the route when navigation changes.

JenniWhitman:
Fixes https://github.com/mitodl/hq/issues/3950

Makes sure filteredPrograms is set earlier on the page to avoid the state where the programs don't show on load of /catalog/programs.
Added routing to update history on change to tab or department

### How can this be tested?
load /catalog/programs and programs should show
load /catalog/programs/department-name and department name should show
When you navigate by clicking a tab or department, the URL should update


annagav (approved):
LGTM 👍

JenniWhitman:
> Can I pull these changes into this larger PR (#2152) I currently have?

Sure - I don't see why not

PR #2154, opened 2024-04-10
Fix enrollment dialog default option

annagav:
Fix https://github.com/mitodl/hq/issues/3944

When selecting "Please select" again it should reset `currentCourseRun` and disable both upgrade and enroll buttons.

Screenshot 2024-04-10 at 1 26 25 PM


annagav:
I am seeing something, that might be a bug:

I select a department that has no propgrams `Department test 1`
Screenshot 2024-04-11 at 8 35 57 AM

Then I click on programs tab. It shows me the programs correctly but the url stays with the previous depeartment selection:
Screenshot 2024-04-11 at 8 35 38 AM

rachellougee (approved):
👍

PR #2157, opened 2024-04-16
Course Product Page with no Runs

annagav:
Fix https://github.com/mitodl/hq/issues/3430

Updates the Course info box to still show info when no active run is available.

Screenshot 2024-04-16 at 3 28 28 PM

To test this make sure you have a course that has no runs that are open for enrollment. The go to that course page, for exapmle: http://mitxonline.odl.local:8013/courses/w-course-29/
Make sure nothing else breaks.

### DIFF
--- a/frontend/public/scss/product-page/product-details.scss
+++ b/frontend/public/scss/product-page/product-details.scss
@@ -270,6 +270,11 @@ body.new-design {
           text-align: center;
         }
 
+        .btn-enrollment-button.disabled {
+          color: $white;
+          background-color: grey;
+        }
+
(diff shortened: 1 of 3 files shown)


JenniWhitman (approved):
Tested both with no open runs to enroll in, runs past date, runs open but with enrollment in the past, and also flipping those runs back and everything shows as expected.

pdpinch:
Does this PR include automated test cases for these conditions? 

> Tested both with no open runs to enroll in, runs past date, runs open but with enrollment in the past, and also flipping those runs back and everything shows as expected.

PR #2183, opened 2024-05-01
Update poetry and yarn lock files

rachellougee:

to fix https://github.com/mitodl/mitxonline/pull/2182

Updating poetry and yarn lock files to fix https://github.com/mitodl/mitxonline/pull/2182. 

### How can this be tested?

automated tests should pass


ecommerce/views_test.py
```diff
@@ -119,7 +119,9 @@ def test_get_products_inactive(user_drf_client, products):
         reverse("products_api-detail", kwargs={"pk": product.id})
     )
 
-    assert_drf_json_equal(resp.json(), {"detail": "Not found."})
+    assert_drf_json_equal(
+        resp.json(), {"detail": "No Product matches the given query."}
+    )
```
rachellougee, line 125: This change is to fix test failure when djangorestframework going to 3.15.1 in this PR.  Alternatively, I can update this line to ` assert resp.status_code == 404` so we don't assert the exact error message.

PR #2190, opened 2024-05-06
Adding management command and function for pulling featured items

JenniWhitman:
Closes https://github.com/mitodl/hq/issues/4028

- Refactors out the filter for enrollable courses to maintain consistency between our catalog API and the homepage featured courses
- Creates a function to create a cached value "CMS_homepage_featured_courses" which holds the queryset, with pages pre-selected, for those featured courses with an age of 1 day. This list is:
  - enrollable
  - The first 2 will be self-paced
  - The rest will be a random assortment
  - This will then be ordered by those with a start date in the future and then those that have already started will be left random

### How can this be tested?
Create courses and run the management command. You will then have to check the memcache database for the key - this can be done in the django shell by importing django.core.cache and using the get method.


annagav (approved):
I was able to successfully run the command and check the cache being set 👍

PR #2193, opened 2024-05-07
3783: dedp course certificates in hubspot

collinpreston:
https://github.com/mitodl/hq/issues/3783

### Description

1. Creates two new checkbox properties, associated with contacts, in Hubspot: course_run_certificates and program_certificates.
2. Adds functionality to populated the new properties with data based on an MITx Online learner's course run and program certificates.  Ensures that the HubSpot contact is updated whenever a certificate is created or deleted in MITx Online.

### How can this be tested?

1. Configure you local MITx Online instance to work with the Hubspot dev environment: (`MITOL_HUBSPOT_API_PRIVATE_TOKEN`, `MITOL_HUBSPOT_API_ID_PREFIX`, `HUBSPOT_PIPELINE_ID`).
3. Create a user and a corresponding course or program certificate.
4. Run `docker-compose run web ./manage.py configure_hubspot_properties` in order to update the Hubspot course_run_certificates and program_certificates properties with the string representation of your course runs and programs.
5.  From the profile page, update your user's first or last name in order to trigger a sync with Hubspot.
6. Verify that the corresponding Contact in Hubspot has the course or program certificate from step 2 defined.

The Hubspot Contact will have properties displayed like the screenshot below:
Screenshot 2024-05-07 at 11 26 46 AM

You can also test the course and program certificate creation, and subsequent triggering of syncing the user's data with the Hubspot contact, by enrolling the user in a course, defining a grade for the user in edx, running the periodic edx sync command that checks for course run completion and grades from MITx Online, which will then create a course and possibly a program certificate.


hubspot_sync/management/commands/configure_hubspot_properties.py
```diff
@@ -637,26 +638,88 @@
 }
 
 
-def upsert_custom_properties():
+def _get_course_run_certificate_hubspot_property():
+    """
+    Creates a dictionary representation of a Hubspot checkbox,
+    populated with options using the string representation of all course runs.
```
pdpinch, line 644: does "string representation" mean "readable id" ?

I'm not sure this is the best place to put it, but an example might help.

collinpreston: For course runs, this would be `courseware_id | title`.

JenniWhitman (approved):
Was able to create a user, sync that, certify said user and once I updated the profile, the certificate was synced for both courses and programs.

collinpreston:
@pdpinch When a new course run or program is created, the command `./manage.py configure_hubspot_properties` must be run from the `web` container in order add that course run or program as an option in the Hubspot certificate property.  How should this process be handled/communicated/documented?

hubspot_sync/management/commands/configure_hubspot_properties.py
pdpinch: Can you give an example? Based on `courseware_id | title` I'm hoping an example would be "course-v1:MITx+14.73x+1T204 | The Challenges of World Poverty"

pdpinch:
> When a new course run or program is created, the command ./manage.py configure_hubspot_properties must be run from the web application in order to add that course run or program as an option in the Hubspot certificate property. 

We discussed this over Slack and it seems like creating the property when a certificate (course or program) is created is preferable to adding another management command that needs to be run manually. 

> How should this process be handled/communicated/documented?

I think we've outgrown this discussion post -- https://github.com/mitodl/hq/discussions/306 -- and should establish something more permant, like the google doc you started. In any case, we should loop in @cachob who will be a key consumer of this kind of documentation.

hubspot_sync/management/commands/configure_hubspot_properties.py
collinpreston: course-v1:MITxT+14.100x+3T2023 | Microeconomics

PR #2199, opened 2024-05-10
restrict duplicate enrollment only if user is verified

annagav:
Fix https://github.com/mitodl/hq/issues/4210
We had a problem where a user was downgrading their enrollment by enrolling again.
So we restricted duplicate enrollments.
However, this caused the issue above.

This PR restricts duplicate enrollment only if user is verified in that course.

### How can this be tested?
Enroll for free in a course run.
Then try to pay for it.
The payment should be successful, and you should see your enrollment in the dashboard.

### DIFF
--- a/courses/views/v1/__init__.py
+++ b/courses/views/v1/__init__.py
@@ -256,7 +256,7 @@ def _validate_enrollment_post_request(
             user=user,
             run=run,
             change_status=None,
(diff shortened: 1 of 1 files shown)


JenniWhitman (approved):
When I am enrolled verified, I can still enroll in other runs.
when I am enrolled verified, I can not enroll in the same run as audit.
when I am enrolled audit, I can enroll as audit.

PR #2200, opened 2024-05-10
add relational dates to ensure all courses end up getting pulled

JenniWhitman:
Fixes https://github.com/mitodl/hq/issues/4215

Changes the dates for the future courses so they remain within range of enrollment

### How can this be tested?
Run `docker compose run --rm web pytest cms/api_test.py::test_home_page_featured_products` multiple times to ensure it does not fail


annagav (approved):
👍

PR #2203, opened 2024-05-13
Add task for creating featured items in cache

JenniWhitman:
Updates https://github.com/mitodl/hq/issues/4029

Adds an offset, single run  cron task that runs every minute to check if the items exist and update if not. This is offset by 30 seconds so it gives some space for the similar google sheets task.

### How can this be tested?
Run the app and see that the cron is running in the celery logs. There should be adequate logging to view this manually.


annagav (approved):
👍

PR #2208, opened 2024-05-14
Updating python version to fix pants

JenniWhitman:
Fixes https://github.com/mitodl/hq/issues/4223

Change the python version on the workflow that calls the pants build

### How can this be tested?
Check out the docs action in github actions https://github.com/mitodl/mitxonline/actions/workflows/docs.yml (I had to manually kick it off using this branch, you can as well to test again)


annagav (approved):
👍

PR #2210, opened 2024-05-15
Adding course topics

annagav:
Fix https://github.com/mitodl/hq/issues/4110
Adding topics in MITx Online. 
The ability to create delete topics through the cms, and add them to courses.

Populates topics with existing department names. I don't think it is necessary, but this is how it was done in mitxpro. And I think it is nice to have some topics to begin with.

Screenshot 2024-05-16 at 8 53 26 AM

### How can this be tested?

Run migrations
Run the app, go to the cms and make sure you can see Course Topics.
Click on it and try to add or delete a course topic.

Go to Course page add topic to a course page.

Also view `http://mitxonline.odl.local:8013/api/v2/parent_course_topics/`


collinpreston:
@annagav hey, my fault, but I think you will need to update this branch from main in order for the application to properly load.  Related to: https://github.com/mitodl/mitxonline/pull/2213

courses/utils.py
```diff
@@ -166,3 +166,27 @@ def get_unenrollable_courses(queryset):
         .filter(courseruns__id__in=courseruns_qs.values_list("id", flat=True))
         .distinct()
     )
+
+
+def get_catalog_course_filter(relative_filter=""):
```
collinpreston, line 171: Do the filters in this method actually align with the filters on our catalog page?

pdpinch:
Can we pre-populate the topics with the ones from this spreadsheet: https://docs.google.com/spreadsheets/d/1Euu4ezPKeYkiTThvmxB5vFONFNG4fOULNW42F6qLuFI/edit?pli=1#gid=431773359

annagav:
> Can we pre-populate the topics with the ones from this spreadsheet: https://docs.google.com/spreadsheets/d/1Euu4ezPKeYkiTThvmxB5vFONFNG4fOULNW42F6qLuFI/edit?pli=1#gid=431773359

Should the first row be included?

courses/utils.py
annagav: I am not sure. Maybe the start date is irrelevant, since we show archived courses. What do you think?

pdpinch:
Yeah, it's a hierarchy. The first row are the top-level topics. The other ones are children.

courses/utils.py
annagav: So it looks like this is the place where courses get filtered for the catalog `class CourseFilterSet`.
The main criteria is_live and is_enrolable.

collinpreston: Do you know what defines "is_enrollable" in this case?

annagav:
@collinpreston make sure reverse all three migrations before checking out the changes:
`./manage.py migrate cms 0035`
`./manage.py migrate courses 0050`

courses/utils.py
annagav: So I looked into a bit, and In order to reuse the the existing filtering I would need to refactor a bunch of existing code, and might introduce a bug. I think it would be better to do that refactor in a separate PR, and test it thoroughly.

collinpreston (commented):
Looks good.  I would just check whether Programs should also have Topics.

annagav:
> Looks good. I would just check whether Programs should also have Topics.

Following the example in xpro, no programs don't have topics.

PR #2211, opened 2024-05-15
Add ol-django olposthog app

collinpreston:
### What are the relevant tickets?
https://github.com/mitodl/mitxonline/issues/2116

### Description (What does it do?)
Adds the ol-django olposthog app to MITx Online in order to centralize Posthog functionality between Open Learning applications.

Adds two new environment variables for controlling Posthog API requests:
POSTHOG_FEATURE_FLAG_REQUEST_TIMEOUT_MS
POSTHOG_MAX_RETRIES
https://docs.google.com/document/d/1W7jYty8HjkaYxSFXoUMF9ZYTOp8eJ4ssmINfgLhQ61Q/edit#heading=h.33ix6lloihnq

### How can this be tested?

1. Posthog feature flags and associated functionality should still perform as they previously did.

PR #2213, opened 2024-05-20
Fix missed feature flag after merge 

collinpreston:
NA

I missed one of the feature flag requests while merging https://github.com/mitodl/mitxonline/pull/2211.  This PR will fix that.

### How can this be tested?
App should load with no errors.


annagav (approved):
👍

PR #2214, opened 2024-05-20
Incorrect count announce

JenniWhitman:

- [ ] Desktop screenshots


annagav:
Just tested this. Unfortunately the issue of the number not being announced when selecting a department and the number of courses is the same.

PR #2216, opened 2024-05-20
Fix flaky test: test_create_featured_courses

JenniWhitman:
Fixes https://github.com/mitodl/hq/issues/4308

test_create_featured_courses performs fine when the dates align, but when the factory produces a random date sometimes that date doesn't align as one would expect. To combat this, I set the dates based on one on another and based on "now" to keep the results consistent.  This was already in place for the future courses since their dates determined their order, but I have also now done this for the in_progress course.  I also made the check that they're all in place into one statement instead of 4 using sets.

### How can this be tested?
Check my logic and run the test... a lot.

### DIFF
--- a/cms/api_test.py
+++ b/cms/api_test.py
@@ -359,6 +359,10 @@ def test_create_featured_items():
         live=True,
         in_progress=True,
     )
+    in_progress_course.enrollment_start = further_past_date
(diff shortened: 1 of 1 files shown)


annagav (approved):
Test pass every time 👍

PR #2228, opened 2024-05-29
Adding topics to course and program api

annagav:
Checks off second box https://github.com/mitodl/hq/issues/4110

Adding course topics to course and program api
Screenshot 2024-05-30 at 10 01 22 AM

### How can this be tested?
Go to http://mitxonline.odl.local:8013/api/v2/courses/
http://mitxonline.odl.local:8013/api/v2/programs/

### DIFF
--- a/courses/admin.py
+++ b/courses/admin.py
@@ -231,6 +231,7 @@ class CourseRunEnrollmentAdmin(AuditableModelAdmin):
         "get_run_courseware_id",
         "enrollment_mode",
         "change_status",
+        "created_on",
     )
     raw_id_fields = (
         "user",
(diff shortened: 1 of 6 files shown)


JenniWhitman (commented):
Shows as expected. 5 more queries isn't ideal, but at least we're keeping track of it. 

Do you think the 5 is because there are 2 course pages being retrieved there and 3 topics? I'm trying to pinpoint what hte 5 is and what the potential increase is for more.

JenniWhitman:
So, the current conftest doesn't actually set any topics, I'm doing that real quick to see how much the impact is. It's possible we can do a prefetch or... something. likely.

JenniWhitman (approved):
OK, I just manually double checked and the addition of additional topics doesn't affect the queries further. Performs as expected locally. I'm going to approve.

PR #2229, opened 2024-05-30
Repair test based on dates (test_create_featured_items) 

collinpreston:
NA

Repairs a flaky test (test_create_featured_items) that is dependent upon Course and CourseRun dates when filtering.

### How can this be tested?
Unit tests should always pass.


annagav (approved):
Looks good 👍

PR #2232, opened 2024-06-03
Fix Featured Items on Homepage

JenniWhitman:
Updates https://github.com/mitodl/hq/issues/3017

Presently, behind a feature flag, the featured items for the homepage are generated and stored in a redis cache. On page load, we check that those courses are still valid and were going from 22 to 7(ish) courses.  This is because the end date was not allowed to be null in the cms.models check.  I've updated this as well as added a test to determine that we're showing courses when we should (and not when we should not)

### How can this be tested?
Have any number of courses that are available to be selected.  
Run the `create_featured_items` management command.  
Do a .get on the redis cache to check how many are stored. - 
```
from django.core import caches
caches["redis"].get("CMS_homepage_featured_items")
```

Load the home page, ensuring that you are in dev/the feature flag is on. Those items in the cache should all appear in the featured carousel.


annagav (approved):
👍

PR #2233, opened 2024-06-03
4244: no error message is shown when education level is not specified in additional details modal

collinpreston:
https://github.com/mitodl/hq/issues/4244
https://github.com/mitodl/hq/issues/4272

Documentation: https://docs.google.com/document/d/1W7jYty8HjkaYxSFXoUMF9ZYTOp8eJ4ssmINfgLhQ61Q/edit#heading=h.ie7st6eu9m5w

### How can this be tested?
Go through each form in the application and verify the following:
- Accessibility with a screen reader.
- https://github.com/mitodl/hq/issues/4244 is resolved.
- Requirements outlined in documentation are defined clearly and addressed: https://docs.google.com/document/d/1W7jYty8HjkaYxSFXoUMF9ZYTOp8eJ4ssmINfgLhQ61Q/edit#heading=h.ppw15bxqq3n7

### DIFF
--- a/frontend/public/src/lib/validation.js
+++ b/frontend/public/src/lib/validation.js
@@ -5,7 +5,8 @@ import * as yup from "yup"
 
 export const passwordFieldRegex = /^(?=.*[0-9])(?=.*[a-zA-Z]).{8,}$/
 
-const newAndConfirmPasswordMatchErrorMessage = "New password and Confirm Password must match."
+const newAndConfirmPasswordMatchErrorMessage =
+  "New password and Confirm Password must match." // pragma: allowlist secret
 
 export const passwordFieldErrorMessage =
   "Password must be atleast 8 characters and contain at least one letter and number."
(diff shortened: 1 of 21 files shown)


frontend/public/src/lib/validation.js
```diff
@@ -3,65 +3,77 @@ import * as yup from "yup"
 
 // Field validations
 
-export const passwordFieldRegex = "^(?=.*[0-9])(?=.*[a-zA-Z]).{8,}$"
+export const passwordFieldRegex = /^(?=.*[0-9])(?=.*[a-zA-Z]).{8,}$/
 
-export const passwordFieldErrorMessage =
-  "Password must be atleast 8 character and contain at least one letter and number."
+const newAndConfirmPasswordMatchErrorMessage =
+  "New password and Confirm New Password must match." // pragma: allowlist secret
```
JenniWhitman, line 9: I think I'm completely missing why this is counting as a secret?

JenniWhitman: (I know you might not know either - just confused)

JenniWhitman (approved):
I ran through the forms trying to break input as much as I could, couldn't find any that didn't respond as I expected. The only thing I have on here is a comment re: why it's counting something as a secret which you may or may not know - that's for my knowledge. Otherwise this looks good to go.

PR #2235, opened 2024-06-04
Fixing CoursePage related name overlap

annagav:
Related to https://github.com/mitodl/mitxonline/pull/2228

Fixing naming overlap between `Course.page` as property and as a related name to `CoursePage`

### How can this be tested?
Create a course without a course page.
Go to `http://mitxonline.odl.local:8013/api/v2/courses/`. Make sure it works.

### DIFF
--- a/courses/models.py
+++ b/courses/models.py
@@ -549,9 +549,9 @@ class Course(TimestampedModel, ValidateOnSaveMixin):
         return self.readable_id.split("+")[-1]
 
     @cached_property
-    def course_page(self):
+    def page(self):
         """Gets the associated CoursePage"""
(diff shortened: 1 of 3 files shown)


JenniWhitman (changes requested):
https://github.com/mitodl/mitxonline/pull/2236

I think the above PR addresses it better - There is a property coursepage adn an attribute coursepage. The above distinctly references the atttribute when building the serializer rather than the property and does so using getattr (the preferred method for checking if an attribute is present). If this makes sense, can merge into your branch and test.

JenniWhitman (approved):
Approved based on changes.

PR #2240, opened 2024-06-07
Update postgres

collinpreston:
https://github.com/mitodl/hq/issues/4418

Update postgres

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,7 +29,7 @@ services:
     ports:
       - "5432"
     environment:
-      POSTGRES_PASSWORD: postgres
+      POSTGRES_PASSWORD: postgres # pragma: allowlist secret
 
   redis:
     image: redis:6.0.5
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -18,16 +18,18 @@ x-environment:
...
   - "host.docker.internal:host-gateway"
   - "local.edly.io:host-gateway"
 
 services:
   db:
-    image: postgres:11.6
+    image: postgres:15.7
     ports:
       - "5432"
+    environment:
```
JenniWhitman, line 31: I was unable to run this as is running `docker compose build` then `docker compose up`. I had to add a volume:
```
    volumes:
      - db-data:/var/lib/postgresql/data
```
to the docker-compose.yml in order to get my local to not look at the previous data. I'm not sure if we want to note that or add it into the docker compose, but when I take it out, mine looks back at the old data directory.

JenniWhitman (changes requested):
I'm going to suggest we add in the volume for postgres to avoid any silliness, but I'm open to discussion since you said yours upgraded without a hitch. Maybe have someone else try as well and see the result?

JenniWhitman (approved):
Works for me!

PR #2243, opened 2024-06-10
Populate profile form occupation checkbox with saved data from user profile

collinpreston:
NA

This PR fixes an issue on the edit profile and additional details forms where the "Are you a" checkboxes were not being populated with the user's saved profile information.  The form would still save the form correctly.

### How can this be tested?
Visit the edit profile page or additional details form, documented here: https://docs.google.com/document/d/1W7jYty8HjkaYxSFXoUMF9ZYTOp8eJ4ssmINfgLhQ61Q/edit?pli=1#heading=h.hi2minfkuu1y
Check or uncheck one of the checkboxes on the form.  Submit the form.  After the page reloads, verify that the changed checkbox represents the change made prior to submitting the form.


annagav (changes requested):
I don't think this is caused by this PR but could we also fix this:
Screenshot 2024-06-11 at 11 08 05 PM

annagav (approved):
👍

PR #2244, opened 2024-06-10
Filter relevant course runs to be live

annagav:
Fix https://github.com/mitodl/hq/issues/4493

Filter relevant course runs to be live

### How can this be tested?
Test course about page,for courses that might have a no runs open for enrollments, and runs that are open but no longer live. Both cases should show the same view. They should not appear in the catalog.
Nothing else should break.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -1204,10 +1204,10 @@ class CoursePage(ProductPage):
 
     def get_context(self, request, *args, **kwargs):
         relevant_run = get_user_relevant_course_run(
-            course=self.product, user=request.user
+            course=self.product
         )
         relevant_runs = list(
-            get_user_relevant_course_run_qset(course=self.product, user=request.user)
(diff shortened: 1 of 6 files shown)


collinpreston (commented):
Just curious, why was the filter (`.exclude(live=False)`) placed in `get_user_relevant_course_run_qset` and not inhttps://github.com/mitodl/mitxonline/blob/0f7d16b453c28650c86bf73ea8903876c63248a7/courses/models.py#L749 ?

annagav:
@collinpreston So the current definition of is_enrollable is not comprehensive enough, there are few more checks that could be added to it. 

For now I removed some unnecessary annotations from relevant course runs and refactored the code a bit.

I think we could put an effort into creating a universal filter for "valid", "enrollable" "live" run, that would be used in the catalog and about page.

collinpreston (approved):

PR #2246, opened 2024-06-10
GA4 Data Layer Push for purchase events

JenniWhitman:
Updates https://github.com/mitodl/hq/issues/4099

This PR does 3 things - 

- Adds a feature flag for the GA work
- Updates the repo to work with GA4 as GTM now goes through GA4, the two are no longer separate entities entirely
- Adds data to the transaction to send forward

### How can this be tested?
Ensure that with the flag off, there's no addition to the purchase page for the purchase event on teh interstitial page.  
If you start a purchase and check the interstitial page (or just get the context of that view programmatically), while the flag is off, the ga template variable should be false and the payload should be none.

Data won't flow correctly in GA until RC, so for now this is mostly to make sure nothing is scathingly wrong + This won't affect data and start sending things wildly.

### Additional Context
Structure of data:
```
event:     "Purchase",
        ecommerce: {
          purchase: {
            transaction_id: this.orderReceipt.reference_number || "", // order or transaction id
            value:          this.orderReceipt.total_price_paid || 0.0, // total purchase value excluding discounts
            tax:            0.0,
            shipping:       0.0,
            currency:       "USD",
            coupon:         discounts[0].discount_code || "",
            items:          [{
              item_id:       cartItem.product.id,
              item_name:     cartItem.description,
              affiliation:   "MITx Online", // always MITx Online
              discount:      discountedPrice,
              item_category: cartItem.product.item_category,
              price:         cartItem.price,
              quantity:      1
            }]
      }
        }
```


annagav (approved):
LGTM 👍

Issue #2201, opened 2024-05-10
Impossible to search for product by course run


annagav:
If you need to search for a product you can't search it by title or by readable_id.

<img width="1592" alt="Screenshot 2024-05-10 at 12 55 08 PM" src="https://github.com/mitodl/mitxonline/assets/7574259/773dc08e-7ce1-4aca-b117-5695c56a8137">

PR #2247, opened 2024-06-11
Fix Django Admin search for Products

collinpreston:
https://github.com/mitodl/mitxonline/issues/2201
https://github.com/mitodl/mitxonline/issues/1895

Fixes the Django Admin search for Products

### How can this be tested?

1. Have an instance of MITx Online with a Product created. 
2. Visit mitxonline.odl.local:8013/admin/ecommerce/product/
3. Use the search bar to search for your Product.  Verify that your Product is displayed if the search matches.
4. Use the search bar to search for a non-existing Product (random string that doesn't match any Product string).  Verify that your Product is not displayed.

### DIFF
--- a/ecommerce/admin.py
+++ b/ecommerce/admin.py
@@ -48,9 +48,9 @@ class ProductAdmin(VersionAdmin):
     model = Product
     search_fields = (
         "description",
-        "courseruns__title",
+        "courserunproducts__title",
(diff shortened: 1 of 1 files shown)


JenniWhitman (approved):
I was able to search and get results as I expected.

Issue #1855, opened 2023-09-01
Remove CMS FacultyBlock model


collinpreston:
The CMS FacultyBlock model is not being used anywhere in the application and should be removed.

PR #2248, opened 2024-06-11
Remove FacultyBlock class

collinpreston:
https://github.com/mitodl/mitxonline/issues/1855

Remove FacultyBlock class


JenniWhitman (approved):
Did a quick check and it's not in use anywhere and my app still loads and runs as expected, good to go.

PR #2275, opened 2024-07-02
Clean up constants

collinpreston:
NA

Removes and uses some constants.


JenniWhitman (approved):
Looks good.

Issue #2225, opened 2024-05-24
Enrollment issue when user unenrolls and enrolls again and pays


annagav:
Related to this issue https://github.com/mitodl/hq/issues/4367

the user has enrolled in audit then unenrolled from the run, then they enrolled in verified. As a result their status in the courseware is verified, but active is False.
### Expected Behavior
When user pays and enrolls in verified track the enrollment in edx should be active.

### Current Behavior
The user can't access the course bc their enrollment is not active.

collinpreston:
@annagav, where you able to reproduce this issue locally or in RC?

annagav:
no, I don't think so

collinpreston:
I have not been able to reproduce this bug following the steps in this issue.

annagav:
If a user has paid for a course run and they want to unenroll:

1. Don't allow. Need to contact CS.
2. Allow to unenroll. Decide what to do with the payment: Order, PaidCourseRun, etc.

Issue #1797, opened 2023-08-04
Enrollment is not activated in edX once you unenroll from a paid mode and then re-enroll


arslanashraf7:
### Steps to Reproduce

- Enroll yourself in a paid mode of a course
- Unenroll yourself through MITxOnline `/dashboard`
- Re-enroll
- You would see enrollment in the MITxOnline dashboard but the enrollment in edX would still be inactive.

### Expected Behavior
- Enrollment in edX should be activated again when a user re-enrolls in a course.

### Actual Behavior
- Enrollment in edX is inactive, User sees `Enroll now` button when they open the course through the MITxOnline dashboard

PR #2300, opened 2024-07-17
When creating enrollment set is_active explicitly in the enrollment api call

annagav:
Fix https://github.com/mitodl/mitxonline/issues/2225 and https://github.com/mitodl/mitxonline/issues/1797

If use enrolls, unenrolls, and enrolls again, the course run enrollment in edx should be active, otherwise the user can't access the course on edx, just like it happened for this user https://github.com/mitodl/hq/issues/4367.

On open edx `is_active = False` is equivalent to being unenrolled.

### How can this be tested?
Locally you can test if by enrolling in a course for an audit track, then unenrolling, then paying again. You can pay with a coupon too.
The api would get called again. This double call of create enrollment api is going to set is_active to true on repeated call.

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -668,6 +668,7 @@ def enroll_in_edx_course_runs(
                     mode=mode,
                     username=username,
                     force_enrollment=force_enrollment,
+                    enrollment_attributes={"is_active": True},
                 )
             results.append(enrollment)
         except HTTPError as exc:  # noqa: PERF203
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -668,8 +668,14 @@ def enroll_in_edx_course_runs(
                     mode=mode,
                     username=username,
                     force_enrollment=force_enrollment,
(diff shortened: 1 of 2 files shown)


pdpinch:
Why?

annagav:
> Why?

If user re-enrolls their enrollment stays inactive and they can't access the course through the courseware.

cp-at-mit (approved):

annagav:
So just adding here more clarification on the issue.

The enrollment stays inactive only following a "verified" enrollment as a result of payment for the course run.

If re-enroll in as audit again, then is works correctly.

annagav:
So I have been playing with the api and what I found is this:
Set up: There is an existing enrollment that has `is_active=False`( meaning the user unenrolled).
Action: 
If I make an api ( `edx_client.enrollments.create_student_enrollment`)call with the same `MODE` is_active gets set to True.
If I make the same api call but the MODE is not the same (changing from `audit` to `verified` or vice versa), then the MODE gets updated but `is_active` stays False.

Issue #2197, opened 2024-05-09
Requests enrollments if user is not authenticated


annagav:
Requests enrollments if user is not authenticated
### Expected Behavior
If I am not logged in it shouldn't request enrollments.

### Current Behavior
Throws an error.

### Steps to Reproduce
<img width="1762" alt="Screenshot 2024-05-09 at 9 14 48 AM" src="https://github.com/mitodl/mitxonline/assets/7574259/c362f9ad-2f1d-4512-b11d-ef3a03064a58">

PR #2304, opened 2024-07-19
Info box slow render fixes

cp-at-mit:
https://github.com/mitodl/hq/issues/3800

Removes a bunch of unused code.  Also removed the enrollment and course run API calls made from the about page since the course and course run information is all contained in the course API call and the enrollment API results were not being used.

The removal of the enrollments API call might also resolve this issue: https://github.com/mitodl/mitxonline/issues/2197

### How can this be tested?

1. The course about page should perform all functions as expected.  This should be tested with one and multiple course runs in different states.


annagav (approved):
👍

PR #2308, opened 2024-07-25
Add availability to Courses API endpoint

JenniWhitman:
Closes https://github.com/mitodl/hq/issues/4994

Adds availability to the courses API endpoint. If a course has any dated run open for enrollment, API will show `dated` for that course. If there are only self paced courses, it will show `anytime`

### How can this be tested?
Hit the endpoint with courses with multiple courseruns. Some good date variability would be to have some runs open for enrollment and some not. Having variability in whether or not they are self-paced will show if the changes meet the need as well.

### DIFF
--- a/courses/utils.py
+++ b/courses/utils.py
@@ -187,6 +187,7 @@ def get_archived_courseruns(queryset):
     """
     now = now_in_utc()
     return queryset.filter(
-        Q(end_date__lt=now)
+        get_enrollable_course_run_filter(now)
+        & Q(end_date__lt=now)
         & (Q(enrollment_end__isnull=True) | Q(enrollment_end__gt=now))
     )
--- a/courses/utils.py
+++ b/courses/utils.py
@@ -173,3 +173,20 @@ def get_unenrollable_courses(queryset):
         .filter(courseruns__id__in=courseruns_qs.values_list("id", flat=True))
         .distinct()
     )
+
+
+def get_archived_courseruns(queryset):
+    """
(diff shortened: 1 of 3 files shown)


courses/utils.py
```diff
@@ -173,3 +173,20 @@ def get_unenrollable_courses(queryset):
         .filter(courseruns__id__in=courseruns_qs.values_list("id", flat=True))
         .distinct()
     )
+
+
+def get_archived_courseruns(queryset):
+    """
+    Returns course runs that are archived. This is defined as:
```
annagav, line 180: An archived course is a course that is enrollable, end date has passed.

JenniWhitman: It was meant to say end. It matches the logic in frontend/public/src/lib/courseApi.js this way. If we think it should differ, it needs to be changed in both.

courses/utils.py
```diff
@@ -173,3 +173,21 @@ def get_unenrollable_courses(queryset):
         .filter(courseruns__id__in=courseruns_qs.values_list("id", flat=True))
         .distinct()
     )
+
+
+def get_archived_courseruns(queryset):
```
annagav, line 178: Could we add a test for this new function?

sovsey: @annagav Do you think we could consider merging this as is, and do the test as a follow-up? It would be great to see this availability update come through on MIT Learn.

annagav (approved):

PR #2310, opened 2024-07-25
Fix GA4 output for purchase

JenniWhitman:
https://github.com/mitodl/hq/issues/4099

Moves to a safer location on the page and changes how prices are formatted

### How can this be tested?
Make sure you can get ot the purchase processor without issue


annagav (approved):
👍

PR #2321, opened 2024-07-31
Add prerequisites to courses API

cp-at-mit:
https://github.com/mitodl/hq/issues/4947

Adds a required_prerequisites field to the courses api

### How can this be tested?
Make a GET call to the courses API for a course with nothing defined in the prerequisite field of the course page in the CMS.  Verify that the required_prerequisites in the API response is false.

Make a GET call to the courses API for a course with text defined in the prerequisite field of the course page in the CMS.  Verify that the required_prerequisites in the API response is true.

### DIFF
--- a/courses/serializers/v1/courses.py
+++ b/courses/serializers/v1/courses.py
@@ -30,7 +30,7 @@ class CourseSerializer(BaseCourseSerializer):
         """Get next run id"""
         run = instance.first_unexpired_run
         return run.id if run is not None else None
-    
+
     def get_required_prerequisites(self, instance):
         """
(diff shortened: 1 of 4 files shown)


annagav:
@cp-at-mit why are we adding this v1 and not v2?

cp-at-mit:
> @cp-at-mit why are we adding this to v1 and not v2?

I can move it to V2 if you think I should

annagav:
As far as I know mit-open is using v2 of the course api.

annagav (approved):
LGTM 👍

PR #2325, opened 2024-08-05
Rectivate enrollments for Differential Calculus

annagav:
Fix problem with https://github.com/mitodl/hq/issues/3780

I accidentally deactivated enrollments for the wrong program.

### How can this be tested?
Make sure that the readable id of the program matches the one in the issue.
which is program-v1:MITxT+18.03x


JenniWhitman (approved):
This will indeed activate enrollments in 18.01 and deactivate in 18.03

PR #2331, opened 2024-08-06
MITx Online receipt displays wrong date

cp-at-mit:
https://github.com/mitodl/hq/issues/5106

Changes the order page to display the datetime for orders using the update_on value instead of the created_on value for the associated order.

### How can this be tested?

1. Begin to purchase a Course.
2. Take a minute or two to complete the payment steps.
3. Verify the order page displays the update_on value which should be equal to the time when you completed the payment steps and were redirected back to MITx Online.


annagav (approved):
👍

PR #2332, opened 2024-08-06
Filtering only enrollable runs on the product page

annagav:
Fix https://github.com/mitodl/hq/issues/5098

In terms of logic there are three places where this is important:
1) I n deciding which "Enroll Now" button to show, (open a dialog or just submit a form)
2) in rendering the enrollment dialog drop down for selecting a run
3) Displaying "more dates" section

Everything should function as usual except for:
1. If there is only one run that is open for enrollment and is upgradable the Enroll Now button open a dialog where the user still has a select dropdown (instead of no dropdown). This makes the choice deliberate and makes the course dates visible.
2. An archived course is not chosen as `next_run_id` if there are other enrollable runs
3. Only enrollable runs show up in the "More Dates" section.
### How can this be tested?

Make sure the "Enroll now" button, the enrollment dialog and "more dates" behaves in a way that is expected.

### DIFF
--- a/courses/models.py
+++ b/courses/models.py
@@ -569,10 +569,11 @@ class Course(TimestampedModel, ValidateOnSaveMixin):
     @cached_property
     def first_unexpired_run(self):
         """
-        Gets the first unexpired CourseRun associated with this Course
+        Gets the first unexpired/enrollable CourseRun associated with this Course. Giving preference to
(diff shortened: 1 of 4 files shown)


JenniWhitman (approved):
Dates make sense for what I have locally.

PR #2354, opened 2024-08-21
Add get_dated_courseruns function for availability to check for dated courses

JenniWhitman:
Fixes https://github.com/mitodl/hq/issues/5260

Rather than relying on the adverse, this PR adds a gunction, get_dated_courseruns, to get any dated, enrollable courseruns for the course being serialized.  I've also added a test to this effect. This will generate the correct availability as we defined dated as having 1 or more dated courseruns, regardless of whether they also had archived or self-paced courses.

### How can this be tested?
Create courseruns that are both enrollable and not, self-paced and not and you should still see dated as long as one is both enrollable and not self-paced.

### DIFF
--- a/courses/utils.py
+++ b/courses/utils.py
@@ -201,11 +201,7 @@ def get_dated_courseruns(queryset):
     - End date can be dated and greater than now or null
     - Enrollable (enrollment start is in the past and enrollment end is in the future or null)
     """
-    now = now_in_utc()
-    return queryset.filter(
-        Q(is_self_paced=False)
-        & Q(start_date__isnull=False)
-        & (Q(end_date__isnull=True) | Q(end_date__gt=now))
-        & (Q(enrollment_end__isnull=True) | Q(enrollment_end__gt=now))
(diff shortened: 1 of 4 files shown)


courses/utils.py
```diff
@@ -191,3 +191,21 @@ def get_archived_courseruns(queryset):
...
+def get_dated_courseruns(queryset):
+    """
+    Returns course runs that are dated meaning they are
+    - Not self-paced
+    - Have a start date
+    - End date can be dated and greater than now or null
+    - Enrollable (enrollment start is in the past and enrollment end is in the future or null)
+    """
+    now = now_in_utc()
+    return queryset.filter(
```
annagav, line 205: could we just call `get_enrollable_courseruns_qs` and then add two requirements: (is_self_paced=False) and (enrollment_end__isnull=True).

annagav (changes requested):
So that we not writing many filters over and over let's reuse the enrollable run filter.

cp-at-mit (approved):

annagav (approved):

Issue #2357, opened 2024-08-23
Catalog shows "Started November 9th" when course is actually archived


annagav:
Here is the course https://mitxonline.mit.edu/courses/course-v1:MITxT+24.118x/

Here is the Catalog: https://mitxonline.mit.edu/catalog/courses/linguistics-and-philosophy

Instead the catalog should say "Start Anytime"

annagav:
If the course is self-paced then we show "Start anytime". If the course is archived but is not self-paced we show the start date.

pdpinch:
@annagav can I get some more detail here about what the issue was and how it was fixed? This is for our future selves, when a question comes up about this in the future.

PR #2359, opened 2024-08-23
Catalog card should show Start Anytime if the course is archived

annagav:
Fix https://github.com/mitodl/mitxonline/issues/2357

This PR changes the front end to use a single source of is_archived and makes the catalog display "Start Anytime" even when the course is not `is_self_paced=True'
Is archived is used in many places in the frontend, but it's value is not dynamic, so it's better to compute it in the backend.

Screenshot 2024-08-23 at 6 26 39 PM

### DIFF
--- a/courses/serializers/v1/base.py
+++ b/courses/serializers/v1/base.py
@@ -35,6 +35,10 @@ class BaseCourseSerializer(serializers.ModelSerializer):
 
 class BaseCourseRunSerializer(serializers.ModelSerializer):
     """Minimal CourseRun model serializer"""
+    is_archived = serializers.SerializerMethodField()
+
+    def get_is_archived(self, instance):
+        return instance.is_enrollable and instance.enrollment_end is None and instance.is_past
 
     class Meta:
         model = models.CourseRun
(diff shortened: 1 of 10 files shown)


courses/serializers/v1/base.py
```diff
@@ -36,6 +36,15 @@ class Meta:
 class BaseCourseRunSerializer(serializers.ModelSerializer):
     """Minimal CourseRun model serializer"""
 
+    is_archived = serializers.SerializerMethodField()
+
+    def get_is_archived(self, instance):
+        return (
+            instance.is_enrollable
+            and instance.enrollment_end is None
```
JenniWhitman, line 44: In the [`get_archived_courseruns` function,](https://github.com/mitodl/mitxonline/pull/2359/files#diff-1891631991481dca5fef7d88592f36f41fe1422ddc3f13e172d8973bc77c8da6R181) we defined enrollment end as able to be null or future.  Here we're ONLY defining it as None. Is this the case? Should the other function be updated? Should this one?

annagav: that condition is not needed there at all.

JenniWhitman (approved):
Good to go

Issue #2137, opened 2024-03-22
Allow self paced courses to have no upgrade deadline


annagav:
As of now if a course offers certificate we show the upgrade deadline in the Course info box and in the upgrade dialog.

For self paced courses the upgrade deadline might not make sense.

Current Behavior:

If upgrade deadline is not set on a course we treat it as if upgrade is not possible.

Desired Behavior:

if the course is self paced, has a product, but no upgrade deadline -> allow users to upgrade (and hide any deadline warnings)

pdpinch:
All of the courses in production currently have an upgrade deadline, so this isn't a high priority fix. 

Peter will bring this up with the MITx team to find out how likely it may be that a course doesn't have an upgrade dealine.

annagav:
If we don't have any implemented UX behavior in case `upgrade_deadline` is null, then we should make upgrade deadline a required field.

PR #2362, opened 2024-08-26
mitxonline dashboard crashes when courserun has no upgrade deadline

cp-at-mit:
https://github.com/mitodl/hq/issues/5284

mitxonline dashboard crashes when courserun has no upgrade deadline

### How can this be tested?

1. Create a course run with no upgrade deadline defined and course run enrollment for that course run.
2. Visit the dashboard and verify that the page loads correctly.
3. Update the course run to have an upgrade deadline in the future.
4. Visit the dashboard and verify that the page loads correctly.


cp-at-mit:
> @cp-at-mit I think this might be a related issue #2137. Does this PR fix this issue?

Yeah it sounds like this would resolve that requirement

annagav (changes requested):
I think this PR should make "upgrade deadline" be required field on the CourseRun model.

cp-at-mit:
> I think this PR should make "upgrade deadline" be required field on the CourseRun model.

Are there CourseRuns with no upgrade deadline?

cp-at-mit:
> @cp-at-mit are there there many course runs in production that are missing it? What type of courses are those?

I think I saw 8 courses without an upgrade deadline.  I think they are just normal courses.

annagav:
So all the upgrade_deadline does is that it prevents a user from enrolling in verified more after the deadline. So if there is no deadline I am assuming no verified mode is available for the course. So we can set the missing upgrade deadline values to any date in the past or now.

pdpinch:
> I think this PR should make "upgrade deadline" be required field on the CourseRun model.

While our current business rule is that all courses should offer certificates or sale and therefore have an upgrade deadline, I'm reluctant to add a validation for this. 

Where would it be applied? How will it complicate the course configuration process, which is already quite messy? What do we do if the upgrade deadline hasn't been determined at the time of courserun creation? And does it tie our hands for offering new types of courses in the future, like SPOCs (small private courses) or other courses without certificates?

annagav (approved):
👍

PR #2414, opened 2024-10-03
fix custom 404 template

cp-at-mit:
https://github.com/orgs/mitodl/projects/21/views/6?pane=issue&itemId=81655244

Viewflow was overriding the 404 error page which then overrided our custom 404 error page.  This PR move the viewflow library lower in on the list of installed apps in the settings.py file, which will allow for our custom 404 error page to take precedence over the 404 error page provided by viewflow. 

### How can this be tested?
Set `DEBUG` equal to `False` in your .env file as well as the docker-compose.yml file.
Restart your application and visit a page like http://mitxonline.odl.local:8013/fake.
Verify that the MITx Online custom 404 page is displayed and not the 404 page from viewflow (you can currently see the 404 page from viewflow if you visit https://rc.mitxonline.mit.edu/fake.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -193,7 +193,7 @@
         "filename": "main/settings.py",
         "hashed_secret": "09edaaba587f94f60fbb5cee2234507bcb883cc2",
         "is_verified": false,
-        "line_number": 955
+        "line_number": 954
       }
     ],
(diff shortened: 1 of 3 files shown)


annagav (approved):
LGTM 👍

PR #2417, opened 2024-10-08
Validate user email on account creation

annagav:
Fix https://github.com/mitodl/hq/issues/5642, https://github.com/mitodl/hq/issues/5931

Adds email as one of the fields to validate.

### How can this be tested? 

Retire an existing user, then try to create an account with the same email address but different username.

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -886,7 +886,7 @@ def unsubscribe_from_edx_course_emails(user, course_run):
     return result
 
 
-def validate_username_with_edx(username, email):
+def validate_username_email_with_edx(username, email):
     """
     Returns validation message after validating it with edX.
 
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -886,12 +886,13 @@ def unsubscribe_from_edx_course_emails(user, course_run):
     return result
 
 
-def validate_username_with_edx(username):
+def validate_username_with_edx(username, email):
     """
     Returns validation message after validating it with edX.
 
     Args:
         username (str): the username
+        email (str): the email
 
     Raises:
         EdxApiRegistrationValidationException: Raised if response status is not OK.
@@ -905,6 +906,7 @@ def validate_username_with_edx(username):
         edx_url(OPENEDX_REGISTRATION_VALIDATION_PATH),
         data=dict(  # noqa: C408
             username=username,
+            email=email,
         ),
     )
     if resp.status_code != status.HTTP_200_OK:
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -912,7 +912,7 @@ def validate_username_email_with_edx(username, email):
     if resp.status_code != status.HTTP_200_OK:
         raise EdxApiRegistrationValidationException(username, resp)
     result = resp.json()
-    return result["validation_decisions"]["username"]
+    return result["validation_decisions"]
 
 
 def bulk_retire_edx_users(usernames):
--- a/users/serializers.py
+++ b/users/serializers.py
@@ -241,7 +241,9 @@ class UserSerializer(serializers.ModelSerializer):
         email = data.get("email")
         if username:
             try:
-                openedx_validation_msg_dict = validate_username_email_with_edx(username, email)
+                openedx_validation_msg_dict = validate_username_email_with_edx(
+                    username, email
+                )
             except (
                 HTTPError,
                 RequestsConnectionError,
@@ -249,7 +251,10 @@ class UserSerializer(serializers.ModelSerializer):
             ) as exc:
                 log.exception("Unable to create user account", exc)  # noqa: PLE1205, TRY401
                 raise serializers.ValidationError(USER_REGISTRATION_FAILED_MSG)  # noqa: B904
-            if openedx_validation_msg_dict["username"] or openedx_validation_msg_dict["email"]:
+            if (
+                openedx_validation_msg_dict["username"]
+                or openedx_validation_msg_dict["email"]
+            ):
                 raise serializers.ValidationError(openedx_validation_msg_dict)
 
         return data
--- a/users/serializers.py
+++ b/users/serializers.py
@@ -17,7 +17,7 @@ from hubspot_sync.task_helpers import sync_hubspot_user
 from mail import verification_api
 from main.constants import USER_REGISTRATION_FAILED_MSG
 from main.serializers import WriteableSerializerMethodField
-from openedx.api import validate_username_with_edx
+from openedx.api import validate_username_with_edx, validate_username_email_with_edx
 from openedx.exceptions import EdxApiRegistrationValidationException
 from openedx.tasks import change_edx_user_email_async
 from users.models import ChangeEmailRequest, LegalAddress, User, UserProfile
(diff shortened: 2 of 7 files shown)


users/serializers.py
```diff
@@ -237,12 +241,17 @@ def validate(self, data):
...
             try:
-                openedx_validation_msg = validate_username_with_edx(username)
-                openedx_validation_msg = OPENEDX_USERNAME_VALIDATION_MSGS_MAP.get(
+                openedx_validation_msg = validate_username_email_with_edx(
+                    username, email
+                )
+                openedx_validation_msg = OPENEDX_VALIDATION_MSGS_MAP.get(
                     openedx_validation_msg, openedx_validation_msg
                 )
             except (
```
asadali145, line 257: While reviewing, A question came to my mind:

For reference, Let's consider:
STEP 1: Users enter an email and we send them the verification email.
STEP 2: Registration form with Name, Usernames, etc


- Why don't we check for the retired email at Step 1?
- With the current implementation, Users can enter a retired email and get an email to register (Pass Step 1). They will fill in the data on page 1 (Step 2) and then this validation will throw an error. Even though we don't ask for email in Step 2, it will throw an error.


@pdpinch What are your thoughts on validating the emails at Step 1 vs Step 2?

openedx/api.py
```diff
@@ -886,12 +886,13 @@ def unsubscribe_from_edx_course_emails(user, course_run):
     return result
 
 
-def validate_username_with_edx(username):
+def validate_username_email_with_edx(username, email):
```
asadali145, line 889: If there is a validation error on edX we are not returning that for the email at the end of this method. It only returns the msg for the username.

users/serializers.py
annagav: yes, I was thinking it makes more sense to check this in initial email validation.

annagav: However in STEP 1 we don't call edx api, and might cause additional issues, with adding extra calls....

annagav: I am hesitant to add the open edx validation api call into STEP 1 because in that step we just verify that the user actually own that email.
I have changed the messaging to come through with the actual validation error so the user will be informed why the account creation did not happen.

asadali145: I think we should confirm this Peter. I am not sure if we should tag him as he is no vacations.

pdpinch:
I'm not sure that I understand exactly what's happening here. Please confirm: 

1.  This PR adds a validation when a user is creating an account and submits their email address for the first time. It checks that the email address hasn't been retired before proceding to the next step of sending the validation email. 

2. For a future PR, I agree that it makes sense to also check with the open edX API that the email address is valid. Please open an issue for that work.

asadali145:
@pdpinch This PR adds a check with the Open edX API for email validation. My reservation is that we are not doing it in the first step before sending the verification email. Instead, we are validating the email in step 2.

Step 2 is where we get the User details and show a success message for email verification. This PR adds a check in this step.
Screenshot 2024-10-24 at 5 49 31 PM

pdpinch:
Thanks for the clarification Asad. 

We will soon be moving email validation to Learn/Keycloak (see https://github.com/mitodl/hq/issues/5664), so it's actually better to do this open edX check closer to Step 2 -- although I recognize that we'll probably end up adjusting this again anyway.

asadali145:
Thanks @pdpinch, I re-review it then.

users/serializers.py
```diff
@@ -252,9 +251,11 @@ def validate(self, data):
...
                 log.exception("Unable to create user account", exc)  # noqa: PLE1205, TRY401
                 raise serializers.ValidationError(USER_REGISTRATION_FAILED_MSG)  # noqa: B904
-
-            if openedx_validation_msg:
-                raise serializers.ValidationError({"username": openedx_validation_msg})
+            if (
+                openedx_validation_msg_dict["username"]
+                or openedx_validation_msg_dict["email"]
+            ):
+                raise serializers.ValidationError(openedx_validation_msg_dict)
```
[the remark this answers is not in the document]
annagav: We should definitely add the validation that at step 1 and show the validation message there.
At this step the user should not be getting validation messages about email. But I think it is still important to validate it here too. 
I am not too worried for the message not to make it to the user. The important change in this PR is not to allow account creation with the same email address.
I have modified the message to be under username field if you feel that is better than no message for now.

users/serializers.py
```diff
@@ -252,10 +254,19 @@ def validate(self, data):
...
+                        "username": OPENEDX_ACCOUNT_CREATION_VALIDATION_MSGS_MAP.get(
+                            openedx_validation_msg_dict["username"]
+                        )
+                    }
+                )
+            if openedx_validation_msg_dict["email"]:
+                # there is no email form field at this point, but we are still validating the email address
+                raise serializers.ValidationError(
+                    {"username": openedx_validation_msg_dict["email"]}
+                )
```
asadali145, line 269: If we do the below, the error message will be displayed in the banner at the top and not the username field. It is better to display this message in the banner.
```suggestion
            if openedx_validation_msg_dict["email"]:
                # there is no email form field at this point, but we are still validating the email address
                raise serializers.ValidationError(openedx_validation_msg_dict["email"])
```

users/serializers.py
```diff
@@ -239,11 +241,11 @@ def validate(self, data):
                 )
 
         username = data.get("username")
+        email = data.get("email")
```
asadali145, line 244: This is unused.

users/serializers.py
```diff
@@ -41,9 +41,11 @@
 USERNAME_ALREADY_EXISTS_MSG = (
     "A user already exists with this username. Please try a different one."
 )
+EMAIL_CONFLICT_MSG = "This email is already associated with an existing account"
```
asadali145, line 44:
```suggestion
EMAIL_CONFLICT_MSG = "This email is associated with an existing account. Please check your email or try a different one."
```

asadali145:
@annagav pre-commit check is failing. Also, I have left a couple of suggestions.

users/serializers.py
annagav: it is used in the api call.

asadali145: Yeah, my bad. My IDE gave a warning so I just commented. Not sure why. It seems good now

users/serializers.py
```diff
@@ -41,9 +41,13 @@
...
 )
+EMAIL_CONFLICT_MSG = (
+    "This email is associated with an existing account. Please try a different one."
+)
 
-OPENEDX_USERNAME_VALIDATION_MSGS_MAP = {
-    "It looks like this username is already taken": USERNAME_ALREADY_EXISTS_MSG
+OPENEDX_ACCOUNT_CREATION_VALIDATION_MSGS_MAP = {
+    "It looks like this username is already taken": USERNAME_ALREADY_EXISTS_MSG,
+    "This email is already associated with an existing account": EMAIL_CONFLICT_MSG,
```
asadali145, line 50: I think we should change this to make it simple:
```suggestion
    "username": USERNAME_ALREADY_EXISTS_MSG,
    "email": EMAIL_CONFLICT_MSG,
```

users/serializers.py
```diff
@@ -252,10 +256,17 @@ def validate(self, data):
...
+                raise serializers.ValidationError(
+                    {
+                        "username": OPENEDX_ACCOUNT_CREATION_VALIDATION_MSGS_MAP.get(
+                            openedx_validation_msg_dict["username"]
+                        )
+                    }
+                )
+            if openedx_validation_msg_dict["email"]:
+                # there is no email form field at this point, but we are still validating the email address
+                raise serializers.ValidationError(openedx_validation_msg_dict["email"])
```
asadali145, line 269: We wanted to display the custom message we defined instead of the message we are getting from Open edX API.

asadali145 (approved):
LGTM!

Just a couple of suggestions regarding the messages.

users/serializers.py
asadali145: And then this would have impact on the usage.

users/serializers.py
annagav: This message is never going to show to the user anyway since we want to add this validation to STEP 1. So I don't see a reason to change this. The message from edX API is informative and accurate.

asadali145: I am unsure if we will do that sooner as per this comment https://github.com/mitodl/mitxonline/pull/2417#issuecomment-2435255463. 

Also, we should have kept it consistent. Now we are displaying a custom message for the username but not for the email. Also, the custom message that we defined and added to the OPENEDX_ACCOUNT_CREATION_VALIDATION_MSGS_MAP is unused now.

annagav: This looks good to me:
Screenshot 2024-11-06 at 9 02 08 AM

PR #2419, opened 2024-10-16
Add required_prerequisites to program API

rachellougee:

https://github.com/mitodl/hq/issues/5751

Adding `required_prerequisites` to the program API

### How can this be tested?

Verifying `required_prerequisites` is true in `/api/v2/programs/id=1` if any course in the program (id=1) has required_prerequisites=true


annagav (approved):
👍

PR #2421, opened 2024-10-22
Don't show course and program ids in financial assistance emails

cp-at-mit:
https://github.com/mitodl/hq/issues/5762

Updates the flexible price email attribute that is used in the template for sending flexible price emails to learners.

### How can this be tested?
This will need to be tested on RC.  There is an issue when loading the staff-dashboard locally using the main branch: https://github.com/mitodl/hq/issues/5837.

### DIFF
--- a/flexiblepricing/mail_api.py
+++ b/flexiblepricing/mail_api.py
@@ -7,7 +7,6 @@ import logging
 from django.core.exceptions import ValidationError
 from mitol.mail.api import get_message_sender
 
-from courses.models import Course
 from flexiblepricing.constants import (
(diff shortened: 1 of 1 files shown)


rachellougee (approved):
The change to remove course and program IDs from the email looks good. I verified it by triggering the email locally using the shell command. 

In the email, the `International Development` track will still be specified for DEDP, as the DEDP program title is not generic (program_id=1,2). This is outlined under 'Expected Behavior' in the linked issue and https://github.com/mitodl/hq/issues/3901

image

Approving it since the title of the linked issue only suggests removing course and program IDs

PR #2422, opened 2024-10-22
add duration and time_commitment to program API

rachellougee:

https://github.com/mitodl/hq/issues/5698

Adding `duration` and `time_commitment` to the program API 

### How can this be tested?

Create a program and populate the length/duration and effort/time_commitment in the program CMS page. These fields should be populated via the endpoint /api/v2/programs/


annagav (approved):
👍

PR #2424, opened 2024-10-22
revert postgres upgrade

cp-at-mit:
Reverts https://github.com/mitodl/mitxonline/pull/2329/files

I initially thought the postgres upgrade worked fine but my local was caching things or something weird.  Anyway, upgrading to Postgres v16 without any modifications to our data does not work, so this PR reverts that upgrade.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,7 @@ x-extra-hosts:
 
 services:
   db:
-    image: postgres:16.4
+    image: postgres:15.8
     ports:
       - "5432"
     environment:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -67,8 +67,10 @@ services:
       << : *py-environment
       PORT: 8011
     env_file: .env
-    links:
+    depends_on:
```
annagav, line 70: This change is not in the initial upgrade PR. Is this by accident?

cp-at-mit: Oh yes, thank you for catching that and picking up the PR quickly!

annagav (approved):
👍

PR #2427, opened 2024-10-23
fix flaky test ordering

cp-at-mit:
NA

Fixes a flaky test that has delayed deployments.  The fix is done by ordering the items in the unit test to ensure that the lists are compared equally.

### How can this be tested?
Tests should pass.

### DIFF
--- a/ecommerce/models_test.py
+++ b/ecommerce/models_test.py
@@ -287,7 +287,9 @@ def test_product_managers():
     active_products = ProductFactory.create_batch(2, is_active=True)
 
     assert list(Product.objects.all().order_by("id")) == list(active_products)
-    assert list(Product.all_objects.all().order_by("id")) == list(inactive_products + active_products)
+    assert list(Product.all_objects.all().order_by("id")) == list(
+        inactive_products + active_products
+    )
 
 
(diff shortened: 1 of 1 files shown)


ecommerce/models_test.py
```diff
@@ -283,11 +283,13 @@ def test_product_managers():
...
-    in_active_products = ProductFactory.create_batch(3, is_active=False)
+    inactive_products = ProductFactory.create_batch(3, is_active=False)
     active_products = ProductFactory.create_batch(2, is_active=True)
 
-    assert list(Product.objects.all()) == active_products
-    assert list(Product.all_objects.all()) == in_active_products + active_products
+    assert set(Product.objects.all().order_by("id")) == set(active_products)
+    assert set(Product.all_objects.all().order_by("id")) == set(
+        inactive_products + active_products
+    )
```
rachellougee, line 292: This works but I think if we are ordering the `Product.objects.all().order_by("id")` then the list makes sense. Take it or leave it
```
assert list(Product.objects.all().order_by("id")) == list(active_products)
assert list(Product.all_objects.all().order_by("id")) == list(
      inactive_products + active_products
 )
```

rachellougee (approved):

PR #2430, opened 2024-10-25
fix: configure_instance command,

marslanabdulrauf:
Fix: configure_instance command, README updated to point new installation docs

- `configure_instance` command updated and `redirect_uris` fixed
- README.md updated to point to handbook's documentation
- All previous docs references updated to point to new handbook's documentation

### How to test it?
`docker-compose run --rm web ./manage.py configure_instance macos` look [here](https://github.com/mitodl/mitxonline/blob/main/docs/source/commands/configure_instance.rst) for more context on how to run it with different params

- [x]  README.md updated to point to new common documentation in [handbook](https://github.com/mitodl/handbook)
- [x] Fix Application creation in `configure_instance` management command


README.md
```diff
@@ -12,18 +12,9 @@ Run through those steps **including the addition of `/etc/hosts` aliases and the
 
 ### Configure mitxonline and Open edX
 
-**Open edX Tutor** can be used with MITx Online instead of the traditional devstack release. See [Local Open edX Tutor and MITx Online Deployment](docs/source/configuration/tutor.rst) for details.
-
-See [MITx Online Quick Start](docs/source/configuration/quickstart.rst) and [Configure Open edX](docs/source/configuration/open_edx.rst)
-
-### Configuring the CMS
```
jkachel, line 19: Why is this being removed? The configure_instance command doesn't configure the CMS and that still needs to be done.

jkachel (changes requested):
Had a comment about the CMS/Wagtail configuration change that seems out of scope for this PR. 

Additionally, if the Tutor and Devstack docs are being moved into the handbook, they should also be removed here. These are in `docs/source/configuration` and there are other docs in here that will need adjustment to point to the new locations for things. (This might just be the quickstart and the index but a quick scan in the `docs` folder for references should be done too.)

README.md
jkachel: The thing GH decided to feature is misleading - I'm referring to the block starting at line 19 (Configuring the CMS). The stuff above it is fine to remove.

marslanabdulrauf: we are configuring wagtail here https://github.com/mitodl/mitxonline/blob/main/main/management/commands/configure_instance.py#L201

jkachel: :+1: missed that looking through the management command code so this is fine

jkachel (approved):
LGTM 👍

PR #2501, opened 2024-12-23
Remove pagination for departments

annagav:
Fix https://github.com/mitodl/hq/issues/6308

Remove pagination for departments api

### How can this be tested?
Create over 12 departments with courses. Go to http://mitxonline.odl.local:8013/api/v2/departments/
You should see all departments on one page.

### DIFF
--- a/courses/views/v2/__init__.py
+++ b/courses/views/v2/__init__.py
@@ -108,7 +108,6 @@ class DepartmentViewSet(viewsets.ReadOnlyModelViewSet):
     """API view set for Departments"""
 
     serializer_class = DepartmentWithCoursesAndProgramsSerializer
-    pagination_class = Pagination
     permission_classes = []
 
     def get_queryset(self):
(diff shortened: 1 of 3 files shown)


pdpinch:
Oh, does this mean that Urban Studies is coming on the 2nd page?

Is removing the pagination the best solution? What about changing the page size? 

Granted, there is a practical limit to the number of departments. The largest possible list is this one, in Learn: https://learn.mit.edu/departments

annagav:
@pdpinch What is the use value for having pagination for departments? It seems redundant.  We want the whole list when rendering the Catalog page.

annagav:
> Oh, does this mean that Urban Studies is coming on the 2nd page?
Yes, and another department "Political Science". Both are missing from the Catalog Page.

annagav:
If 31 is largest number of Departments and we need all of the departments on first rendering I don't see a reason to have pagination.

pdpinch:
Ok, if you and Collin are in agreement that's good enough for me. 

What do you think about releasing this today? It's a pretty serious bug.

PR #2520, opened 2025-01-30
removing order from InstructorPageLink

annagav:
fix https://github.com/mitodl/hq/issues/6415

Removing the ordering of instructor link page. It doesn't do anything useful.

When the admin re-chooses instructors the order is unpredictable, or at least was unpredictable.

### How can this be tested?
1. Go to cms page. Start editing a course page by adding a bunch of instructors to the page. Check the order on the page.
2. Then go ahead and change some of the selection to another instructor. Check the order of instructors on the page again. (this used to mess up the order)

3. Then try deleting and adding them in the desired order.
Check that the order is as expected.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -1105,7 +1105,7 @@ class ProductPage(VideoPlayerConfigMixin, MetadataPageMixin):
         InlinePanel(
             "linked_instructors",
             label="Faculty Members",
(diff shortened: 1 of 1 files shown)


jkachel (approved):
Looks OK to me 👍

pdpinch:
Sorry, I don't understand. Does this make the order of instructors follow the order in the CMS?

annagav:
yes.

PR #2549, opened 2025-02-24
Open api

cp-at-mit:
https://github.com/mitodl/hq/issues/6791

Adds automatically generated OpenAPI specification documentation for many of the APIs in MITx Online.  The OpenAPI specification documentation is created using the drf-spectacular library.

This PR also removes some unused serializers

### How can this be tested?
Execute the following command from the root directory: `./scripts/generate_openapi.sh`

Verify the generated files.

### Additional Context
This PR does not add functionality to the Github actions to verify that the OpenAPI specification is up-to-date, as we currently do in other applications.  That will be added in a later PR.


jkachel (changes requested):
First run at this: as the app can now generate an OpenAPI client, we should have checks for it in CI. These should be sourceable from another project - there's some tasks that need to get added to `ci.yaml` itself and there's another script to pull. 

Still working through review, though; will have more comments later.

ChristopherChudzicki:
@cp-at-mit I was using the the schema in this PR to generate some types for work within MIT Learn (as a temporary measure for https://github.com/mitodl/hq/issues/6870, until we have the API client as a package).

I noticed that some of the types were wrong—e.g., the spec says string when it's actually a nullable string. I addressed some of these issues in this branch: https://github.com/mitodl/mitxonline/compare/open-api...open-api-cc-tweaks

BTW: The schema definitely doesn't need to be perfect—I expect that as we work on the dashboard, we'll discover issues, and make appropriate fixes then.

jkachel:
> Yeah I figured that should probably be done in a different PR but up to you.

The schema generation CI steps are pretty straightforward - shouldn't require any changes to work here so it'd save us an additional PR to get that in now.

jkachel (commented):
I went back and forth on this some - when I run the Spectacular commands to generate the spec, I get a bunch of errors and warnings (37 unique warnings and 17 unique errors specifically). Looking through the output, a lot of this seems to be because it's trying to pull in stuff that it shouldn't - there are Wagtail endpoints and login API stuff (including OIDC-specific things) that really don't belong in the client anyway. (There's no reason for the OIDC configuration endpoint to be in there, for instance.) 

However, I'm not sure if this is a problem with my local setup or what. I did review the configuration settings and everything generally looks OK, except that there's a bunch of APIs with versions `v1` and `v2` which aren't in the drf allowed list. (But changing that setting didn't change anything either.) So I'm not sure what the problem is. I will ping you later to discuss - I'm not convinced this is a problem with the setup (given that at least you were able to generate a spec).

cms/serializers.py
```diff
@@ -27,19 +31,23 @@ def get_feature_image_src(self, instance):
...
+    @extend_schema_field(serializers.URLField)
     def get_page_url(self, instance):
         return instance.get_url()
 
+    @extend_schema_field(str)
     def get_description(self, instance):
         return bleach.clean(instance.description, tags=[], strip=True)
 
+    @extend_schema_field(str)
     def get_effort(self, instance):
```
ChristopherChudzicki, line 43:
```suggestion
    def get_effort(self, instance) -> str | None:
```

typo annotations are good enough in simple cases, and this one should be nullable.

A downside of drf-spectacular is that we do need to manually provide some types, like here.

That said, we do a lot less manually than we would if we just wrote TS definitions for all the types in frontend files.

courses/serializers/v1/base.py
```diff
@@ -37,7 +38,16 @@ class BaseCourseRunSerializer(serializers.ModelSerializer):
     """Minimal CourseRun model serializer"""
 
     is_archived = serializers.SerializerMethodField()
-
+    courseware_url = serializers.URLField()
+    is_upgradable = serializers.BooleanField()
+    is_enrollable = serializers.BooleanField()
+    is_self_paced = serializers.BooleanField()
+    run_tag = serializers.CharField()
```
ChristopherChudzicki, line 45: I don't think you should need to specify `run_tag = serializers.CharField()` here. drf-spectacular should infer it from the model.

Additionally, specifying it here has some disadvantages—which is why i removed a bunch in https://github.com/mitodl/mitxonline/commit/efa29419304ccf3ef6c5924800211c4a48f5c23a. For example, if you set 
```python
start_date = serializers.DateTimeField()
```

here, then drf-spectacular will think `start_time` is required and non-nullable.

If you let drf-spectacular infer it from the model, then it can see `start_time` is nullable, and so correctly types the response property as nullable.

I'm not 100% sure, but I think that for model properties, a type annotation on the property is enough, at least in simple cases (int, bool, etc).

courses/serializers/v1/programs.py
```diff
@@ -151,6 +226,7 @@ class UserProgramEnrollmentDetailSerializer(serializers.Serializer):
     enrollments = CourseRunEnrollmentSerializer(many=True)
     certificate = serializers.SerializerMethodField(read_only=True)
 
+    @extend_schema_field(ProgramCertificateSerializer)
```
ChristopherChudzicki, line 229:
```suggestion
    @extend_schema_field(ProgramCertificateSerializer(allow_null=True))
```

courses/serializers/v2/courses.py
```diff
@@ -63,6 +66,7 @@ def get_duration(self, instance):
 
         return None
 
+    @extend_schema_field(str)
     def get_time_commitment(self, instance):
```
ChristopherChudzicki, line 70:
```suggestion 
    def get_time_commitment(self, instance) -> str | None:
```
looks like a bunch of other properties should have `| None` too. (or be Optional?)

ChristopherChudzicki (commented):
Left a few comments, some of which apply many places.

I agree w/ James btw, we don't need all the endpoints. The enrollment ones are particularly useful.

jkachel (commented):
Generated with no issues 🎉 

Not approving quite yet - need the spec check parts of it that we discussed in Slack, but once that's here it should be good to go.

jkachel (approved):
LGTM 👍

ChristopherChudzicki (approved):
Saw you did the Decimal change, did you want to add `CourseRunGradeSerializer(many=True)` ?  Or happy to improve the spec later.

Either way, 👍

PR #2550, opened 2025-02-24
Add OIDC login via social-auth

jkachel:

Closes mitodl/hq#6586

Configures the authentication pipeline to work with OIDC (Keycloak) based authentication.

A Keycloak instance is necessary for testing this.  See the `README-keycloak.md` for setup instructions. (It is not the same as it is elsewhere! Hopefully it's better/easier to follow.) The sample instance that exists elsewhere has been migrated here to make testing easier. Or, you can also use a different Keycloak instance, including one that is part of some other app (like Learn or Unified Ecommerce).

When completed and enabled, users will be given the option of logging in via email or Keycloak:
- Email login works as it did. Users can sign up, run forgot password, sign in via email, etc. all as before.
- Keycloak login will bounce the user over to Keycloak to authenticate, and then they will return to MITx Online. If the user's account doesn't exist, it will be created. Users will _not_ have to verify their email addresses in MITx Online. Users _will_ have to complete the MITx Online-specific profile loading process when they try to do certain operations (enroll, etc.).

We don't have design for this but the choice screen just gives you a couple buttons to direct the user.

This does _not_ depend on APISIX at all. However, if this is sitting behind a correctly configured APISIX instance and the user's already logged in, the user should just get bounced straight back from Keycloak. But the app will not pick up an existing APISIX session.

Login interstitial:

image

### How can this be tested?

- Set `EXPOSE_OIDC_LOGIN` to True in your `.env`.
- Configure the OIDC settings appropriately. See the `README-keycloak.md` for instructions.
- Follow the instructions for configuring the provided Keycloak instance _or_ configure the app to use an external one. _See the notes in `README-keycloak.md`_ if you're going to use a separate Keycloak instance on the same machine.
- Go to the app, and click Login.
- You should see the interstitial above.
- Clicking Sign In with Email should take you to the normal sign-in workflow - this should work as expected, with no changes.
- Clicking Sign In with Keycloak should redirect you to Keycloak to sign in. Upon returning to the app, you should be logged in. If your account didn't exist in MITx Online before, it should be created.
- Clicking Log Out should redirect you into Keycloak for logout.

### Additional Context

This doesn't tackle grabbing an APISIX session. Different middleware is needed for this.


users/management/commands/promote_user.py
```diff
@@ -0,0 +1,85 @@
...
+            "--demote",
+            action="store_true",
+            help="Demote the user from staff.",
+        )
+
+        parser.add_argument(
+            "--superuser",
+            action="store_true",
+            help="Promote the user to superuser.",
+        )
```
rhysyngsun, line 37: I would suggest structuring this using subparsers because it'll handle some of the validation you're doing below automatically. I recently did this in open-discussions for a similar type of command and it simplifies things substantially: https://github.com/mitodl/open-discussions/blob/master/authentication/management/commands/validate_user_emails.py#L30

frontend/public/src/containers/pages/login/LoginPages.js
```diff
@@ -26,7 +27,8 @@ const ForgotPasswordPages = () => (
 
 const LoginPages = () => (
   <Switch>
-    <Route exact path={routes.login.begin} component={LoginEmailPage} />
+    <Route exact path={routes.login.begin} component={LoginChoicePage} />
```
rhysyngsun, line 30: I don't know that we want to present UI to the user to choose which way to login, but it should instead be feature flagged as still being `LoginEmailPage` unless the flag is enabled. If the flag is enabled then the "Login" button should send them directly to Keycloak to login.

authentication/pipeline/user.py
```diff
@@ -30,6 +33,55 @@
...
+        raise RequireEmailException(backend, None)
+
+    if "email" in details and is_user_email_blocked(details["email"]):
+        raise EmailBlockedException(backend, None)
+
+    retval = create_user(strategy, details, backend, user, *args, **kwargs)
+
+    if retval.get("is_new"):
+        # the backend should be setting the is_active flag to True (and it seems
+        # to be doing that!) but it isn't so we need to set it to active here.
```
rhysyngsun, line 75: Once we switchover completely to keycloak, we can set the default on the `User` model to `True` and remove this code.

frontend/public/src/containers/pages/login/LoginPages.js
jkachel: Refactored this so there's no choice page. I left the Create Account button, but also set its URL to the login URL - thinking that should probably be collapsed into Sign In as a Sign In / Create Account but also feel like that needs some design input.

rhysyngsun (commented):
Code looks good, I need to do a bit of setup locally to be able to test

docker-compose.yml
```diff
@@ -174,8 +175,36 @@ services:
...
+    environment:
+      - KEYCLOAK_SVC_KEYSTORE_PASSWORD=${KEYCLOAK_SVC_KEYSTORE_PASSWORD:-supersecret123456}
+      - KEYCLOAK_SVC_HOSTNAME=${KEYCLOAK_SVC_HOSTNAME:-kc.odl.local}
+      - KEYCLOAK_ADMIN=${KEYCLOAK_SVC_ADMIN:-admin}
+      - KEYCLOAK_ADMIN_PASSWORD=${KEYCLOAK_SVC_ADMIN_PASSWORD:-admin}
+      - KEYCLOAK_PORT=${KEYCLOAK_PORT:-7080}
+      - KEYCLOAK_SSL_PORT=${KEYCLOAK_SSL_PORT:-7443}
+    ports:
+      - ${KEYCLOAK_PORT}:${KEYCLOAK_PORT}
+      - ${KEYCLOAK_SSL_PORT}:${KEYCLOAK_SSL_PORT}
```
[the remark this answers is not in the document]
jkachel: resolved in https://github.com/mitodl/mitxonline/pull/2550/commits/95dec4aa7ba6210907144c34f28063bf4900afab

authentication/views.py
```diff
@@ -163,6 +164,43 @@ def confirmation_sent(request, **kwargs):  # pylint: disable=unused-argument  #
...
+            string: The URL to redirect the user to in order to logout.
+        """
+        strategy = load_strategy()
+        storage = strategy.storage
+        user_social_auth_record = storage.user.get_social_auth_for_user(
+            user, provider=OlOpenIdConnectAuth.name
+        ).first()
+
+        if not user_social_auth_record:
+            return False
```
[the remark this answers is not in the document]
jkachel: resolved in https://github.com/mitodl/mitxonline/pull/2550/commits/647bf3268b426421a38a4197f85c8283c499d15f

PR #2551, opened 2025-02-24
Ag/upgrade python

annagav:
Related to https://github.com/mitodl/hq/issues/5951

Upgrade python to 3.12
uwsgi 

### How can this be tested?
Nothing should break

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.11.9 as base
+FROM python:3.12.9 as base
 LABEL maintainer "ODL DevOps <mitx-devops@mit.edu>"
 
 # Add package files, install updated node and pip
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10.16 as base
+FROM python:3.11.9 as base
 LABEL maintainer "ODL DevOps <mitx-devops@mit.edu>"
 
 # Add package files, install updated node and pip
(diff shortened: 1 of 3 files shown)


jkachel:
FWIW, I opened an issue in the upstream CyberSource repo for the `SyntaxWarning` issue we discussed: https://github.com/CyberSource/cybersource-rest-client-python/issues/143

I also figured out a workaround. You can use the `warnings` module to filter out `SyntaxWarning` when importing the payment_gateway library (or this can be migrated up to `ol-django` - that might be easier since less imports to deal with). This would obvs be temporary but it would make it work for the short term:

```python
import warnings

with warnings.catch_warnings():
	warnings.filterwarnings("ignore", category=SyntaxWarning)
	from CyberSource import ApiClient, OrdersApi
```
(this will need updates for the actual imports and such)

I tested this in a test app I did up (see the issue) and this was sufficient to make pytest stop failing.

pdpinch:
Need some help from @rhysyngsun on a hypothesis test error

rhysyngsun:
I poked at this for about an hour. An upgrade to `celery-redbeat` is needed too because I can't build the docker container without doing that first. It's probably a good day or so to dig into what changed with hypothesis as it's several major version upgrades all at once.

I'd suggest instead holding off on this until we finish switching over to keycloak at which point we'll just be removing the failing code here anyway.

PR #2571, opened 2025-03-24
Seed data for testing courses locally and RC

annagav:
Fix https://github.com/mitodl/hq/issues/5336
Based on Scenarios from https://github.com/mitodl/hq/issues/5335

Populate or update test courses and course runs, for testing product page enrollment UI states.
This courses are not connected to OpenedX instance.

The data should be valid for a whole calendar year.

### How can this be tested?

Run a management command:
`./manage.py populate_course_data`

Creates 4 Scenarios:

Screenshot 2025-03-26 at 3 22 13 PM

Screenshot 2025-03-26 at 2 40 07 PM
Screenshot 2025-03-26 at 2 40 29 PM

Screenshot 2025-03-26 at 3 22 38 PM
Screenshot 2025-03-26 at 3 22 25 PM

### DIFF
--- a/courses/management/commands/populate_course_data.py
+++ b/courses/management/commands/populate_course_data.py
@@ -1,21 +1,20 @@
+from datetime import timedelta
 
 import reversion
-from datetime import timedelta
+from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
+from django.core.management.base import BaseCommand
 from mitol.common.utils import now_in_utc
 
-from django.contrib.auth import get_user_model
-from django.core.management.base import BaseCommand
-
-from cms.models import CoursePage, CourseIndexPage
+from cms.models import CourseIndexPage, CoursePage
 from courses.management.utils import load_json_from_file
-from courses.models import CourseRun, Course, Department
+from courses.models import Course, CourseRun, Department
 from ecommerce.models import Product
 
 User = get_user_model()
 
-COURSE_DATA_PATH = 'courses/management/courses.json'
-FAKE_COURSE_DESC_PREFIX = '[FAKE] '
+COURSE_DATA_PATH = "courses/management/courses.json"
+FAKE_COURSE_DESC_PREFIX = "[FAKE] "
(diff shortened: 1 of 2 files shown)


ChristopherChudzicki (commented):
Creating the test courses was easy, and they show up in the fake department 👍 

Two questions:

1. I noticed that on http://mitxonline.odl.local:8013/catalog/courses the fake department courses don't show up. Is that intentional?
2. When I try to enroll in courses, I get the following error:
    ```
    null value in column "product_version_id" of relation "ecommerce_line" violates not-    null constraint
    DETAIL:  Failing row contains (5, 2025-03-27 17:20:39.12618+00, 2025-03-27 17:20:39.126186+00, 1, 5, null, 59, 1).
    ```

(2) occurs both for audit track and verified track. For audit track, the enrollment seems to work—I get the django error screen, but if I visit my dashboard afterward, I can see the enrollment.

I'm not sure if I'm missing some setup, or if (2) is an issue.

If it matters, I have `FEATURE_IGNORE_EDX_FAILURES=True` set.

annagav:
> I noticed that on http://mitxonline.odl.local:8013/catalog/courses the fake department courses don't show up. Is that intentional?

It should show up on the side bar:
Screenshot 2025-03-26 at 3 22 13 PM

annagav:
> When I try to enroll in courses, I get the following error:

Good catch! I forgot to add the `with reversion.create_revision():`. So I added that.
Enrollment should not really succeed bc there is no corresponding course run on open edx. 
I am not sure what we expect in terms of enrollment for these fake courses.

ChristopherChudzicki (approved):
With new change and adding the cybersource env vars from https://github.com/mitodl/mitxonline/blob/main/docs/source/configuration/ecommerce.rst, enrolling in both verified and non-verified worked great 👍 

The only oddity I noticed was the courses not showing up in "All" on first load. That definitely seems like an existing frontend issue, though, since the API is returning good data and there are no frontend changes here.

PR #2572, opened 2025-03-24
6810 update financial assistance to use ue discount codes

cp-at-mit:
https://github.com/mitodl/hq/issues/6810

Adds signal to create a Discount in UE when an approved FlexiblePrice object is created in MITx Online.

### How can this be tested?

1. You will need to be running UE locally with this, https://github.com/mitodl/unified-ecommerce/pull/235, UE PR checked out.
2. Create a new integrated system called "mitxonline"
3. Create a new integration system API key http://ue.odl.local:9080/admin/system_meta/integratedsystemapikey/
4. Set the `UNIFIED_ECOMMERCE_API_KEY` environment variable in MITx Online to the API key from step 3.
5. Update one of the products in EU to have a SKU of `course-v1:edX+E2E-101+course` and set the system to the system created in step 2.
6. Configure flexible price tiers in MITx Online for the course `course-v1:edX+E2E-101+course`.
7. Ensure that a user exists in MITx Online and EU with the same email address in both systems.
8. Create an approved flexible price object for the user in MITx Online.  Once saved, an API request to the UE system should be made.  Verify that a Discount is created for the user/product/system in UE.

### Additional Context
The discount codes created in UE should automatically be applied when checking out (ref: https://github.com/mitodl/unified-ecommerce/blob/1cffc3c57239897aab27f9d49487d8692e3a1495/payments/api.py#L466)

### DIFF
--- a/flexiblepricing/tasks_test.py
+++ b/flexiblepricing/tasks_test.py
@@ -385,7 +385,7 @@ class TestFlexiblePriceDiscountProcessing(TestCase):
         assert mock_process.call_count == len(self.program.courses)
         for i, course in enumerate(self.program.courses):
             args, _ = mock_process.call_args_list[i]
-            assert args[0] == course
+            assert args[0] == course[0]
             assert args[1] == instance
 
     @patch("flexiblepricing.tasks._process_flexible_price_discount")


jkachel (changes requested):
This seemed to work with a standalone course that has a single current course run and has financial assistance configured for it. The discount applied correctly in UE when I added the item to the basket there too.

It did not work when I created a course that was in a program, though. I noted this in the comments - the immediate root cause is that it's looking at the program, and the program model doesn't have an API to pull all the products associated with the program. However, this highlighted something else: even if it's looking at the course, it's only going to create a discount for the first product it finds. It really needs to create discounts for everything that the learner is eligible to take.

Not sure what the best way to do this is - the easiest solution is just to loop through and create a new discount for every eligible enrollable course that has a product. That will create a lot of discounts but that's probably OK for MVP. (This definitely needs to be queued.)

jkachel (changes requested):
For a single course, this worked OK. When I tested with a program, I got an error:

```
[2025-04-10 18:16:59,070: ERROR/ForkPoolWorker-8] Error processing flexible price discount
Traceback (most recent call last):
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/fields/__init__.py", line 2053, in get_prep_value
    return int(value)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'Course'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/app/flexiblepricing/tasks.py", line 260, in process_flexible_price_discount_task
    _process_flexible_price_discount(instance)
  File "/app/flexiblepricing/tasks.py", line 130, in _process_flexible_price_discount
    _process_course_discounts(course, instance)
  File "/app/flexiblepricing/tasks.py", line 142, in _process_course_discounts
    course_runs = get_enrollable_courseruns_qs(valid_courses=[course])
  File "/app/courses/utils.py", line 114, in get_enrollable_courseruns_qs
    return CourseRun.objects.filter(q_filters)
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/query.py", line 1436, in filter
    return self._filter_or_exclude(False, args, kwargs)
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/query.py", line 1454, in _filter_or_exclude
    clone._filter_or_exclude_inplace(negate, args, kwargs)
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/query.py", line 1461, in _filter_or_exclude_inplace
    self._query.add_q(Q(*args, **kwargs))
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1546, in add_q
    clause, _ = self._add_q(q_object, self.used_aliases)
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1577, in _add_q
    child_clause, needed_inner = self.build_filter(
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1405, in build_filter
    return self._add_q(
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1577, in _add_q
    child_clause, needed_inner = self.build_filter(
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1492, in build_filter
    condition = self.build_lookup(lookups, col, value)
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/sql/query.py", line 1319, in build_lookup
    lookup = lookup_class(lhs, rhs)
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/lookups.py", line 27, in __init__
    self.rhs = self.get_prep_lookup()
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/fields/related_lookups.py", line 92, in get_prep_lookup
    self.rhs = [target_field.get_prep_value(v) for v in self.rhs]
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/fields/related_lookups.py", line 92, in <listcomp>
    self.rhs = [target_field.get_prep_value(v) for v in self.rhs]
  File "/opt/venv/lib/python3.10/site-packages/django/db/models/fields/__init__.py", line 2055, in get_prep_value
    raise e.__class__(
TypeError: Field 'id' expected a number but got <Course: course-v1:MITxT+PhysX+101 | PhysX 101>.
```

This sort of appears to start in `flexiblepricing/tasks.py` on line 130; I'm not sure why this is complaining about a type mismatch, though. (I thought this was because I didn't have a product in MITx Online for the 2nd course in the program, but this is erroring on the first one, which was fine to begin with.)

One other change: the `UNIFIED_ECOMMERCE_API_KEY` setting isn't in `settings.py` - I had to put that in to make it work.

jkachel (approved):
LGTM 👍 tested with a program and got all of the discounts I was expecting.

PR #2573, opened 2025-03-25
Fix varnish config for arbitrary hostnames

rhysyngsun:

N/A

This simplifies and fixes the varnish configuration so that it doesn't do the extra steps of setting a hardcoded host name.

Instead, the `docker-compse.yml` has been updated to use the forwarded host/port, which accomplishes the same result but will work for arbitrary url bases.

I also made a bit of a breaking change to the way that `PUBLIC_PATH` is set because I need to be able to override the entire string (e.g. `https://mitxonline.odl.local', different protocol and no port). This shouldn't be an issue for most people unless they doing something super custom.

### How can this be tested?

You should be able to `docker compose up` and the app should function like normal.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -66,6 +66,9 @@ services:
     environment:
       << : *py-environment
       PORT: 8011
+      # these are necessary to get the hostname from the varnish proxy
+      USE_X_FORWARDED_HOST: true
+      USE_X_FORWARDED_PORT: true
     env_file: .env
     links:
       - db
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -86,7 +89,7 @@ services:
     ports:
       - "8012:8012"
     environment:
-      PUBLIC_PATH: http://${MITX_ONLINE_HOSTNAME:-mitxonline.odl.local}:8012/
+      PUBLIC_PATH: ${MITX_ONLINE_WATCH_BASE_URL:-http://mitxonline.odl.local:8012}
```
jkachel, line 92: This (and line 134) need a trailing slash on the default. As-is, the JS app doesn't load.

jkachel (changes requested):
Generally looks good - I tested with a random hostname and that was OK too. I did identify a small issue with the defaults for the new base URL settings, though.

jkachel (approved):
LGTM 👍

PR #2574, opened 2025-03-25
Fix API issue caused by non-versioned API

cp-at-mit:
NA
Ref: https://mitodl.slack.com/archives/C02649X2P1V/p1742331049168759

MITx Online included a courses API that was not versions (v1, v0) which was removed in a previous PR as it was causing issues when generating the OpenAPI specification automatically.  The API that was removed was an exact copy of the v1 API.  This PR updates code that relied on the removed courses API to now use the v1 courses API.

### How can this be tested?
Enrolling into a course from the course page should function as expected.


annagav (approved):
👍

PR #2583, opened 2025-04-02
A few small schema fixes (optional fields)

ChristopherChudzicki:
For dashboard work, noticed during https://github.com/mitodl/hq/issues/7004

Marks the a few nullable properties as optional.

### How can this be tested?
This should have absolutely no runtime effects.

### DIFF
--- a/courses/serializers/v1/courses.py
+++ b/courses/serializers/v1/courses.py
@@ -32,7 +32,6 @@ class CourseSerializer(BaseCourseSerializer):
     page = CoursePageSerializer(read_only=True)
     programs = serializers.SerializerMethodField()
 
-    @extend_schema_field(int)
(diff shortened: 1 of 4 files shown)


courses/serializers/v1/courses.py
```diff
@@ -32,7 +32,6 @@ class CourseSerializer(BaseCourseSerializer):
     page = CoursePageSerializer(read_only=True)
     programs = serializers.SerializerMethodField()
 
-    @extend_schema_field(int)
     def get_next_run_id(self, instance) -> int | None:
```
ChristopherChudzicki, line 35: This change was the motivation for PR, but I noticed a few other nullable serializer method field annotations to update, too.

annagav (approved):
👍

PR #2585, opened 2025-04-02
Add xdist so tests can be run faster

rhysyngsun:

N/A

Python tests have been taking an increasing amount of time to run, definitely locally and I suspect github actions are suffering as well. This adds the plugin `pytest-xdist` which allows the tests to run in parallel. Each test worker gets its own db so they don't stomp on each other.

It also fixes the ecommerce tests which starting failing when run in parallel because some the settings fixture (moved to `conftest.py`) was setting the settings on the global django settings instead of the pytest `settings` fixture so it was leaking across test boundaries, but in such a way that it allowed other tests to pass that otherwise wouldn't have the correct configuration if they ran earlier.

### How can this be tested?

You can run this a few ways (you don't need to test them all, but test at least one `-n` option):

- `pytest` - same as before, everything is run serially
- `pytest -n auto` - run 1 process per physical cpu
- `pytest -n logical` - run 1 process per logical cpu
- `pytest -n N` - run N processes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,6 +45,7 @@ repos:
           - ".*_test.js"
           - --exclude-files
           - "config/keycloak/*"
+        additional_dependencies: ["gibberish-detector"]
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: "v0.7.0"
     hooks:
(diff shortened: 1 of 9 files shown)


annagav (approved):
👍

PR #2546, opened 2025-02-21
Prevent linking existing user if uid is different

annagav:
### What are the relevant tickets?
Fix https://github.com/mitodl/hq/issues/6588

### Description (What does it do?)
If a user is currently logged in under email address, social auth should reject an attempt to link the accounts.

PR #2586, opened 2025-04-02
Check that if social auths already exist we dont create another

annagav:
Similar to this reverted PR https://github.com/mitodl/mitxonline/pull/2546
But makes sure not to create extra auths going forward, rather than checking if multiple already exist (which plenty of them do)
Makes sure not to create another social auth for same provider

### How can this be tested?
Try to create multiple auths for the same provider.


annagav:
We decided not to bother with this since there were too many auth duplicates already in MITxOnline.

PR #2594, opened 2025-04-08
Switch to using edx_username

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/7014

- I switched `User.USERNAME` back to `"username"` because this was a) breaking the `createsuperuser` command and b) a user can't really provide a valid value for `global_id` when creating a user via that command anyway.
- Switches reads to use `edx_username`. A helper property was added to `User` to make this easier.
- Changes were made in django-admin so that displays the value from `OpenEdxUser` and write to that value as well:
Screenshot 2025-04-22 at 15-30-12 User edx_username testlearner email testlearner@mitxonline odl local Change user Django site admin
 

### How can this be tested?

- You should be able to create a new user, the user gets created in openedx with the correct username, and you can register for a course
- You should be able to create a new user if your local openedx is down and when you bring it back up the user should get created with the correct username
- You should be able to update the username in django-admin and it will update the value in `OpenEdxUser.edx_username` and leave the value in `User.username` alone.


jkachel:
Did quick sanity-check read-through and did not see anything that looked out of place.

PR #2599, opened 2025-04-15
7056 apisix integration for mitx online

cp-at-mit:
https://github.com/mitodl/hq/issues/7056

Adds APISix to the app similar to: https://github.com/mitodl/unified-ecommerce/pull/241/

### How can this be tested?

- App should function as normal.
- Tests should pass.


jkachel:
I did some quick testing of this and I have a couple of small things while this is in progress:
- Settings: 
   - The `MITOL_APIGATEWAY_USERINFO_CREATE` and `MITOL_APIGATEWAY_USERINFO_UPDATE` flags should probably be exposed in the `.env`. (You can roll these into a single one.. like `MITXONLINE_SCIM_ENABLED` or something.) When there's no SCIM support, the app needs to make/update the user records on its own.
   -  Also, the defaults for `MITOL_APIGATEWAY_ALLOWED_REDIRECT_HOSTS` should be changed (or also exposed in the `.env` or both) - they're the Unified Ecommerce ones currently.
- User model: we need to set the `is_active` flag in the User model to default to True. Otherwise, users will log in, the app will create their accounts, and they will continue to be logged out, because the resulting record isn't active.

Otherwise, seemed to work OK for me.

jkachel (changes requested):
I tested this and added some comments right as you were marking this ready for review, so they're in the main conversation thread. Just a couple of small changes and then I think this is OK.

jkachel (approved):
LGTM 👍 I was able to log in with a known account, an unknown SSO account, and using the regular email workflow.

I will note that logging out gave me the `id_token_hint` error when logging out using the email workflow - this is because APISIX takes over the `/logout` URL. So, for normal local development, we should continue to hit the app directly on port 8013; when deployed behind APISIX, we should configure it to not do things with the logout URL until the SSO integration stuff is done. (Nothing to change here, just configuration notes for deployment, in other words.)

PR #2604, opened 2025-04-22
dockerfile enhancements

Ardiea:
Productionalize the docker container in preparation for moving mitxonline into Kubernetes. 

### How can this be tested?
Verify that development workflows still behave as expected and docker compose up works like it should

Follows the same patterns as this mit-learn PR https://github.com/mitodl/mit-learn/pull/2161


Dockerfile
```diff
@@ -1,70 +1,78 @@
...
+COPY . /src
+WORKDIR /src
 
 USER root
-RUN apt-get clean && apt-get purge
 
 # Set pip cache folder, as it is breaking pip when it is on a shared volume
 ENV XDG_CACHE_HOME /tmp/.cache
 
-FROM base as django
```
jkachel, line 57: The docker-compose file references this still, in the `celery` service.  Probably fine to change the target there to `django-server`.

Dockerfile
```diff
@@ -1,70 +1,78 @@
...
 
 USER mitodl
```
jkachel, line 68: The user gets set in a few places that we probably don't need - it gets set to mitodl in the poetry target, then gets set to root to set the pip cache folder, then gets set to mitodl again, then gets set to mitodl at the very end (in the poetry target). The last one (line 79, at the end) is the one that stood out the most but the others don't appear to be really necessary either. Once the user gets set to mitodl before installing poetry, that should be sufficient I think?

apt.txt
```diff
@@ -6,3 +6,5 @@ zlib1g-dev
 net-tools
 libpq-dev
 postgresql-client
+build-essential
+libcairo2-dev
```
jkachel, line 10: `python:3.10.16-slim` doesn't include `wget` so the `web` container won't start. It does a health check to see if the `watch` and `refine` containers are up using a script that depends on it - [we got that script from elsewhere](https://github.com/mitodl/mitxonline/blob/main/scripts/wait-for.sh) so it would probably be easiest to just add `wget`.

jkachel (changes requested):
A couple of minor changes needed. Tested OK when I made the changes myself locally, so should be OK once they're knocked out, but local dev doesn't work as is.

jkachel (approved):
LGTM :+1: everything started up OK and app appeared to work just fine. The pre-commit changes shouldn't require a re-review so this good to go.

PR #2605, opened 2025-04-23
Make email settings more clear

jkachel:

mitodl/hq#4051

A handful of changes that make the email configuration settings more clear:

- Set the Mailgun From override to the `MITX_ONLINE_FROM_EMAIL` setting so we don't have >1 setting for this
- Set `MITX_ONLINE_REPLY_TO_ADDRESS` to default to `MITX_ONLINE_FROM_EMAIL` (since this seems like a reasonable choice)
- Update the description of `EMAIL_SUPPORT` to note that the setting is only a display setting

Following this, the only environment setting you really have to set will be `MITX_ONLINE_FROM_EMAIL`. (Mostly, this is just to make things slightly less confusing.)

### How can this be tested?

All tests should pass. If you have a usable Mailgun setup, you can try things that send email and see what you get back. There should be no functional changes.

### Additional Context

There will be a related PR to make sure the settings for deployment are correct too.

### DIFF
--- a/app.json
+++ b/app.json
@@ -190,10 +190,6 @@
       "description": "Maximum number of emails to send in a batch",
       "required": false
     },
-    "MAILGUN_FROM_EMAIL": {
-      "description": "Email which mail comes from",
-      "required": false
-    },
     "MAILGUN_KEY": {
       "description": "The token for authenticating against the Mailgun API",
       "required": true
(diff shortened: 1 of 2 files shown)


annagav (approved):
LGTM 👍  (some tests are still failing)

jkachel:
I'm going to go ahead and merge this - the test that's failing is in the financial assistance code, and it's not new (I can see the same test failure in other runs that predate this code). Fixing it is kind of out of scope for this PR though.

PR #2612, opened 2025-04-26
add redirect for user onboarding

annagav:
Fix https://github.com/mitodl/hq/issues/7099
When a user logs in via APISIX, if they haven't completed onboarding they should be redirected to it.

### How can this be tested?

1. Checkout this branch
2. run migrations
3. Assuming you have an existing user locally, manually set user.profile.completed_onboarding to False (migrations automatically set to True for all existing users)
4. login and note that you are automatically redirected to the onboarding flow
5. upon completing the onboarding flow you will be redirected to the "next" parameter.


annagav:
There are still couple of things we want to decide:

1. When do we enable this redirect
2. What is the criteria for having completed onboarding. In mit-learn it looks like the criteria is having any interaction with the form. From my perspective, I would think maybe checking if the required fields have been filled.

pdpinch:
> 1. When do we enable this redirect

I thought we had a flag for APISIX that we were using to determine when to turn on the new authentication paths. 

> 2. What is the criteria for having completed onboarding.

This should be the same as it is currently, which I believe checks for some required fields.

rhysyngsun (approved):
LGTM

PR #2615, opened 2025-04-28
Remove override of User.USERNAME

rhysyngsun:

N/A

This updates the apigateway dependency to use https://github.com/mitodl/ol-django/pull/237

Also cleaned up some lint errors in Dockerfile

### How can this be tested?

You should be able to create new accounts and the `createsuperuser` command should work again.


annagav (approved):
Works for me 👍

PR #2616, opened 2025-04-28
 Add a migration to copy existing email addresses into the username field

cp-at-mit:
https://github.com/mitodl/hq/issues/7173

Adds a migration that populates all User's username fields with their current email address value.

### How can this be tested?
run the migration on your local environment and ensure that the username is set to the value of the email address.  If the email address is longer than the max username length then the username will be set to the max length of the email address possible.

### DIFF
--- a/users/migrations/0027_alter_user_username.py
+++ b/users/migrations/0027_alter_user_username.py
@@ -5,7 +5,7 @@ from django.db import migrations, models
 
 class Migration(migrations.Migration):
     dependencies = [
-        ("users", "0027_set_username_to_email"),
+        ("users", "0026_alter_user_is_active"),
     ]
 
     operations = [
--- a/users/migrations/0028_set_username_to_email.py
+++ b/users/migrations/0028_set_username_to_email.py
@@ -27,7 +27,7 @@ def reverse_migration(apps, schema_editor):
 
 class Migration(migrations.Migration):
     dependencies = [
-        ("users", "0026_alter_user_is_active"),
+        ("users", "0027_alter_user_username"),
     ]
 
     operations = [


users/migrations/0027_set_username_to_email.py
```diff
@@ -0,0 +1,30 @@
+from django.db import migrations
+
+
+def update_usernames_to_email(apps, schema_editor):
+    """Update usernames to match user email addresses"""
+    User = apps.get_model("users", "User")
+
+    # Update usernames to match emails but truncate at max length
+    # Get max_length from username field
+    max_length = User._meta.get_field("username").max_length  # noqa: SLF001
```
rhysyngsun, line 10: I forgot this column had a lower max_length - if we have to truncate the data there isn't much point in doing this work anyway - so can we widen the column to match `email` here too so you don't have to truncate it?

users/migrations/0027_set_username_to_email.py
```diff
@@ -0,0 +1,30 @@
...
+
+    users = User.objects.all()
+    for user in users:
```
rhysyngsun, line 13: This will load all the users into memory which will have a chance of crashing in production with the large user count. Better run a bulk `update()` call over medium-sized slices of the user table.

rhysyngsun (changes requested):

users/migrations/0027_set_username_to_email.py
rhysyngsun: That should be plenty, looks like there's a default of 254 for `EmailField` in django based on specs: https://github.com/django/django/blob/stable/5.2.x/django/db/models/fields/__init__.py#L1923-L1924

users/migrations/0028_alter_user_username.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.20 on 2025-04-29 13:08
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
rhysyngsun, line 6: The migrations need to swap order so you widen the column and then populate the data over without truncating it.

rhysyngsun (changes requested):

rhysyngsun (approved):
LGTM

PR #2620, opened 2025-04-29
Make gemini optional on PRs

rhysyngsun:

N/A

This makes Gemini optional but folks can still request it manually to cut down on unwanted PR review noise.


annagav (approved):
👍

PR #2621, opened 2025-04-30
Remove unused username and user methods

cp-at-mit:
NA

Gemini provides a pretty clean summary below, but I removed these methods since I couldn't find them being used anywhere.  There are also other functions for the age related stuff that can be used instead.


jkachel (approved):
LGTM 👍

PR #2624, opened 2025-05-02
fix: Show only Login and update login view for proper redirection with apisix

arslanashraf7:
https://github.com/mitodl/hq/issues/7232

- Shows only Single `Login` button instead of separate `Sign In` and `Create Account` buttons when APISIX login is enabled (`MITOL_APIGATEWAY_DISABLE_MIDDLEWARE=False`)
- Updated the Login view to handle the logic for the proper redirection of the user based on the authentication. See how this should work https://github.com/mitodl/hq/issues/7232#issuecomment-2847193968
- Adds an additional setting (`_JAVA_OPTIONS`) to the Keycloak environment in `docker-compose.yml`. The Keycloak container does not work for me locally on ARM64 based m4 machine. So, this fixes that. To make it work I have set `JAVA_OPTIONS=-XX:UseSVE=0` in my .env file. But this this work fine for other even without setting this.

**With MITOL_APIGATEWAY_DISABLE_MIDDLEWARE disabled:**
image

WithMITOL_APIGATEWAY_DISABLE_MIDDLEWARE **enabled**
image

NOTE: The screenshots show the Login/Sign in text in Red because I hovered over them to capture the URL they call too. At the bottom left.

- [ ] Desktop screenshots

### How can this be tested?

**NOTE:** I personally was not able to setup the APISIX and Keycloak locally so a complete testing of this PR can be done by someone who already has it all set up.

- First, try to set `MITOL_APIGATEWAY_DISABLE_MIDDLEWARE=False` in your .env on the main branch
- When you open the http://mitxonline.odl.local:8013/ you should only see `Catalog` and `Login` buttons and no create account/sign in buttons. Also, make sure that `EXPOSE_OIDC_LOGIN=False` while testing this, otherwise you will see the create account and sign in buttons that take the user to oidc login URL
- Clicking the `Login` button should redirect the user to `/login`, which should be taken by APISIX once that is completely set up. See https://github.com/mitodl/hq/issues/7232#issuecomment-2847190952 and https://github.com/mitodl/hq/issues/7232#issuecomment-2847193968

- Also, make sure that your Keycloak container runs fine to test `JAVA_OPTIONS` that I added in the container environment


jkachel (approved):
LGTM 👍 
Also, good catch about `JAVA_OPTS` - I have not had that issue but I'm on an M1 Pro still. (I'm kinda curious what the difference is.)

PR #2650, opened 2025-05-09
Create profile and legal address on apisix auth

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/7307

Adds `Profile` and `LegalAddress` to the set of created models when a user is authenticated via APISIX.

### How can this be tested?

Create a new user and verify these models get created.


jkachel (approved):
LGTM 👍

PR #2652, opened 2025-05-12
feat: Add health check endpoint

blarghmatey:
N/A

Adds a health-check package and configuration to the application.

### How can this be tested?

Launch the app and verify that the `/health`, `/health/startup/`, etc. checks return successfully.

### Additional Context

In order to effectively manage Kubernetes applications we want to have startup, liveness, and readiness checks. This adds a package for exposing healthchecks that include information about the runtime status of the application.


rhysyngsun (approved):
LGTM

PR #2654, opened 2025-05-12
B2B Provisioning: Create enrollment codes when contracts are in place

jkachel:

Closes mitodl/hq#7242

Depending on what the settings are for a B2B contract, we may need to create enrollment codes for the target learners:

| Integration Type | Enrollment Price | Learner Cap | Result |
|---|---|---|---|
| SSO | Free | No | No Codes |
| SSO | Free | Yes | No Codes |
| SSO | Non-Free | No | Unlimited Use Codes |
| SSO | Non-Free | Yes | One-Time Use Code |
| Non-SSO | Free | Yes | One-Time Use Codes |
| Non-SSO |  Non-Free | Yes | One-Time Uses Codes |

(Non-SSO contracts that don't have a learner cap aren't considered valid. SSO contracts that are not-free and have no learner cap are supported but this configuration is discouraged.)

This PR implements code that creates codes in the system when the contract is saved, or updates existing, unredeemed codes for the contract.

We also want to be able to get the codes _out_ to send to the client for distribution. This adds a `codes-out` option to the `b2b_list` command that will grab the codes and drop them into a CSV file. 

### How can this be tested?

You will need CyberSource credentials - completing check out isn't strictly necessary but you'll likely have issues getting to that point if you don't have credentials.

1. Create a contract and org.
2. Create a course and attach it to the org. 
3. Get the codes out of the system: `b2b_list contracts --codes-out <contract id>` will put them in a file called `codes-<contract id>.csv`.
4. The codes CSV contains a field with a link to add the product to the cart. Use one of them to add to cart.
5. You should be able to check out with one of the supplied codes.

Repeat these steps for contracts with and without SSO; with and without a learner cap; and with and without a set price. If you use the same contract, it should simply update the discounts, rather than creating new ones each time.

### Additional Context

I'm not sure the SSO contract case that I noted above should exist - it does this for now but this seems pretty dangerous, since we don't really have a way to check for organization when checking a code. (We can maybe add that later though?)


annagav (approved):
👍

PR #2657, opened 2025-05-13
Filter courses based on orgs for user

cp-at-mit:
Part of https://github.com/mitodl/hq/issues/7280

Filters the courses based on the orgs that the authenticated user is associated with.

### How can this be tested?

1. Associate your user with a contract.
2. Associate a course with the contract.
3. Visit http://mitxonline.odl.local:8013/api/v2/courses/?org_id=<ORG_ID> using the org ID associated with the contract, which is associated with the user.
4. Ensure that the expected courses appear.
5. Logout, and visit http://mitxonline.odl.local:8013/api/v2/courses/?org_id=<ORG_ID>.  Ensure that all courses that are not associated with a contract are shown.
6. Remove the contract association from your user and login again.
7. Visit http://mitxonline.odl.local:8013/api/v2/courses/?org_id=<ORG_ID> and ensure that all courses that are not associated with a contract are shown.

### DIFF
--- a/courses/views/v2/__init__.py
+++ b/courses/views/v2/__init__.py
@@ -77,7 +77,11 @@ class CourseFilterSet(django_filters.FilterSet):
         org_id = request.query_params.get("org_id") if request else None
 
         if org_id:
-            if user and user.is_authenticated and user.b2b_organizations.filter(id=org_id).exists():
+            if (
+                user
+                and user.is_authenticated
+                and user.b2b_organizations.filter(id=org_id).exists()
+            ):
                 queryset = queryset.filter(
                     courseruns__b2b_contract__organization_id=org_id,
                     courseruns__b2b_contract__active=True,


courses/views/v2/__init__.py
```diff
@@ -73,12 +73,17 @@ class CourseFilterSet(django_filters.FilterSet):
...
     def filter_queryset(self, queryset):
         request = self.request
-        user = request.user if request else None
+        user = getattr(request, "user", None)
         org_id = request.query_params.get("org_id") if request else None
 
-        if not user or user.is_anonymous:
-            queryset = queryset.filter(courseruns__b2b_contract__isnull=True)
-        elif org_id:
+        show_contracted = (
```
jkachel, line 79: I think we want this to display nothing if the learner's not in the org. As is, they won't get the org's courses back but they will get everything that's public, which seems like it'd be problematic for the dashboard view. (This would be a weird state for the learner but not one that's impossible, I think.)

jkachel (changes requested):
Filtering on org_id works well when the learner's in the org. I think the output needs to be adjusted if org_id is there but isn't one the learner's in, though.

jkachel (approved):
LGTM 👍

PR #2659, opened 2025-05-13
Update Users API to be versioned (at v0), so it shows up in the OpenAPI spec

jkachel:

Closes mitodl/hq#7326

Adds the users API to the OpenAPI spec by mounting it at `/api/v0` in the urlconf (in addition to `/api`, where it is now).

Also added a couple of small changes to the UserSerializer so that Spectactular knows what the types are for a handful of fields.

### How can this be tested?

Automated tests should pass.

The Users API should appear in the spec.


gumaerc (approved):
:+1:

PR #2660, opened 2025-05-14
Add API key for B2B data

jkachel:

Closes mitodl/hq#7244

Other systems will likely need to access the B2B data (orgs and contracts) stored in the system. This adds API key authentication for those endpoints, so you can be either an admin user or supply an API key to access the data.

This also updates the `b2b_list` command to add an option to list out the users that are associated with an org or contract.

### How can this be tested?

Automated tests should pass.

You should be able to create an API key in the Django Admin. Then, you should be able to access the b2b endpoints using the key: `curl -H "Authorization: Api-Key <key>" http://mitxonline.odl.local:9080/api/v0/b2b/organizations/` (and `contracts`). 

You should be able to run `b2b_list learners` and specify either `--contract` or `--org` to list users in a specific org. You may need to go in and add a relationship for a user before you'll see anything - you can do this in Django Admin; you don't have to run through the enrollment code checkout process.

### DIFF
--- a/b2b/management/commands/b2b_list.py
+++ b/b2b/management/commands/b2b_list.py
@@ -2,7 +2,7 @@
 
 import logging
 
-from django.core.management import BaseCommand
+from django.core.management import BaseCommand, CommandError
 from rich.console import Console
 from rich.table import Table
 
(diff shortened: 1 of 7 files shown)


jkachel:
The tests were failing because it generated a username that was > 30 characters long, and the edX username can't be longer than 30. So, updated the post-save stuff so that it grabs the first 30 characters of the username for the Open edX user.

PR #2663, opened 2025-05-14
refactor: Remove syslog logging handler

blarghmatey:
N/A

Removes syslog handler from Django logging and relies solely on console logging.

### How can this be tested?

Verify that logs still show up in docker-compose

### Additional Context

The syslog handler won't work under Docker/Kubernetes. Removing it now that this app is no longer running on Heroku.

### DIFF
--- a/app.json
+++ b/app.json
@@ -430,14 +430,6 @@
       "description": "E-mail to use for the from field",
       "required": false
     },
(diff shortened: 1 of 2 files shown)


annagav (approved):
👍

PR #2665, opened 2025-05-15
B2 b provisioning limit programs with org id specified

cp-at-mit:
https://github.com/mitodl/hq/issues/7280

Adds filtering by org_id to the programs API endpoint.

### How can this be tested?
- Associate your user with a contract.
- Associate a course with the contract.
- Add the course as a requirement for a program
- Visit http://mitxonline.odl.local:8013/api/v2/programs/?org_id=<ORG_ID> using the org ID associated with the contract, which is associated with the user.
- Ensure that the expected programs appear.
- Logout, and visit http://mitxonline.odl.local:8013/api/v2/programs/?org_id=<ORG_ID>. Ensure that all programs that are not associated with a contract are shown.
- Remove the contract association from your user and login again.
- Visit http://mitxonline.odl.local:8013/api/v2/programs/?org_id=<ORG_ID> and ensure that all programs that are not associated with a contract are shown.

### DIFF
--- a/courses/views/v2/views_test.py
+++ b/courses/views/v2/views_test.py
@@ -16,7 +16,12 @@ from rest_framework.test import APIClient
 
 from b2b.api import create_contract_run
 from b2b.factories import ContractPageFactory, OrganizationPageFactory
-from courses.factories import CourseFactory, CourseRunFactory, DepartmentFactory, ProgramFactory
+from courses.factories import (
+    CourseFactory,
+    CourseRunFactory,
+    DepartmentFactory,
+    ProgramFactory,
+)
 from courses.models import Course, Program
 from courses.serializers.v2.courses import CourseWithCourseRunsSerializer
 from courses.serializers.v2.departments import (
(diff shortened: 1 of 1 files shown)


jkachel:
I think this works OK. I got kinda stuck on some potential situations that I think are ultimately things that don't affect this PR or the way this was defined. (Maybe we can discuss during the MITx Online project meeting today? @pdpinch @sovsey )

The thing that I had a concern about was programs that are just for B2B contracts. The first few contracts will have their own courses and program(s), and as such we shouldn't display those to users. The filter that this PR defines only filters by organization if `org_id` is specified - without it, you get all programs and as such you'll get courses that users can't enroll in. (As you would anyway - the course being in a program doesn't mean you can enroll; there may not be a valid course run for you to enroll in yet.) 

The frontend won't show programs that don't have CMS pages though, so if we don't create pages for the B2B-specific programs, then regular users won't see the B2B program in MITx Online's catalog. But they still show up in the API, so we should probably add a flag to the program to specify that it's for B2B (I don't think it matters _which_ org/contract, just that it _is_) so we can exclude those by default for anonymous users.

jkachel (changes requested):
Seems to work OK; could use some tests, though.

jkachel (approved):
LGTM 👍

PR #2667, opened 2025-05-19
Remove configure hubspot command

annagav:
N/a

This is blocking certificate generation

### How can this be tested?
nothing should break.


rhysyngsun (approved):
LGTM

PR #2669, opened 2025-05-19
Rename the realm import file

jkachel:

n/a

Keycloak can import realm information from a file, and it wants that file to be named for the realm that's being imported. It has flagged this as a warning in the past but it seems the most recent version has been updated to flag it as a fatal error. So, I had to rename it to get the instance going.

### How can this be tested?

It's probably easiest to do this in a new Compose project so you don't have to break your existing setup.

Make sure you `docker compose pull` so you have the latest images. The docker-compose file specifies `latest` but for clarity my system ended up pulling down `26.2.4`, which is image ID `4a81762677f8`. 

Bring the keycloak service up: `docker compose -p kc-import-test --profile keycloak keycloak up` - this will bring Keycloak and the DB up only. It should succeed and you should be able to get into the default realm; it should look the same otherwise.

There are a few minor UI changes in 26.2.4 (or maybe earlier, I don't know what version I was running) - the realm dropdown is gone now and you have to go to Manage Realms to change realms. Everything else looked largely the same, and I ran into no issues with APISIX integration, but I haven't done a deep dive into whatever else has changed.


annagav:
I think this PR fixes this https://github.com/mitodl/hq/issues/7212

annagav (approved):

PR #2672, opened 2025-05-21
Update urlconf to add accessible schema/Swagger URLs, add better typing for B2B data, update OpenAPI spec

jkachel:

n/a

The `extend_schema_field` for `b2b_organizations` in the User serializer was just `list[str]`; this changes it so that it should return the proper organization type instead, and fixes the nested contract serializer so that it's typed correctly too.

This also adds the frontends for the schema - we didn't expose the schema download or Swagger UI endpoints before, and this adds that too.

### How can this be tested?

Check the type of the `b2b_organizations` member in the users API (specifically, the `/api/v0/users/me` one).

You should also be able to get to Swagger at `/api/schema/swagger-ui/` and download the spec yaml at `/api/schema/` .

### DIFF
--- a/main/urls.py
+++ b/main/urls.py
@@ -18,6 +18,7 @@ from django.conf import settings
 from django.conf.urls.static import static
 from django.contrib import admin
 from django.urls import include, path, re_path
+from drf_spectacular.views import SpectacularAPIView, SpectacularSwaggerView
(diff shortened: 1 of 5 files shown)


jkachel:
Test failure for flaky test, not related to this code.

gumaerc (approved):
:+1:

PR #2694, opened 2025-05-30
Nl/fix global

rhysyngsun:

N/A

Updates mitol-django-scim to pick up changes in https://github.com/mitodl/ol-django/pull/262


jkachel (approved):
👍

PR #2696, opened 2025-05-30
Rework program filtering when an org_id is specified

jkachel:

Related to mitodl/mit-learn#2256

The filtering was relying on the request user to be filtered through in a manner that wasn't actually workable. Reworked to use the request option that gets passed into the FilterSet anyway.

Also added some code to default filter out programs that require any courses that have B2B courseruns, and update the programs viewset to allow viewing un-authenticated (courses is public access, so should programs, I think, but not B2B ones).

### How can this be tested?

0. Get some courses in place. The test ones are fine.
1. Get a contract set up.
2. Attach at least one course to the contract.
3. Set up two programs.
   4. One should have _just_ courses that have no B2B contract course runs.
   5. One should have _at least_ one course that has a B2B contract course run.
   6. You can delete the Electives out and just add a single Required course for expediency.
7. Hit the programs v2 API: `/api/v2/programs/` - you should only see the program that has the non-B2B course runs.
8. Set a user up so that it is in the contract.
9. Hit the programs v2 API again, and filter by the org_id that the contract is in: `/api/v2/programs/?org_id=<id>` - you should see _just_ the program that belongs to the contract
10. You can also create another contract, and then assign it to your user while removing the first assignment. You should now _not_ see the program in the API results.

### DIFF
--- a/courses/views/v2/__init__.py
+++ b/courses/views/v2/__init__.py
@@ -8,6 +8,7 @@ from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.utils import extend_schema
 from rest_framework import viewsets
 from rest_framework.pagination import PageNumberPagination
+from rest_framework.permissions import IsAuthenticatedOrReadOnly
 
 from courses.models import (
     Course,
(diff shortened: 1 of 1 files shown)


gumaerc (approved):
👍 LGTM

PR #2697, opened 2025-06-02
Fix redirect urls when api gateway is enabled

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/7423

Switches the redirect path based on whether the api gateway is enabled or not.

### How can this be tested?

- With `MITOL_APIGATEWAY_DISABLE_MIDDLEWARE=False` (gateway enabled), navigate to `/dashboard`, you should be redirected to `/login/?next=/dashboard/`.
- 
- With `MITOL_APIGATEWAY_DISABLE_MIDDLEWARE=True` (gateway disabled), navigate to `/dashboard`, you should be redirected to `/signin/?next=/dashboard/`.


jkachel (approved):
👍 LGTM

PR #2707, opened 2025-06-05
Remove full name on the password prompt

annagav:
Fix https://github.com/mitodl/hq/issues/7415

Removes full name

Before: 

Screenshot 2025-06-05 at 10 19 50 AM

After:
Screenshot 2025-06-05 at 10 20 16 AM


annagav:
@pdpinch could you approve this change.

rhysyngsun (changes requested):
Can you remove the name from the response in the API too? https://github.com/mitodl/mitxonline/blob/main/authentication/serializers.py#L59-L60

PR #2736, opened 2025-06-16
Fix logout redirects

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/7564

This re-adjusts the logout flow when the apisix middleware is enabled to ensure the user is logged out of openedx, keycloak, apisix, and mitxonline.

I refactored the `urls.py` files a bit too because in order to write tests for the new logout endpoint I needed to be able to load a specific set of `urlpatterns` under pytest.

### How can this be tested?

- Ensure in opendex that you have this or something similar set:
```
IDA_LOGOUT_URI_LIST.append("https:/mitxonline.odl.local/logout")
```
- Login as a user and ensure you're also logged in on openedx
- Click logout in mitxonline, you should land on the mitxonline home page and have no session there or in openedx or keycloak
- Login as a user and ensure you're also logged in on openedx
- Click logout in openedx, you should land on the mitxonline home page and have no session there or in openedx or keycloak

Then, turn off the API gateway middleware and retest all of the above, it should still work.


jkachel (approved):
I think this is fine. Running the flow runs things into edX for logout, which then hits the `/logout/oidc` route that APISIX consumes, and that kills all the sessions. So, LGTM 👍 

The issues I was having testing this amounted to differences in hostnames. I think you have your Tutor setup configured to exist at `openedx.odl.local` and I was using the default `local.openedx.io`, and I had none of these services set up to use SSL. So, when the iframe in the edX logout interstitial starts pinging things to try to log you out, it does hit `/logout/?no_redirect=1` as it should but the cookies (and thus the session) don't transmit in that transaction, because it's an iframe in a different origin. So, it was acting correctly; it actually didn't have a session.

With Tutor set up to use a `.odl.local` domain and the cookie domains set in apisix to be `.odl.local`, the cookies _do_ come over in the iframe and things work properly. (The other option for fixing this is getting APISIX set up with a certificate; we then have to set the cookies to be SameSite=None and Secure so they will transfer over (Lax won't do it).) But, this is a problem for local deployments and shouldn't be an issue on RC I think, since the origins aren't totally different there.

PR #2738, opened 2025-06-16
Update next_run_id to add B2B course run awareness

jkachel:

Closes mitodl/hq#7563

If you're pulling data for a course that has runs that meet the time requirements for enrollability (start date before now, enrollment start date before now, end dates after now or not set) but are a mix of B2B course runs and regular ones, the `next_run_id` field in the course may return an invalid course run:

- If you're not filtering for a particular B2B organization, you might still get a B2B course run as the next ID if it's the next available run
- If you are filtering on organization, you might get a course run that's not the B2B run

To fix this, this PR updates `first_unexpired_run` to filter out B2B course runs entirely. It also adds a second method, `get_first_unexpired_org_run`, that filters specifically on the user's contracts (as we want to see things the user is allowed to get to - if there are runs in contracts the user isn't in, we don't want to consider those, either). The view is also updated to pull the user's contract IDs out and add them to the serializer context so it will know to use the new method if the org filter is enabled.

### How can this be tested?

Set up a course with a regular course run that can be enrolled in. Ensure you can see the course in the v2 courses API (`/api/v2/courses/`) and that the `next_run_id` points to the regular course run.

Add a contract, then add the course to the contract. Reload the courses API. You should still see the course you created, and the `next_run_id` should be the regular course run still.

Set the regular course run's enrollment start date to a future date (or, clear it), and reload the course API. You should see null for the `next_run_id`.

Add the contract to the current user. Then, hit the API with the org filter - `/api/v2/courses/?org_id=<org>`. Now, you should see the B2B course run's ID as `next_run_id`.

Set the B2B run's enrollment date to a future date and reload the course API. You should see null for `next_run_id`.

Create a new contract within the same org, and add the course to the new contract. Set the enrollment dates for the B2B runs up so that the new contract's run starts before the existing contract's run. _Do not_ add the user to the new contract. Then, reload the course API again - you should see either null or the existing contract's run ID in `next_run_id` depending on whether or not the existing run's enrollment start is in the future or not.

### DIFF
--- a/b2b/factories.py
+++ b/b2b/factories.py
@@ -9,7 +9,7 @@ from b2b.models import ContractPage, OrganizationIndexPage, OrganizationPage
 from cms.factories import HomePageFactory
 from cms.models import HomePage
 
-FAKE = faker.Factory.create()
+FAKE = faker.Faker()
 
 
 class OrganizationIndexPageFactory(wagtail_factories.PageFactory):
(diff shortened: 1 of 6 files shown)


annagav (approved):
👍

PR #2739, opened 2025-06-17
save edx_username when creating profile

annagav:
edx username not saved properly
and Fix https://github.com/mitodl/hq/issues/7609

Add create_edx_user to profile creation
Onboarding API should call create_edx_user when the user submits the profile information.

Now it passes the provided username to create_user .

### How can this be tested?
create a new account , fillout  the profile 
You edx user should have a edx_username saved correctly.
The edx api call to create user should succeed.

### DIFF
--- a/authentication/api_gateway/serializers.py
+++ b/authentication/api_gateway/serializers.py
@@ -7,6 +7,7 @@ from django.db import transaction
 from rest_framework import serializers
 
 from hubspot_sync.task_helpers import sync_hubspot_user
+from openedx.api import create_user
 from users.serializers import (
     LegalAddressSerializer,
     UserProfileSerializer,
@@ -24,7 +25,7 @@ class RegisterDetailsSerializer(serializers.Serializer):
     legal_address = LegalAddressSerializer(write_only=True)
     user_profile = UserProfileSerializer(write_only=True)
 
-    def save(self, validated_data):
+    def create(self, validated_data):
         """Save user legal address and user profile"""
         request = self.context["request"]
         user = request.user
@@ -33,9 +34,14 @@ class RegisterDetailsSerializer(serializers.Serializer):
         legal_address_data = validated_data.pop("legal_address")
         user_profile_data = validated_data.pop("user_profile", None)
         with transaction.atomic():
-            user.openedx_user.edx_username = username
             user.name = name
             user.save()
+            if user.openedx_user is None:
+                create_user(user)
+            else:
+                user.openedx_user.edx_username = username
+                user.openedx_user.save()
+
             if legal_address_data:
                 legal_address = LegalAddressSerializer(
                     user.legal_address, data=legal_address_data
@@ -49,7 +55,6 @@ class RegisterDetailsSerializer(serializers.Serializer):
                 )
                 if user_profile.is_valid():
                     user_profile.save()
-
         sync_hubspot_user(user)
         return user
 
--- a/authentication/api_gateway/serializers.py
+++ b/authentication/api_gateway/serializers.py
@@ -37,7 +37,7 @@ class RegisterDetailsSerializer(serializers.Serializer):
             user.name = name
             user.save()
             if user.openedx_user is None:
-                create_user(user)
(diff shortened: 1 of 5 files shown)


authentication/api_gateway/serializers.py
```diff
@@ -33,9 +34,14 @@ def create(self, validated_data):
         legal_address_data = validated_data.pop("legal_address")
         user_profile_data = validated_data.pop("user_profile", None)
         with transaction.atomic():
-            user.openedx_user.edx_username = username
             user.name = name
             user.save()
+            if user.openedx_user is None:
+                create_user(user)
```
jkachel, line 40: Noting this for me later - this calls the openedx create_user, but the user may not have an OpenEdxUser so they may not have a username yet.. not 100% sure here yet; have to fix my local Tutor setup, so this is a note for when I look at this next (prob Friday).

jkachel, line 40: When I tested this with a brand-new user, I got an error back from edX because there wasn't a username specified. I had supplied one in `/create-profile/` in mitxonline but there wasn't yet an `OpenEdxUser` record for my user, so no username got passed on to edX. So, I think we should explicitly create the `OpenEdxUser` if there's not one already. (`create_user` will do `get_or_create` but since the username's not there in a new record, it fails.)

jkachel (changes requested):
Ran into an error testing with a new account - think it's _probably_ an easy fix, though.

openedx/api.py
```diff
@@ -97,6 +99,13 @@ def create_edx_user(user):
             user=user, platform=PLATFORM_EDX
         )
 
+        if open_edx_user.edx_username is None:
+            if edx_username is None:
+                # If we don't have an edx username provided, generate one from the user's email
+                edx_username = user.email
+            open_edx_user.edx_username = edx_username
+            open_edx_user.save()
```
jkachel, line 107: This ends up sending `None` as the username below (on line 132). Probably best to reference the `open_edx_user` directly there; that seemed to work best in testing.

I'm still getting a 403 error from Open edX when testing this locally so not sure what's going on there. I did check to make sure the payload that is going to edX looks more-or-less right but I can't get it to be happy with the request. Not sure if this is a misconfiguration issue on my end or something else.

jkachel (changes requested):
One more minor-ish change (while I figure out whether or not my edX instance is broken or not).

openedx/api.py
```diff
@@ -97,6 +99,13 @@ def create_edx_user(user):
             user=user, platform=PLATFORM_EDX
         )
 
+        if open_edx_user.edx_username is None:
+            if edx_username is None:
+                # If we don't have an edx username provided, generate one from the user's email
+                edx_username = user.email
```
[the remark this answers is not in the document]
pdpinch: Like something that uses https://github.com/mitodl/ol-django/pull/266 ?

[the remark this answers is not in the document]
annagav: so are we talking about adding username generation to this PR?

openedx/api.py
annagav: I added the reference to the edx user. I think that cached property doesn't get updated properly. I had same issue  in the tests.

authentication/api_gateway/serializers.py
```diff
@@ -33,9 +34,14 @@ def create(self, validated_data):
...
         user_profile_data = validated_data.pop("user_profile", None)
         with transaction.atomic():
-            user.openedx_user.edx_username = username
             user.name = name
             user.save()
+            if user.openedx_user is None:
+                create_user(user, username)
+            else:
+                user.openedx_user.edx_username = username
+                user.openedx_user.save()
```
rhysyngsun, line 43: I _think_ this can just be collapsed down to an unconditional call to `create_user` like this:
```suggestion
            create_user(user, username)
```

rhysyngsun (changes requested):

rhysyngsun (approved):
LGTM

PR #2740, opened 2025-06-17
7509 improve error message when user doesnt enter an enrollment code for a b2b product

cp-at-mit:
https://github.com/mitodl/hq/issues/7509

Changes the error message shown when a user has a product associated with a contract in their cart, and proceeds to checkout without populating the coupon code field.

Also automatically applies any coupon code entered in the coupon code field when the proceed to payment button is clicked.

### How can this be tested?

1. Attempt to checkout with a product that is associated with a contract, without entering in the discount code.  Verify that the error message matches what is described in the linked issue.
2. Attempt to checkout with a product that is associated with a contract, enter a discount code, and click the proceed with payment button.  Verify that the cost of the product is reduced by the discount associated with the discount code.


annagav (approved):
👍

PR #2747, opened 2025-06-23
Fix a warning for an unnecessary null=True

rhysyngsun:

N/A

This fixes the following annoying warning as well as adds a check for these kinds of errors/warnings to prevent them in the future:

```
System check identified some issues:

WARNINGS:
cms.CoursePage.topics: (fields.W340) null has no effect on ManyToManyField.
```

### How can this be tested?

Tests should pass


jkachel (approved):
LGTM 👍

PR #2751, opened 2025-06-24
additional details modal from dashboard should display country when state is required

cp-at-mit:
https://github.com/mitodl/hq/issues/7508

The linked issue affects users who have a country of "US" or "CA" but no state and no year of birth.  The modal would only display the Year of birth field and not a state field, so the form validation would fail on the value for state.

### How can this be tested?

1. Create a user with a country of "US", no state, and no year of birth.
2. Enroll into a course.
3. From the dashboard, click go to course button.
4. Verify that the modal shows the year of birth, country, and state fields.  The country field should be populated.
5. Enter info into all fields and save.  You should be redirected to the course page and the information should save to the user object.

### DIFF
--- a/authentication/api_gateway/serializers.py
+++ b/authentication/api_gateway/serializers.py
@@ -7,6 +7,7 @@ from django.db import transaction
 from rest_framework import serializers
 
 from hubspot_sync.task_helpers import sync_hubspot_user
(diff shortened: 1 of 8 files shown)


jkachel (approved):
LGTM 👍

PR #2752, opened 2025-06-24
Fix lazy_fixture error

rhysyngsun:

N/A

Fixes an import error from a previous commit.

### How can this be tested?

Tests should pass.


annagav (approved):
👍

PR #2753, opened 2025-06-24
Repair faulty users when OpenEdxUser exists already

annagav:
Fix https://github.com/mitodl/hq/issues/7610

Update the FaultyOpenEdxUserManager manager to also include users who have edx_username set but have not been synced.
### How can this be tested?

Tests should pass

You can test this by updating your OpenEdxUser `has_been_synced` to false and then making sure that the task has picked it up and tried to sync with edx.

### DIFF
--- a/users/models.py
+++ b/users/models.py
@@ -225,7 +225,7 @@ class FaultyOpenEdxUserManager(BaseUserManager):
     """User manager that defines a queryset of Users that are incorrectly configured in the openedx"""
 
     def get_queryset(self):  # pylint: disable=missing-docstring
-        return (
+        faulty_users = (
             super()
             .get_queryset()
             .select_related("openedx_api_auth")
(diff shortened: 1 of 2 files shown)


users/models.py
```diff
@@ -225,20 +225,24 @@ class FaultyOpenEdxUserManager(BaseUserManager):
     """User manager that defines a queryset of Users that are incorrectly configured in the openedx"""
 
     def get_queryset(self):  # pylint: disable=missing-docstring
-        return (
+        faulty_users = (
             super()
             .get_queryset()
             .select_related("openedx_api_auth")
-            .prefetch_related("openedx_users")
+            .prefetch_related("openedx_users", "openedx_users__has_been_synced")
```
rhysyngsun, line 232: This new prefetch isn't needed since the record is already prefetched with the first arg.

rhysyngsun (changes requested):

rhysyngsun:
Needs a rebase to pick up those test fixes too.

PR #2754, opened 2025-06-24
Add a command to fix missing edx_username values

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/7652

This adds a management command that updates a user's `OpenEdxUser.edx_username` from their `User.username`. By default it does this for all users that have either not `OpenEdxUser` record or a null `OpenEdxUser.edx_username`. A specific user can be specified as well.

### How can this be tested?

- Ensure you have at least two users with a null `edx_username`. If you create one, be sure they have a `LegalAddress` and `UserProfile` and that `User.name` is populated.
- Run `./manage.py repair_openedx_username`, you should see you users in the list. Choose no (n).
- Run `./manage.py repair_openedx_username --user-id ID`, you should see only that user. Choose yes (y), the user should have `edx_username` set and be created in openedx. 
- Run `./manage.py repair_openedx_username`, you should see the remaining users. Choose yes (y) and the remaining users should be repaired.

### DIFF
--- a/openedx/management/commands/repair_openedx_usernames.py
+++ b/openedx/management/commands/repair_openedx_usernames.py
@@ -1,0 +1,60 @@
+"""
+Repairs the openedx username
+"""
+
+import sys
+
+from django.contrib.auth import get_user_model
+from django.core.management import BaseCommand
+from django.db.models import F, Q
+from tabulate import tabulate
+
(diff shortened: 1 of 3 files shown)


jkachel (approved):
LGTM :+1:

PR #2757, opened 2025-06-25
Fix course api courserun ordering

rhysyngsun:

N/A

Fixes a test regression in the courses API by reintroducing natural sorting by id explicitly.

### How can this be tested?

Tests should pass.


annagav (approved):
👍

PR #2760, opened 2025-06-27
Fix another set of flaky course api tests

rhysyngsun:

N/A

Fixes some other course api tests that were failing due to inconsistent run ordering.

### How can this be tested?

Tests should pass


annagav (approved):
👍

PR #2762, opened 2025-06-30
fix: Set custom CSRF cookie name

blarghmatey:
https://github.mit.edu/mitxonline/mitxonline-issues/issues/942

The csrftoken cookie is now conflicting with CSRF in the edx-platform application. This sets a different cookie name to avoid those conflicts since the cookie is being set on a shared domain namespace.

### How can this be tested?

Verify that the CSRF cookie name is set to the custom value

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -110,6 +110,12 @@ CSRF_COOKIE_DOMAIN = get_string(
     description="Domain to set the CSRF cookie to.",
 )
 
+CSRF_COOKIE_NAME = get_string(
+    name="CSRF_COOKIE_NAME",
+    default="csrf_mitxonline",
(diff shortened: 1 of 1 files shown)


main/settings.py
```diff
@@ -110,6 +110,12 @@
     description="Domain to set the CSRF cookie to.",
 )
 
+CSRF_COOKIE_NAME = get_string(
+    name="CSRF_COOKIE_NAME",
+    default="csrf_mitxonline",
+    description="Domain to set the CSRF cookie to.",
```
pdpinch, line 116: Is this the right description?

pdpinch (approved):
Left a question, but otherwise I think it's ok?

PR #2768, opened 2025-07-01
Added the private auth complete route for no apisix

rhysyngsun:

N/A

Adds the private auth complete view to a url that apisix won't intercept for auth.

### How can this be tested?

Run the app, the url `/_/auth/complete` should resolve to a HTTP 200 response (there will be no content)


pdpinch:
@blarghmatey is adding this route as an allowed redirect URL in production now.

PR #2772, opened 2025-07-02
Allow program as requirement or elective on another program

cp-at-mit:
https://github.com/mitodl/hq/issues/7697

Allows a Django Admin user to add a program as an elective or requirement within another program.

### How can this be tested?

1. Create two programs.
2. Within Django Admin, add one of the programs as an elective or requirement within the other program.


annagav (approved):
LGTM

PR #2782, opened 2025-07-09
Revert "Revert recent CSRF-related changes"

gumaerc:
Reverts mitodl/mitxonline#2776


pdpinch:
How will we avoid the problems we had two weeks ago with setting the csrftoken on a wildcard domain?

PR #2788, opened 2025-07-11
/api/users/me API appears to be returning email address

cp-at-mit:
https://github.com/mitodl/hq/issues/7792

Updates the user serializer to use the edx_username serializer method for the user's username.

### How can this be tested?

1. Visit http://mitxonline.odl.local:8013/admin/users/user/ and take not of the user's edx username.
2. Visit http://mitxonline.odl.local:8013/api/users/me and ensure that the username shown matches the edx username from step 1.

### DIFF
--- a/cms/api_test.py
+++ b/cms/api_test.py
@@ -238,7 +238,15 @@ def test_home_page_featured_products_sorting(mocker):
     )
     featured_products = home_page.products
     assert len(featured_products) == 2
-    assert featured_products == page_data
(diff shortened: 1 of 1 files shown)


gumaerc:
I'm not sure if I'm doing something wrong, but I'm logged in as the admin user (`admin@odl.local`) and I'm seeing an email address instead of my edx username:

chrome_UuxQ3fvhhl
chrome_kvWXrsRF9J

Do I need to create a new user to do this?

PR #2793, opened 2025-07-15
Handle edx_username outside transaction

rhysyngsun:

N/A

This solves a deadlock situation where we were updating the `edx_username` inside of the transaction but that causes openedx's registration API to not be able to see that value when it makes a call to `/user/me` because we haven't committed it to the DB. This moves that code into an equivalent `update()` query outside of the transaction. It also adds a check inside that transaction to bail if `edx_username` is `None` and a similar check to `create_edx_auth_token` that we now need too because it's not guaranteed that method gets called and the user exists in openedx anymore because the username isn't set nor has the user been synced.

### How can this be tested?

You should be able to register for a user locally, the user should be created in openedx, and their social auth uid should match their `edx_username` and NOT be `"None"`.

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -108,7 +108,7 @@ def create_edx_user(user, edx_username=None):
             user=user, platform=PLATFORM_EDX
         )
 
-        if open_edx_user.edx_username:
+        if not open_edx_user.edx_username:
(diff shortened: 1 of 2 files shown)


jkachel (approved):
👍

PR #2802, opened 2025-07-17
fix: handle duplicate username errors from edX by retrying user creation

asajjad2:
[#7748](https://github.com/mitodl/hq/issues/7748)

This PR improves user creation in Open edX by handling cases where the requested username is already taken in edX. When the edX API returns a `409` response with a `duplicate-username` error and provides username suggestions, the system will now automatically retry user creation using the first suggested username.

- Enhanced create_edx_user logic:
    - Detects 409 responses from the edX registration endpoint indicating a duplicate username.
    - If the response includes username_suggestions, the function updates the username to the first suggestion and retries the registration once.
    - If the retry also fails or no suggestions are provided, the original error is raised as before.

### How can this be tested?

1. Create a new user in edX directly with a unique username and email
2. Register a new user in MITxOnline with the same username as the one you just created in Open edX, but use a different email
3. Verify that a new user is created in both MITxOnline and edX but with a slightly different username than the original one


openedx/api.py
```diff
@@ -84,6 +84,76 @@ def create_user(user, edx_username=None):
...
+        OpenEdxUserCreateError: if user creation fails
+    """
+    req_session = requests.Session()
+    if settings.MITX_ONLINE_REGISTRATION_ACCESS_TOKEN is not None:
+        req_session.headers.update(
+            {ACCESS_TOKEN_HEADER_NAME: settings.MITX_ONLINE_REGISTRATION_ACCESS_TOKEN}
+        )
+
+    tried_suggestion = False
+    while True:
```
rhysyngsun, line 109: We should limit this to a certain number of attempts because we don't want this to run indefinitely. It should run some configurable number of times (a default of 5 is probably fine) and if it hasn't been able to successfully register the user with a suggested username log an error.

asajjad2: edX only returns us 3 suggested usernames similar to a response I've posted [here](https://github.com/mitodl/hq/issues/7748#issuecomment-3056428881). In the current implementation, I'm assuming that all of the suggested usernames are valid responses from edX so we just try again once with the first suggested username and if that also fails, we throw an error.

If you want I could update it to try all 3 before it throws an error?

rhysyngsun: I guess I didn't quite read your implementation over well enough to see that. My concern is more about just the `while True` loop itself, they're always an opportunity for some small logical tweak later to cause an infinite loop and I think my concern is because this function is just complicated enough that could happen.

asajjad2:
@rhysyngsun @cp-at-mit I've now updated the code to loop over all suggestions if one keeps failing after the other - and log the error if all options are exhausted

rhysyngsun (approved):

PR #2804, opened 2025-07-17
Collections model

cp-at-mit:
https://github.com/mitodl/hq/issues/7697

Adds in the ProgramCollection model which can be managed through the CMS and Django Admin.

### How can this be tested?

1. Use this branch and run your migrations.
2. Create two programs
3. Visit http://mitxonline.odl.local:8013/cms/snippets/courses/programcollection/
4. Create a new ProgramCollection.  Select both programs, provide a title and description.
5. Visit http://mitxonline.odl.local:8013/api/v2/program-collections/ and ensure that the program collection is displayed.
6. You can also view the ProgramCollection Django Admin page to ensure that it's displaying correctly.


jkachel (changes requested):
The basic functionality here looks good. I think this needs to be a full-fat Wagtail Page though; that way non-engineer people can manage these through the Wagtail UI. The root container page can just live under the Home page.

jkachel (approved):
LGTM :+1:

PR #2809, opened 2025-07-21
feat: enable wagtail API

asadali145:

https://github.com/mitodl/hq/issues/7865

Enables Wagtail APIs

### How can this be tested?

- Checkout this branch
- Create some data like Courses, Course Runs, Programs, Program Runs, Course Pages in CMS with certificates, instructors, program pages etc
- Now the pages API is available at `/api/v2/pages/`
  - You can follow [Wagtail API Usage Guide ](https://docs.wagtail.org/en/stable/advanced_topics/api/v2/usage.html)
  - This base URL will return all pages.
  - To get a list of CoursePage objects with all fields
    - `/api/v2/pages/?fields=*&type=cms.CoursePage`
  - To get a list of ProgramPage objects with all fields
    - `/api/v2/pages/?fields=*&type=cms.ProgramPage`
  - To get a list of CertificatePage objects with all fields
    - `/api/v2/pages/?fields=*&type=cms.CertificatePage`
  - To get certificate data of specific revision
    - `/api/v2/pages/<CertificatePageID>/?fields=*&revision_id=<RevisionID>`
  - To filter CoursePage and ProgramPage on readable_ids, you can pass Readable ID in query param `readable_id=<ID>`
- Images API is `/api/v2/images/`
- Documents API is `/api/v2/documents/`

### DIFF
--- a/cms/wagtail_api/schema/views.py
+++ b/cms/wagtail_api/schema/views.py
@@ -8,7 +8,6 @@ from rest_framework.views import APIView
 
 from .serializers import (
     CertificatePageListSerializer,
-    CertificatePageSerializer,
     CoursePageListSerializer,
     PageListSerializer,
     ProgramPageListSerializer,
@@ -134,7 +133,13 @@ class WagtailPagesRetrieveSchemaView(APIView):
                 location=OpenApiParameter.PATH,
                 required=True,
                 description="ID of the Wagtail page",
-            )
+            ),
+            OpenApiParameter(
+                name="revision_id",
+                required=False,
+                type=int,
+                description="Optional revision ID to retrieve a specific revision of the page",
+            ),
         ],
         responses={
             200: OpenApiResponse(
@@ -149,37 +154,6 @@ class WagtailPagesRetrieveSchemaView(APIView):
                 description="Returns a page of any known Wagtail page type",
             )
         },
-    )
-    def get(self, request, id, *args, **kwargs):  # noqa: ARG002, A002
-        return Response(
-            {
-                "id": id,
-                "title": "Sample Page",
-                "meta": {
-                    "total_count": 1,
-                },
-            }
-        )
-
-
-class WagtailCertificatePagesRetrieveSchemaView(APIView):
-    """
-    View for retrieving details of a specific Certificate Page with schema documentation.
-    """
-
-    @extend_schema(
-        summary="Get Certificate Page and Revision Details",
-        description="Returns details of a specific certificate page by ID. Can be used to retrieve a specific revision if `revision_id` is provided.",
-        parameters=[
-            OpenApiParameter(
-                name="id",
-                type=int,
-                location=OpenApiParameter.PATH,
-                required=True,
-                description="ID of the Certificate page",
-            ),
-        ],
-        responses=CertificatePageSerializer,
     )
     def get(self, request, id, *args, **kwargs):  # noqa: ARG002, A002
         return Response(
--- a/cms/wagtail_api/schema/views.py
+++ b/cms/wagtail_api/schema/views.py
@@ -1,0 +1,93 @@
+"""
+Views for Wagtail API Schema
+"""
+
+from drf_spectacular.utils import OpenApiParameter, extend_schema
+from rest_framework.response import Response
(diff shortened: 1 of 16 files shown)


asadali145:
## Updates:

### OpenAPI Schema
- I had to add custom OpenAPI schema views using the custom view and serializers. Additionally, I had to override the default pages endpoint with a custom view for the schema. I have added the schema endpoints for Pages, Certificate pages, and Course Pages. I can add the program list schema view once this is approved.

### Certificate Revisions
**Question:** Why do we need to create an endpoint that can return revision data? If we need it for a specific user certificate, then why don't we add the revision data in the User Certificate API response?

Current Implementation: For now, I have used the page detail API, and now revision of a certificate page can be fetched like:
`/api/v2/pages/21/?fields=*&revision_id=18`

Here `21` is the ID of the Certificate page and `18` is the revision ID.

@jkachel Can you look this comment and review the PR again?

cms/wagtail_api/schema/views.py
```diff
@@ -0,0 +1,93 @@
...
+from rest_framework.views import APIView
+
+from .serializers import (
+    CertificatePageListSerializer,
+    CoursePageListSerializer,
+    PageListSerializer,
+)
+
+
+class WagtailPagesSchemaView(APIView):
```
jkachel, line 16: One (hopefully) minor thing: this doesn't expose a retrieve view in the spec. We may not need it necessarily for the certificates APIs (see review comment) but we will for other things.

jkachel (changes requested):
One more change and then I think we're good to go.

Regarding the certificate data: thank you for adding the revision parameter; it'll be useful in other cases too, I think. For certificates specifically, the plan was to make that API super basic and just return the certificate data, but given that we have the FK to the revision and that your code here adds the appropriate serializers, it wouldn't be a bad idea to include the revision page data in its response. So, will do that.

asadali145:
@jkachel, I have added retrieve view for Page and Certificate Page (Including revisions). Also, I have added the program list view schema as well.

Can you please review this once more?

openapi/hooks.py
```diff
@@ -114,3 +117,100 @@ def exclude_paths_hook(endpoints, **kwargs):  # noqa: ARG001
...
+        (
+            "/api/v2/pages/?fields=*&type=cms.ProgramPage",
+            "^api/v2/pages/?fields=*&type=cms.ProgramPage$",
+            "GET",
+            program_view,
+        )
+    )
+    endpoints.append(
+        (
+            "/api/v2/pages/<int>/",
```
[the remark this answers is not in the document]
asadali145: Added in https://github.com/mitodl/mitxonline/pull/2809/commits/b508b6e443865c97e2855431c1d46e5d78a756c9

openapi/hooks.py
```diff
@@ -114,3 +117,100 @@ def exclude_paths_hook(endpoints, **kwargs):  # noqa: ARG001
...
+        (
+            "/api/v2/pages/?fields=*&type=cms.ProgramPage",
+            "^api/v2/pages/?fields=*&type=cms.ProgramPage$",
+            "GET",
+            program_view,
+        )
+    )
+    endpoints.append(
+        (
+            "/api/v2/pages/<int:pk>/",
```
jkachel, line 202: Still issues - I was actually able to get this to work by using `{pk}` for the placeholder and then manually setting the `operation_id` in the `extend_schema` decorator (otherwise it complains about a conflict). Something similar should also work for the certificate retrieve below with the revision ID in it. As is, the client that gets generated ends up with this code in it:

```
const localVarPath = `/api/v2/pages/<int:pk>/`
                .replace(`{${"pk"}}`, encodeURIComponent(String(pk)));
```
which doesn't work (and neither does the Swagger UI). (For what it's worth, spectacular will replace pk with id - not sure why that is.) It might be helpful to grab the mitxonline-api-clients repo and build a client to see what it outputs too.

asadali145: @jkachel Can you take a look at the changes in https://github.com/mitodl/mitxonline/pull/2809/commits/d056a2e389aedf00da31a7d203e4aaee892ee98a?

PR #2811, opened 2025-07-21
 Disable create-account page when keycloak login is enabled

cp-at-mit:
https://github.com/mitodl/hq/issues/7841

Redirects the user to the Keycloak login page from the create account page when the following are set:
```
MITOL_APIGATEWAY_DISABLE_MIDDLEWARE=False
EXPOSE_OIDC_LOGIN=False
```

### How can this be tested?
Visit http://mitxonline.odl.local:9080/create-account/.
Ensure that you are redirected to the keycloak login page.


annagav:
I don't seem to get redirected:
Screenshot 2025-07-22 at 9 09 35 AM

PR #2816, opened 2025-07-22
Optimize create_featured_items

cp-at-mit:
NA

The create_featured_items function is being called every 10 minutes.  This PR optimizes the code so that it is more performant.

Specifically:
- Use set() when possible
- Add prefect for course__page and select_related for get_enrollable_courseruns_qs
- delete redis cache entry without first checking if it exists
- use random instead of database random which is more complex
- no need for list conversion of self_paced_course_ids
- Added error handling


annagav (approved):
👍

PR #2820, opened 2025-07-23
Fix overwriting of name field with blank values

rhysyngsun:

N/A

This fixes an issue where we were overwriting a non-empty `User.name` values with a blank string when the field was missing from the SCIM payload. It prevents data corruption by misconfiguration of the SCIM remote.

### How can this be tested?

You can test this by removing the `fullName` attribute from the scim remote in keycloak and performing a sync. In this branch is should not result in the name being set as blank.

### DIFF
--- a/users/adapters.py
+++ b/users/adapters.py
@@ -63,7 +63,7 @@ class LearnUserAdapter(UserAdapter):
         """
         super().from_dict(d)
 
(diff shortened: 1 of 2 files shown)


jkachel (approved):
LGTM 👍  with caveat that I don't have a SCIM setup but this looks pretty straightforward

PR #2828, opened 2025-07-24
add collections to program, show program type requirements

gumaerc:
Closes https://github.com/mitodl/hq/issues/7916
Closes https://github.com/mitodl/hq/issues/7915

This PR:
- Adds `collections` to the API response for `Program` objects, which contains an array of integer ids matching `ProgramCollection` objects that list the `Program` in its `programs` field
- Restructures the `requirements` field in the API response to include keys for both `courses` and `programs`, including `required` and `elective` under that for each type

### How can this be tested?
- Make sure you have some test `Program` objects in your database, at least one of which has another `Program` listed as a requirement
- Create a program collection in Wagtail at Root -> Home Page -> Program Collections
- Add some programs to the collection and click Publish
- Make sure your user is enrolled in the programs
- Visit the `Program` API endpoint at `/api/v2/programs/`
- Verify that on the programs that are part of collections, you see a `collections` array with the integer ID's of the associated collections
- Verify that the `requirements` node has the `Program` requirements you added on the appropriate `Program`

### DIFF
--- a/courses/serializers/v2/programs.py
+++ b/courses/serializers/v2/programs.py
@@ -86,6 +86,7 @@ class ProgramSerializer(serializers.ModelSerializer):
     """Program Model Serializer v2"""
 
     courses = serializers.SerializerMethodField()
+    collections = serializers.SerializerMethodField()
     requirements = serializers.SerializerMethodField()
     req_tree = serializers.SerializerMethodField()
     page = serializers.SerializerMethodField()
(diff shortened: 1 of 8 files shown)


courses/serializers/v2/programs.py
```diff
@@ -104,35 +105,75 @@ class ProgramSerializer(serializers.ModelSerializer):
     def get_courses(self, instance) -> list[int]:
         return [course[0].id for course in instance.courses if course[0].live]
 
+    def get_collections(self, instance) -> list[int]:
+        return [
+            collection.id
+            for collection in ProgramCollection.objects.filter(programs__id=instance.id)
+        ]
+
     @extend_schema_field(
```
gumaerc, line 114: I'm not 100% certain that changing the structure of `requirements` here doesn't have any side effects. The unit tests are passing, but I'm particularly interested if someone with more experience in this codebase knows if this will be problematic. If so, we can change the approach.

PR #2832, opened 2025-07-25
Add certificate retrieval APIs

jkachel:

Closes mitodl/hq#7867

Add certificate retrieval APIs to MITx Online.

There are two, because there are two kinds of certificate:
- `/api/v2/course_certificate/` retrieves a course certificate
- `/api/v2/program_certificate/` retrieves a program certificate

These take the UUID that is returned from the enrollment API and return the certificate data. They should not require authentication. Both the user's specific certificate information is returned and the Wagtail page information for the page.

### How can this be tested?

Automated tests should pass.

Manual tests will require you to have a certificate page created, and a certificate issued for a user (for both a course and a program). Once you have that, you should be able to hit the API in the Swagger UI and get the certificate data back as expected. You should not need a user session to get data back - certificates should be viewable by anonymous users.

This won't return a rendered page; it's intended to be used by the front end to assemble the rendered certificate in the browser.

### DIFF
--- a/courses/serializers/v2/certificates.py
+++ b/courses/serializers/v2/certificates.py
@@ -1,10 +1,123 @@
 """Serializers for certificate data."""
 
 from rest_framework import serializers
-from wagtail.api.v2.serializers import PageSerializer
+from rest_framework.reverse import reverse
+from wagtail.api.v2.serializers import (
+    PageHtmlUrlField,
+    PageLocaleField,
+)
+from wagtail.models import Page
 
+from cms.models import CertificatePage
+from cms.wagtail_api.schema.serializers import (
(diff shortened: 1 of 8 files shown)


jkachel:
Made a couple of minor changes - the API wasn't set to AllowAny permission, so did that. I was looking at the output again too and realized that there's way too much user data being sent back, so swapped out the `UserSerializer` for `PublicUserSerializer`, which emits far less data (but still enough for rendering the certificate).

gumaerc (approved):
Based on our discussion, testing, and everything you see here, I'm approving this. As far as I can tell, everything is here that we need.

PR #2856, opened 2025-08-01
fix course creation in configure_instance

gumaerc:
Closes https://github.com/mitodl/hq/issues/7985

This PR fixes a bug in the `configure_instance` script in the part that creates the Edx Demonstration Course and E2E Test Course. Currently, these commands pass a single string to the `depts` argument on `create_courseware`. On the other end, in `create_courseware`, this is treated as an array. This causes the string to be split into its individual characters, and a `Department` object is created for each letter. Once it gets to the second course, it hits a letter that has already been used and a collision occurs. This PR simply passes arrays instead.

This is what Django admin for `Department` looks like currently after running `configure_instance`:
image
After this fix:
image

### How can this be tested?
- Spin up a fresh instance of `mitxonline`
- Set up OpenEdx and run `configure_instance` in the way described in the guide: https://github.com/mitodl/handbook/blob/master/openedx/MITx-edx-integration-tutor.md#mit-application-setup
- The command should not create errant one letter departments as shown in the above screenshot

### DIFF
--- a/main/management/commands/configure_instance.py
+++ b/main/management/commands/configure_instance.py
@@ -210,7 +210,7 @@ class Command(BaseCommand):
             "program-v1:MITx+DEDP",
             "Data, Economics and Development Policy",
             live=True,
-            depts="Economics",
+            depts=["Economics"],
             create_depts=True,
         )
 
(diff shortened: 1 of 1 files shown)


jkachel (approved):
LGTM 👍

rhysyngsun (approved):

PR #2861, opened 2025-08-05
Fix some issues with openedx user creation

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/7996
Fixes https://github.com/mitodl/hq/issues/7990

This makes a few significant changes to decrease errors reported to Sentry and fix conflicting state not allowing users to be created with suggested usernames.

- Removes the locking of the OpenEdxUser record so that we can update the `edx_username` and openedx can see that update. Opted for a redis-based lock instead to prevent a thundering herd of cron tasks.
- Reworked the username suggestion code to keep appending usernames (this simplified the code a bit)
- Put the number of suggested username attempts behind a configurable variable just in case.
- Added a `desired_edx_username` field to track the data the user entered or the one we generated separately from the one that's being used.
- Updated the implementation to set `edx_username` with suggested username before we call the Openedx registration API so that OpenedX sees that value in our userinfo API too.

### How can this be tested?

- Create a user and select a username that conflicts with an existing user


asajjad2:
I haven't been able to figure out APISIX issue on my local MITxOnline setup so haven't tested it but the logic looks good to me. The only issue I could think of is the current implementation (might just be a rare case) may crash with an `IntegrityError` if Open edX suggests a username that already exists locally. When the code saves 
```
open_edx_user.edx_username = current_username
```
before the API call, it doesn't check local uniqueness which would cause the function to crash instead of gracefully continuing to the next suggestion.

PR #2862, opened 2025-08-05
remove id requirement from program enrollment listing view

gumaerc:
Closes https://github.com/mitodl/hq/issues/8012

This PR adds an explicit `retrieve` view to `UserProgramEnrollmentsViewSet`, setting the required `id` parameter on that and the `destroy` view. The `list` endpoint had the parameter removed, so you can query it without arguments and get all of the program enrollments for the current user.

### How can this be tested?
- Spin up `mitxonline` on this branch
- Create some program enrollments for your user
- Visit `/api/schema/swagger-ui/#/program_enrollments/program_enrollments_list`
- Verify that the listing view requires no parameters and returns all your program enrollments
- Verify that the retrieval and destroy endpoints still require ID and function

### DIFF
--- a/courses/views/v1/__init__.py
+++ b/courses/views/v1/__init__.py
@@ -475,21 +475,21 @@ class UserEnrollmentsApiViewSet(
             raise NotImplementedError
 
 
-@extend_schema(
-    responses={200: UserProgramEnrollmentDetailSerializer},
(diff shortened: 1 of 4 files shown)


annagav (approved):
👍

PR #2868, opened 2025-08-08
fix: ensure unique edx_username before save

asajjad2:
[#7980](https://github.com/mitodl/hq/issues/7980)

The [reconcile_edx_username](https://github.com/mitodl/mitxonline/blob/a513e3f320584989c7cb6ad215bd1c022266b468/openedx/api.py#L257) function was throwing `IntegrityError` when trying to save a username that already existed in the local OpenEdxUser table. This occurred because the function didn't check for local uniqueness before saving, causing crashes instead of graceful handling.

This PR adds `_ensure_unique_edx_username` helper function to centralize username uniqueness logic and updated both `reconcile_edx_username` and `create_edx_user` to use it, preventing IntegrityError crashes.

### How can this be tested?
1. Register a user with name "Mike Smith" and email "mike.smith@example.com"
   - Manually set their edx_username to "mike-smith" in Django admin or database
   - Enroll them in a course
   - Verify enrollment succeeds and they have edx_username "mike-smith"

2. Register another user with name "Mike Smith" and email "mike.smith1@example.com"
    - Important: Before enrolling, ensure they don't have an edx_username set

3. Login as the second user (mike.smith1@example.com)
    - Try to enroll in a course
    - Verify enrollment succeeds
    - Check that they now have edx_username "mike-smith1" (or similar variant)

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -129,6 +129,26 @@ def _extract_username_suggestions(data, suggestions_extracted):
     return [], suggestions_extracted
 
 
+def _ensure_unique_edx_username(desired_username):
+    """
+    Ensure the desired username is unique in the OpenEdxUser table.
+
+    Args:
+        desired_username (str): The desired username
+
+    Returns:
+        str: A unique username, either the original or a modified version
+    """
+    if OpenEdxUser.objects.filter(edx_username=desired_username).exists():
+        return _find_available_username(
+            desired_username,
+            model=OpenEdxUser,
+            username_field="edx_username",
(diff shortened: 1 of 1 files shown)


rhysyngsun:
Note: I also opened https://github.com/mitodl/mitxonline/pull/2869 which is probably what ultimately caused this but I think what you have here is still worth it for an extra layer to ensure the user gets some kind of username.

asajjad2:
@rhysyngsun can we merge this?

rhysyngsun:
@asajjad2 go for it

PR #2881, opened 2025-08-18
Stop sync on openedx user if validation errors occur

rhysyngsun:

Closes https://github.com/mitodl/hq/issues/8187

If a validation error occurs this means the user has to change their data before we can possibly get a different outcome. This PR adds fields and code to track when an error occurs and skip that user on the next attempt. Admin interfaces added to manage this and reset the errors so it will retry. 

I also removed some of the existing validation on the mitxonline end so that we consistently just rely on the validation responses from openedx.

I also refactored the task, command, and api function that loops over the users to a) not return the list of users and b) chunk iterate over these users. This is to address memory issues we will hit in production where there are ~300k users that we need to attempt a repair for but likely will get an error for.

"Clear Sync Errors" button on top right corner:
Screenshot 2025-08-27 at 14-19-18 rhysyngsun errortest1@gmail com Change user Django site admin
Display of error response:
Screenshot 2025-08-27 at 14-19-08 rhysyngsun errortest1@gmail com Change user Django site admin

### How can this be tested?

- Create a user in django admin, add a Profile and LegalAddress record (you need to fill out at least one field for django-admin to save the record). Purposely leave a field like email or full name blank
- The celery task should then run, you can kick it off manually or set `REPAIR_OPENEDX_USERS_FREQUENCY` to a low value like 60 (seconds).
- You should see the celery task run and write out a debug line saying it synced your user
- There should now be an error on your OpenEdxUser and the celery task will not reattempt to sync that user.
- You can click the "Clear Sync Errors" button to clear that state and the task will retry the user


jkachel (approved):
LGTM 👍

PR #2884, opened 2025-08-19
Add flag to disable user repair task

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/8207

This adds an environment variable that disables the repair task because it's being problematic right now and unable to complete the bulk of its work until https://github.com/mitodl/mitxonline/pull/2881 lands and we can catch up.

### How can this be tested?

Set `REPAIR_OPENEDX_USERS_FREQUENCY` to something low like `60` (seconds) and when `DISABLE_USER_REPAIR_TASK` is set to true you should see the message "Skipping repair_faulty_openedx_users task as it is disabled" in the logs, if it's false then you should see the task run successfully

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -1503,3 +1503,9 @@ OPENTELEMETRY_EXPORT_TIMEOUT_MS = get_int(
     default=5000,
     description="Timeout for opentelemetry export",
 )
+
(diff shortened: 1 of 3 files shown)


jkachel (approved):
LGTM :+1:

PR #2896, opened 2025-08-25
feat: use courselist for course sync

asajjad2:
[#7160](https://github.com/mitodl/hq/issues/7160), [#2864](https://github.com/mitodl/mitxonline/pull/2864)

This PR adds an additional check on top of the changed made in this [reverted PR](https://github.com/mitodl/mitxonline/pull/2864)

#### Context:

edX API rejects malformed course keys with 400 errors. Many courses in MITxOnline have invalid courseware IDs that don't follow the required format `course-v1:ORGANIZATION+COURSE_ID+RUN_ID`. This PR adds the validation (on top of the original changes to enable use of courselist API) to filter out malformed course keys before making API calls, preventing 400 errors and allowing the sync to continue with valid courses.

### How can this be tested?

In addition to the original testing steps, the following steps can be taken to verify system behavior in the presence of invalid keys:

1. Add a course with an invalid course key via django-admin e.g. `course-v1:NotMITx-0.000x-4T2029`
2. Run management command: Execute `python manage.py sync_courseruns` to test bulk sync functionality
    - Verify all valid course keys are sent to edX and some courses are updated 
    - Verify invalid course keys are logged for identification
3. Check Celery task: Trigger the `sync_courseruns_data` task and verify it also skips the invalid course


annagav (approved):
LGTM 👍

PR #2900, opened 2025-08-26
Update hypothesis to 4.57.1

annagav:
N/A

 - Installing sortedcontainers (2.4.0)
  - Updating edx-api-client (1.11.0 -> 1.12.0)
  - Updating hypothesis (4.23.9 -> 4.57.1)
  - Installing python-redis-lock (4.0.0)

### How can this be tested?
nothing should break


rhysyngsun:
@annagav if it's easier, we should just rip all the legacy auth code out which uses hypothesis for the tests

PR #2903, opened 2025-08-27
Add an IsAdminOrReadOnly permission set for contracts/orgs

jkachel:
N/A

https://github.com/mitodl/mit-learn/pull/2453 in Learn updates the org dashboard to hit the contracts page API, which requires an admin user; didn't catch this in testing but it breaks the org dashboard if you're not an admin. So, adding a permissions class to allow read-only if authed, read/write if staff permissions.

### How can this be tested?

Should be able to hit `/api/b2b/v0/contracts` and `organizations` as a non-admin, but won't be able to write to them.
Should be able to write to the above if an admin.


jkachel:
fwiw it's probably less important that the admin read/write ability is working; doubt it will have issues but if admins can only read, that's fine for now

PR #2906, opened 2025-08-27
Save generated username in desired field too since locking uses this

jkachel:

#2893 

`create_edx_user_request` uses the `desired_edx_username` field but the changes I made to `reconcile_edx_username` doesn't update that field, just the actual edx_username field, so it seems to get confused a bit.

### How can this be tested?

Test should pass, creating a user in edX that had to be generated should also work.

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -368,9 +368,11 @@ def reconcile_edx_username(user, *, desired_username=None):
                     )
 
                 if not OpenEdxUser.objects.filter(
-                    edx_username=username_to_try
+                    edx_username=username_to_try,
+                    desired_edx_username=username_to_try,
                 ).exists():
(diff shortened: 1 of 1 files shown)


openedx/api.py
```diff
@@ -368,9 +368,11 @@ def reconcile_edx_username(user, *, desired_username=None):
                     )
 
                 if not OpenEdxUser.objects.filter(
-                    edx_username=username_to_try
+                    edx_username=username_to_try,
+                    desired_edx_username=username_to_try,
```
rhysyngsun, line 372: I don't think you need to filter on this because in your previous PR the unique constraint was removed so it's expected there might be duplicates here.

rhysyngsun (changes requested):

PR #2912, opened 2025-08-28
Skip onboarding if the user has any enrollments

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/8311

This adds code that skips onboarding if the user has any enrollments.

### How can this be tested?

For an existing user that has at least 1 enrollment, set `completed_onboarding` to `False` on their profile and login. They should not be redirected to onboarding. Optionally, you can test this behavior with the same scenario but by clicking the "Continue to Course" button on the learn dashboard. Both are logically identical they just differ in the user's destination.

### DIFF
--- a/authentication/api_gateway/views.py
+++ b/authentication/api_gateway/views.py
@@ -129,16 +129,18 @@ class GatewayLoginView(View):
         GET endpoint for logging a user in.
         """
         redirect_url = get_redirect_url(request)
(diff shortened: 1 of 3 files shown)


jkachel (approved):
LGTM 👍 set up a new user and ran through the one-click process to generate an enrollment via Learn; did not get sent back into MITx Online for onboarding

PR #2926, opened 2025-09-09
fix: create edx username when user enrolls without edx_username

asadali145:

Fixes https://github.com/mitodl/hq/issues/8356

Creates open edx user object with the edx_username when user enrolls without filling the profile form.

### How can this be tested?

- Create a user and do not fill profile form
- Enroll in a course and confirm that the Open edx user exists in mitxonline app with a valid edx_username at `/admin/openedx/openedxuser/`

### DIFF
--- a/openedx/api_test.py
+++ b/openedx/api_test.py
@@ -56,7 +56,6 @@ from openedx.exceptions import (
     EdxApiEmailSettingsErrorException,
     EdxApiEnrollErrorException,
     EdxApiRegistrationValidationException,
-    OpenEdxUserMissingError,
     UnknownEdxApiEmailSettingsException,
     UnknownEdxApiEnrollException,
     UserNameUpdateFailedException,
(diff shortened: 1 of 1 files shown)


openedx/api.py
```diff
@@ -931,10 +930,10 @@ def enroll_in_edx_course_runs(
...
 
     if not user.openedx_users.exists() or not user.edx_username:
-        msg = (
-            f"User {user} has no Open edX user (or no username): '{user.edx_username}'"
-        )
-        raise OpenEdxUserMissingError(msg)
+        from openedx.api import reconcile_edx_username
+
+        if reconcile_edx_username(user):
+            user.refresh_from_db()
```
rhysyngsun, line 936: This code was actually recently removed in https://github.com/mitodl/mitxonline/pull/2893 because it was causing errors with the b2b enrollment code, so bringing it back here will cause a regression with that. The consensus at the time was that whatever code was calling `enroll_in_edx_course_runs` needed to independently ensure that `openedx.api.create_user` (as oppossed to `reconcile_edx_username`) should be called by whatever code was calling `enroll_in_edx_course_runs`, ensuring it isn't called from within a transaction.

asadali145: Thank you, we have a couple of options to resolve this current issue.

- Call `reconcile_edx_username` in the [enrollment API](https://github.com/mitodl/mitxonline/blob/main/courses/views/v1/__init__.py#L330) before calling `create_run_enrollments`. This will fix the issue for the enrollments.
- Call `reconcile_edx_username` in the `create_edx_enrollments`. This has multiple uses, such as commands and old APIs.
- Catch the current OpenEdxUserMissingError exception and ask the user to fill out the profile form.

What do you suggest?

@pdpinch what are your thoughts on this?

rhysyngsun:
I put up this PR which does what I think we need the API to do, I haven't had a chance to test because I'm actually having trouble reproducing the bug locally though: https://github.com/mitodl/mitxonline/pull/2928

PR #2928, opened 2025-09-09
Fix user creation on enrollment

rhysyngsun:

- [ ] Desktop screenshots

### DIFF
--- a/users/models.py
+++ b/users/models.py
@@ -193,7 +193,7 @@ class UserManager(BaseUserManager):
         fields = {
             **extra_fields,
             "email": email,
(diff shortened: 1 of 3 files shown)


users/models.py
```diff
@@ -378,6 +379,12 @@ def should_skip_onboarding(self):
             or self.courserunenrollment_set(manager="all_objects").exists()
         )
 
+    @property
+    def openedx_user_exists(self):
```
asadali145, line 383: I have renamed from `edx_user_exists` to `openedx_user_exists ` and also fixed the usages.

PR #2932, opened 2025-09-11
Fix django admin

rhysyngsun:

N/A

This fixes an issue with trying to set the username for users who haven't had their openedx user created yet. It always overrides `desired_edx_username` to be the value of `edx_username`.

### How can this be tested?

- Create a user but don't input an edx username, check their record in django-admin and enter an edx_username.
- Check the database directly and confirm that the `desired_edx_username` field also updated to the same value.


jkachel (approved):
LGTM :+1:

PR #2942, opened 2025-09-15
fix: handle non-ascii username collisions

asajjad2:
[#8527](https://github.com/mitodl/hq/issues/8527)

When OpenEdX returns empty username suggestions for usernames containing non-ASCII characters (like "Nguyễn"), the system was raising an `OpenEdxUserCreateError` instead of falling back to local username generation. This fix implements a fallback mechanism that generates unique usernames locally using random number suffixes when OpenEdX suggestions are unavailable.

#### Changes

- Added `_generate_unique_username()` helper function to modularize username generation logic
- Updated `_create_edx_user_request()` to fall back to local username generation when OpenEdX returns empty suggestions
- Refactored `reconcile_edx_username()` to use the same modular username generation function

### How can this be tested?

1. Create a user with non-ASCII username: `Nguyễn`. Verify the user gets created both in MITxOnline and edX.
2. Try to create another user with the same username
3. Verify the behavior:
    - The second user should be created successfully (not fail with an error)
    - The second user should get a unique username with a random suffix
    - No OpenEdxUserCreateError should be raised
    - Both users should exist in the system with different usernames

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -127,6 +127,39 @@ def _extract_username_suggestions(data, suggestions_extracted):
     return [], suggestions_extracted
 
 
+def _generate_unique_username(base_username, max_length=OPENEDX_USERNAME_MAX_LEN):
+    """
+    Generate a unique username by appending random numbers to the base username.
+    
+    Args:
+        base_username (str): The base username to use
+        max_length (int): Maximum length for the generated username
+        
+    Returns:
+        str or None: A unique username, or None if no unique username could be generated
+    """
+    ranges = (
+        (0, 9),
+        (10, 99),
+        (100, 999),
+        (1000, 9999),
+        (10000, 99999),
+    )
+    
+    for intrange in ranges:
+        random_int = random.randint(intrange[0], intrange[1])  # noqa: S311
+        username_to_try = f"{base_username}_{random_int}"
+        
+        if len(username_to_try) > max_length:
+            amount_to_truncate = len(username_to_try) - max_length
+            username_to_try = f"{base_username[:amount_to_truncate]}-{random_int}"
+        
+        if not OpenEdxUser.objects.filter(edx_username=username_to_try).exists():
+            return username_to_try
+    
+    return None
+
+
 def _create_edx_user_request(open_edx_user, user, access_token):  # noqa: C901
     """
     Handle the actual user creation request to Open edX with retry logic for duplicate usernames.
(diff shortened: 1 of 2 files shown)


openedx/api.py
```diff
@@ -127,6 +127,81 @@ def _extract_username_suggestions(data, suggestions_extracted):
...
+        data, suggestions_extracted
+    )
+    if suggestions:
+        suggested_usernames = suggestions
+
+    if not suggested_usernames:
+        log.info(
+            "OpenEdX returned empty username suggestions, falling back to local generation"
+        )
+        base_username = open_edx_user.desired_edx_username or user.username
```
rhysyngsun, line 193: `User.username` will be the email address, but I don't think we want to use the entire thing because I think openedx rejects usernames with `@` in them. We could use the local part though, but I would also probably access this through `email` just so it's clear to readers of this code why the split on `@` is happening.

openedx/api.py
```diff
@@ -316,31 +396,13 @@ def reconcile_edx_username(user, *, desired_username=None):
...
-                    edx_user.desired_edx_username = username_to_try
-                    edx_user.save()
-                    break
+            unique_username = _generate_unique_username(edx_username)
+            if unique_username:
+                edx_user.edx_username = unique_username
+                edx_user.desired_edx_username = unique_username
+                edx_user.save()
+            else:
+                log.warning("Could not generate unique username for %s", edx_username)
```
rhysyngsun, line 405: This should be `error` so it shows up in sentry.

openedx/api_test.py
```diff
@@ -24,6 +24,7 @@
     ACCESS_TOKEN_HEADER_NAME,
     OPENEDX_AUTH_DEFAULT_TTL_IN_SECONDS,
     OPENEDX_REGISTRATION_VALIDATION_PATH,
+    _generate_unique_username,
```
rhysyngsun, line 27: Since we're testing this directly, it should drop the `_` prefix for private.

rhysyngsun (changes requested):

jkachel (approved):
This seems to work? I created two accounts, both with Nguyễn in the username:
- student30@odl.local had desired edX username Nguyễn30
- student31@odl.local had desired edX username Nguyễn30

For the first account, I wasn't able to get the account to sync to edX without manually changing the desired username to "Nguyen30". I did this before creating the second account

When I created the second account, the system seemed to grasp that there was a collision and lowercased and appended `_4` to the username. This still didn't sync though because it had a character that was invalid for edX in it.

So, long story short, this seems to generate the username OK but there's some string cleaning that (I think) used to be there that isn't now that makes this sort of hard to test. Not sure if that's in scope for this PR or if there's something else out there to take care of the sanitization part.

PR #2945, opened 2025-09-16
Make validation for username field work

annagav:
Fix https://github.com/mitodl/hq/issues/8498

The validation for username field was not working properly before.
This PR fixes it, and adds more special character validation.

Screenshot 2025-09-16 at 11 22 52 AM

### DIFF
--- a/frontend/public/src/components/forms/new_ProfileFormFields.js
+++ b/frontend/public/src/components/forms/new_ProfileFormFields.js
@@ -34,7 +34,7 @@ export const legalAddressValidation = yup.object().shape({
     .min(3, usernameFieldErrorMessage)
     .max(30, usernameFieldErrorMessage)
     .matches(
-      /^[^!@#$%^&*()+=[]{};':"\\|,<>\/?~]*$/,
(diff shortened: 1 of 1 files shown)


frontend/public/src/components/forms/new_ProfileFormFields.js
```diff
@@ -25,7 +26,17 @@ export const NAME_REGEX_FAIL_MESSAGE =
...
-  name:          yup.string().required().label("Full Name").min(2).max(254),
+  name:     yup.string().required().label("Full Name").min(2).max(254),
+  username: yup
+    .string()
+    .required()
+    .label("Public Username")
+    .min(3, usernameFieldErrorMessage)
+    .max(30, usernameFieldErrorMessage)
+    .matches(
+      /^[^!@#$%^&*()+=[]{};':"\\|,<>\/?~]*$/,
```
jkachel, line 37: This specifies a character class that contains `[]` - the closing square bracket isn't escaped, though, so the regex fails.

jkachel (changes requested):
Needs a backslash, otherwise seems to work OK.

PR #2946, opened 2025-09-16
Use absolute URLs for feature_image_source

ChristopherChudzicki:
- For https://github.com/mitodl/hq/issues/8414

Makes `feature_image_src` a fully qualified URL so we can use it on other sites (like Learn)

### How can this be tested?
Check that:
- images still work as expected within MITxOnline
- images in the wagtail API responses, e.g., http://api.learn.odl.local:8065/mitxonline/api/v2/pages/?fields=*&type=cms.CoursePage&readable_id=course-v1%3AMITxT%2B7.28.1x-fake, include fully qualified URLs.

### DIFF
--- a/cms/api_test.py
+++ b/cms/api_test.py
@@ -92,13 +92,14 @@ def test_ensure_home_page_and_site():
 def test_get_wagtail_img_src(settings):
     """get_wagtail_img_src should return the correct image URL"""
     settings.MEDIA_URL = "/mediatest/"
-    img_path = "/path/to/my-image.jpg"
+    settings.SITE_BASE_URL = "http://mitxonline.test"
+    img_filename = "path/to/my-image.jpg"
     img_hash = "abc123"
     home_page = HomePageFactory.build(
-        hero__file__filename=img_path, hero__file_hash=img_hash
+        hero__file__filename=img_filename, hero__file_hash=img_hash
     )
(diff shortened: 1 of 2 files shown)


cms/api_test.py
```diff
@@ -92,13 +92,14 @@ def test_ensure_home_page_and_site():
 def test_get_wagtail_img_src(settings):
     """get_wagtail_img_src should return the correct image URL"""
     settings.MEDIA_URL = "/mediatest/"
-    img_path = "/path/to/my-image.jpg"
+    settings.SITE_BASE_URL = "http://mitxonline.test"
+    img_filename = "path/to/my-image.jpg"
     img_hash = "abc123"
     home_page = HomePageFactory.build(
```
ChristopherChudzicki, line 98: **Question:** In case anyone knows...

My understanding is that factory boy `build(...)` method does not save a model to the DB, so it shouldn't require database access. And, indeed, this test is not marked with `@pytest.mark.django_db`

But when I ran this test all by itself via
```
pytest cms/api_test.py::test_get_wagtail_img_src -sv --no-cov
```
I got an error that the test was accessing the db but not marked with `@pytest.mark.django_db`. But again, the test file passes if all the tests run...

```
pytest cms/api_test.py -sv --no-cov
```

Curious if anyone knows why.

jkachel (approved):
LGTM 👍

PR #2948, opened 2025-09-17
Add Keycloak organization sync

jkachel:

https://github.com/mitodl/hq/issues/7376
https://github.com/mitodl/hq/issues/8022

In order to completely move organization data into Keycloak, we need to be able to get it back _out_ of Keycloak. SCIM is not appropriate for this as it doesn't really understand the concept of organizations. So, instead, we need to use the Keycloak admin API.

This PR adds a handful of things:
- A set of dataclasses derived from the not necessarily release quality OpenAPI spec for the Keycloak admin API
- A relatively simple client class that provides a reasonable interface to the API
- Some helper methods and APIs to pull Keycloak organization data via the API, and then update/create records as appropriate in MITx Online

### How can this be tested?

There are new settings that need to be set:

For scheduled tasks:
- `KEYCLOAK_ORG_SYNC_FREQUENCY` - Controls the period for the organization sync task. Defaults to 1 day (86400).
- `KEYCLOAK_ORG_SYNC_OFFSET`  - Adds an offset to prevent collisions. Defaults to 43200 (1/2 day). 

For the API itself:
- `KEYCLOAK_BASE_URL` - the base URL for the Keycloak instance
- `KEYCLOAK_REALM_NAME` - the realm name to use - for local deployments, this will almost definitely be `ol-local`
- `KEYCLOAK_DISCOVERY_URL` - the OIDC discovery URL for the realm (available in Realm Settings in the Keycloak admin)
- `KEYCLOAK_ADMIN_CLIENT_ID` - the client ID to use for the admin client. May be the same as the client that APISIX uses, but you may also want a separate one.
- `KEYCLOAK_ADMIN_CLIENT_SECRET` - the client secret cor the admin client
- `KEYCLOAK_ADMIN_CLIENT_SCOPES` - OpenID scopes to request. This can generally be pretty minimal - `openid profile email` is good enough unless you have something specific you want to do.
- `KEYCLOAK_ADMIN_CLIENT_NO_VERIFY_SSL` - set to True to skip SSL certificate verification for the API. Good for local deployments; otherwise, should be set to False.

Then, create some orgs (at least one or two) in Keycloak (and ideally add members to them). In MITx Online, create some orgs - these should be different than the Keycloak ones - and at least one org that matches (by UUID) an org in Keycloak. 

Run `sync_keycloak_orgs` management command. The non-Keycloak orgs should be left alone; you should have one new one; and changes between Keycloak and MITx online should be resolved. 

Then, change one of the org's details in Keycloak and run the command again. The changes should be reflected in MITx Online.

### Additional Context

We map the name to the `title` and `name` attributes, the alias to `org_key`, the description to `description`, and we use the Keycloak UUID for the `sso_organization_id` field. We may want to add more fields or do something with the key/value attributes that Keycloak supports later.

This doesn't change any membership details for any users - it just makes sure the orgs are squared up between Keycloak and MITx Online. 

This doesn't sync any orgs that don't have a UUID. 

We will still have some work to do to make data model changes to complete 8022, and we'll need to add contract attachment updates (which are also not in this PR). (Additionally, the Keycloak client code should end up in ol-django as other apps will need to use it, but it will be easier to build out in situ.)


b2b/keycloak_admin_api.py
```diff
@@ -0,0 +1,427 @@
...
+            urljoin(self.base_url, url_path)
+            if skip_realmify
+            else self.realmify_url(url_path)
+        )
+
+        return (
+            self.oauth_session.request(method, request_url, **kwargs)
+            if self.oauth_session
+            else None
+        )
```
rhysyngsun, line 125: i think this might be worth splitting this into a `request`. `realm_request`, and `_request` (for the common stuff) so the behavior is split at the function level rather than needing to pass an arg, I think that documents the intent better.

b2b/keycloak_admin_api.py
```diff
@@ -0,0 +1,427 @@
...
+            client_id=client_id,
+            client_secret=client_secret,
+            token_endpoint=self.openid_configuration["token_endpoint"],
+            scope=settings.KEYCLOAK_ADMIN_CLIENT_SCOPES,
+            verify=not self.skip_verify,
+        )
+        self.token = self.oauth_session.fetch_token(
+            self.openid_configuration["token_endpoint"],
+            grant_type="client_credentials",
+        )
```
rhysyngsun, line 95: I don't know if this could be potentially long-running, but if it might be it's probably worth hooking into the automatic refreshing mechanism: https://requests-oauthlib.readthedocs.io/en/latest/oauth2_workflow.html#third-recommended-define-automatic-token-refresh-and-update

b2b/tasks.py
```diff
@@ -12,3 +12,11 @@ def queue_enrollment_code_check(contract_id: int):
 
     contract = ContractPage.objects.get(id=contract_id)
     ensure_enrollment_codes_exist(contract)
+
+
+@app.task()
```
rhysyngsun, line 17: Should this be `acks_late=True` because if the worker gets killed in the middle of a task it won't rerun for possibly another 24 hours?

pyproject.toml
```diff
@@ -117,6 +119,7 @@ pytest-xdist = {extras = ["psutil"], version = "^3.6.1"}
 [tool.ruff]
 target-version = "py310"
 line-length = 88
+extend-exclude = [ "b2b/keycloak_admin_dataclasses.py" ]
```
rhysyngsun, line 122: Was this temporary or is there a reason to exclude the entire file from formatting/linting?

b2b/api.py
```diff
@@ -793,3 +794,51 @@ def reconcile_user_orgs(user, organizations):
...
+    """
+
+    org_model = get_keycloak_model(*KCAM_ORGANIZATIONS)
+    orgs = org_model.list()
+    parent_org_page = OrganizationIndexPage.objects.first()
+    created = 0
+    updated = 0
+
+    for org in orgs:
+        page = OrganizationPage.objects.filter(sso_organization_id=org.id).first()
```
rhysyngsun, line 818: Since this field isn't unique, should this just be a `get()` so we can catch if there's multiple pages w/ the same org id and log an error because that's an indication data needs to be fixed? Or should this field on `OrganizationPage` be unique?

rhysyngsun (changes requested):
Feedback on code review

PR #2951, opened 2025-09-18
Add support for filling the country of residence and income fields for financial assistance forms

jkachel:

Closes mitodl/hq#8337

Exposes the `country_of_income` and `country_of_residence` fields on the financial assistance form by adding a Country Code field and adjusting the submit and view workflows to pull the user's prior submissions for auto-filling these two fields.

Also, renames the existing Country field - which is _really_ a currency code field - to Country Currency.

### How can this be tested?

Automated tests should pass.

These two fields need to be added to the financial assistance form before they will appear:
- Country of Income
- Country of Residence

These fields should be set to the "Country Code" type, and named **exactly** as shown. 

Then, load the form. (You may need to reset any prior attempts in the database.) You should see the two new fields, and you should get a dropdown with the full name of the country in it in each field. 

Submit the request and make sure the fields are set as expected.


annagav (approved):
👍

PR #2953, opened 2025-09-18
respect the order of ID's passed as query string args to the courses API

gumaerc:
Closes https://github.com/mitodl/hq/issues/8626

This PR adds functionality to the `get_queryset` function of `CoursesViewSet`. To request multiple course details from this API, you need to pass an array of integer ID's to the `id` query string param. The `get_queryset` function was blanket ordering all results with `order_by("title")`, which would give alphabetical ordering. This PR ensures that if ID's are passed that the returned courses follow the order in which they were passed in.

### How can this be tested?
- Make sure you have 10+ courses labeled with numbers 1-10+ in their title. You can use the test courses created by `populate_course_data` for this.
- Get the integer `id` properties of these courses by looking in Django admin
- Visit the API at `/api/v2/courses/?id=` and after the equals sign add in the id's of your courses separated by commas
- Ensure that the courses are returned in the order you specified
- Swap around the ID's and ensure this remains the case

### DIFF
--- a/courses/views/v2/__init__.py
+++ b/courses/views/v2/__init__.py
@@ -5,7 +5,7 @@ Course API Views version 2
 import contextlib
 
 import django_filters
-from django.db.models import Count, Prefetch
+from django.db.models import Case, Count, Prefetch, When
 from django.shortcuts import get_object_or_404
 from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.types import OpenApiTypes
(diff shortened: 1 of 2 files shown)


courses/views/v2/__init__.py
```diff
@@ -254,15 +254,29 @@ class CourseViewSet(viewsets.ReadOnlyModelViewSet):
...
             .distinct()
         )
 
+        # If IDs are provided via filter, preserve their order
+        id_filter = self.request.query_params.get("id")
+        if id_filter:
+            try:
+                ids = [int(id_str.strip()) for id_str in id_filter.split(",")]
+                queryset = queryset.filter(id__in=ids).order_by(
+                    Case(*[When(id=pk, then=pos) for pos, pk in enumerate(ids)])
```
asadali145, line 271: nit: we can use postgres array_position like:
```suggestion
                    Func(ids, F("id"), function="array_position")
```

It will reduce execution time slightly and will also reduce the query size.

asadali145 (approved):
LGTM!

Just a small suggestion on the PR.

PS: Not related to this PR, the courses API performs quite a few N+1 queries, For my local setup I had 13 courses and it performed 148 queries and about 10 per course are repeated ones.

PR #2954, opened 2025-09-19
Add third title to SignatoryPage

annagav:
Fix https://github.com/mitodl/hq/issues/8610

Add third title to SignatoryPage

Screenshot 2025-09-19 at 1 38 49 PM

### DIFF
--- a/cms/migrations/0052_signatorypage_title_3.py
+++ b/cms/migrations/0052_signatorypage_title_3.py
@@ -4,15 +4,19 @@ from django.db import migrations, models
 
 
 class Migration(migrations.Migration):
-
     dependencies = [
-        ('cms', '0051_alter_formfield_field_type'),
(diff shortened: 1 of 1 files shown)


cms/migrations/0052_signatorypage_title_3.py
```diff
@@ -0,0 +1,22 @@
...
+        ("cms", "0051_alter_formfield_field_type"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="signatorypage",
+            name="title_3",
+            field=models.CharField(
+                blank=True,
+                help_text="Specify signatory thrid title in organization.",
```
asadali145, line 17: We need to recreate the migration.

asadali145 (changes requested):
There are failing tests, but I'm not sure if they are related to this PR. We need to update/recreate the migration.

PR #2967, opened 2025-09-24
make first_name / last_name not required in legal_address updates

gumaerc:
Closes https://github.com/mitodl/hq/issues/8690

This PR updates `LegalAddressSerializer` to allow partial `PATCH` requests to API endpoints that use it not specifying `first_name` or `last_name`. 

### How can this be tested?
- Spin up `mitxonline` on this branch
- Visit the home page, then click the menu in the top right and select Profile
- Open the development console in your browser and go to the Network tab
- Make sure the "preserve log" functionality is checked
- Fill out the required fields in the form and submit it
- In your network tab, you should see a `PATCH` request to `/api/users/me`
- Right click on this and click Copy -> Copy as fetch
- In the console section, paste the request we copied
- Edit the `body` part of the request to only update the country (and state if your country is US or CA) like `"body": "{\"legal_address\":{\"country\":\"US\",\"state\":\"US-MA\"}}",`
- Hit enter to submit the request and make sure it succeeds
- Run the request again, but this time change the country or the state or both
- Verify that the request succeeds and that when you refresh the profile page, the info is updated

### DIFF
--- a/users/serializers.py
+++ b/users/serializers.py
@@ -120,8 +120,9 @@ class LegalAddressSerializer(serializers.ModelSerializer):
         """Validate the legal address data"""
         errors = {}
 
-        # For CREATE operations (not partial updates), require first_name and last_name
-        if not self.partial:
+        # For POST operations (not partial updates), require first_name and last_name
+        request = self.context.get("request", None)
+        if request and request.method == "POST":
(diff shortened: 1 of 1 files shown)


jkachel (approved):
LGTM :+1:

PR #2970, opened 2025-09-25
Fix openedx user sync issues

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/8707
Fixes https://github.com/mitodl/hq/issues/8708

This fixes two issues that came up in RC:
- If a user has a `desired_edx_username` that conflicts with an existing `edx_username` value the whole operation fails due to the unique constraint
- If a user has both an email and an edx_username that conflicts, a different error is returned. The code correctly doesn't try any futher, but an error is logged to sentry instead of being recorded to the `OpenEdxUser` error data.
- Refactored a bit because the `create_edx_user_request` function was getting so big it was failing lint errors

### How can this be tested?

Test https://github.com/mitodl/hq/issues/8708
- Create a new user, skip the onboarding form. 
- Go to django admin and set a `desired_edx_username` value that conflicts with an existing `edx_username` value.
- Run the `repair_missing_courseware_records` command on that user, they should get a different numerically suffixed username.

Test  https://github.com/mitodl/hq/issues/8707
- Create a new user, skip the onboarding form.
- Change the email address to conflict with another user in openedx and do the same with `desired_edx_username`
- Run the `repair_missing_courseware_records` command on that user, they should get `has_sync_error` set to true and `sync_error_data` populated with the error message from openedx. This state will then exclude the user from any further attempts to repair them until their state is fixed (no need to test that, just an fyi).

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -113,9 +113,20 @@ def _is_duplicate_username_error(resp, data):
     )
 
 
-def _is_bad_request(resp):
+def _is_duplicate_email_username_error(resp, data):
+    """Check if the response indicates a duplicate username error."""
+    return (
+        resp.status_code == status.HTTP_409_CONFLICT
+        and data.get("error_code") == "duplicate-email-username"
+    )
+
+
+def _is_bad_request(resp, data):
     """Check if the response indicates a bad request."""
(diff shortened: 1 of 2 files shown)


annagav (approved):
LGTM

PR #2952, opened 2025-09-18
feat: update wagtail api permissions

asadali145:
### What are the relevant tickets?

https://github.com/mitodl/hq/issues/8627

### Description (What does it do?)

- All **_non-B2B_** `cms.coursepage` and `cms.programpage` pages are allowed anonymous access.
- B2B `cms.coursepage` and `cms.programpage` pages are NOT allowed anonymous access.
  - Both these types are accessible to authenticated users.
- ALL `cms.certificatepage` pages, **_including B2B_**, are ALLOWED anonymous access.
- All of this means that when Wagtail API is accessed with a specific page type query parameter, i.e., `?type=cms.coursepage|programpage|certificatepage`, then anonymous access will be allowed. Otherwise, we do not allow anonymous access.
- ALL other pages are accessible to the Authenticated users ONLY.

### How can this be tested?

- Test that anonymous user can access non-B2B course and program pages at `/api/v2/pages/?type=cms.courspage&fields=*` & `/api/v2/pages/?type=cms.programpage&fields=*`
  - Test that the authenticated user can access all course and program pages, including B2B.
  - B2B programs have `b2b_only=True` in the program object.
  - B2B course pages have `include_in_learn_catalog=True` in the CMS page. 
- Test that anonymous user can access all certificate pages at `/api/v2/pages/?type=cms.certificate&fields=*`, including B2B.
- Test that all other pages are only accessible to authenticated users.

PR #2975, opened 2025-09-29
Wagtail permissions and url fix

ChristopherChudzicki:
Small followup to https://github.com/mitodl/mitxonline/pull/2952

Makes url casing more consistent. See code comment.

### How can this be tested?
1. Check that `/api/v2/pages/?type=cms.courspage&fields=*` and `/api/v2/pages/?type=cms.CoursPage&fields=*` are both accessible to anonymous users, and that `/api/v2/pages/?type=cms.courspage&fields=*` shows up int he OpenAPI spec.
2. Similar for programs/certifies.

### DIFF
--- a/openapi/specs/v0.yaml
+++ b/openapi/specs/v0.yaml
@@ -1130,9 +1130,9 @@ paths:
                 - $ref: '#/components/schemas/CertificatePage'
                 - $ref: '#/components/schemas/Page'
           description: Returns a page of any known Wagtail page type
-  /api/v2/pages/?fields=*&type=cms.CertificatePage:
+  /api/v2/pages/?fields=*&type=cms.certificatepage:
     get:
-      operationId: pages_?fields=*&type=cms.CertificatePage_retrieve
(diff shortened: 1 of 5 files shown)


openapi/specs/v0.yaml
```diff
@@ -1130,9 +1130,9 @@ paths:
                 - $ref: '#/components/schemas/CertificatePage'
                 - $ref: '#/components/schemas/Page'
           description: Returns a page of any known Wagtail page type
-  /api/v2/pages/?fields=*&type=cms.CertificatePage:
+  /api/v2/pages/?fields=*&type=cms.certificatepage:
     get:
-      operationId: pages_?fields=*&type=cms.CertificatePage_retrieve
```
ChristopherChudzicki, line 1135: The URLs included in the API spec required admin permission because they weren't lowercase.

Now the permission check lowercases like the filter does, but also updated these to be lowercase

asadali145 (approved):
LGTM!

PR #2979, opened 2025-09-29
Courses API w/ org= handle anonymous

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/8750

While it's not expected, sometimes the courses API is called with `?org_id=` for an anonymous user, probably due to a session timeout coinciding with a dashboard load. This prevents a 500 error. Will address how to handle the anonymous use case more resiliently in a separate PR since it'll require a new API.

### How can this be tested?

This is a bit complicated to test, so I think I'd say the test I wrote is enough to cover it.


gumaerc (approved):
I wasn't able to reproduce this locally, but looking at the code based on our discussion about this I believe this is safe to merge. The change is small and straightforward, and will prevent an error in the case the MITx Online session somehow expires and returns an anonymous user.

jkachel (approved):
LGTM :+1:

PR #2986, opened 2025-10-01
feat: management commands to create signatories in bulk and delete a signatory

arslanashraf7:
https://github.com/mitodl/hq/issues/8666 & https://github.com/mitodl/hq/issues/8761

- Adds a management command to read the signatories from a CSV file and create them in CMS (https://github.com/mitodl/hq/issues/8666)
- Adds another management command that takes a signatory page id and removes it from all the existing certificate pages and saves a new revision  (https://github.com/mitodl/hq/issues/8761)
- Adds --dry-run option in both the commands so that we can see what will be created or updated via these commands without making any actual changes. This will help verify data before we actually change it.

- [ ] Desktop screenshots

### How can this be tested?
 - Setup MITx Online locally on this branch
 - Copy the sheet mentioned https://github.com/mitodl/hq/issues/8666 in CSV format in MITx Online web container
 - Run these commands to test bulk signatory create:
     -  ```./manage.py bulk_populate_signatories --csv-file=<CSV_FILE_PATH> --dry-run``` to test the dry run
     -  ```./manage.py bulk_populate_signatories --csv-file=<CSV_FILE_PATH> ```. You should see signatories being created in the terminal. Verify them in CMS, The signatory pages should be created under the Signatories Index page and all these newly created signatory pages should be live/published as well.
     - To test different scenarios, You should create a Signatory page from the sheet with the same naming convention the command will, And then run the command, and it should not be re-created/updated.
 - Run these commands to test the Delete Signatory command:
     - Create a certificate page with multiple signatories in it, This means that you have a working Certificate page that you can view in `http://mitxonline.odl.local:8013/certificate/<certificate_page_uuid>` . 
     - Now pick a signatory from the certificate page and get It's page ID in CMS (The ID can be found in the URL when you open the page)
     - ```./manage.py remove_signatory --signatory-id=<ID> --dry-run``` Verify the results, but there should be no actual changes made
     - ```./manage.py remove_signatory --signatory-id=<ID>``` Verify that the certificate page is updated and published
     - Now open the `http://mitxonline.odl.local:8013/certificate/<certificate_page_uuid>` again, It should still show you the old signatory in it. This is because we are removing the signatory for the future certificates. The ones generated already would still have it.

### Additional Context
I didn't add tests for these commands because I believe these might turn out to be one time things.

### DIFF
--- a/cms/management/commands/remove_signatory.py
+++ b/cms/management/commands/remove_signatory.py
@@ -1,0 +1,92 @@
+from django.core.management.base import BaseCommand
+from wagtail.blocks import StreamValue
+
+from cms.models import CertificatePage
+
+
+class Command(BaseCommand):
+    """Django management command to remove a signatory from all certificate pages.
+
+    This command finds all certificate pages that contain a specific signatory
+    and removes that signatory from their signatories field, creating new
+    revisions and publishing the updated pages.
+    """
+
+    help = "Remove a signatory from all certificate pages and create new revisions"
+
+    def add_arguments(self, parser):
+        """Parses command line arguments."""
+        parser.add_argument(
+            "--signatory-id",
+            type=int,
+            help="ID of the signatory page to remove",
+            required=True,
+        )
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Show which certificate pages would be updated without making changes",
+        )
+
+    def handle(self, *args, **options):  # noqa: ARG002
+        """Handles the command execution."""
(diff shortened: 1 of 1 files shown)


rachellougee (approved):
Looks good to me. I've tested both management commands, and they work fine. 

For `remove_signatory`, the only thing I see is if the removed signatory is the only one on the certificate page - this would leave the certificate page without any signatories. But it seems like there shouldn't be any certificate page with only Chris C's signatory, so this shouldn't be a problem

PR #3133, opened 2025-12-05
Upgrade django to 5.2

annagav:
Fix https://github.com/mitodl/hq/issues/8953

  - Updating django  5.2
  
### How can this be tested?
everything should work as expected

### DIFF
--- a/courses/models.py
+++ b/courses/models.py
@@ -2334,7 +2334,7 @@ class ProgramRequirement(MP_Node):
     path = models.CharField(max_length=255, unique=True, db_collation="C")
 
     node_type = models.CharField(  # noqa: DJ001
-        choices=ProgramRequirementNodeType.choices,
+        choices=ProgramRequirementNodeType,
         max_length=len(max(ProgramRequirementNodeType.values, key=len)),
         null=True,
     )
@@ -2344,7 +2344,7 @@ class ProgramRequirement(MP_Node):
         MIN_NUMBER_OF = "min_number_of", "Minimum # of"
 
     operator = models.CharField(  # noqa: DJ001
-        choices=Operator.choices,
+        choices=Operator,
         max_length=len(max(Operator.values, key=len)),
         null=True,
     )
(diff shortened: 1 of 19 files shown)


courses/models.py
```diff
@@ -2456,7 +2456,7 @@ class Meta:
             # validate the fields based on the node 'type'
             CheckConstraint(
                 name="courses_programrequirement_node_check",
-                check=(
+                condition=(
```
rhysyngsun, line 2432: There are 2 migrations - 0022 and 0063 that are still using `check=`

pyproject.toml
```diff
@@ -15,7 +15,7 @@ dependencies = [
     "click>=8.1,!=8.2.0",
     "deepdiff>=8.0.0,<9",
     "dj-database-url>=0.5.0,<0.6",
-    "django==5.1.15",
+    "django==5.2",
```
rhysyngsun: I think going forward it might be better to have this be defined as a constraint within a minor version, so `django>=5.2,<5.3`. That way whenever a new patch comes down only the lockfile needs to be updated.

rhysyngsun (changes requested):
Code review, will test shortly.

main/middleware.py
```diff
@@ -24,6 +24,10 @@ def process_response(self, request, response):
 
         return response
 
+    async def aprocess_request(self, request):
+        # Call the sync version for compatibility
+        return self.process_request(request)
+
```
Anas12091101: Why was this method added, and how is it expected to be invoked? I don’t think `aprocess_request` is a recognized hook in Django middleware.

Anas12091101 (commented):
I started reviewing this, but I see Nathan has already provided feedback and is testing it. I just have one small question; otherwise, I’ll leave the rest to him:

main/middleware.py
rhysyngsun: Good catch, I agree, I think it's something specific to `RemoteUserMiddleware` , it looks like `MiddlewareMixin` just calls `sync_to_async` on `process_request()` anyway: https://github.com/django/django/blob/stable/5.2.x/django/utils/deprecation.py#L125-L142

annagav: yes, that was one of the attempts to fix the error that were showing up, and carried over by accident.

rhysyngsun (approved):
LGTM

PR #3376, opened 2026-03-11
Prefer enrollable course runs for start date

cp-at-mit:
https://github.com/mitodl/hq/issues/10496

When computing the start date label for multi-run courses, prefer runs that are still open for enrollment (is_enrollable). If no enrollable runs exist, fall back to all runs. This avoids showing start dates from long-past, closed runs while preserving existing behavior for single-run courseware and still using compareCourseRunStartDates for sorting.

### How can this be tested?
Create a course with two course runs, one in the past and one currently open for enrollment.  Visit the catalog page and verify that the start date is for the course run that's currently open for enrollment.

### DIFF
--- a/frontend/public/src/lib/util.js
+++ b/frontend/public/src/lib/util.js
@@ -351,9 +351,24 @@ export const getStartDateText = (
     return ""
   }
 
-  const courseRun = courseware.courseruns ?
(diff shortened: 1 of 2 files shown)


annagav:
This is no longer an issue. Can we close this?

PR #3378, opened 2026-03-11
Only return enrollable course runs

cp-at-mit:
https://github.com/mitodl/hq/issues/10496

Adds a filter to only return enrollable course runs as part of the response from CourseWithCourseRunsSerializer

### How can this be tested?
Create a course with two course runs, one in the past and one currently open for enrollment. Visit the catalog page and verify that the start date is for the course run that's currently open for enrollment.


annagav:
could we close this?

PR #3426, opened 2026-03-26
fix: make page a nullable property and specs updated

marslanabdulrauf:
https://github.com/mitodl/hq/issues/10652

This pull request updates the course and program serializers to properly handle cases where a course or program does not have an associated CMS page. The serializers now return `null` for the `page` field instead of a default object, and the OpenAPI specs are updated to reflect this. Corresponding tests are added to ensure this behavior.

**Serializer improvements:**

* Updated `CourseSerializer` and `ProgramSerializer` to allow the `page` field to be `null` if there is no associated CMS page, returning `None` instead of a default object. (`courses/serializers/v2/courses.py` [[1]](diffhunk://#diff-766793b89a3fc4a84d693e68ead2b833ac2b70cd38c9da615776a9486476cc66L36-R36) `courses/serializers/v2/programs.py` [[2]](diffhunk://#diff-2b70908f850c80a4e0010b3fc13b1777a83315bd267737095e19a848d3bb7c45L454-R455)
* Cleaned up imports in `programs.py` by removing unused `get_thumbnail_url`. (`courses/serializers/v2/programs.py` [courses/serializers/v2/programs.pyL15-R15](diffhunk://#diff-2b70908f850c80a4e0010b3fc13b1777a83315bd267737095e19a848d3bb7c45L15-R15))

**OpenAPI specification updates:**

* Added `nullable: true` to the `page` fields for courses and programs in all OpenAPI spec versions (`v0.yaml`, `v1.yaml`, `v2.yaml`) to indicate that these fields can be `null`. 

**Testing:**

* Added tests to verify that the serializers return `page: None` when a course or program does not have a CMS page. (`courses/serializers/v2/courses_test.py` [[1]](diffhunk://#diff-13a749a5929261cfbfec9907a6f96d63af64935ddc49efe07ec5d08d4310530fR166-R175) `courses/serializers/v2/programs_test.py` [[2]](diffhunk://#diff-c16959541f568075874846b3276449847ef9f8a0d37c3e3c25b21b838e4d6eb2R162-R170)


marslanabdulrauf:
@ChristopherChudzicki should we bump the version for breaking change which is causing `github` check failure ?

ChristopherChudzicki:
> @ChristopherChudzicki should we bump the version for breaking change which is causing github check failure ?

Sorry I didn't have time to review this more carefully today, but IMO this isn't a breaking change—it's just making the spec accurate. The API can return null now, but the spec doesn't reflect that. We aren't changing the API, we are fixing the spec.

marslanabdulrauf:
Oh, I checked the [V2Course serializer](https://github.com/mitodl/mitxonline/pull/3426/changes) didn't had `allow_null=True` which I added and generated the v2 specs.

Also had similar requirement for Programs as well:

> Additionally, api/v2/programs page has an incorrect type in the spec. Sometimes it's the full page object, sometimes it's just { feature_image_src }. My preference would be to make a breaking change to the api/v2/programs api to make it nullable, rather than the { feature_image_src } fallback which is never used on MITxOnline or Learn. (The fallback for v1 is used, I believe.)

So made changes around that as well and updated the v2 specs.

ChristopherChudzicki:
> Oh, I checked the [V2Course serializer](https://github.com/mitodl/mitxonline/pull/3426/changes) didn't had allow_null=True which I added and generated the v2 specs.

👍 That sounds good to me. Again, not a breaking change—the serializer already returned null for page, e.g., https://mitxonline.mit.edu/api/v2/courses/course-v1:UAI_NTUA_SOURCE+NTUA.UAI.VT.3/ (since DRF serializers don't validate responses, only requests).

My suggestion is:
1. Fix tests / linting (the linting failure is probably pre-existing, but the line number changed, so it's not getting excluded by drf_lint_baseline.json anymore...may need to regenerate that.)
2. Before we merge, let's get a PR ready for MIT Learn for the updated client. The change to nullable page is accurate—reflects reality—but will trigger some new TS warnings in MIT Learn. We can generate a API client for the branch via https://github.com/mitodl/mitxonline-api-clients#branch-clients-local-development and prepare the Learn PR

ChristopherChudzicki (approved):

ChristopherChudzicki:
👍 Thanks!

PR #3459, opened 2026-04-06
feat: add edx certificate webhook

Anas12091101:
https://github.com/mitodl/hq/issues/10777

This PR adds a webhook API endpoint (POST /api/certificate_webhook/) that receives certificate creation events from an Open edX plugin. When edX creates a certificate, the plugin (WIP, see: https://github.com/mitodl/open-edx-plugins/pull/684) sends the user's email and course run ID; this endpoint fetches the grade from edX, syncs it locally, and creates the corresponding certificate in MITx Online.

- [ ] Desktop screenshots

### How can this be tested?

**1. Set up OAuth2 credentials**

Create an OAuth2 access token in Django admin (`/admin/oauth2_provider/accesstoken/`) or use an existing access token.

**2. Prerequisites**

Ensure the user has a verified enrollment in the course run, and the course run exists in edX with a passing grade for the user.

**3. Test certificate creation (201)**

```bash
curl -s -w "\nHTTP_STATUS: %{http_code}\n" -X POST \
  http://localhost:9080/api/process_certificate_webhook/ \
  -H "Authorization: Bearer <your_access_token>" \
  -H "Content-Type: application/json" \
  -d '{"user_id": "<user_email>", "course_id": "<course_run_courseware_id>"}'
```

Expected response:
```json
{"certificate_status": "created"}
HTTP_STATUS: 200
```

**4. Test idempotency — duplicate call (200)**

Run the same curl command again. Expected response:
```json
{"certificate_status": "exists"}
HTTP_STATUS: 200
```

**5. Test unauthenticated request (401)**

```bash
curl -s -w "\nHTTP_STATUS: %{http_code}\n" -X POST \
  http://localhost:9080/api/process_certificate_webhook/ \
  -H "Content-Type: application/json" \
  -d '{"user_id": "test@example.com", "course_id": "course-v1:MITx+1+2024"}'
```

Expected: `HTTP_STATUS: 401`

**6. Test missing fields (400)**

```bash
curl -s -w "\nHTTP_STATUS: %{http_code}\n" -X POST \
  http://localhost:9080/api/process_certificate_webhook/ \
  -H "Authorization: Bearer <your_access_token>" \
  -H "Content-Type: application/json" \
  -d '{"user_id": "test@example.com"}'
```

Expected: `HTTP_STATUS: 400` with error about missing `course_id`.

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -933,33 +933,32 @@ def get_certificate_grade_eligible_runs(now):
 def generate_course_run_certificates(  # noqa: C901
     user=None,
     course_run=None,
-    is_webhook=False,  # noqa: FBT002
+    force=False,  # noqa: FBT002
 ):
     """
     Hits the edX grades API and generates the certificates and grades for users for course runs.
 
+    Each parameter works independently:
+    - If course_run is provided, only that course run is processed (skipping eligibility filtering).
+    - If user is provided, only that user's grade/certificate is processed.
+    - If force is True, certificate date/eligibility checks are bypassed.
+
     When called without arguments (periodic task path), it fetches all eligible course runs and
     processes grades/certificates for all users in each run.
-
-    When called with user, course_run, and is_webhook=True (webhook path), it processes a single
-    user's grade and certificate for a specific course run, bypassing certificate_available_date
-    and course run eligibility checks since edX has already determined the certificate is ready.
 
     Args:
         user (User or None): If provided, process only this user's grade/certificate.
         course_run (CourseRun or None): If provided, process only this course run.
-        is_webhook (bool): If True, bypass eligibility checks.
+        force (bool): If True, bypass certificate_available_date and eligibility checks.
 
     Returns:
         None
     """
     now = now_in_utc()
 
-    # Webhook path: use the provided course run directly, no eligibility filtering
-    if is_webhook and user and course_run:
+    if course_run:
         course_runs = [course_run]
     else:
-        # Periodic task path: fetch all eligible runs
         course_runs = get_certificate_grade_eligible_runs(now)
 
         if course_runs is None or course_runs.count() == 0:
@@ -986,7 +985,7 @@ def generate_course_run_certificates(  # noqa: C901
             except ValidationError:
                 msg = f"Can't save grade {edx_grade} for {run_user} in {run}, skipping certificate generation"
                 log.exception(msg)
-                if is_webhook:
+                if len(course_runs) == 1:
                     return
                 continue
 
@@ -996,14 +995,15 @@ def generate_course_run_certificates(  # noqa: C901
                 updated_grades_count += 1
 
             # Check certificate generation eligibility
-            # For webhooks: bypass checks since edX has already determined the certificate is ready
-            # For periodic task:
+            # When force is True: bypass checks (e.g. webhook path where edX already
+            #   determined the certificate is ready)
+            # Otherwise:
             #   1. For self_paced course runs we generate certificates right away irrespective
             #      of certificate_available_date
             #   2. For other course runs we generate certificates if the certificate_available_date
             #      has passed
             if (
-                is_webhook
+                force
                 or run.is_self_paced
                 or (
                     run.certificate_available_date
--- a/courses/api.py
+++ b/courses/api.py
@@ -930,34 +930,70 @@ def get_certificate_grade_eligible_runs(now):
     return course_runs  # noqa: RET504
 
 
-def generate_course_run_certificates():
+def generate_course_run_certificates(  # noqa: C901
+    user=None,
+    course_run=None,
+    is_webhook=False,  # noqa: FBT002
+):
     """
-    Hits the edX grades API for eligible course runs and generates the certificates and grades for users for course runs
+    Hits the edX grades API and generates the certificates and grades for users for course runs.
+
+    When called without arguments (periodic task path), it fetches all eligible course runs and
+    processes grades/certificates for all users in each run.
+
+    When called with user, course_run, and is_webhook=True (webhook path), it processes a single
+    user's grade and certificate for a specific course run, bypassing certificate_available_date
+    and course run eligibility checks since edX has already determined the certificate is ready.
+
+    Args:
+        user (User or None): If provided, process only this user's grade/certificate.
+        course_run (CourseRun or None): If provided, process only this course run.
+        is_webhook (bool): If True, bypass eligibility checks.
+
+    Returns:
+        str or None: The certificate status - "created", "exists", or None.
     """
     now = now_in_utc()
-    course_runs = get_certificate_grade_eligible_runs(now)
 
-    if course_runs is None or course_runs.count() == 0:
-        log.info("No course runs matched the certificates generation criteria")
-        return
+    certificate_status = None
+
+    # Webhook path: use the provided course run directly, no eligibility filtering
+    if is_webhook and user and course_run:
+        course_runs = [course_run]
+    else:
+        # Periodic task path: fetch all eligible runs
+        course_runs = get_certificate_grade_eligible_runs(now)
+
+        if course_runs is None or course_runs.count() == 0:
+            log.info("No course runs matched the certificates generation criteria")
+            return certificate_status
 
     for run in course_runs:
-        edx_grade_user_iter = exception_logging_generator(
-            get_edx_grades_with_users(run)
-        )
+        if is_webhook:
+            # Webhook: fetch grade for the specific user only
+            edx_grade_user_iter = get_edx_grades_with_users(run, user=user)
+        else:
+            edx_grade_user_iter = exception_logging_generator(
+                get_edx_grades_with_users(run)
+            )
         created_grades_count, updated_grades_count, generated_certificates_count = (
             0,
             0,
             0,
         )
-        for edx_grade, user in edx_grade_user_iter:
+        for edx_grade, run_user in edx_grade_user_iter:
             try:
                 course_run_grade, created, updated = ensure_course_run_grade(
-                    user=user, course_run=run, edx_grade=edx_grade, should_update=True
+                    user=run_user,
+                    course_run=run,
+                    edx_grade=edx_grade,
+                    should_update=True,
                 )
             except ValidationError:
-                msg = f"Can't save grade {edx_grade} for {user} in {run}, skipping certificate generation"
+                msg = f"Can't save grade {edx_grade} for {run_user} in {run}, skipping certificate generation"
                 log.exception(msg)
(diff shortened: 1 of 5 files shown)


openedx/views.py
```diff
@@ -1,10 +1,84 @@
...
+    API view for receiving certificate creation events from Open edX.
+
+    When Open edX creates a certificate for a user, it sends a POST request
+    to this endpoint with the user's email and the course ID. This view then
+    fetches the grade from edX, syncs it locally, and creates the corresponding
+    certificate in MITx Online.
+    """
+
+    authentication_classes = [OAuth2Authentication]
+    permission_classes = [IsAuthenticated]
```
arslanashraf7: Should be admin level permissions. `IsAdminUser`

courses/api.py
```diff
@@ -895,34 +895,70 @@ def get_certificate_grade_eligible_runs(now):
...
-        edx_grade_user_iter = exception_logging_generator(
-            get_edx_grades_with_users(run)
-        )
+        if is_webhook:
+            # Webhook: fetch grade for the specific user only
+            edx_grade_user_iter = get_edx_grades_with_users(run, user=user)
+        else:
+            edx_grade_user_iter = exception_logging_generator(
+                get_edx_grades_with_users(run)
+            )
```
arslanashraf7: Don't need this condition if the passed user is `None`. So, in both cases, we should be able to do `get_edx_grades_with_users(run, user=user)` where when user=None, it would automatically run for every user.

courses/api.py
```diff
@@ -895,34 +895,70 @@ def get_certificate_grade_eligible_runs(now):
...
         created_grades_count, updated_grades_count, generated_certificates_count = (
             0,
             0,
             0,
         )
-        for edx_grade, user in edx_grade_user_iter:
+        for edx_grade, run_user in edx_grade_user_iter:
```
arslanashraf7, line 977: nit: this rename isn't needed really.

courses/api.py
```diff
@@ -931,30 +967,58 @@ def generate_course_run_certificates():
...
+                        "Certificate created for user %s and course_run %s",
+                        run_user,
+                        run,
                     )
                     generated_certificates_count += 1
+
+                if created:
+                    certificate_status = "created"
+                elif certificate:
+                    certificate_status = "exists"
```
arslanashraf7: Aren't they always going to be available together? I checked `process_course_run_grade_certificate` and it would only return a `certificate` object when it creates one.

openedx/views.py
```diff
@@ -1,10 +1,84 @@
...
+            user=user,
+            course_run=course_run,
+            is_webhook=True,
+        )
+
+        response_status = (
+            status.HTTP_201_CREATED
+            if certificate_status == "created"
+            else status.HTTP_200_OK
+        )
```
arslanashraf7: The idea of the webhook is to trigger the processing of the certificate. I don't think we specifically need to return the created/deleted status from the actual certificate creation. So, if a valid request is received, it would return 200 that the request if being processed.
Additionally, Open edX is not going to look for the actual status and do something about it unless it is 40x or 50x.

courses/api.py
```diff
@@ -895,34 +895,70 @@ def get_certificate_grade_eligible_runs(now):
...
+    Returns:
+        str or None: The certificate status - "created", "exists", or None.
     """
     now = now_in_utc()
-    course_runs = get_certificate_grade_eligible_runs(now)
 
-    if course_runs is None or course_runs.count() == 0:
-        log.info("No course runs matched the certificates generation criteria")
-        return
+    certificate_status = None
```
arslanashraf7: No need for this `certificate_status = None` and the related complexity, The view should just call this method and assume the certificate processing started.

openedx/urls.py
```diff
@@ -15,4 +15,9 @@
         views.openedx_private_auth_complete,
         name="openedx-private-oauth-complete-no-apisix",
     ),
+    path(
+        "api/certificate_webhook/",
```
arslanashraf7:
```suggestion
        "api/process_certificate_webhook/",
```

openedx/views.py
```diff
@@ -1,10 +1,84 @@
...
+log = logging.getLogger(__name__)
 
 
 def openedx_private_auth_complete(request):  # noqa: ARG001
     """Responds with a simple HTTP_200_OK"""
     # NOTE: this is only meant as a landing endpoint for api.create_edx_auth_token() flow
     return HttpResponse(status=status.HTTP_200_OK)
+
+
+class CertificateWebhookView(APIView):
```
arslanashraf7:
```suggestion
class ProcessCertificateWebhookView(APIView):
```

openedx/views.py
```diff
@@ -1,10 +1,84 @@
...
+                status=status.HTTP_404_NOT_FOUND,
+            )
+
+        log.info(
+            "Certificate webhook received for user %s, course run %s",
+            user_email,
+            course_run_id,
+        )
+
+        certificate_status = generate_course_run_certificates(
```
arslanashraf7: At this point, the view can check for an existing certificate before calling `generate_course_run_certificates` for idempotency.

arslanashraf7 (changes requested):

courses/api.py
```diff
@@ -930,34 +930,64 @@ def get_certificate_grade_eligible_runs(now):
...
     """
     now = now_in_utc()
-    course_runs = get_certificate_grade_eligible_runs(now)
 
-    if course_runs is None or course_runs.count() == 0:
-        log.info("No course runs matched the certificates generation criteria")
-        return
+    # Webhook path: use the provided course run directly, no eligibility filtering
+    if is_webhook and user and course_run:
+        course_runs = [course_run]
```
arslanashraf7: Now that I think more about this, it would be kind of an anti-pattern. e.g., if someone is passing a course_run explicitly, we should still let them do the desired thing. So, if a run or a user is passed, the function should work on that independently. The webhook check is just to bypass the conditions below. In fact, I think the name of this boolean should be `bypass_validation` or `force`, which will allow running the cert generation without the dates and eligibility check.

In the end, I think we should handle every parameter separately:
1. If a course run is passed, we process only that course run
2. If a user is passed, we process only that user
3. If `bypass_validation` or `force` is passed, we bypass the dates, eligibility, etc. checks
This way the method would be more flexible.

arslanashraf7 (changes requested):

openedx/views.py
```diff
@@ -119,3 +120,68 @@ def edx_enrollment_webhook(request):
...
+    to this endpoint with the user's email and the course ID. This view then
+    fetches the grade from edX, syncs it locally, and creates the corresponding
+    certificate in MITx Online.
+    """
+
+    authentication_classes = [OAuth2Authentication]
+    permission_classes = [IsAdminUser]
+
+    def post(self, request):
+        user_email = request.data.get("user_id")
```
arslanashraf7: Also let's call it `email`:

```suggestion
        user_email = request.data.get("email")
```

arslanashraf7 (changes requested):

openedx/views.py
arslanashraf7: @Anas12091101 This still needs to be applied.

courses/api.py
```diff
@@ -930,34 +930,63 @@ def get_certificate_grade_eligible_runs(now):
...
+    user=None,
+    course_run=None,
+    force=False,  # noqa: FBT002
+):
     """
-    Hits the edX grades API for eligible course runs and generates the certificates and grades for users for course runs
+    Hits the edX grades API and generates the certificates and grades for users for course runs.
+
+    Each parameter works independently:
+    - If course_run is provided, only that course run is processed (skipping eligibility filtering).
```
arslanashraf7:
```suggestion
    - If course_run is provided, only that course run is processed (skipping course runs eligibility filtering).
```

courses/api.py
```diff
@@ -930,34 +930,63 @@ def get_certificate_grade_eligible_runs(now):
...
+                    course_run=run,
+                    edx_grade=edx_grade,
+                    should_update=True,
                 )
             except ValidationError:
-                msg = f"Can't save grade {edx_grade} for {user} in {run}, skipping certificate generation"
+                msg = f"Can't save grade {edx_grade} for {run_user} in {run}, skipping certificate generation"
                 log.exception(msg)
+                if len(course_runs) == 1:
+                    return
```
arslanashraf7: No need for this check. If there is just one course run in the loop, it will automatically stop at this point after continue.

openedx/urls.py
```diff
@@ -20,4 +20,9 @@
         views.edx_enrollment_webhook,
         name="openedx-enrollment-webhook",
     ),
+    path(
+        "api/process_certificate_webhook/",
```
arslanashraf7: Let's make the URL pattern similar to `"api/openedx_webhook/enrollment/"` above, to keep it standard with other webhook(s).

openedx/views.py
```diff
@@ -119,3 +120,68 @@ def edx_enrollment_webhook(request):
...
+    When Open edX creates a certificate for a user, it sends a POST request
+    to this endpoint with the user's email and the course ID. This view then
+    fetches the grade from edX, syncs it locally, and creates the corresponding
+    certificate in MITx Online.
+    """
+
+    authentication_classes = [OAuth2Authentication]
+    permission_classes = [IsAdminUser]
+
+    def post(self, request):
```
arslanashraf7: Let's add more log statements in different code paths. Currently there is no log in the application that tells what happened to the request.

openedx/views.py
```diff
@@ -119,3 +120,68 @@ def edx_enrollment_webhook(request):
...
+            return Response(
+                {"error": f"Course run with id '{course_run_id}' not found."},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+
+        log.info(
+            "Certificate webhook received for user %s, course run %s",
+            user_email,
+            course_run_id,
+        )
```
arslanashraf7: This log should be at the start of the view, indicating that the webhook request has been received. In this place, we can log something like "Validation completed, the webook will be processed. etc"

openedx/views.py
```diff
@@ -119,3 +120,68 @@ def edx_enrollment_webhook(request):
         },
         status=status.HTTP_201_CREATED if created else status.HTTP_200_OK,
     )
+
+
+@extend_schema(exclude=True)
+class ProcessCertificateWebhookView(APIView):
```
arslanashraf7: nit: let's use decorator based functional view here as well just like we did in enrollment webhook.

arslanashraf7 (changes requested):

arslanashraf7 (approved):
LGTM 👍 Thanks for making all the changes.

PR #3473, opened 2026-04-08
Adds action to populate empty verifiable creds by courserun or program

dsubak:
https://github.com/mitodl/hq/issues/10692

- Adds an action on the Program and CourseRun pages to generate verifiable credentials for certificates issued for those programs or courseruns which don't already have a verifiable credential record.
  - Indicates how many certificate backfills were attempted. In the event of a failure, outputs the UUID for the certificates which resulted in errors..
    - N.B. Certificates for which backfills were attempted but didn't complete due to not having `should_provision_verifiable_credential` set DO NOT count as failures.
- Adds a super thin admin panel for verifiable creds. This is because previously deleting those records required shell access. Now that individual records can be deleted, couple with the above change, admins can delete and regenerate verifiable credentials if needed.
- Fixes a small issue with program backfills

Screenshot 2026-04-08 at 12 11 09 PM
Screenshot 2026-04-08 at 12 12 35 PM

Verifiable creds admin view
Screenshot 2026-04-08 at 2 01 25 PM

### How can this be tested?
This is most easily tested if you already have a certificate provisioned. If you do, the only steps required are:
- Set the following env vars
```
VERIFIABLE_CREDENTIAL_SIGNER_URL=http://dcc.odl.local:4005/instance/test/credentials/issue
VERIFIABLE_CREDENTIAL_BEARER_TOKEN='test'
```
- Ensure that `COMPOSE_PROFILES` has `verifiable-creds` in addition to whatever else you normally have set up.
- Go to the CMS page for the certificate, add any value in the new `Verifiable credential criteria` field and check `Should provision verifiable credential`. Publish your changes
- If you have already provisioned a verifiable credential for it, delete it. This can be done by visiting the new admin page for Verifiable Credentials and deleting the object with the same UUID, or by deleting it via django or DB shell.
- Go to the courserun or program admin page (depending on which type of certificate you've got) and select the corresponding row.
- Select the backfill action from the drop down and select `Go`.
- Observe that a verifiable credential has been provisioned for the corresponding certificate.

If you don't have a certificate provisioned, you'll need a fair amount of boilerplate set up, principally correctly set up courses, course runs and/or programs, along with their corresponding CMS pages. The `test_datagen` script sets up a lot of that if you are able to use it - once that's done you should only need to:
- Create a Program or CourseRunEnrollment for the user you'll create the cert for (I did this using the admin page)
- Create the certificate of the corresponding type (Also performed via admin)

### DIFF
--- a/courses/admin.py
+++ b/courses/admin.py
@@ -41,6 +41,26 @@ from main.utils import get_field_names
 from openedx.tasks import retry_failed_edx_enrollments
 
 
+class VerifiableCredentialBackfillAdminMixin:
+    def populate_verifiable_credentials_for_certificate(self, request, certificates):
+        """Helper method to create and associate a verifiable credential for a given certificate"""
+        failed_certificates = []
+        for certificate in certificates:
+            try:
+                create_verifiable_credential(certificate, raise_on_error=True)
+            except Exception:  # noqa: PERF203, BLE001
+                failed_certificates.append(certificate)
+
+        message = f"Successfully requested verifiable credential backfill for {len(certificates)} course run certificates."
+        level = messages.INFO
+        if failed_certificates:
+            # We indicate IDs, but errors should also be logged to sentry from within create_verifiable_credential
+            level = messages.WARNING
+            message = f"Successfully requested verifiable credential backfill for {len(certificates)} course run certificates, but encountered errors for certificates with IDs: {[cert.id for cert in failed_certificates]}"
+
+        self.message_user(request, message, level=level)
+
+
 class ProgramContractPageInline(admin.TabularInline):
     """Inline for contract pages"""
 
@@ -52,7 +72,7 @@ class ProgramContractPageInline(admin.TabularInline):
 
 
 @admin.register(Program)
-class ProgramAdmin(admin.ModelAdmin):
+class ProgramAdmin(VerifiableCredentialBackfillAdminMixin, admin.ModelAdmin):
     """Admin for Program"""
 
     model = Program
@@ -83,7 +103,7 @@ class ProgramAdmin(admin.ModelAdmin):
                 program_id__in=program_ids, verifiable_credential__isnull=True
             )
         )
-        populate_verifiable_credentials_for_certificate(self, request, certificates)
+        self.populate_verifiable_credentials_for_certificate(request, certificates)
 
 
 @admin.register(ProgramRun)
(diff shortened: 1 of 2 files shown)


courses/admin.py
```diff
@@ -686,3 +730,22 @@ class EnrollmentModeAdmin(admin.ModelAdmin):
     list_filter = [
         "requires_payment",
     ]
+
+
+def populate_verifiable_credentials_for_certificate(admin, request, certificates):
```
asadali145: nit: could be moved to a common admin mixin that can be made parent for the ProgramRun and CourseRun admins.

courses/admin.py
```diff
@@ -686,3 +730,22 @@ class EnrollmentModeAdmin(admin.ModelAdmin):
...
+    """Helper method to create and associate a verifiable credential for a given certificate"""
+    failed_certificates = []
+    for certificate in certificates:
+        try:
+            create_verifiable_credential(certificate, raise_on_error=True)
+        except Exception:  # noqa: PERF203, BLE001
+            failed_certificates.append(certificate)
+
+    message = f"Successfully requested verifiable credential backfill for {len(certificates)} course run certificates."
```
asadali145:
```suggestion
    message = f"Successfully requested verifiable credential backfill for {len(certificates)} {admin.model._meta.model_name} certificates."
```

asadali145 (changes requested):
LGTM! We need to fix the messages.

courses/admin.py
```diff
@@ -635,6 +687,18 @@ def get_queryset(self, request):  # noqa: ARG002
         return self.model.all_objects.get_queryset().select_related("user", "program")
 
 
+@admin.register(VerifiableCredential)
+class VerifiableCredentialAdmin(TimestampedModelAdmin):
+    """Admin for VerifiableCredential"""
+
+    model = VerifiableCredential
+    include_timestamps_in_list = True
+    list_display = ["uuid", "programcertificate", "courseruncertificate"]
```
asadali145, line 696: nit: I think we should override queryset method to select_related program and course certificates to avoid N+1 queries.

```
def get_queryset(self, request):
    qs = super().get_queryset(request)
    return qs.select_related("programcertificate", "courseruncertificate")
```

asadali145 (approved):
LGTM!

PR #3474, opened 2026-04-08
10745 learn   hubspot should be updated when user adds item to cart

cp-at-mit:
https://github.com/mitodl/hq/issues/10745

Creates hubspot deals when a user adds a UAI course to their cart.

### How can this be tested?
Create a course that will be considered a UAI course (courseware ID course-v1:UAI.....)
Configure Hubspot.
Add the course to your cart and verify that your contact and deal are created in Hubspot.


ecommerce/views/legacy/__init__.py
```diff
@@ -722,6 +724,15 @@ def add_to_cart(self, request):
                 BasketItem.objects.create(basket=basket, product=product)
                 message = "Product added to cart"
 
+            if message == "Product added to cart" and isinstance(
+                product.purchasable_object, CourseRun
+            ):
+                sync_hubspot_cart_add(
```
jkachel: This probably needs to move. This endpoint is only used by MITx Online directly now; Learn uses one of the other 3 APIs we have for this (the general one in `v0`, the B2B-specific one, or the verified program-specific one).

Also: this seems to match only on CourseRun. Should it also run for Program? Most of the actual buying going on will be for the UAI program. We ought to see records come through for the individual runs as they enroll in them, though (once the verified program enrollment code is updated for this), so not sure on this front.

edit to add: not sure we need to put this in all 3 of those APIs but when Learn adds things to the cart, it uses the `v0` API. So that one probably needs to be covered. The other APIs generate a cart and then immediately fulfill it, which should generate some HubSpot traffic, so dunno that they need the additional HubSpot reporting.

jkachel (changes requested):
Noticed an issue with the location of this. But I think maybe someone else should test for functionality - my HubSpot integration isn't set up right so kinda stuck on that regard.

I did look at the issue that Sentry came up with and i don't believe it's valid, since the conversion from basket to order will look for a matching pending order.

ecommerce/views/v0/__init__.py
```diff
@@ -369,6 +371,13 @@ def create_basket_with_products(request):
             BasketItem.objects.update_or_create(
                 basket=basket, product=product, defaults={"quantity": quantity}
             )
+            # Sync with HubSpot for CourseRun products
```
jkachel, line 382: This needs to be here but it also moreso needs to be in `_create_basket_from_product` above. That one gets hit more frequently, usually we hit the single-product API. (Really this view function needs to be refactored to use `_create_basket_from_product` but that doesn't need to happen in this PR.)

jkachel (changes requested):
Did see the task hit when I hit Start Module on a B2B course so that was good. Still need to test verified programs. But also noticed an issue with the changes to the `v0` API. Should be pretty straightforward fix though. Will still need to complete my review but figured no sense in leaving this sitting until that's done.

jkachel (approved):
LGTM 👍  still prob worth getting someone with a working HubSpot setup to test - I can see that the call happens at the appropriate times but I can't validate the results of those calls.

gumaerc:
@jkachel @cp-at-mit I was able to successfully test this and see my contact and deal created end to end. It rejects the default dev test users (admin@odl.local, student@odl.local, etc.) because they look like fake email addresses. You have to create a user with a realistic looking email address in order for it to work, but once I did that I was able to see it come thru. James's last piece of feedback there to make sure it was called on `_create_basket_from_product` was crucial, as that's ultimately the endpoint that Learn ends up using when it goes to add an item to the cart via the API. I suppose it's worth noting that we don't call it from the legacy UI if someone somehow adds to cart that way, but I'm guessing we intentionally don't want that. Either way, this seems good to go as far as I'm concerned.

image

PR #3482, opened 2026-04-10
Add admin search by edx_username

annagav:

N/A 

Wanted to to search edx user by edx_username

### How can this be tested?
try search your user by edx username


cp-at-mit (approved):
Looks good.  Would it make sense to add the edx_username to the columns listed, so that you can see if your search is matching the edx_username or some other searchable field?

PR #3487, opened 2026-04-14
Add fields for configuring the How You'll Learn cards on product pages

jkachel:

Part of mitodl/hq#10856

### Screenshot
Screenshot 2026-04-15 at 1 06 24 PM

Adds some checkbox fields for the cards that we'd like to display on the product pages in Learn. These include:

- REAL-WORLD LEARNING
        Learn from faculty experts who emphasize practical application over theory.
- LEARN BY DOING
        Practice core competencies through case studies, simulations, and hands-on tools.
- LEARN FROM OTHERS
        Connect with an international community of professionals working on real-world projects.
- LEARN ON DEMAND
        Access all course content online with complete flexibility to study at your own pace.
- AI-ENABLED SUPPORT
        Get personalized help on assignments from AskTIM, powered by advanced AI.
- STACKABLE CREDENTIALS
        Earn certificates at key milestones—module, course, and program—building a portfolio of expertise.

The text for these may change depending on what kind of course this is (UAI, MITx, etc.) but that will be controlled by the component.

### How can this be tested?

The fields should show up in the course pages and program pages within the CMS. The fields should work as expected. The API output should include these new fields as well.

### Additional Context

There will be a complimentary PR that will update the Learn frontend to make the cards in the How You'll Learn section dependent on these checkboxes.

There will need to be a further PR to set these to reasonable values for the courses that are there, but I don't think we have that codified at time of writing.


pdpinch:
Can I get a screenshot of the CMS UI for this?

ChristopherChudzicki:
@pdpinch 
> Can I get a screenshot of the CMS UI for this?

I just added one to the PR description. (@jkachel fyi... didn't edit anything else)

ChristopherChudzicki (approved):
Approving, but, consider the following change:

**Suggestion:** I was expecting the API to return an array like `{key, title, description}[]`. I think this would be better because:
- Wagtail title/description and item order would automatically be in-sync with frontend.
- if we get rid of the "How You'll Learn" images, it is entirely driven by the API data, frontend doesn't need to know anything else.
    - removing the images is something that has been discussed as a possibility 
    - even with `{key,title,description}[]` response, if we keep the images, frontend needs to have a key->image map.

http://mitxonline.odl.local:8013/api/v2/pages/?type=cms.programpage&readable_id=program-v1%3AUAI%2BB2C&fields=* is the API i am talking about... (Annoying: need to encodeURICoponent the readable ids)

ChristopherChudzicki:
@jkachel Actually, Also... Should we add a data migration that:
- populates the relevant 4 for UAI courses/programs
- leaves all the others blank

jkachel:
> @jkachel Actually, Also... Should we add a data migration that:
> 
> - populates the relevant 4 for UAI courses/programs
> 
> - leaves all the others blank

Mentioned this in the PR - not sure what we want for the defaults so didn't tackle this, so it doesn't hold up the changes for the front end. But I have the front end changes done so if those are the defaults we want I can add a step to the migration to set them.

jkachel:
> Suggestion: I was expecting the API to return an array like {key, title, description}[]. I think this would be better because:

I kept this simple because the upstream issue indicated that these weren't to be editable - and wrt [the frontend changes I have in process](https://github.com/mitodl/mit-learn/pull/3198), these are just flags; the content is defined wholly in the React code. (I just have that mocked API response y'all had in the HowYoullLearnSection component matched up to the fields. You get cards for whichever ones are True.)

However, my initial thought on this was to set this up as a linked page, like we do for signatories in contracts. The cards would be another set of pages within Wagtail and we'd link them into the product page. Then we send whatever the details are for that card as part of the API response. Potential tricky bit with this is that the card content can be dependent on the course type (UAI, MITx, etc.) and so we'd have to design for that. But this seemed like it'd be out of spec for the issue as it was written so I went with the simpler route. Thoughts @pdpinch ?

ChristopherChudzicki:
> The cards would be another set of pages within Wagtail and we'd link them into the product page. 

I think that's more complicated than we need now, as you say, not planning to edit them much.

What I had in mind was something like
- the six flags you have now
- additional model property (or serializer) like
    ```python
    @property
    def how_you_will_learn(self):
        return [
            {"key": key, "title": title, "description": description}
            for key, title, description in HOW_YOU_LEARN_CHOICES
            if getattr(self, f"hyl_choice_{key}", False)
        ]
    ```

so that the title/description stay synced between mitxlone and mit-learn

jkachel:
> > The cards would be another set of pages within Wagtail and we'd link them into the product page.
> 
> I think that's more complicated than we need now, as you say, not planning to edit them much.
> 
> What I had in mind was something like
> 
> * the six flags you have now
> * additional model property (or serializer) like
>   ```python
>   @property
>   def how_you_will_learn(self):
>       return [
>           {"key": key, "title": title, "description": description}
>           for key, title, description in HOW_YOU_LEARN_CHOICES
>           if getattr(self, f"hyl_choice_{key}", False)
>       ]
>   ```
> 
> so that the title/description stay synced between mitxlone and mit-learn

https://github.com/mitodl/mitxonline/pull/3487/commits/8d0b1df17ca6fabab510b6ef6b048bb40a47b6ab  updates the API to coalesce the choice fields in this manner. Will update https://github.com/mitodl/mit-learn/pull/3198 appropriately - should make that a simpler implementation, really.

ChristopherChudzicki (approved):
👍

PR #3491, opened 2026-04-15
Change field to match what you learn

dsubak:

We added a new field for verifiable creds in [this PR](https://github.com/mitodl/mitxonline/pull/3373) - at the time I just pattern matched with the other fields on the certificate page, which were charfields of length 250. Now that we're starting to populate the field, we are gonna mostly be copying from `what_you_learn` on the product page with some formatting tweaks, and that text can be much longer as it's a rich text field.

This simply alters the backing column to be a TextField. Note that we expressly DO NOT want to use a wagtail RichTextField, as that saves the text with markup, which has caused us issues in the past when we were attempting to generate verifiable creds based on the `what_you_learn` field.

Screenshot 2026-04-15 at 12 01 53 PM

### How can this be tested?
Simply visit any certificate page in your CMS and save string over 250 characters long. It ought to work

### Additional Context
- While the alter table statement should work just fine and be non-destructive, I did also verify that at the moment neither RC nor prod have any non-null values for the column in question.


cms/migrations/0059_alter_certificatepage_verifiable_credential_criteria.py
```diff
@@ -0,0 +1,21 @@
...
+        ("cms", "0058_programpage_include_in_learn_catalog"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="certificatepage",
+            name="verifiable_credential_criteria",
+            field=models.TextField(
+                blank=True,
+                help_text="For verifiable credentials issued for this certificate, this is the criteria narrative field. It should be something descriptive, like a list of completed courses, and may be plaintext or markdown. If it is not supplied, no verifiable credential will be provisioned for those certificates.",
```
dsubak, line 17: @pdpinch Not making any changes to the help text to include the max length as there functionally isn't gonna be one, just as an FYI.

jkachel (approved):
LGTM 👍

PR #3494, opened 2026-04-15
Make enrollment_mode field non-nullable

ChristopherChudzicki:
Fixes
- https://github.com/mitodl/hq/issues/10938

Makes the `enrollment_mode` field on the abstract `EnrollmentModel` (inherited by `CourseRunEnrollment` and `ProgramEnrollment`) non-nullable. The field already has a default value (`"audit"`) and is never null in RC or production, so this removes an unnecessary nullable state that was causing spurious null-check requirements in frontend code and tests.

### How can this be tested?
1. Migrations / tests should pass.
2. Try enrolling in something.

### Checklist

Before deploying, probably worth verifying still none on production/rc:
```python
from courses.models import CourseRunEnrollment, ProgramEnrollment
for Model in [CourseRunEnrollment, ProgramEnrollment]:
    print(f"{Model.__name__}: {Model.all_objects.filter(enrollment_mode__isnull=True).count()}")
```
Both counts should be 0.


jkachel (approved):
LGTM 👍

PR #3497, opened 2026-04-16
3488 mitx online   update xpro hubspot when course purchased is uai

### LINKED ISSUE #3488: MITx Online - Update Xpro hubspot when course purchased is UAI
### Description/Context
When a course is purchased, we create a `deal` in Hubspot in an account that's linked to MITx Online.  We want to update this to add logic that will create a `deal` in Hubspot in an account that's linked to Xpro when the course purchased is a UAI course.

### Plan/Design
Update the existing fulfilled order/hubspot logic to check whether the course is for a UAI course.

cp-at-mit:
https://github.com/mitodl/mitxonline/issues/3488

Adds the deal to Xpro hubspot account when the course is a UAI course.

### How can this be tested?
UAI_MITOL_HUBSPOT_API_PRIVATE_TOKEN equal to the xpro hubspot API key
Create all that is required to checkout with a UAI course.
Complete a purchase of the UAI course and verify that a deal is created in the xpro Hubspot account.  All line items should be on the deal.


arslanashraf7:

hubspot_sync/task_helpers.py
```diff
@@ -34,17 +35,31 @@ def sync_hubspot_deal(order: Order):
...
+        if token:
+            try:
+                tasks.sync_deal_with_hubspot_targeted.apply_async(
+                    args=(order.id, token), countdown=10
+                )
+            except:  # noqa: E722
+                log.exception(
+                    "Exception calling sync_deal_with_hubspot_targeted for order %d",
+                    order.id,
+                )
```
Anas12091101, line 66: It’s better not to pass the secret token to the task, since it would be serialized into Redis and could be exposed in Celery monitoring tools or Sentry. Instead, we can pass a boolean `is_uai` flag and have the task read the token from settings at execution time in api.py.

hubspot_sync/api.py
```diff
@@ -1332,6 +1332,65 @@ def sync_deal_with_hubspot(order: Order) -> SimplePublicObject:
...
+    deal_input = _build_target_deal_message(order, hubspot_client)
+    contact_id = _ensure_hubspot_contact_for_user(order.purchaser, hubspot_client)
+    if not contact_id:
+        contact_body = make_contact_sync_message_from_user(order.purchaser)
+        wait_for_hubspot_rate_limit()
+        contact_result = hubspot_client.crm.objects.basic_api.create(
+            object_type=HubspotObjectType.CONTACTS.value,
+            simple_public_object_input_for_create=contact_body,
+        )
+        contact_id = contact_result.id
```
Anas12091101: This is redundant and unnecessary; `_ensure_hubspot_contact_for_user` already creates the contact if it doesn’t exist.

hubspot_sync/api.py
```diff
@@ -1332,6 +1332,65 @@ def sync_deal_with_hubspot(order: Order) -> SimplePublicObject:
...
+            simple_public_object_input_for_create=line_item_input,
+        )
+
+        wait_for_hubspot_rate_limit()
+        hubspot_client.crm.associations.v4.basic_api.create_default(
+            from_object_type=HubspotObjectType.LINES.value,
+            from_object_id=line_item_result.id,
+            to_object_type=HubspotObjectType.DEALS.value,
+            to_object_id=result.id,
+        )
```
Anas12091101: These are not idempotent. If the task retries for any reason, it could create duplicate entries on HubSpot.

Anas12091101 (changes requested):
Some changes:

hubspot_sync/api.py
```diff
@@ -1347,46 +1348,86 @@ def sync_deal_with_hubspot_targeted(order: Order, token: str) -> SimplePublicObj
...
+            pass  # Association may already exist
+
+    # Get existing line items for this deal
+    existing_line_item_ids = _find_target_line_items_for_deal(hubspot_client, result.id)
+    expected_line_count = order.lines.count()
+    
+    # Only create line items if we don't have the expected number
+    if len(existing_line_item_ids) != expected_line_count:
+        for line in order.lines.all():
+            line_item_input = _build_target_line_item_message(line, hubspot_client)
```
Anas12091101: If a 429 fires after creating 1 of 2 line items, the retry sees 1 != 2 and re-creates all lines (including the one already created), resulting in 3 total. The guard checks the count but doesn't know which lines already exist, so it blindly creates all of them again.

Fix: search by `unique_app_id` per line before creating (same pattern used for products in `_ensure_target_hubspot_product_for_line`). Something like:

```python
 # Create line items, skipping any that already exist (by unique_app_id)
    for line in order.lines.all():
        line_item_input = _build_target_line_item_message(line, hubspot_client)
        unique_app_id = line_item_input.properties.get("unique_app_id", "")

        # Check if this specific line item already exists
        if unique_app_id:
            existing_line_item_id = _find_target_line_item_by_unique_app_id(
                hubspot_client, unique_app_id
            )
            if existing_line_item_id:
                continue
      # create...
```

hubspot_sync/api.py
```diff
@@ -1719,6 +1829,54 @@ def _find_target_product_id_by_unique_app_id(
...
+    hubspot_client: HubspotApi, deal_id: str
+) -> list[str]:
+    """Find line item ids associated with a deal in the target account."""
+    wait_for_hubspot_rate_limit()
+    try:
+        response = hubspot_client.crm.associations.v4.basic_api.get_page(
+            from_object_type=HubspotObjectType.DEALS.value,
+            from_object_id=deal_id,
+            to_object_type=HubspotObjectType.LINES.value,
+        )
```
Anas12091101: The get_page method in the base API doesn’t accept `from_object_type` or `from_object_id` parameters. As a result, this call will fail silently and return an empty list.

Here’s the current definition of `get_page`:

```python
    def get_page(self, object_type, object_id, to_object_type, **kwargs):  # noqa: E501
        """List  # noqa: E501

        List all associations of an object by object type. Limit 500 per call.  # noqa: E501
        This method makes a synchronous HTTP request by default. To make an
        asynchronous HTTP request, please pass async_req=True

        >>> thread = api.get_page(object_type, object_id, to_object_type, async_req=True)
        >>> result = thread.get()

        :param object_type: (required)
        :type object_type: str
        :param object_id: (required)
        :type object_id: str
        :param to_object_type: (required)
        :type to_object_type: str
        :param after: The paging cursor token of the last successfully read resource will be returned as the `paging.next.after` JSON property of a paged response containing more results.
        :type after: str
        :param limit: The maximum number of results to display per page.
        :type limit: int
        :param async_req: Whether to execute the request asynchronously.
        :type async_req: bool, optional
        :param _preload_content: if False, the urllib3.HTTPResponse object will
                                 be returned without reading/decoding response
                                 data. Default is True.
        :type _preload_content: bool, optional
        :param _request_timeout: timeout setting for this request. If one
                                 number provided, it will be total request
                                 timeout. It can also be a pair (tuple) of
                                 (connection, read) timeouts.
        :return: Returns the result object.
                 If the method is called asynchronously,
                 returns the request thread.
        :rtype: CollectionResponseMultiAssociatedObjectWithLabelForwardPaging
        """
        kwargs["_return_http_data_only"] = True
        return self.get_page_with_http_info(object_type, object_id, to_object_type, **kwargs)  # noqa: E501
```
To fix this, we should either rename from_object_type → object_type and from_object_id → object_id, or pass the values as positional arguments instead.

hubspot_sync/api.py
```diff
@@ -1332,6 +1332,116 @@ def sync_deal_with_hubspot(order: Order) -> SimplePublicObject:
...
+    # Ensure contact exists and is associated
+    contact_id = _ensure_hubspot_contact_for_user(order.purchaser, hubspot_client)
+
+    # Check if deal-contact association exists
+    try:
+        wait_for_hubspot_rate_limit()
+        associations = hubspot_client.crm.associations.v4.basic_api.get_page(
+            from_object_type=HubspotObjectType.DEALS.value,
+            from_object_id=result.id,
+            to_object_type=HubspotObjectType.CONTACTS.value,
```
Anas12091101: Same as https://github.com/mitodl/mitxonline/pull/3497/changes#r3109727819

Anas12091101 (changes requested):

hubspot_sync/api.py
```diff
@@ -1719,6 +1830,133 @@ def _find_target_product_id_by_unique_app_id(
...
+            ),
+        )
+        if response.results:
+            return response.results[0].id
+    except Exception:
+        log.exception("Failed to search for deal by dealname: %s", dealname)
+    return None
+
+
+def _find_target_line_items_for_deal(
```
Anas12091101: The function is not being used anywhere now. We can remove it.

hubspot_sync/api.py
```diff
@@ -1719,6 +1830,133 @@ def _find_target_product_id_by_unique_app_id(
...
+                ],
+                properties=["dealname"],
+                limit=1,
+            ),
+        )
+        if response.results:
+            return response.results[0].id
+    except Exception:
+        log.exception("Failed to search for deal by dealname: %s", dealname)
+    return None
```
Anas12091101, line 1866: If a 429 exception occurs here, we simply log it and return None. Wouldn’t that risk creating duplicates? Should we instead raise the exception so the task can retry rather than potentially introducing duplicates?

hubspot_sync/api.py
```diff
@@ -1719,6 +1830,133 @@ def _find_target_product_id_by_unique_app_id(
...
+        if response.results:
+            return response.results[0].id
+        else:
+            return None
+    except (ApiException, TooManyRequestsException):
+        log.exception(
+            "Failed to search for line item with unique_app_id %s",
+            unique_app_id,
+        )
+        return None
```
Anas12091101, line 1909: Same as https://github.com/mitodl/mitxonline/pull/3497/changes#r3118660627

Anas12091101 (changes requested):
Looks good. Still some minor suggestions

Anas12091101 (approved):
LGTM 👍 
Thanks for making all the changes

cp-at-mit:
Solid review @Anas12091101 !

PR #3499, opened 2026-04-17
fix: if user is not sync'd with edX yet, don't raise

marslanabdulrauf:
https://github.com/mitodl/hq/issues/10624#issuecomment-4259248487

This pull request improves the reliability and safety of Open edX user synchronization by adding explicit checks for user existence in edX and handling cases where the user is not yet synced. It introduces a helper function to verify user existence, updates several functions to gracefully handle missing users, and ensures that API clients are only created when appropriate.

**API client creation and error handling:**

* Modified `get_edx_api_client` to return `None` (instead of raising an error) when a user is not yet synced with edX, if the `IGNORE_EDX_FAILURES` feature is enabled; otherwise, it raises a specific exception.

### How can this be tested?
1. Ensure a user exists who has not been synced to edX (`has_been_synced=False` on their `OpenEdxUser` record)
2. Go to Django admin → OpenEdx Users → find the user → confirm `has_been_synced` is unchecked
3. In `.env`, ensure `IGNORE_EDX_FAILURES` is set to True
4. Get into django shell `docker compose run --rm web python manage.py shell` and run:
```python
from openedx.api import get_edx_api_client

user = get_user_model().objects.get(email="your-unsynced-user@example.com")
get_edx_api_client(user)  # should raise NoEdxApiAuthError if IGNORE_EDX_FAILURES is set to False
```
5. WARNING "get_edx_api_client: user <email> is not yet synced with edX, skipping"

### Additional Context
There are scenarios when `change_edx_user_name_async` tasks fails when the user is not yet sync'd with edX.
We have confirmed it only happens once for a single user.

### DIFF
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -407,6 +407,19 @@ def _create_edx_user_request(open_edx_user, user, access_token):  # noqa: C901, 
         lock.release()
 
 
+def _edx_user_exists(user):
+    """
+    Returns True if the user can be verified to exist in edX, False otherwise.
+    """
+    try:
+        client = get_edx_api_client(user)
+        if client is not None:
+            client.user_info.get_user_info()
+    except:  # noqa: E722
+        return False
+    return True
+
+
 def create_edx_user(user, edx_username=None):
     """
     Makes a request to create an equivalent user in Open edX
@@ -442,17 +455,11 @@ def create_edx_user(user, edx_username=None):
         log.warning("create_edx_user: skipping create for %s, no edx_username", user)
         return False
 
-    if open_edx_user.has_been_synced:
-        # Here we should check with edx that the user exists on that end.
-        try:
-            client = get_edx_api_client(user)
-            client.user_info.get_user_info()
-        except:  # noqa: S110, E722
-            pass
-        else:
-            open_edx_user.has_been_synced = True
-            open_edx_user.save()
-            return False
+    if open_edx_user.has_been_synced and _edx_user_exists(user):
+        # User already exists in edX, mark as synced and skip creation
+        open_edx_user.has_been_synced = True
+        open_edx_user.save()
+        return False
 
     try:
         return _create_edx_user_request(open_edx_user, user, access_token)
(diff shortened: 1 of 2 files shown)


openedx/api.py
```diff
@@ -441,17 +455,11 @@ def create_edx_user(user, edx_username=None):
...
-        except:  # noqa: S110, E722
-            pass
-        else:
-            open_edx_user.has_been_synced = True
-            open_edx_user.save()
-            return False
+    if open_edx_user.has_been_synced and _edx_user_exists(user):
+        # User already exists in edX, mark as synced and skip creation
+        open_edx_user.has_been_synced = True
+        open_edx_user.save()
```
Anas12091101: This seems to be redundant, as this code will only execute when open_edx_user.has_been_synced = True. Is there a reason we are doing this?

openedx/api.py
```diff
@@ -1257,6 +1274,8 @@ def sync_enrollments_with_edx(
 ) -> SyncResult[courses.models.CourseRunEnrollment]:
     """Syncs enrollment records so that local enrollments match the enrollment data in edX"""
     client = get_edx_api_client(user)
+    if client is None:
+        return SyncResult()
```
Anas12091101, line 1317: Can we add test for this?

openedx/api.py
```diff
@@ -1333,6 +1352,8 @@ def subscribe_to_edx_course_emails(user, course_run):
         EdxApiChangeEmailSettingsException: Raised if an unknown error was encountered during the edX API request
     """
     edx_client = get_edx_api_client(user)
+    if edx_client is None:
+        return None
```
Anas12091101, line 1395: Can we add a test for this?

openedx/api.py
```diff
@@ -1359,6 +1380,8 @@ def unsubscribe_from_edx_course_emails(user, course_run):
...
```
Anas12091101, line 1423: Can we add a test for this?

Anas12091101 (changes requested):
Looks good.. just a few requests:

Anas12091101 (approved):
LGTM 👍

PR #3500, opened 2026-04-17
Add helpers and include Program in HubSpot sync

cp-at-mit:
NA

Creates deals in Hubspot when a user adds a UAI course to their cart.

### How can this be tested?
Copy the same process as described in https://github.com/mitodl/mitxonline/pull/3474 but for a course.


arslanashraf7:

courses/utils.py
```diff
@@ -183,6 +183,26 @@ def is_uai_course_run(course_run):
...
+    Returns:
+        bool: True if the program is UAI, False otherwise
+    """
+    if not program or not program.readable_id:
+        return False
+
+    readable_id = program.readable_id
+    # UAI programs use the format "program-v1:UAI+..." or just "UAI+..."
+    return readable_id.startswith(
+        (f"program-v1:{UAI_COURSEWARE_ID_PREFIX}", UAI_COURSEWARE_ID_PREFIX)
```
Anas12091101: This will also return false for UAI programs. The prefix is defined as UAI_COURSEWARE_ID_PREFIX = "UAI_", while UAI program IDs start with "program-v1:UAI+".

We should instead update the check to match "program-v1:UAI" so that UAI programs are correctly identified.

ecommerce/api.py
```diff
@@ -1091,12 +1092,19 @@ def create_verified_program_course_run_enrollment(request, courserun, program):
         redeemed_basket=basket,
     )
 
-    # Sync with HubSpot for CourseRun products
-    if isinstance(product.purchasable_object, CourseRun):
+    # Sync with HubSpot for CourseRun and Program products
+    if is_product_courserun(product) or is_product_program(product):
```
Anas12091101: This check seems redundant. A product can only be either a course run or a program, and in both cases, this condition will evaluate to True.

ecommerce/api.py
```diff
@@ -1091,12 +1092,19 @@ def create_verified_program_course_run_enrollment(request, courserun, program):
...
             product,
-            is_uai_course=is_uai_course_run(product.purchasable_object),
+            is_uai=(
+                is_product_courserun(product)
+                and is_uai_course_run(product.purchasable_object)
+            )
+            or (
+                is_product_program(product)
+                and is_uai_program(product.purchasable_object)
+            ),
```
Anas12091101: I’m not sure why we’re checking `is_product_courserun` and `is_product_program` here. On L1076, the product is explicitly fetched using the course run content type, so it should always be a course run.

Is there something I’m missing here?

```suggestion
            is_uai=is_uai_course_run(courserun) or is_uai_program(program),
```

Anas12091101 (commented):
I ran the following command to test the program creation flow:

```bash
docker compose run --rm web python manage.py shell -c "
from users.models import User
from ecommerce.models import Product
from hubspot_sync.api import track_cart_add_with_hubspot
user = User.objects.get(email='muhammad.anas@arbisoft.com')
product = Product.objects.get(id=6)
result = track_cart_add_with_hubspot(user, product, is_uai_course=True)
print(f'Result: {result}')
"
```
But it failed with the error:
```
ValueError: Cannot query "program-v1:UAI_MITx+TestProg | UAI Test Program": 
Must be "CourseRun" instance.
```

It's breaking here: https://github.com/mitodl/mitxonline/blob/sync-programs-with-hubspot-cart-add/hubspot_sync/serializers.py#L69

Anas12091101:
@cp-at-mit, pre-commit is failing. Could you please fix that?

cp-at-mit:
@Anas12091101 I don't think I'm gonna be able to fix those pre-commit issues easily and still get this PR released today.  I think most of those pre-commit errors are unrelated to my changes but are showing up because I changed the file.  Do you agree?

Anas12091101:
In that case, we can update [drf_lint_baseline.json](https://github.com/mitodl/mitxonline/blob/main/drf_lint_baseline.json) and include the lines causing the issue

cp-at-mit:
@Anas12091101 looks like updating that json file has allowed the pre-commit check to pass.

b2b/api.py
```diff
@@ -1016,12 +1042,19 @@ def _prepare_basket_for_b2b_enrollment(request, product: Product) -> Basket:
     item = BasketItem.objects.create(product=product, basket=basket, quantity=1)
     item.save()
 
-    # Sync with HubSpot for CourseRun products
-    if isinstance(product.purchasable_object, CourseRun):
+    # Sync with HubSpot for CourseRun and Program products
+    if is_product_courserun(product) or is_product_program(product):
```
Anas12091101: I think this check is redundant, since a product can only be either a course or a program. Please let me know if I’m missing something.

The same concern applies to all other places where this check is used.

Anas12091101 (approved):
LGTM 👍 Just a small non-blocker concern:

PR #3501, opened 2026-04-17
Force explicit calls to generate enrollment codes for B2B contracts

jkachel:

Closes mitodl/hq#10708

It seems like the problem we have with enrollment code generation is caused by excessive calls to the reconciliation code. The `save` method on the `ContractPage` calls the enrollment code check and we do that pretty frequently especially when we're adding courses to contracts.

Instead of doing that, this PR removes the call to `ensure_enrollment_codes_exist` in the `ContractPage.save` function. If we need to generate codes for a contract, we'll do it explicitly.

This PR also adds a `queue_codes` arg to the `create_contract_run` function; if set to True, the function will queue up the enrollment code check call. Additionally, there's a new `--make-codes` flag for the `b2b_courseware add` command which will also queue the code check at the end of the process if specified. 

However, the `b2b_codes validate --fix` command has existed for a bit now; we should change our process to add the courseware items to the contract, and then call that command when we've completed that task. This ensures the code check call happens once, when it's necessary.

### How can this be tested?

Automated tests should pass. Specifically, some of the automated tests for the `b2b` app call `create_contract_run` with the code check option turned on, so they should trigger the enrollment code check. (See the diff for specifics.)

Any operation that involves saving the contract should no longer trigger the enrollment code check - so, creating or modifying a contract (either on the command line or via Wagtail), or adding courseware to a contract, should be tested.

Test the `b2b_courseware add --make-codes` flag - add a course to the contract with that flag specified and it should trigger the code check. Without the `--make-codes` flag, it shouldn't try to queue it.

Test that `b2b_codes validate --fix` works as expected. It was not changed but it'd be good to sanity check it now, since we'll depend on it once this gets released.

### DIFF
--- a/b2b/api.py
+++ b/b2b/api.py
@@ -32,7 +32,7 @@ from b2b.models import (
     OrganizationPage,
     UserOrganization,
 )
-from b2b.tasks import queue_contract_sheet_update_post_save
+from b2b.tasks import queue_contract_sheet_update_post_save, queue_enrollment_code_check
 from cms.api import get_home_page
 from courses.constants import UAI_COURSEWARE_ID_PREFIX
 from courses.models import Course, CourseRun, Department, EnrollmentMode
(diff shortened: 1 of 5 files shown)


annagav (approved):
Works for me 👍

PR #3505, opened 2026-04-21
Add more logging for B2B enroll/attach error states

jkachel:

mitodl/hq#10988

The above issue is not the only place where this has been an issue - we've gotten a few reports of users trying to attach to a code membership contract with no seat limit, and receiving an "invalid code" error. We just send the error state back to the user, though, so we don't have anything we can use to trace the error back on our end. This adds some error/exception logging in a few places so we will have that data in the future.

### How can this be tested?

Hit the enroll and attach APIs with some bad data - you should start seeing error logs from these things. Some scenarios to try include enrolling in a B2B course with no session; enrolling in a B2B course with an invalid product; and attaching to a contract that's full.

### DIFF
--- a/b2b/api.py
+++ b/b2b/api.py
@@ -980,19 +980,35 @@ def _validate_b2b_enrollment_prerequisites(user, product: Product) -> Union[dict
         dict with error result if validation fails, None if validation passes.
     """
     if not user.is_authenticated:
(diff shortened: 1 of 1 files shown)


annagav (approved):
👍

PR #3511, opened 2026-04-21
Add skip_certificates flag to HubSpot property setup

cp-at-mit:
Introduce a skip_certificates keyword flag to _ensure_target_hubspot_custom_properties and thread it through the backward-compatible wrapper _ensure_target_hubspot_contact_properties. When skip_certificates is True the contact certificate properties are not added to the target account (used for UAI courses that don't need them). Default behavior remains unchanged (flag defaults to False) and the call in track_cart_add_with_hubspot now passes is_uai_course to avoid creating certificate properties for UAI deals.

Introduce a skip_certificates keyword flag to _ensure_target_hubspot_custom_properties and thread it through the backward-compatible wrapper _ensure_target_hubspot_contact_properties. When skip_certificates is True the contact certificate properties are not added to the target account (used for UAI courses that don't need them). Default behavior remains unchanged (flag defaults to False) and the call in track_cart_add_with_hubspot now passes is_uai_course to avoid creating certificate properties for UAI deals.

### How can this be tested?
Not really sure.  I'm not sure why the data is so large in mitx online when syncing with hubspot.

### DIFF
--- a/hubspot_sync/api.py
+++ b/hubspot_sync/api.py
@@ -1699,7 +1699,7 @@ def _normalize_line_item_properties_for_target_account(
             line_item_properties["status"] = resolved_status
 
 
-def _ensure_target_hubspot_custom_properties(hubspot_client: HubspotApi) -> None:
+def _ensure_target_hubspot_custom_properties(hubspot_client: HubspotApi, *, skip_certificates: bool = False) -> None:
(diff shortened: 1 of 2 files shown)


annagav (approved):
LGTM 👍

PR #3516, opened 2026-04-22
feat: add language and is_primary_language fields to CourseRun for multilingual support

jkachel:
_Co-authored with Copilot_

mitodl/hq#3516

This adds the necessary backend support for course run internationalization in MITx Online. That is:
- Adds a `language` field to the `CourseRun`
- Adds a `is_primary_language` field to the `CourseRun` to designate a particular run as the main language for the run tag group
- Updates the APIs to return this data
- Updates the `Course` APis to return back some nicely formatted data for available languages
- Updates the B2B system to ensure it clones course runs appropriately when they have >1 language
- Updates `import_courserun` to allow the language to be specified (since this data doesn't exist in edX) and a _specific_ run tag to be specified (since we want these to stay the same)

### How can this be tested?

Automated tests should pass. There are a handful of specific tests for language functionality.

It's easiest to test this by starting in edX:
- Create a _main course run_ that can be a source course. This can be something like `course-v1:INTLTEST+TestCourse.001+1T2026`. 
- Re-run the course in another language - leave the parameters the same, but set the run tag to something appropriate, like `1T2026_SP`. Resulting courseware ID should be something like `course-v1:INTLTEST+TestCourse.001+1T2026_SP` depending on how you set up the first course.
- This is sufficient but add more of these if you want. 

Import the course run into MITx Online:

`import_courserun --courserun course-v1:INTLTEST+TestCourse.001+1T2026 --live --create-cms-page -d "Translations" --create-depts --publish-cms-page --source-course --language en --primary-lang` 

This will create the main course run in the system. You should get a warning message about the run tag - in this case, it is safe to ignore (for a new course, you do want to use the default generally).

Import the second run into MITx Online:

`import_courserun --courserun course-v1:INTLTEST+TestCourse.001+1T2026_SP --live -d "Translations" --source-course --language sp --run-tag 1T2026` 

This will create the second course run in MITx Online. Some things to check for:
- The run tag should be `1T2026`, _not_ `1T2026_SP`.
- The courseware ID should reflect what you passed in (including the run tag difference). 
- If you add the CMS flags, you should get an error - it will try to make the CMS page for the course and will hit a wall there, because the first command already made it.

Since these courses are `live=True`, you should be able to see them in the API output. Specifically: http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/courses/api_v2_courses_list and then specifying the readable ID for the main course should only show you one course run - the first, primary one - but you should now see a "langauge_options" property in the output that should contain both runs.

Additionally, you should be able to search by language code - note that you will still get the _primary_ course run back as the main data for the course run, but it should show you runs that have the specified language in them.

Then, test with a B2B contract:

Create a new B2B contract/org (or select one to use, doesn't really matter). Then, add the course you created to the contract. It should create new course runs for _both_ languages and queue reruns of each in edX as well.

Add yourself to the org and contract, then search for courses in the contract. You should see the proper run options for the contract, including the additional languages.

### Additional Context

This essentially shoehorns an additional level of granularity into the course run hierarchy. We're regarding all of the runs that are tagged in a particular fashion as a single run - in the test examples, all of those courses are 1T2026, we just happen to have one in English and one in Spanish. But it's the same content and requirements and dates, so we keep the same run tag within MITx Online.

The FK between course runs in MITx Online and edX is the `courseware_id` field. edX will need a different courseware ID for each run, so the run tag part of that will include the language code where appropriate.

The language code defaults to empty. The code should regard either a course with a language and the primary flag to be the primary, or a course with no language code set at all to be the primary. But we should probably backfill the courses we have out there at some point, especially if we're planning to add translations for the particular course.

The DRF lint baseline was updated as well here because the changes made moved some stuff around - but there should not be any _new_ things added to it, the changes are just where the existing things moved after adding some more code elsewhere.


courses/management/commands/import_courserun.py
```diff
@@ -297,6 +310,21 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument  # noqa: C
 
             run, page, product = run_data
 
+            if kwargs.get("language", False):
+                run.language = kwargs.pop("language")
+                run.is_primary_language = kwargs.get("primary_lang", False)
```
gumaerc, line 314: If the user doesn't set `--primary-lang` when running this, won't it result in an invisible courserun? The `--language` argument defaults to "en", so that if statement above will never evaluate false. The `first_unexpired_run` filter on the `Course` model filters with `.filter(Q(language__isnull=True) | Q(is_primary_language=True))`. In this case, it will set the language to "en" and set `primary_lang` to `False`, resulting in no `next_run_id`.

courses/management/commands/import_courserun.py
```diff
@@ -297,6 +310,21 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument  # noqa: C
 
             run, page, product = run_data
 
+            if kwargs.get("language", False):
+                run.language = kwargs.pop("language")
```
gumaerc: Popping the language here means that subsequent runs being processed in the loop get `None` for language.

b2b/api.py
```diff
@@ -278,6 +275,77 @@ def import_and_create_contract_run(  # noqa: PLR0913
...
+        List of distinct source CourseRun objects, one per language (or one
+        for the "no language" legacy case).
+    Raises:
+        SourceCourseIncompleteError: if require_designated is True and no source
+            run is found, or if more than one source run exists for the same
+            language.
+    """
+
+    primary_source_run = (
+        course.courseruns.filter(is_source_run=True)
```
gumaerc: Are we phasing out matching on `SOURCE` being in the run ID? I noticed we aren't filtering on those here, but it's still part of the filter on the model.

courses/models.py
```diff
@@ -1206,9 +1222,86 @@ class CourseRun(TimestampedModel):
         default=False,
         help_text='Designate this run as a "source" run for contract re-runs of the course.',
     )
+    language = models.CharField(  # noqa: DJ001
+        max_length=8,
+        blank=True,
```
gumaerc, line 1263: We allow both blank and null here, but filtering logic downstream is only checking `isnull`. This means that a blank language is being treated as specified, but blank. Is that intentional?

drf_lint_baseline.json
```diff
@@ -1,4 +1,10 @@
 [
+  ".venv/lib/python3.11/site-packages/django/core/serializers/python.py:31:8:ORM001",
+  ".venv/lib/python3.11/site-packages/djoser/serializers.py:178:24:ORM001",
+  ".venv/lib/python3.11/site-packages/djoser/webauthn/serializers.py:30:11:ORM001",
+  ".venv/lib/python3.11/site-packages/djoser/webauthn/serializers.py:59:31:ORM001",
+  ".venv/lib/python3.11/site-packages/rest_framework_simplejwt/serializers.py:116:20:ORM001",
+  ".venv/lib/python3.11/site-packages/rest_framework_simplejwt/serializers.py:177:15:ORM001",
```
gumaerc: Should this file be referencing a specific venv?

courses/management/commands/create_courseware.py
```diff
@@ -670,6 +672,19 @@ def add_arguments(self, parser) -> None:
             const="course",
         )
 
+        parser.add_argument(
+            "--language",
+            "--lang",
+            type="str",
```
gumaerc: If you try and run this now you get `ValueError: 'str' is not callable`. I think this needs to be the actual string type and not a string literal.

gumaerc (changes requested):
I was able to test this end to end from OpenEdx all the way down into Learn and see the language options come through. The main issues I noticed were some syntax with the management commands themselves, along with a few other questions:

b2b/api.py
jkachel: Generally: yes we favor the flag over the run tag being `SOURCE`. I added that filter back here, though; I don't really want to have to tackle that transition in this PR.

drf_lint_baseline.json
jkachel: no.. but this is a good reminder to check your shell globs when updating baselines because the `find` command I was using doesn't filter out `.venv`

(I don't think it ultimately matters because `.venv` is in `.gitignore` but still shouldn't be in here)

courses/management/commands/import_courserun.py
```diff
@@ -297,6 +309,21 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument  # noqa: C
 
             run, page, product = run_data
 
+            if not kwargs.get("language", False):
```
gumaerc: Isn't this conditional inverted? It says "If `language` (default false) is *not* set, set the language on the run to the `language` argument being passed in, which we just determined to not be set?

courses/management/commands/create_courseware.py
```diff
@@ -670,6 +672,19 @@ def add_arguments(self, parser) -> None:
             const="course",
         )
 
+        parser.add_argument(
+            "--language",
+            "--lang",
+            type=str,
+            help='Set the language for the course run. (Default "en".)',
+            default="en",
```
gumaerc: We're setting the default to "en" here like we were in `import_courserun`, is that correct? When running the command, we default `primary_lang` to `False` during the `get` (as well as `language` to "en" again) so if neither are specified again we end up with a run with a language set but no default. It seems that again `first_unexpired_run` will not come back with anything as it filters with `.filter(Q(language="") | Q(is_primary_language=True))`.

gumaerc (changes requested):
A lot was fixed, but I think there are still a couple more issues:

courses/management/commands/import_courserun.py
jkachel: fixed in https://github.com/mitodl/mitxonline/pull/3516/commits/b6fbf7565a7feba4096aba1d5ac10361e261af02

gumaerc (approved):
👍 LGTM!

PR #3517, opened 2026-04-22
Make basket serializer test deterministic

cp-at-mit:
Replace non-deterministic factory defaults with explicit values and consistent rounding to avoid flaky failures. The test now creates a Product with a fixed price Decimal("100.00") and a 20% percent-off discount (DISCOUNT_TYPE_PERCENT_OFF, amount=20), attaches the product to the BasketItem, and computes the discounted price via DiscountType.for_discount(...). The raw result is quantized to Decimal("0.01") to match the serializer's rounding, and assertions verify the serialized total_price, discounted_price, and that the final price equals Decimal("80.00"). Comments were added to clarify the deterministic intent and the rounding behavior.

- [ ] Desktop screenshots

### DIFF
--- a/ecommerce/serializers/serializers_test.py
+++ b/ecommerce/serializers/serializers_test.py
@@ -214,21 +214,20 @@ def test_basket_with_product_serializer():
     """
     Tests serialization of a basket with the attached products (and any
     discounts applied).
-    
+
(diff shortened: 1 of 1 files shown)


rhysyngsun (approved):
LGTM

PR #3518, opened 2026-04-22
Fix transaction race condition with hubspot sync

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/11012

This fixes an issue that happens when there's a race condition between the middleware and SCIM and the transaction that created a user in the middleware hits a conflict and rolls back. Will deal with the rollback separately but this will make this area of code more robust to this kind of error should it ever come up in another way.

### How can this be tested?

Tests should pass. This is incredibly hard to reproduce because it requires sub-second timing.

### DIFF
--- a/users/signals.py
+++ b/users/signals.py
@@ -1,3 +1,4 @@
+from django.db import transaction
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
I'm going to go ahead and approve this just based on looking at the code, because as you mentioned it's going to be basically impossible to reproduce locally, especially on a fast computer.

PR #3519, opened 2026-04-22
Fixes 500 errors when viewing an Organization in the Django Admin

jkachel:

n/a

Noticed that the Django Admin consistently returns a 500 when viewing a B2B organization in the Django Admin in production. This hopefully fixes that by prefetching the user and specifically displaying only the user email for the `UserOrganizationAdminInline`. 

This also removes the add/change/delete permission from the inline. If you want to manage admins, you can do that from the individual user's edit screen in the Django Admin. 

### How can this be tested?

View an organization page in the Django Admin. It should work as expected. Ideally, use an org that has some managers set up in it. You can set that up in the user record.

### Additional Context

Assuming this will help - the page takes a while to load and then errors out with a 502, but the Django process doesn't ever fail really so there's no Sentry error pointing to something specific. I can't reproduce this on RC or locally. But, the User model can be pretty unwieldy to load so this seemed like a good place to start.

### DIFF
--- a/b2b/models.py
+++ b/b2b/models.py
@@ -643,11 +643,6 @@ class UserOrganization(models.Model):
 
         return f"UserOrganization: {self.user} in {self.organization}"
 
-    def user_email(self):
-        """Return the user's email address."""
-
-        return self.user.email
-
 
 def is_organization_manager(user, org_id):
     """
--- a/b2b/models.py
+++ b/b2b/models.py
@@ -643,6 +643,11 @@ class UserOrganization(models.Model):
 
         return f"UserOrganization: {self.user} in {self.organization}"
 
+    def user_email(self):
+        """Return the user's email address."""
+
+        return self.user.email
+
 
 def is_organization_manager(user, org_id):
     """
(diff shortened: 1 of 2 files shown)


b2b/models.py
```diff
@@ -643,6 +643,11 @@ def __str__(self):
 
         return f"UserOrganization: {self.user} in {self.organization}"
 
+    def user_email(self):
```
rhysyngsun: I think this is a better served with `@admin.display` (see the third example [here](https://docs.djangoproject.com/en/6.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display)), since this is very specific to django-admin.

rhysyngsun (changes requested):

b2b/models.py
jkachel: Fixed in https://github.com/mitodl/mitxonline/pull/3519/commits/3ac1482b10c5d3795a109f1d793267ef70cde585

rhysyngsun (approved):
LGTM

PR #3520, opened 2026-04-23
Add skip_certificates for UAI HubSpot sync

cp-at-mit:
Introduce skip_certificates flag to contact sync functions so certificate properties can be excluded for UAI HubSpot accounts. Propagate the flag to make_contact_sync_message_from_user and _ensure_hubspot_contact_for_user, detect UAI target accounts by comparing the token to the UAI_MITOL_HUBSPOT_API_PRIVATE_TOKEN setting, and skip program/course certificate properties when requested. Improve logging for contact lookups/creates and deal-contact associations, update local HubspotObject mapping and user.hubspot_sync_datetime on create, and add error handling that logs failures and returns None when contact ensure fails. Also update track and targeted sync call sites to pass the skip flag.

NA

Adds helpful logging to debug hubspot API calls since we make many calls during user actions.  Also fixes and issue when creating a contact in UAI hubspot (xpro hubspot for now).

### How can this be tested?
Checking out with a UAI course should successfully create your user in the UAI/Xpro hubspot.  Triggering a failure of this locally is a bit tough since you would need to create a lot of data for this to fail.  But ensuring that it works is helpful to make sure we're not breaking anything that currently works.

### DIFF
--- a/hubspot_sync/api.py
+++ b/hubspot_sync/api.py
@@ -2175,13 +2175,18 @@ def track_cart_add_with_hubspot(
 
     try:
         hubspot_client = HubspotApi(access_token=token)
+        
+        # Determine which account is actually being used for the API call
+        # This ensures skip_certificates reflects the actual HubSpot account, not just course type
+        is_uai_account = token == getattr(settings, "UAI_MITOL_HUBSPOT_API_PRIVATE_TOKEN", "")
+        
         _ensure_target_hubspot_contact_properties(
-            hubspot_client, skip_certificates=is_uai_course
+            hubspot_client, skip_certificates=is_uai_account
         )
 
         # UAI deals must have a contact in the same HubSpot account.
         contact_id = _ensure_hubspot_contact_for_user(
-            user, hubspot_client, skip_certificates=is_uai_course
+            user, hubspot_client, skip_certificates=is_uai_account
         )
         if not contact_id:
             return False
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
This worked for me. I was able to create a new test user, see the contact get created, see the deal get created when adding to cart, and see the deal subsequently closed after checkout. I don't really have any meaningful feedback aside from AI recommending we add a test like this:

```
def test_sync_deal_with_hubspot_targeted_uai_omits_certificate_contact_properties(
    settings, mocker, hubspot_order
):
    """Checkout-targeted UAI sync should create contacts without certificate properties."""
    settings.UAI_MITOL_HUBSPOT_API_PRIVATE_TOKEN = "uai-token"  # noqa: S105

    mock_client_cls = mocker.patch("hubspot_sync.api.HubspotApi")
    mocker.patch("hubspot_sync.api.wait_for_hubspot_rate_limit")
    mocker.patch(
        "hubspot_sync.api._build_target_deal_message",
        return_value=SimplePublicObjectInput(properties={"dealname": "deal-1"}),
    )
    mocker.patch("hubspot_sync.api._find_target_deal_id_by_dealname", return_value=None)
    mocker.patch("hubspot_sync.api._find_hubspot_contact_id_by_email", return_value=None)
    mocker.patch("hubspot_sync.api._ensure_target_line_item_for_line", return_value=None)

    mock_client = mock_client_cls.return_value
    mock_client.crm.objects.basic_api.create.side_effect = [
        SimpleNamespace(id="deal-1"),
        SimpleNamespace(id="contact-1"),
    ]
    mock_client.crm.associations.v4.basic_api.get_page.return_value = SimpleNamespace(
        results=[]
    )
```

I'm not sure if it's super necessary though.

PR #3524, opened 2026-04-24
Skip HubSpot sync for B2B users

cp-at-mit:
Avoid syncing B2B users to HubSpot by checking user.b2b_contracts.exists() in sync_contact_with_hubspot and sync_hubspot_user; log an informational message and return early. Update the Celery task wrapper to handle a None result (return result.id if present). Add tests covering skip and normal sync behavior and required factory imports.

https://github.com/mitodl/hq/issues/11009

Avoid syncing B2B users to HubSpot by checking user.b2b_contracts.exists() in sync_contact_with_hubspot and sync_hubspot_user; log an informational message and return early. Update the Celery task wrapper to handle a None result (return result.id if present). Add tests covering skip and normal sync behavior and required factory imports.

Also avoids creating deals for b2b users.


arslanashraf7:

hubspot_sync/task_helpers_test.py
```diff
@@ -93,6 +94,74 @@ def test_sync_hubspot_user(mocker, mock_exception_log, user, raise_exc):
...
+
+    # Should not call the sync task
+    mock_sync.assert_not_called()
+
+    # Should log that user was skipped
+    mock_info_log.assert_called_once_with(
+        "Skipping HubSpot sync for B2B user %s (user_id=%d)",
+        user.edx_username,
+        user.id,
+    )
```
Anas12091101, line 125: Instead of patching django.db.models.query.QuerySet.exists globally, use real factory data to set up the B2B contract relationship. The global mock overrides .exists() for every queryset in the test, not just user.b2b_contracts.exists(), which can hide bugs and cause subtle failures in unrelated code paths:


```suggestion
def test_sync_hubspot_user_skips_b2b_users(mocker, settings):
    """sync_hubspot_user should skip B2B users and not call HubSpot sync"""
    settings.MITOL_HUBSPOT_API_PRIVATE_TOKEN = "faketoken"  # noqa: S105

    mock_sync = mocker.patch(
        "hubspot_sync.task_helpers.tasks.sync_contact_with_hubspot.delay"
    )
    mock_info_log = mocker.patch("hubspot_sync.task_helpers.log.info")

    user = UserFactory.create()
    contract = ContractPageFactory.create()
    user.b2b_contracts.add(contract)

    # Reset mocks after user creation to clear any calls during setup
    mock_sync.reset_mock()
    mock_info_log.reset_mock()

    sync_hubspot_user(user)

    # Should not call the sync task
    mock_sync.assert_not_called()

    # Should log that user was skipped
    mock_info_log.assert_called_once_with(
        "Skipping HubSpot sync for B2B user %s (user_id=%d)",
        user.edx_username,
        user.id,
    )
```

hubspot_sync/api.py
```diff
@@ -1363,8 +1357,18 @@ def sync_deal_with_hubspot_targeted(order: Order, token: str) -> SimplePublicObj
...
+    # Skip sync for B2B users to avoid errors
+    if order.purchaser.b2b_contracts.exists():
+        log.info(
+            "Skipping HubSpot deal sync for B2B user %s (user_id=%d, order_id=%d)",
+            order.purchaser.edx_username or order.purchaser.email,
+            order.purchaser.id,
+            order.id,
+        )
+        return None
+
```
Anas12091101, line 1389: Instead of adding this here, we should move the check into the [`sync_hubspot_deal`](https://github.com/mitodl/mitxonline/blob/hubspot-b2b-users/hubspot_sync/task_helpers.py#L43) function so the task only runs when a B2B contract does not exist.

Anas12091101 (changes requested):
Some changes

Anas12091101 (approved):
LGTM 👍

PR #3528, opened 2026-04-28
Search deals by unique_app_id to avoid duplicates

cp-at-mit:
When syncing deals, read unique_app_id from the built deal input and, if no match is found by dealname, also search the target HubSpot account by unique_app_id to prevent duplicate deals. Adds _find_target_deal_id_by_unique_app_id which uses the search API (returns the first match or None), re-raises TooManyRequestsException so callers can retry, and logs other failures.

This was the error response we were getting from Hubspot:
```
HTTP response body: {"status":"error","message":"Cannot set PropertyValueCoordinates{portalId=20596155, objectTypeId=ObjectTypeId{legacyObjectType=DEAL}, propertyName=unique_app_id, value=mitxonline-421782} on 59724531777. 59724604020 already has that value.","correlationId":"019dd074-190d-77c5-8845-07addbd588ef","category":"VALIDATION_ERROR"} 
```

When syncing deals, read unique_app_id from the built deal input and, if no match is found by dealname, also search the target HubSpot account by unique_app_id to prevent duplicate deals. Adds _find_target_deal_id_by_unique_app_id which uses the search API (returns the first match or None), re-raises TooManyRequestsException so callers can retry, and logs other failures.

Noticed this issue when we're trying to create new "add to cart" deals in hubspot from MITx Online.  This mainly occurs when a customer is adding a UAI course to their cart which we try create a deal in the Xpro hubspot account.

### How can this be tested?
You should be able to create a deal in Hubspot by adding a UAI course to your cart in MITx Online. Then go into hubspot and rename the deal.  When you checkout within MITx Online, then the same deal should be updated with a close date in Hubspot.


hubspot_sync/api.py
```diff
@@ -1369,10 +1369,17 @@ def sync_deal_with_hubspot_targeted(order: Order, token: str) -> SimplePublicObj
...
     deal_input = _build_target_deal_message(order, hubspot_client)
     dealname = deal_input.properties.get("dealname")
+    unique_app_id = deal_input.properties.get("unique_app_id")
 
-    # Check if deal already exists
+    # Check if deal already exists by dealname first
     existing_deal_id = _find_target_deal_id_by_dealname(hubspot_client, dealname)
 
+    # If not found by dealname, also check by unique_app_id to prevent duplicates
+    if not existing_deal_id and unique_app_id:
```
gumaerc, line 1378: Can we add a regression test for this new functionality? It should check that update is called instead of create in this specific instance where the deal name doesn't match.

gumaerc (changes requested):
This worked for me, but I have one request regarding testing. Beyond that, We should either create a new issue for this or at least link to the PR that this is following up on so that the history in Github ties it together. I'm still a little unclear on the original story on why this was a problem. Is marketing renaming deals in HubSpot, or is this just an issue with old data not matching the deal name it's expecting now?

hubspot_sync/api.py
cp-at-mit: I think it's because we're not using Hubspot as expected and running into issues of our creation.  We're creating deals, for the same contacts, across two different hubspot accounts.

gumaerc (approved):
👍

PR #3532, opened 2026-04-28
fix: make DiscountFactory defaults evaluate per-instance to prevent flaky test_create_basket_with_product

Anas12091101:
https://github.com/mitodl/hq/issues/10549

#### Issue

`DiscountFactory` and its variants (`OneTimeDiscountFactory`, `OneTimePerUserDiscountFactory`, `UnlimitedUseDiscountFactory`, `SetLimitDiscountFactory`) used `random.randrange()` to set defaults for `amount`, `discount_type`, and `redemption_type`. These evaluate once at module import time, freezing a single value for every factory instance in the process. This made `test_create_basket_with_product` flaky because the frozen `redemption_type` (which the test never pinned explicitly) could vary across test runs, affecting discount validation and comparison behavior.

#### Solution

- Replace import-time `random.randrange()` calls in all five discount factories with `fuzzy.FuzzyInteger` / `fuzzy.FuzzyChoice`, which generate fresh values per instance.
- Pin explicit `discount_type` and `redemption_type` values in `test_create_basket_with_product` so the test controls all variables that affect discount comparison.
- Fix product price to `$1000.00` so product price randomness doesn't factor into discount comparisons.
- Fix unrelated test breakage in `flexiblepricing/api_test.py` (tier linkage) and `openedx/api_test.py` (caplog level).

- [ ] Desktop screenshots

### DIFF
--- a/ecommerce/factories.py
+++ b/ecommerce/factories.py
@@ -1,5 +1,3 @@
-import random
-
 import faker
 from factory import SubFactory, fuzzy
 from factory.django import DjangoModelFactory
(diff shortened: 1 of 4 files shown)


jkachel (approved):
LGTM 👍

PR #3535, opened 2026-04-29
Reuse cart-add orders; prefer unique_app_id

cp-at-mit:
Fixes an issues that's related to a couple issues.
https://github.com/mitodl/hq/issues/11103
Add logic and logging to reuse existing cart-add PendingOrder instances during checkout (ecommerce/models.py) by searching for individual product-based pending orders when a multi-product match isn't found; log when reusing or finding individual orders. Update HubSpot sync (hubspot_sync/api.py) to look up deals by unique_app_id (consistent across cart-add and checkout flows) and remove the deprecated dealname search helper; adjust tests (hubspot_sync/api_test.py) to expect only the unique_app_id lookup. Update serializer (hubspot_sync/serializers.py) to derive unique_app_id from the first line item's purchasable content (mitxonline-<model>-<id>), falling back to the legacy order-based app id for compatibility.

- [ ] Desktop screenshots

### How can this be tested?
1. Add a UAI course to your cart and verify that the deal is created in the xpro Hubspot.
2. PRoceed with the checkout process and verify that the same deal is updated in Hubspot.

I think this was not seen in local environments since our deployed environments have many more orders for the same user, along with deals for the same UAI product that already exist in MITx Online's hubspot that were created prior to us starting work on using the xpro hubspot account for UAI orders.


courses/views/v2/__init__.py
```diff
@@ -410,6 +410,7 @@ def get_queryset(self):
             queryset=CourseRun.objects.order_by("id")
             .select_related("b2b_contract")
             .prefetch_related(modes_prefetch, products_prefetch),
+            to_attr="prefetched_courseruns",
```
rhysyngsun: This appears unrelated to the issue?

ecommerce/models.py
```diff
@@ -791,36 +793,11 @@ def send_ecommerce_order_receipt(self):
 class PendingOrder(Order):
     """An order that is pending payment"""
 
-    @transaction.atomic
-    def _get_or_create(
```
rhysyngsun, line 795: I'm not entirely sure if this is a required change to fix the issue at hand and I'd have a preference to do this level of refactoring in its own PR at any rate.

rhysyngsun (changes requested):

ecommerce/models.py
cp-at-mit: Yeah this was to try and solve the issue.  I can maybe revert it and create a hubspot specific function

annagav:
I was trying to functionally test the PR, and I think I am missing something. I see that we sync deals on order creation, but when I add a course run to cart there is no order created at that point and I can't see where a hubspot sync would get triggered at that point.

cp-at-mit:
Closing as Nathan suggested that the refactoring included in this PR adds too much of a change and should be done in a separate PR.  I will create a new PR to address the Hubspot issue with fewer changes.

"if you need to do the refactoring that really needs to happen in a separate PR, it's too complex of a delta to review for a rushed friday release"

PR #3537, opened 2026-04-30
Fix some issues with the multiple language support functionality

jkachel:

Closes mitodl/hq#11100

Makes some changes to the multi-language support in MITx Online: 

- Constraints on the `CourseRun` model are refactored to be simpler/more effective
- Consider B2B contract and source run flags separately for language and primary run flags
- Add more extensive tests to make sure run objects can be created as expected

This also adds in this related change, which will help the multi-language issues but is also a necessary larger change:

- By default, filter out source runs from the CourseRun dataset.

### How can this be tested?

Automated tests should pass. Tests should be reviewed to ensure they're not missing anything.

You should now be able to create sets of languages within a run tag for a course runs that are:
- flagged as source runs
- FKed to a B2B contract
- neither of these

You will still need a unique `courseware_id` for each run so they link to a specific course run in edX. But you should be able to set up (e.g.) `course-v1+MITxT+12.345x` with runs that have run tag `1T2026` with languages `en, jp, no` and have one of those be a primary language, and then you should be able to do that again with the runs set as source runs, and then do that again with the runs set to belong to a B2B contract (any number of times).

You should also not see source runs in the Courses API output, or anywhere else. So, given two runs for `course-v1:MITxT+12.345x+1T2026`, one marked as language `en` with the primary flag and the source run flag on, and the other marked as language `en` with neither flag, you should only see the _second_ one. (And you should be able to create the second one; you weren't able to before.)


courses/api.py
```diff
@@ -1371,7 +1371,7 @@ def import_courserun_from_edx(  # noqa: C901, PLR0913
                     if dept:
                         root_course.departments.add(dept.id)
 
-    new_run = CourseRun.objects.create(
+    new_run = CourseRun.all_objects.create(
```
gumaerc, line 1374: Above on line 1331 we are not using `all_objects` to filter. This causes source runs to be filtered out of that query response, which causes this line to try and create the run when it might already exist, making this not idempotent. I tried this and got an `IntegrityError`. On line 241 we also have:

```
run_qs = CourseRun.objects.filter(courseware_id=course_run_id)
```

This will similarly filter out source runs, and re-importing them will cause the same failure.

courses/models.py
```diff
@@ -166,6 +166,17 @@ def with_text_id(self, text_id):
...
 
 
+class UsableCourseRunManager(models.Manager):
+    """Simple manager to filter out source runs."""
+
+    def get_queryset(self):
+        """Return a queryset that ignores all source runs."""
+
+        return (
+            super().get_queryset().filter(Q(is_source_run=False) | Q(run_tag="SOURCE"))
```
gumaerc: This allows any run with the `SOURCE` tag to pass through regardless of what `is_source_run` is set to. I validated this with:

- A run with `is_source_run=True` AND `run_tag="SOURCE"` appears in `CourseRun.objects`
- API response includes the legacy SOURCE-tagged run in both courseruns and language_options
- Violates the requirement: "source runs don't appear in the API output"

Was the `Q(run_tag="SOURCE")` fallback intentional? If so, should it require is_source_run=False to also be True? For example:

```
return super().get_queryset().filter(
    Q(is_source_run=False) | Q(run_tag="SOURCE", is_source_run=False)
)
```

courses/views/v2/views_test.py
```diff
@@ -2359,3 +2359,79 @@ def test_program_products_prefetch_query_count(
...
+
+    resp_json = resp.json()
+
+    seen_run_ids = [run["id"] for run in resp_json["courseruns"]]
+
+    assert resp_json["next_run_id"] == (
+        translated_regular_run.id if primary == "transreg" else regular_run.id
+    )
+    assert source_run.id not in seen_run_ids
+    assert translated_regular_run not in seen_run_ids
```
gumaerc: This assertion compares a `CourseRun` model instance to a list of integers, so it always passes and is not asserting what it's supposed to.

gumaerc (changes requested):
I found some issues:

courses/models.py
jkachel: no, it is backwards. changing to `Q(is_source_run=False) | ~Q(run_tag="SOURCE")`

gumaerc (approved):
👍

PR #3539, opened 2026-05-01
Deduplicate deals and use consistent unique_app_id

cp-at-mit:
Prevent duplicate cart-add deals by looking up existing deals by dealname and a new consistent unique_app_id before creating; update existing deals when found. Change OrderToDealSerializer.get_unique_app_id to generate a stable ID based on purchaser email and the purchasable object (fallback to order id for orders without lines) so cart-add and checkout flows use the same identifier. Update tests to assert the new unique_app_id generation logic.

https://github.com/mitodl/hq/issues/11103

Hoping that this fixes an issue where multiple new deals were being created between the mitx online and xpro hubspot accounts, and sometimes duplicated on the same hubspot account, when a user adds a UAI course to their cart and completes the checkout process.

### How can this be tested?
1. Create a UAI course that will satisfy the logic in `is_uai_course_run`.
2. Login through Keycloak.
3. Add the UAI course from step 1 to your cart.
4. Verify that the logs show a celery task kicked off to create the hubspot deal in the xpro account.
5. Verify that the deal was created in the hubspot xpro account.
6. Complete the purchase of the course.
7. Verify that the logs show a celery task kicked off to update the hubspot deal in xpro account.
8. Verify that the same deal that was created in step 5 was updated and now has a "close" date.
9. Verify that no other hubspot deals were created.


hubspot_sync/serializers.py
```diff
@@ -187,7 +187,19 @@ def _get_discount(self, instance):
...
 
     def get_unique_app_id(self, instance):
-        """Get the app_id for the object"""
+        """Get the app_id for the object - based on user and purchasable object for consistency"""
+        # For consistency between cart-add and checkout, base the unique_app_id on
+        # the user and purchasable object rather than the order ID
+        first_line = instance.lines.first()
+        if first_line and first_line.purchased_object:
+            # Generate consistent ID based on user email and purchasable object
+            user_identifier = instance.purchaser.email.lower()
```
rhysyngsun: Will this be ok over time if the user changes their email or are we just concerned about generating an id that's unique enough for this moment in time? `User.global_id` might be a safer bet unless there's value in using `email`.

cp-at-mit: I think just unique enough for this moment in time.  I guess the order could be lost if the user adds it to their cart, changes their email, then completes the order.

annagav (approved):
I think this is working. I was able to verify that the deal was created on "add_to_cart" and updated on checkout. However, I had to use coupon for checkout in order to close the order. 

I saw some strange things which I think were happening because I could add multiple courses to cart, so I hope that is not an issue.

Another, thing is that we usually don't create an order on "add_to_cart" (at least that was the behavior I see locally), but when we sync this "event" with hubspot an order was created in order to sync, and when I further checkout out I verified that it didn't get duplicated.

PR #3542, opened 2026-05-01
Fixes some additional issues with the language support

jkachel:

Related to mitodl/hq#11100

There were a couple of other issues that needed fixing for course translations:

- The course run clone task wasn't firing - this was because the task hadn't been updated to use `CourseRun.all_objects` and we only really clone source courses, so it wasn't seeing anything to clone.
- The `create_contract_run` code needed a couple of updates to consider the language when checking for an existing run
- The `create_contract_run` code also needed to continue on if `no_reruns` is set (which it is for programs) - if you added a program, it'd end up adding one, then looping and determining that the course run already was there and then stopping

Also, adds some minor creature comfort things:
- The Course Run admin in Django Admin now shows primary language and source run flags, and the contract if there is one
- The Course admin in Django Admin now lists out the course's runs
- Using `b2b_courseware add <course>`, you can now specify `--allow-reruns` to explicitly tell it to re-run a particular course (it would default to doing this whether you wanted to or not; it now defaults to _not_ rerun courses).

### How can this be tested?

Automated test should work.

Adding a course or a program to a contract should kick off a clone task, and that should actually succeed now. You should at least see it be queued if you don't have edX set up.

Adding a course that has multiple languages via a program should work properly. You should be able to re-add a program to an existing contract where it hasn't worked right, and it should create the runs you're missing for it.

### DIFF
--- a/b2b/management/commands/b2b_courseware.py
+++ b/b2b/management/commands/b2b_courseware.py
@@ -137,14 +137,14 @@ Specifying a program will only unlink the program from the contract, unless "--r
         add_subparser.add_argument(
             "--no-create-runs",
             help="Don't create contract runs in edX for the specified course, just add it to the contract.",
-            dest="create_runs",
-            action="store_false",
+            dest="no_create_runs",
+            action="store_true",
         )
         add_subparser.add_argument(
             "--allow-reruns",
-            help="Don't re-run courses that already have a run, skip over them instead. (Use if adding a language to an existing course, or cleaning up.)",
+            help="Allow courses to be re-run.",
             dest="allow_reruns",
-            action="store_false",
+            action="store_true",
         )
         add_subparser.add_argument(
             "--force",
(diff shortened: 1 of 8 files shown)


b2b/management/commands/b2b_courseware.py
```diff
@@ -292,7 +301,11 @@ def handle_add(self, contract, coursewares, **kwargs):  # noqa: PLR0915, C901
                 # This is a course, so create a run (unless we've been told not to).
 
                 if self.create_run(
-                    contract, courseware, skip_edx=create_runs, org_prefix=org_prefix
+                    contract,
+                    courseware,
+                    skip_edx=create_runs,
```
gumaerc: The `create_runs` argument defaults to true, and when `skip_edx` is true `clone_courserun` is skipped. I was able to prove this by creating a fresh contract and trying to add a single course to it. The contract run was created, but when I patched `openedx.tasks.clone_courserun.delay` the call count was 0.

b2b/management/commands/b2b_courseware.py
```diff
@@ -138,6 +140,12 @@ def add_arguments(self, parser):
             dest="create_runs",
             action="store_false",
         )
+        add_subparser.add_argument(
+            "--allow-reruns",
+            help="Don't re-run courses that already have a run, skip over them instead. (Use if adding a language to an existing course, or cleaning up.)",
```
gumaerc: This language is reversed and a bit confusing. The argument is called `--allow-reruns` and down below we store the value of `allow_reruns` in a var called `no_reruns`. Since the action is `store_false`, setting this flag disables `no_reruns` and allows reruns. We should update the help text to reflect that.

b2b/models.py
```diff
@@ -561,11 +560,8 @@ def add_program_courses(self, program, order=None):
             models.Q(courseruns__is_source_run=True)
             | models.Q(courseruns__run_tag="SOURCE")
         ).all():
-            try:
-                create_contract_run(self, course, no_reruns=True)
-                managed += 1
-            except TargetCourseRunExistsError:  # noqa: PERF203
-                skipped_run_creation += 1
```
gumaerc, line 568: This function now always returns 0 for `skipped_run_creation`. We should either remove it from the logging or increment it where run creation is skipped.

gumaerc (changes requested):
Adding a program called `clone_courserun` as expected, but adding a single course did not:

jkachel:
https://github.com/mitodl/mitxonline/pull/3542/commits/2a7d2157f58aa66de54a2cb73866629d2327104f adds some tests for the `b2b_courseware` command - this isn't intended to be a complete set, just a sufficient amount to cover the changes made in this PR. (It doesn't test any of the remove functionality or code creation flags for instance.)

gumaerc (approved):
👍 LGTM!

PR #3548, opened 2026-05-05
feat: removes the MITx site header from all checkout-related views

zamanafzal:
https://github.com/mitodl/hq/issues/11122

Removes the MITx site header from all checkout-related views unconditionally (without depending on ?ecom-service=true), so branding stays consistent through cart/checkout flows including payment-cancel return paths.

Screenshot 2026-05-05 at 2 31 37 PM

Screenshot 2026-05-05 at 2 31 09 PM

Screenshot 2026-05-05 at 2 31 00 PM

### How can this be tested?
Confirm MITx site header is not visible at all the following urls.

- Open /cart/ (without ecom-service query param)

- Open /orders/history

   

- Open /orders/receipt/<any-id>

- Trigger checkout from cart (Place your order)

- In local, CyberSource may return 403 due to gateway auth/session config.
- Even after returning to /cart/ without query params, header remains hidden (regression target).

**In the end:**

Open /dashboard/ (control route)

_Confirm MITx site header is visible._

### DIFF
--- a/frontend/public/src/containers/App.js
+++ b/frontend/public/src/containers/App.js
@@ -73,7 +73,7 @@ export class App extends React.Component<Props, void> {
     return searchParams.has("ecom-service")
   }
 
-  isCheckoutPage() {
+  isCheckoutRelatedPage() {
     const { match, location } = this.props
     return (
       !!matchPath(location.pathname, {
@@ -104,7 +104,7 @@ export class App extends React.Component<Props, void> {
 
     return (
       <div className="app" aria-flowto="notifications-container">
-        {!this.isEcomServiceMode() && !this.isCheckoutPage() && (
+        {!this.isEcomServiceMode() && !this.isCheckoutRelatedPage() && (
           <Header
             currentUser={currentUser}
             cartItemsCount={currentUser.is_authenticated ? cartItemsCount : 0}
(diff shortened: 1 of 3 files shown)


frontend/public/src/containers/App.js
```diff
@@ -73,12 +73,26 @@ export class App extends React.Component<Props, void> {
...
+        path:  urljoin(match.url, String(routes.cart)),
+        exact: false
+      }) ||
+      !!matchPath(location.pathname, {
+        path:  urljoin(match.url, String(routes.orderHistory)),
+        exact: false
+      }) ||
+      !!matchPath(location.pathname, {
+        path:  urljoin(match.url, String(routes.orderReceipt)),
+        exact: false
```
dsubak, line 89: It looks like we used to pass `exact: true` to the matchPath call for orderReceipt - is changing it to false intentional?

Totally fine if so, just wanted to double check!

dsubak (approved):
Does what it says on the tin - I had to fake a bit of data so it's wasn't a precise test of the regression, but I was able to verify that all the listed pages no longer display an mitx site header. LGTM

frontend/public/src/containers/App.js
zamanafzal: Yes, it's used on purpose to keep matching any nested/extended checkout-related paths, not just a single exact URL.

PR #3550, opened 2026-05-05
Update pytest to not ignore all deprecations

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/11155

Updates the pytest configuration so it doesn't ignore all deprecations and instead selectively ignores known deprecations with the code at the moment. The django deprecation is already fixed in the django 5.2 upgrade PR.

### How can this be tested?

Tests should pass.


annagav (approved):
LGTM

PR #3551, opened 2026-05-05
only respond with enrolled contract on enrollment code redemption

gumaerc:
Pre-requisite for https://github.com/mitodl/hq/issues/11112

This PR changes the B2B enrollment code attach endpoint to return only the contract that you have enrolled in that is tied to the enrollment code you are redeeming. Previously it returned a list of all the users contracts every time.

### How can this be tested?
- Generate an enrollment code using the b2b management commands
- Redeem the code using the API and ensure that you only get the contract tied to the code in the response

### DIFF
--- a/b2b/views/v0/__init__.py
+++ b/b2b/views/v0/__init__.py
@@ -121,7 +121,7 @@ class AttachContractApi(APIView):
 
     @extend_schema(
         request=None,
-        responses=ContractPageSerializer(many=True),
+        responses=ContractPageSerializer,
     )
     def post(self, request, enrollment_code: str, format=None):  # noqa: A002, ARG002
         """
@@ -144,7 +144,7 @@ class AttachContractApi(APIView):
         - 200: Code valid but user already attached to all associated contracts
         - 404: Invalid or expired enrollment code
         - 409: Code valid but no available seats in associated contract(s)
-        - list of ContractPageSerializer - the contracts for the user
+        - ContractPageSerializer (or null) - the active contract associated with the code
         """
         now = now_in_utc()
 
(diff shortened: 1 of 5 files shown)


b2b/views/v0/__init__.py
```diff
@@ -144,7 +144,7 @@ def post(self, request, enrollment_code: str, format=None):  # noqa: A002, ARG00
         - 200: Code valid but user already attached to all associated contracts
         - 404: Invalid or expired enrollment code
         - 409: Code valid but no available seats in associated contract(s)
-        - list of ContractPageSerializer - the contracts for the user
+        - ContractPageSerializer (or null) - the active contract associated with the code
```
ChristopherChudzicki: @gumaerc This is a breaking change in the API, right? Previously returned an array of objects, now returns a single object. I think that will break Learn here: https://github.com/mitodl/mit-learn/blob/6063a3d2dab070229ce3ff3dbb4f235eeb64ad5b/frontends/main/src/app-pages/EnrollmentCodePage/EnrollmentCodePage.tsx#L45-L47

I agree that returning a single object is more natural.

But to avoid breaking the schema—which is consequential here—I would suggest we return `[singleObject]`

ChristopherChudzicki (changes requested):
Will leave rest of review to someone else, but i was just looking at the API schema concern and have a concern. Let me know if i am incorrect!

dsubak (approved):
Functionality looks good to me, was able to verify that I only get back the contract related to the code passed in in cases where I would've gotten everything back before. 

Looks like @ChristopherChudzicki feedback has been addressed as well, so if ya'll are happy with it I think this'll play

ChristopherChudzicki (approved):
I haven't tested (left to dan + you) but the spec change is fixed, thanks!

PR #3552, opened 2026-05-06
Add --make-codes flag to b2b_courseware call in test_datagen and language fixes

dsubak:
This script has been busted for a bit - there's been a fair number of changes around language handling, filtering out source runs in some cases and changing the implicit provisioning of enrollment codes in certain codepaths. This attempts to fix the showstopping errors.

test_datagen is of limited utility at the moment, but it can still be useful to provision a fresh database with some structured data. It did provision data for use with e2e testing over in learn, but unfortunately since that doesn't run anywhere, both the tests and this script have not been kept up to date w/ behavioral changes. These changes probably aren't perfect and make no attempt to verify and fix the e2e tests, but they unblock use for bootstrapping local devs.

### How can this be tested?
- On a fresh local, run `test_datagen`. You should observe that after the queued task runs you have enrollment codes for the provisioned contract. You should be able to see it at `/staff-dashboard/discounts` if you've set up a user with admin access

### DIFF
--- a/courses/management/commands/create_courseware.py
+++ b/courses/management/commands/create_courseware.py
@@ -676,6 +676,7 @@ class Command(BaseCommand):
             "--language",
             "--lang",
             type=str,
+            default="en",
             help='Set the language for the course run. (Default "en".)',
         )
         parser.add_argument(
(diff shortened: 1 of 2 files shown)


asadali145 (commented):
LGTM!

asadali145 (approved):

PR #3554, opened 2026-05-06
Log if theres a misconfigured enrollment code given in attach API, return 404

dsubak:
Adds a log message for the misconfiguration encountered [here](https://mitodl.slack.com/archives/C03K5HYGPT9/p1778008858760089)

Essentially, the code existed, but it was not linked up such that Discount.b2b_contracts() returned an empty list. In this codepath, that's not really expected, so it dropped through after performing no attachment and returned a 200. 

With this change we emit an error to sentry and return a 404.

Code in question on production:
Screenshot 2026-05-05 at 5 18 39 PM

### How can this be tested?
- Add a Discount in the admin page. I used the following minimal settings.
Screenshot 2026-05-06 at 12 36 33 PM
- Visit `http://mitxonline.odl.local:9080/api/v0/b2b/attach/test/` and hit POST. 
- You should get back 404 with a descriptive message and in the web container's UI, we should see an error like the following:
Screenshot 2026-05-06 at 12 39 27 PM

### Additional Context
- There were a few calls to `code.b2b_contracts().values_list("id", flat=True)` repeated in helper functions. Since `b2b_contracts()` isn't cached, I figured I'd pull that logic into the caller to try and reduce duplicative DB calls. That said, a quick look at the codebase makes it seem like Discount.b2b_contracts is only used in this api end point and a single management command, so an alternative approach would be to just make it a cached_property

### DIFF
--- a/b2b/views/v0/__init__.py
+++ b/b2b/views/v0/__init__.py
@@ -248,6 +248,15 @@ class AttachContractApi(APIView):
     def _get_eligible_contracts(self, user, code, now):
         """Return contracts associated with the code that the user can join."""
         contract_ids = list(code.b2b_contracts().values_list("id", flat=True))
+        if not contract_ids:
+            # Log an error here. We found a code, but it isn't associated with any contracts, which is generally confusing for operators.
+            # We could also surface this to the user by checking in the caller for an empty return.
+            log.error(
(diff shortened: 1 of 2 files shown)


annagav (approved):
LGTM 👍

PR #3555, opened 2026-05-06
Cache the results of _get_filtered_runs

rhysyngsun:

N/A

Caches the return value of some possibly costly looping in the serializer. The return value would be the same every time it's called.

### How can this be tested?

Nothing functional should change.

### DIFF
--- a/courses/views/v2/__init__.py
+++ b/courses/views/v2/__init__.py
@@ -206,9 +206,9 @@ class ProgramViewSet(ReadableIdLookupMixin, viewsets.ReadOnlyModelViewSet):
             added_context["include_programs"] = True
 
         if qp.get("org_id"):
-            added_context["org_id"] = qp.get("org_id")
+            added_context["org_id"] = int(qp.get("org_id"))
         if qp.get("contract_id"):
-            added_context["contract_id"] = qp.get("contract_id")
+            added_context["contract_id"] = int(qp.get("contract_id"))
 
         return {**super().get_serializer_context(), **added_context}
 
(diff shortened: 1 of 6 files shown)


asadali145:
@rhysyngsun do you think copilot review is a valid one?

pyproject.toml
```diff
@@ -82,6 +82,7 @@ dependencies = [
     "mitol-django-olposthog>=2026.3.6,<2027",
     "django-removals>=1.1.6,<2",
     "django-prefetch>=1.2.3,<2",
+    "lru-method-cache>=0.2.0",
```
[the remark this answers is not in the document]
rhysyngsun: Fine as is

rhysyngsun:
@asadali145 all set

courses/views/v2/__init__.py
```diff
@@ -206,9 +206,9 @@ def get_serializer_context(self):
             added_context["include_programs"] = True
 
         if qp.get("org_id"):
-            added_context["org_id"] = qp.get("org_id")
+            added_context["org_id"] = int(qp.get("org_id"))
         if qp.get("contract_id"):
-            added_context["contract_id"] = qp.get("contract_id")
+            added_context["contract_id"] = int(qp.get("contract_id"))
```
asadali145: This is not the right place for these changes. This is ProgramViewSet. We need these changes in CourseViewSet below as CourseWithCourseRunsSerializer is used in CourseViewSet

asadali145 (changes requested):

courses/views/v2/__init__.py
rhysyngsun: fixed

asadali145 (approved):
LGTM

PR #3572, opened 2026-05-11
Skip certificate validation if page revision missing

cp-at-mit:
https://github.com/mitodl/hq/issues/11150

Return early from CourseRunCertificate.clean and ProgramCertificate.clean when certificate_page_revision is None to avoid validation errors during revocation flows that clear the revision. Added a unit test to ensure CourseRunCertificate.clean does not raise when certificate_page_revision is intentionally unset.

### How can this be tested?

1. Open a course run certificate in Django Admin
2. Clear the "Certificate page revision" field (or check a "revoke" checkbox if one exists)
3. Save the form
4. Verify it saves without the AttributeError

### DIFF
--- a/courses/models.py
+++ b/courses/models.py
@@ -1685,6 +1685,10 @@ class CourseRunCertificate(TimestampedModel, BaseCertificate):
     def clean(self):
         from cms.models import CertificatePage, CoursePage  # noqa: PLC0415
 
+        # Skip validation if certificate_page_revision is not set (e.g., when revoking)
+        if self.certificate_page_revision is None:
+            return
+
         certpage = CertificatePage.objects.filter(
             pk=int(self.certificate_page_revision.object_id),
         )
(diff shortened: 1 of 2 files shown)


courses/models.py
```diff
@@ -1778,6 +1782,10 @@ def __str__(self):  # noqa: DJ012
     def clean(self):
         from cms.models import CertificatePage, ProgramPage  # noqa: PLC0415
 
+        # Skip validation if certificate_page_revision is not set (e.g., when revoking)
```
annagav, line 1785: Do we want to allow certificate_page_revision not set only when revoking? How do we know if we are revoking or something else?

cp-at-mit:
@pdpinch can we remove the validation that enforces course run certificates must have a certificate page revision defined?

annagav (approved):
I think this should be fine, since certificate_page_revision is not required

PR #3575, opened 2026-05-12
Filter out certificates which have future values for their issue_date

dsubak:
Fixes https://github.com/mitodl/hq/issues/11235
Hooks into the ActiveCertificates manager to additionally filter out certificates which have an issue_date in the future. This does not touch admin pages, which use the unfiltered certificate model managers instead.

### How can this be tested?
Tests should continue passing.

Manual testing:
- Provision a program or courseruncertificate. You should be able to observe it both via the API endpoint (http://mitxonline.odl.local:9080/api/v2/program_certificates/<UUID>/) as well as the wagtail serve page (http://mitxonline.odl.local:9080/certificate/program/<UUID>/)
- In the admin, set the issue_date to a time in the future.
- Observe that those URLs now return 404s.

### Additional Context
- @annagav pointed out that we check CourseRunCertificates in a signal to see if we should issue a new ProgramCertificate - filtering out things in the future would affect that process, but luckily we already use the unfiltered `all_objects` manager in that context. Additionally, when checking course requirements we rely on raw joins between the Course and CourseRunCertificates, so we should bypass the manager in that context too.
Screenshot 2026-05-12 at 11 42 02 AM
Screenshot 2026-05-12 at 11 44 44 AM

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -1144,10 +1144,7 @@ def _has_earned_program_cert(user, program):
         elif node.is_program:
             # has earned certificate for the required sub-program
             return ProgramCertificate.all_objects.filter(
-                user=user,
-                program=node.required_program,
-                is_revoked=False,
-                is_active=True,
+                user=user, program=node.required_program, is_revoked=False
             ).exists()
         return False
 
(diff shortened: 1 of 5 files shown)


annagav (approved):
LGTM 👍

PR #3578, opened 2026-05-13
Set cart-add dealstage to checkout_abandoned

cp-at-mit:
https://github.com/mitodl/hq/issues/11228

Introduce CART_ADD_DEAL_STAGE constant and force the dealstage on cart-add deals to that value in _sync_cart_add_deal_with_hubspot. This ensures cart-add events are always tracked as the HubSpot "checkout_abandoned" stage. Added a unit test that mocks HubSpot client and verifies the created deal's dealstage matches the new constant.

### How can this be tested?

- Add any course to your cart and verify that a hubspot deal is created with the deal stage of "checkout abandoned"
- Pay for the course and verify that the same hubspot deal is updated so the deal stage is "processed".

### DIFF
--- a/hubspot_sync/api.py
+++ b/hubspot_sync/api.py
@@ -61,6 +61,9 @@ from users.models import (
 
 log = logging.getLogger(__name__)
 
+# HubSpot internal option value for the "Checkout Abandoned" deal stage.
+CART_ADD_DEAL_STAGE = "checkout_abandoned"
+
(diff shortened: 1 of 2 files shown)


dsubak (approved):
Code is straightforward, and I've verified that the it does add the expected value when you hit the right codepath. I will note that I wasn't able to test end to end since I don't have hubspot set up or an account, but the properties look correct in my (slightly hacked up) test!

PR #3594, opened 2026-05-19
Detect UAI programs; prefer purchased_object

cp-at-mit:
Fixes issue with handling UAI program deals being synced to the correct hubspot account

Extend is_uai_order to detect UAI programs as well as course runs. The function now prefers line.purchased_object (for fulfilled lines) and falls back to line.product.purchasable_object, skips lines without a purchasable object, and checks both courseware_id (course runs) and readable_id (programs). Tests updated: import is_uai_order and add parametrized tests covering detection from purchasable_object for course runs and programs, plus a test ensuring purchased_object is used when present.

### How can this be tested?
Add a UAI program to your cart and ensure that a hubspot deal is created.  Complete the purchase of the UAI program and ensure that the same deal is updated with a closed date.


courses/utils.py
```diff
@@ -203,19 +203,31 @@ def is_uai_program(program):
...
-            if hasattr(course_run, "courseware_id") and is_uai_course_run(course_run):
-                return True
+        purchasable_object = getattr(line, "purchased_object", None)
+        if not purchasable_object and hasattr(line.product, "purchasable_object"):
+            purchasable_object = line.product.purchasable_object
+
+        if not purchasable_object:
+            continue
+
+        if hasattr(purchasable_object, "courseware_id") and is_uai_course_run(
```
dsubak: I see that in the b2b/api.py file we have a [few helper functions which use isinstance checks](https://github.com/mitodl/mitxonline/blob/adcad84d63e97a1c7b057523278c896744563e55/b2b/api.py#L551) instead of hasattr checks over columns to detect which model we've got - should we prefer isinstance checks here as well?

dsubak (approved):
I'm fighting with my local to test this, but the PR seems safe enough - as long as you feel good about it, I'm gonna preemptively approve it!

PR #3597, opened 2026-05-21
Fix consideration for program certificate generation

jkachel:

Closes mitodl/hq#11403

Fixes an issue with program certificate generation - if the user is enrolled in a course run that's audit-only, but the course has runs that have a verified mode, the user wouldn't get a program certificate.

### How can this be tested?

Automated tests should pass and the test for this specifically should be updated to account for this case.

To test manually, create a program with courses and audit-only course runs. Ensure at least one course has a separate, verified run. Enroll the user in the program with a verified enrollment and enroll them in the audit-only runs. Enter passing grades for the course runs, then attempt to generate the program certificate. It should work - you should get a program certificate for the program and no course run certificates.

This should also work if some of the course run enrollments are verified enrollments too. The rule is that we should generate you a program certificate if you have a verified enrollment in the program, a passing grade in the course runs that are audit-only, and a certificate in the course runs that have verified modes. So, any permutation of this should work. The problematic case is when the course has some runs that are audit-only and some that have verified modes.

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -1115,18 +1115,25 @@ def _has_earned_program_cert(user, program):
         bool: True if a user has earned all the course certificates required
               for a given program else False
     """
-    program_course_ids = [course[0].id for course in program.courses]
+
+    user_courseruns = CourseRun.objects.filter(
+        enrollments__user=user,
+        enrollments__active=True,
+        enrollments__change_status__isnull=True,
+    ).filter(course__in=program.courses_qset)
 
     cert_courses = Course.objects.filter(
-        id__in=program_course_ids,
+        courseruns__in=user_courseruns,
         courseruns__courseruncertificates__user=user,
         courseruns__courseruncertificates__is_revoked=False,
     )
     grade_courses = Course.objects.filter(
-        id__in=program_course_ids,
+        courseruns__in=user_courseruns.exclude(
+            enrollment_modes__mode_slug=EDX_ENROLLMENT_VERIFIED_MODE
+        ),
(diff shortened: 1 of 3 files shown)


dsubak (approved):
I think this tracks!

With the setup in the PR description, on main we get no certificate issued.
```
In [2]: from django.contrib.auth import get_user_model
   ...: from courses.api import generate_program_certificate
   ...: from courses.models import *
   ...: User = get_user_model()
   ...: user = User.objects.get(email="admin@odl.local")
   ...: program = Program.objects.get(id=10)
   ...: generate_program_certificate(user, program)
Out[2]: (None, False)
```

Conversely, when on this branch we correctly generate a certificate having made no database changes.
```
In [3]: from django.contrib.auth import get_user_model
   ...: from courses.api import generate_program_certificate
   ...: from courses.models import *
   ...: User = get_user_model()
   ...: user = User.objects.get(email="admin@odl.local")
   ...: program = Program.objects.get(id=10)
   ...: generate_program_certificate(user, program)
2026-05-21T18:02:04.326113Z [info     ] Program certificate for [test-admin] in program [Audit-Only Test Program] is created. [courses.api]
2026-05-21T18:02:04.330649Z [info     ] Skipping HubSpot sync for B2B user test-admin (user_id=1) [hubspot_sync.task_helpers]
Out[3]: 
(<ProgramCertificate: ProgramCertificate for user=test-admin, program=program-v1:MITx+AuditTest (ca5a9969-6ae7-4d40-ad0d-5269585c8b3d)">,
 True)
```

PR #3605, opened 2026-05-22
Initial variant support

jkachel:

mitodl/hq#11401

Gets variant support built out for courses and runs.

Courses get a "possible variants" field that tells you what the course _should_ have. This is exposed in the `v2/courses` API.  A list of variant runs isn't in here; this is covered later.

Course runs get a couple new fields, `variant_length` and `variant_industry`, which are enums. The `v2/courses` API includes filtering for these like it does for language. (These fields are wrapped up in a mixin so that they can be put elsewhere and all this can be consistent.)

There's a new API in `v3/courses/variants/` that takes a list of course IDs, a contract ID, and the variant options. Assuming the user is in the contract, this will return the runs that match the request _or_ the default for each course.

### How can this be tested?

The migrations set up a default set of variant options. The default is `language = "en", industry = "", length = ""`. 

You can add more variants through the Course page in the Django Admin. There's a new inline that should work for this. You can also set the variants in the Course Runs in Django Admin too. They're at the top of the page because the fields come from a mixin now (including language; I will fix this but it's not fixed now). 

You will need courses in a B2B contract to test the variant stuff. This is technically B2B only other than the languages. The language fields that were there in the API are still there for that reason.

When you hit the `v2/courses` API, you should see a list of the possible variants that the course should support. You can then feed that into the `v3/courses/variants/` API and get back the appropriate course runs for a given contract. See below for notes/questions.

### Additional Context

API at this point should be stable.

I added the language validation stuff in for this - digging through the translation effort, it looks like we are tagging the runs using the language codes from openedx-translations. These are _technically_ ISO 639-1 two-letter codes plus a locale code (`es-419`, `de-DE`, `zh-HK`). There's some code in here that decodes those using a lookup table and using `pycountry` for the ISO code - we should probably do something similar for how we display these in the frontend (since AFAIK it's just a hard-coded LUT there?).

_It's really important that you fix up some of your course runs for this stuff to work._ The default variant set expects an "en" language course run at least within your given contract, so make sure you have one. Setting the languages on these via a migration or something is actually kind of non-trivial given the constraints we have on the course run model.

**Questions:** 
- The variants API returns the course runs that match _and_ the default runs. Should it just do one or the other?
- The variants API is an "and" for the variant options - if you want english, healthcare, and short, then you will only get runs that match that (plus the defaults). I am assuming this is what we want, since the dropdown will be specific combinations, but is there a reason we might want these to be an "or" instead?


courses/constants.py
```diff
@@ -56,3 +56,182 @@
...
+    (
+        "",
+        "Normal",
+    ),
+    (
+        "S",
+        "Short",
+    ),
+    (
+        "L",
```
asadali145: I am using F/Full in the course keys on the Open edX side.

courses/constants.py
```diff
@@ -56,3 +56,182 @@
...
+        "Long",
+    ),
+]
+
+# These are mostly ISO 639-1 codes, except sometimes they have a locale code
+# appended. The list is from openedx-translations.
+COURSE_VARIANT_LANGUAGE = [
```
asadali145, line 92: Language code and locale looks fine here. FYI that we have less languages in the course translations plugin https://github.com/mitodl/open-edx-plugins/blob/main/src/ol_openedx_course_translations/ol_openedx_course_translations/settings/common.py#L77

variants/models.py
```diff
@@ -0,0 +1,144 @@
...
+    """Return a Q object of objects that support variants."""
+    return models.Q(app_label="courses", model="course") | models.Q(
+        app_label="b2b", model="contractpage"
+    )
+
+
+class VariantOptionsModel(models.Model):
+    """Mixin for variant option types, so they're the same wherever they're needed."""
+
+    language = models.CharField(
```
asadali145, line 34: Can we use language choices?

variants/models.py
```diff
@@ -0,0 +1,144 @@
...
+        max_length=8,
+        blank=True,
+        default="",
```
asadali145: Should this default to `en`?

courses/views/v2/__init__.py
```diff
@@ -427,6 +430,19 @@ def get_queryset(self):
...
 
+        # Prefetch the variant sets but filter B2B according to the query params.
+        qp = self.request.query_params
+        variant_b2b_enabled = "org_id" in qp or "contract_id" in qp
+
+        queryset = queryset.prefetch_related(
+            Prefetch(
+                "possible_variant_sets",
+                queryset=SupportedVariant.objects.filter(
+                    Q(b2b_only=variant_b2b_enabled) | Q(default_variant=True)
```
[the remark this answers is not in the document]
jkachel: fixed with a constraint, default_variant=True and b2b_only=True isn't a valid state

courses/views/v3/__init__.py
```diff
@@ -359,3 +364,143 @@ def get_course_outline(request, course_id):  # noqa: ARG001
...
+        content_type=course_ct, object_id__in=course_ids, default_variant=True
+    ).all()
+
+    # construct filter
+    variant_filter = (
+        Q(language=language) & Q(variant_length=length) & Q(variant_industry=industry)
+    )
+
+    output = []
+
```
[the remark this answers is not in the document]
jkachel: empty values are allowed for the variant option fields and have meaning, so this is desired behavior

courses/management/commands/import_courserun.py
```diff
@@ -309,20 +336,30 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument  # noqa: C
...
+            try:
                 run.save()
+            except ValidationError:
+                self.stderr.write(
+                    self.style.ERROR(
+                        f"ERROR: Language code {run.language} specified is invalid, cannot set."
+                    )
+                )
 
             success_count += 1
```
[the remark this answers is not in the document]
jkachel: it still makes the run

courses/views/v3/__init__.py
```diff
@@ -359,3 +364,143 @@ def get_course_outline(request, course_id):  # noqa: ARG001
...
+    for course in courses:
+        default_variant = [
+            variant for variant in default_variants if variant.object_id == course.id
+        ]
+        runs_qs = get_enrollable_courseruns_qs(valid_courses=[course]).filter(
```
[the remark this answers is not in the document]
jkachel: This is on purpose. I am calling `get_enrollable_courseruns_qs` explicitly for this reason - if the run isn't enrollable it doesn't belong in this output.

courses/management/commands/import_courserun.py
```diff
@@ -309,20 +341,51 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument  # noqa: C
...
+                log.exception(
+                    "import_courserun: ValidationError reached importing %s",
+                    run.courseware_id,
+                )
+                self.stderr.write(
+                    self.style.ERROR(
+                        f"ERROR: Could not set options on {run.courseware_id}, so variant options will be blank."
+                    )
+                )
+                continue
```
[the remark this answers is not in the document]
jkachel: I am ok with that

courses/management/commands/import_courserun.py
```diff
@@ -309,20 +341,51 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument  # noqa: C
...
 
+            try:
+                run.clean_language()
+            except ValidationError:
+                self.stderr.write(
+                    self.style.ERROR(
+                        f"ERROR: Language code {run.language} specified is invalid, cannot set."
+                    )
+                )
+                continue
```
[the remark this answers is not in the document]
jkachel: the command notes this in the output
the run is _already there_ at that point because the import call has already created it
it's noted so you can fix it manually later

asadali145 (commented):
Just some of the stuff around variants and language codes.

courses/models.py
```diff
@@ -1381,6 +1401,8 @@ class Meta:
                     "language",
                     "is_source_run",
                     "b2b_contract",
+                    "variant_length",
+                    "variant_industry",
                 ],
                 condition=~Q(language=""),
                 nulls_distinct=False,
```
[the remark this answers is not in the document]
jkachel: those workflows will need to be updated, language should now always be set, the field def doesn't change because we have to backfill the field values before making the field not-null. we should never have >1 `en` run even if there's no language actually set because that also means we're re-using a run tag, which is also not allowed.

courses/views/v2/__init__.py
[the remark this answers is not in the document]
jkachel: yup, it sure does

if we're in a B2B context, the only variants that apply are B2B-only ones and the default. that's it. works as expected.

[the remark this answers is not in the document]
gumaerc: Wild that it would miss this because it's so obvious. The filter is literally called `b2b_only`... In my experience AI agents seem to interpret the courses API as a way for users to search and add courses to their contract. It doesn't seem to be able to grok that the contracts are built through management commands and that the contract runs are associated before people are really hitting this API endpoint in that context.

variants/models.py
jkachel: Changed the default but we (I) have some work to do to backfill - need to be a bit careful to ensure the constraints don't get tripped up. (Bringing this up to make a note of it - really this should have been defaulted to `en` from the start, not sure why I didn't.)

courses/views/v3/__init__.py
```diff
@@ -359,3 +365,146 @@ def get_course_outline(request, course_id):  # noqa: ARG001
...
+            runs_qs = runs_qs.filter(variant_filter)
+
+        output.append(
+            {
+                "id": course.id,
+                "courseruns": runs_qs.all(),
+            }
+        )
+
+    return Response(CourseVariantRunsResponseSerializer(output, many=True).data)
```
gumaerc, line 510: There seems to be an issue with this. In the resulting API client, we get:

`courseVariantRunsV3(requestParameters: CoursesApiCourseVariantRunsV3Request, options?: RawAxiosRequestConfig): Promise<import("axios").AxiosResponse<CourseVariantRunsResponse, any, {}>>;`

Notice how the return type is `CourseVariantRunsResponse`. The serializer is for some reason not telling OpenAPI that it should return an array.

jkachel: should be fixed in https://github.com/mitodl/mitxonline/pull/3605/commits/f389395e6194b817c459b67006ba4204288bcafc

gumaerc: Confirmed fixed.

courses/management/commands/check_courseware_variant.py
```diff
@@ -0,0 +1,117 @@
+"""
+Check course variants for validity.
```
annagav: Could you please add here a bit more information about what this is for for a person who would be new to this codebase?
Maybe also an example line would be nice with possible options.

courses/views/v3/__init__.py
```diff
@@ -359,3 +365,146 @@ def get_course_outline(request, course_id):  # noqa: ARG001
...
+            {
+                "detail": "No courses found.",
+                "course_ids": course_ids,
+            },
+            status=status.HTTP_404_NOT_FOUND,
+        )
+
+    default_variants = SupportedVariant.objects.filter(
+        content_type=course_ct, object_id__in=course_ids, default_variant=True
+    ).all()
```
annagav: .all() is not necessary here.

annagav (commented):
Just few minor comments. Going to do functional testing now.

annagav:
So far my understanding is that the definition of variant sets are somewhat unrelated to the course runs definitions of variants. So I can define course runs with variants that do not exist on the Course it would still show up in the api. Is that expected?

annagav (approved):
Just left few minor comments. Functionality LGMT 👍

jkachel:
> So far my understanding is that the definition of variant sets are somewhat unrelated to the course runs definitions of variants. So I can define course runs with variants that do not exist on the Course it would still show up in the api. Is that expected?

So you can have variant runs that don't match the "supported set".. they'll still show up in the general APIs (i.e. `v2/courses/`) but the `v3/courses/variant_runs` API shouldn't show them (because it should be looking for ones that match specifically). Will test that out - not sure the algorithm I used for generating the filter is all that good.

jkachel:
FWIW - I tested the variant API; you can pull the variants that aren't officially supported (the API doesn't validate what you've asked for) but you have to know they exist. Otherwise, you don't see the stray variants in the output.

PR #3608, opened 2026-05-26
fix: handle missing signature image in CertificatePage signatory_items

rachellougee:

https://github.com/mitodl/hq/issues/11458#issuecomment-4544706989

Sentry error https://mit-office-of-digital-learning.sentry.io/issues/7292013765/?query=is%3Aunresolved&referrer=issue-stream
```
Got AttributeError when attempting to get a value for field `signatory_items` on serializer `CertificatePageModelSerializer`.
The serializer field might be named incorrectly and not match any attribute or key on the `CertificatePage` instance.
Original exception text was: 'NoneType' object has no attribute 'file'.
```

Fixes a 500 error when a signatory on a Certificate Page has no signature image. signatory_items now returns null for signature_image instead of raising an AttributeError in API `/api/v2/program_certificates/{UUID}/`

### How can this be tested?

Automatic test should pass

Manual test: In the CMS, create a Certificate Page with at least one signatory that has no signature image set. Then call GET `/api/v2/program_certificates/{UUID}/`  — it should return 200 with "signature_image": null instead of a 500 error.

```
---e.g. api/v2/program_certificates/{UUID}/`

   "signatory_items": [
      {
        "name": "XXX",
        "title_1": null,
        "title_2": null,
        "title_3": null,
        "organization": null,
        "signature_image": null
      }
```

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -450,7 +450,9 @@ class CertificatePage(CourseProgramChildPage):
                 "title_2": page.title_2,
                 "title_3": page.title_3,
                 "organization": page.organization,
-                "signature_image": page.signature_image.file.url,
+                "signature_image": page.signature_image.file.url
+                if page.signature_image
+                else None,
             }
             for page in self.signatory_pages
         ]
--- a/cms/models_test.py
+++ b/cms/models_test.py
@@ -619,6 +619,23 @@ def test_certificate_for_program_page():
         assert signatory.value.signature_image.title == "Image"
 
 
+def test_signatory_items_with_no_signature_image():
+    """
+    signatory_items should return None for signature_image when none is set,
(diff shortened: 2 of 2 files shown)


rachellougee:
I've fixed the missing signatory image in https://mitxonline.mit.edu/cms/pages/420/edit/ but this PR is still good to fix

pdpinch:
What happens with this PR if the image is missing?

rachellougee:
> What happens with this PR if the image is missing?

If the image is missing, this PR makes the /api/v2/program_certificates/{UUID}/ API return `200` instead of `500` error. Certificate URL will be accessible on Learn instead of a broken link.

annagav (approved):
👍

PR #3609, opened 2026-05-27
Add variant support to contracts

jkachel:

mitodl/hq#11283
mitodl/hq#11401

Adds support for variant options to contracts.

Like `Course`, you can assign `SupportedVariant` to a `ContractPage`. In this instance the B2B flag is ignored (it's a contract, it's always B2B). `ContractPage` should also have a default set and there's new methods in the model to pull out the courses and runs that apply to the configured set of variants.

A `check_contract_variant` command lets you check (and fix the default) variants for a given contract, similar to `check_courseware_variant` (mostly the same command actually).

The `b2b_courseware` command adds additional filtering options and also checks the configured contract variant sets when adding courseware and creating new runs. Variants will have a run tag that represents the variant options that the run is for. _TBD: haven't done this part yet_

### How can this be tested?

You need the setup from 11401 going already since this depends on courses having variants. You should have source runs for some variant sets.

Set up a contract and add some variants to it. Then add a course or program that contains a course that supports those variants and has appropriate source runs. It should create an appropriate number of course runs for the contract. If it can't find a matching variant run, then it should complain at you.

You can also use the filters to do something else, probably add a variant that's not in the contract. _TBD: haven't done this either_

### Additional Context

this will probably change a bunch since this is a branch off the branch for 11401


jkachel:
Superseded by https://github.com/mitodl/mitxonline/pull/3618

PR #3617, opened 2026-05-28
Remove django-hijack integration

cp-at-mit:
Drop django-hijack usage across the project: remove package from pyproject, delete hijack notification template and its include in base.html, remove hijack URL route and admin mixin usage, and remove hijack-related settings, middleware and context variable. Also adjust cms test to no longer expect hijack logout redirect key, and remove the session-based hijack_history 403 guard in ProfileDetailsAPIView to reflect the removed hijack behavior.

https://github.com/mitodl/hq/issues/10910


jkachel (approved):
LGTM 👍 

The test failure is something that got fixed in #3618 which merged late yesterday so shouldn't be an issue once this merges.

PR #3618, opened 2026-05-28
Add variant support to contracts

jkachel:

mitodl/hq#11283
mitodl/hq#11401

Adds support for variant options to contracts.

Like `Course`, you can assign `SupportedVariant` to a `ContractPage`. In this instance the B2B flag is ignored (it's a contract, it's always B2B). `ContractPage` should also have a default set and there's new methods in the model to pull out the courses and runs that apply to the configured set of variants. This includes a migration that backfills the default variants for current contracts.

A `check_contract_variant` command lets you check (and fix the default) variants for a given contract, similar to `check_courseware_variant` (mostly the same command, actually).

The `b2b_courseware` command adds additional filtering options and also checks the configured contract variant sets when adding courseware and creating new runs.  It additionally adds filtering options for variant sets, so only certain variants can be built out at a time if necessary.

Contract variants can be managed through the Django Admin for contracts. This is the only writable thing in the Django Admin for contracts - this will get moved into Wagtail later but skipped over that for now to get this ready to go.

### How can this be tested?

Note that having the test course runs exist in edX is not necessary for this. (Though that will mean the attempt to create the runs in edX will fail.)

Ensure you have some courses set up with variant options:
- Create a course and make sure it has the usual default variant options in place. (Either use `check_courseware_variant --fix-default` or just add one that has language = `en` and the default flag checked.)
- Make sure there's a matching source run for the default variant. If you're using an existing course, whichever run is there is probably fine; just make sure it has the language set to `en` and the primary flag checked.
- Create additional variants for the course. Make sure the B2B-only flag is checked for these. (Ideally, create an overlapping set that doesn't have the B2B flag checked - i.e. have one that's just `fr` language, and one non-B2B that's `dk` and one B2B one that's `hi`,Short,Finance or something.) 
- Create source runs for these variants. 
- Optionally repeat these steps and create a program that includes courses with variant setups.

Then get a contract set up properly:
- Create a contract and ensure it has the default option set. Same process for a course, except the command is `check_contract_variant --fix-default` if you're using that.

Run the `check_contract_variant` and `check_courseware_variant` commands to make sure the courses show up as expected. (Skip the `--fix-default` flag this time.)

Add the courses/programs to the contract using the usual command: `b2b_courseware add <contract> <readable id>`.

You can also test filtering: for `b2b_courseware add`, specify `--variant <var spec>` as many times as desired. `<var spec>` is in the help but it is a comma-separated list of `language,industry code,length code`. You can omit parameters as long as they're in the correct order (`en,HC` is OK, `en,,S` is OK but `en,S` would not be).

Language filtering was added in https://github.com/mitodl/mitxonline/pull/3610 and that should continue to work, but it has been updated to respect variant options. 

There are new APIs to check as well:
[`v0/b2b/contracts/{slug}/all_variant_runs/`](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_contracts_all_variant_runs_list) - lists out the variant runs for the contract with filtering by course ID(s)
[`v0/b2b/contracts/{slug}/`](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_contracts_retrieve) - should now include the variant options in the list

Automated tests should pass - this has more of them than #3605 did and should fix some test issues that were caused in between that PR and this one.


annagav (approved):
Works for me 👍

PR #3621, opened 2026-05-29
B2B/variants: add some documentation and clean up some things

jkachel:

Closes mitodl/hq#11548

This is a grab bag of some minor fixes and cleanups that were necessary after getting the initial run of variant code pushed out:
- Adds some documentation
- Adds an active manager to the B2B `ContractPage` model, and changes some things over to use it (API endpoints, mostly)
- Changes the default length to "Full" - this is now the blank option that was "Normal" - and update any records that were marked as "F" to be "" instead
- Updates `check_courseware_variant` to display runs that are "hidden" (i.e., that don't match up with the configured variants) and to add a `--b2b` flag that checks B2B-only variants and only displays source runs
- Updates the CourseRun admin so that the fields are grouped more logically
- Updates the Course admin so that the course runs can be navigated to and fixes the supported variants inline to not have any extras (which was kind of confusing)
- Updates some viewsets to filter out inactive variant options in the v2/courses and v3/courses APIs

### How can this be tested?

Documentation: proofread

API changes: You should generally not see inactive contracts or inactive variant options now.

`check_courseware_variant`: Create a course with some B2B-only variants, and create some runs for it that don't match up with any configured variants, and then run the command. If `--b2b` is set, you should just see the B2B variants (and the default) and it should just give you source runs back; you should also see it list out runs that don't match up with the configured set for the course.

Admin changes: should just need to be looked at - Course Run in particular is a good bit different

### Additional Context

Automated tests should mostly pass _except for this one_: `courses/views/v2/views_test.py::test_filter_with_org_id_inactive_contract_excluded` 

I fixed that in #3626 so not fixing it here too. Also, the `active` flags were removed where we're no longer including inactive records; this is removing a field so the spec checker is unhappy about that.


b2b/views/v0/__init__.py
```diff
@@ -78,7 +78,7 @@ class ContractPageViewSet(viewsets.ReadOnlyModelViewSet):
 
     def get_queryset(self):
         """Filter to only return active contracts by default."""
-        return ContractPage.objects.filter(active=True).prefetch_related(
+        return ContractPage.active_objects.filter(active=True).prefetch_related(
```
annagav: do we still need the filter filter(active=True)?

annagav (approved):
LGTM 👍 Just a small comment below

jkachel:
Fixed the nit - the tests pass as expected (one failure that was fixed elsewhere, and OpenAPI complains as noted) so going ahead with merging.

PR #3623, opened 2026-06-01
Make SignatoryPage.signature_image required

cp-at-mit:
https://github.com/mitodl/hq/issues/11542

Enforce a mandatory signature image on SignatoryPage: change the field to null=False, blank=False and on_delete=PROTECT (replacing SET_NULL). Add a migration that deletes rows with NULL signature_image before altering the field. Update tests to reflect the new required field (rename test and assert a signature image is present/properly configured).

### How can this be tested?
Go to the CMS and attempt to create a new signatory page within uploading an image.  You should get a validation error when publishing.


cms/migrations/0063_alter_signatorypage_signature_image.py
```diff
@@ -0,0 +1,30 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("cms", "0062_alter_certificatepage_signatories"),
+        ("wagtailimages", "0027_image_description"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql="DELETE FROM cms_signatorypage WHERE signature_image_id IS NULL;",
+            reverse_sql=migrations.RunSQL.noop,
```
dsubak: We're okay deleting these records? Do we have any other cases where the signatory is used but isn't broken by not having an image? 

As far as I can tell the only actual models that directly reference a SignatoryPage are the CertificatePage and SignatoryObjectIndexPage - if we delete a SignatoryPage referenced by the CertificatePage, is that going to make anything worse?

dsubak (approved):
Image
Definitely prevents me from adding a page w/o an image.

I had one open question, but as long as we know what'll happen in that case and are okay with it, I think this is good to go!

cms/migrations/0063_alter_signatorypage_signature_image.py
cp-at-mit: Added a placeholder image.  Just a single pixel.

dsubak (approved):

PR #3625, opened 2026-06-01
Model changes for discount code pre-assignment

dsubak:

First step in supporting write actions and preassignment for allocating contract codes to emails.

The bulk of this PR is model changes for DiscountContractAttachmentRedemption. This is so that we can create records ahead of time, potentially for learners who have not yet created users. 
A discount can be in one of three statuses from the /codes API going forward: "unassigned" if there is no DiscountContractAttachmentRedemption record, "assigned" if there is a record, but it is not linked to a user and redeemed_on is null, and "redeemed" if it is linked to a user and has a redeemed_on value.

Specific technical changes
  - Makes the foreign key to user nullable.
  - Adds assigned_email, assigned_name, assigned_by, redeemed_on, last_reminder_sent_on.
  - Add a migration which sets `assigned_email`, `assigned_name`, `redeemed_on` values for existing rows.
  - Accessor to derive a status based on whether or not the `redeemed_on` field is set. 
  - Changes to serializer to remove the boolean `is_redeemed` in favor of `redemption_status` derived field, as well as adding new fields.
  - Changes to serializer to pull values directly from DiscountContractAttachmentRedemption record when appropriate.
  - Preliminary scaffolding for the write endpoints - `assign`, `revoke`, `remind` and `bulk_assign`. All of these are subject to complete revision and currently return a 405.

### How can this be tested?
- Set up a contract and organization and add your user to the organization as a manager (by changing your `UserOrganization` record to have `is_manager=True`. Make sure the contract has a max_learners value.
- Provision discount codes for the contract using `ensure_enrollment_codes_exist` 
  - Redeem at least one code (this can be done by visiting /enrollmentcode if you have Learn connected - otherwise create one DiscountContractAttachmentRedemption linked to a proper user with redeemed_on set.)
  - Assign at least one code by creating a DiscountContractAttachmentRedemption record that has an assigned name and email
- Visit `http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/`
  - You should observe that the response has row for the redeemed code, one for the assigned code and the rest should be unassigned.

Screenshot 2026-06-03 at 10 57 10 AM

TODO:
- [x] Figure out if we need to worry about the migration being reversible.
- [x] Figure out how to manage the breaking changes to the API. I assume I need to just suppress it since nothing is using the tool atm?
- [x] Resolve https://github.com/mitodl/mitxonline/pull/3625#discussion_r3351951440, this may necessitate changes to attach other than what is currently in build
- [x] Tests for new functionality
- [x] Make sure /enrollementcode still works as expected


b2b/serializers/v0/manager.py
```diff
@@ -93,40 +93,63 @@ class ManagerEnrollmentCodeSerializer(serializers.ModelSerializer):
     """Serializer for enrollment codes available to a contract."""
 
     code = serializers.CharField(source="discount_code")
-    is_redeemed = serializers.SerializerMethodField()
-    redeemed_by = serializers.SerializerMethodField()
+    redemption_status = serializers.SerializerMethodField()
+    assigned_to = serializers.SerializerMethodField()
```
dsubak, line 97: Need to decide if it makes sense to keep `redeemed_by` and `is_redeemed`. `redeemed_by` is semantically replaced by `assigned_to`, while `is_redeemed` is now more accurately represented by `redemption_status`

b2b/views/v0/manager.py
```diff
@@ -226,30 +226,13 @@ def codes(self, request, **kwargs):  # noqa: ARG002
...
-                )
-            )
-            .annotate(
-                last_redemption_date=Subquery(
-                    DiscountContractAttachmentRedemption.objects.select_related("user")
-                    .filter(discount=OuterRef("pk"))
-                    .order_by("-created_on")
-                    .values("created_on")[:1]
-                )
+        discounts = contract.get_discounts().prefetch_related(
```
dsubak, line 240: I felt that this approach was somewhat clearer than slapping together a big CASE statement for the status annotations, but if folks disagree, I can keep it more in line w/ the previous approach.

b2b/migrations/0024_discountcontractattachmentredemption_assigned_by_and_more.py
```diff
@@ -0,0 +1,94 @@
...
+            name="redeemed_on",
+            field=models.DateTimeField(
+                blank=True,
+                help_text="When the code was claimed. None indicates that it has been assigned, but not yet claimed.",
+                null=True,
+            ),
+        ),
+        migrations.RunPython(
+            code=backfill_redemption_fields,
+            reverse_code=migrations.RunPython.noop,
```
dsubak, line 80: This isn't meaningfully reversible - we'd need to either remove records which have null values for the user or associate it with some other user.

b2b/views/v0/manager.py
jkachel: This is a "do it in the frontend" versus "do it in the database" issue I think. Either is fine but I will note that if we have a `managed` contract that has no seat limit, we could have literally thousands of redemptions on an enrollment code, so we maybe want to add a limit to the prefetch? (FWIW we do have some code-managed contracts out in prod with no seat limits but at least the couple I clicked through were also pre-pilot ones that don't count anymore.. but also thinking of the HHC org, which is turning into a bunch of edge cases.)

b2b/serializers/v0/manager.py
jkachel: `is_redeemed` seems like an reasonable thing to get rid of. Status is increasing in complexity so it'll be better to have the one field that exposes that complexity.

`redeemed_by` is a bit more thorny. We have to make some changes to the attach API logic as part of this. Part of that will be deciding whether or not we want to allow users that have a code, but aren't apparently assigned to said code, to be able to redeem it. From a customer service/UX point of view, the answer to that is "yes we should", because people may be assigned a code and then opt to use a different email address. In this case it'd be good to see both, so the admin can be aware of this. (The org did pay for seats - compliance checking is part of the goal for this, I think.) The other path is "no" and that ensures that only the assigned apparent user can use the code, but if the user wants to use a different email address, they've gotta work that out with someone to update the assignment (which probably ends in a support ticket for us).

b2b/models.py
```diff
@@ -711,6 +717,38 @@ class DiscountContractAttachmentRedemption(TimestampedModel):
         related_name="code_redemptions",
     )
 
+    assigned_email = models.EmailField(  # noqa: DJ001
```
jkachel: Nit - for both of these, any reason why `null=True` and not `default=""` ? The latter means we only ever have one kind of empty state for these fields (just empty string).

jkachel (changes requested):
Did just a read-through for now - looks pretty good so far. Some comments below and one nit.

b2b/serializers/v0/manager.py
dsubak: Hmmm, I kinda feel that there's a security risk by allowing users with mismatched emails to use codes. That would mean that anyone with access to read an invite email would be able to go redeem the code - great if I get my code on `dansubak@mit.edu` and want to use it on `daniel.subak@gmail.com`, not as good if I leave my laptop open and someone copies the link and redeems it without my knowledge (or worse, has compromised my email).

If I'm assessing this correctly, what is our general stance on access controls for these codes? I'm typically on the side of stricter access controls, but if this is an acceptable risk and we do explicitly want to enable folks to use it across emails, we can do that as long as everyone's on the same page.

b2b/views/v0/manager.py
dsubak: Agreed - I thought the query was starting to look a bit difficult to parse which is why I went this way. Let me see about getting a prefetch limit on there and as long as nobody feels very strongly, I'll keep it as is!

b2b/models.py
dsubak: No good rationale, just copied it from somewhere and it didn't matter much for the purposes of this PR. Happy to make that change!

b2b/views/v0/manager.py
dsubak: Adding a limit slice to the prefetch queryset does seem to work in practice.

The query it generates is truly something to behold (it uses window functions and looks gnarly) but if it causes issues, we can always come back to it. Frankly, jamming it into prefetch isn't gonna save us much vs just doing a different query and wiring em up in Python if necessary, so I'm happy to take the bet that this'll be fine for now. Resolving!

b2b/serializers/v0/manager.py
dsubak: @pdpinch or @skurchak - James and I spoke a bit and it sounds like this might be a product behavior question. Do we want to enforce that only users who have emails matching the one assigned to them can redeem a code or do we want to allow any user to use an assigned code?

dsubak:
~We do still need to chase https://github.com/mitodl/mitxonline/pull/3625#discussion_r3351951440 to ground, but I think this is safe to review and release in its current form, so I've pulled it out of draft for now!~ 

This is done, we've readded the redeemed_by serializer field

b2b/serializers/v0/manager.py
pdpinch: No enforcement. We have experience from running xPRO that a significant number of users prefer to use a different email address than the one that was invited into the contract. 

For example, if you already have an account from when you were in school, and then your current employer invites you. You would prefer to continue using the existing account that already has access to your courses and grades. 

I know it was on the bubble, but we had considered a UI feature to indicate when an invitation was redeemed by a different email address.

dsubak: Very well, we can do that. As long as we have considered it and think the use case outweighs the risk of unauthorized access, easy enough to implement. 

I'll make the necessary changes and write a test over this behavior

dsubak: I've readded `redeemed_by` to the serializer - it continues to reflect the value of the attached `user.email`, while `assigned_to` reflects the new model field value. 

I've written a (fairly useless) test for a non-matching case, but once we have the assign endpoint implemented in a subsequent PR, I'll make it a full integration tests of the assign/attach/codes workflow to make sure everything's linked up.

Resolving this for now!

jkachel (approved):
LGTM :+1:

PR #3626, opened 2026-06-01
Move where variant options are set during import_courserun

jkachel:

N/A

This was in #3621 but is now separated out so it can be reviewed more quickly. 

`import_courserun` will try to set language, industry, and length options on the imported run. However, it was doing this after the run was imported, and the underlying `import_courserun_from_edx` call just sets these fields to the default which causes a collision when importing the variants. So, instead, the variant options are now set in `import_courserun_from_edx`.

### How can this be tested?

Testing requires a working edX installation. This doesn't affect B2B contracts, so those don't need to be set up.

_Example starting course run: `course-v1:UAI+UAI.99+1T2026`_

In edX Studio:

1. Create a new course run. Using the example run, the org would be `UAI`, the course number would be `UAI.99`, and the run tag would be `1T2026`.
2. Re-run the course run. Change the run tag to indicate that it's a variant. For example, this would indicate an English/healthcare/short variant: `1T2026_en_HC_S`.

In MITx Online:

1. Import the initial run: `import_courserun --courserun course-v1:UAI+UAI.99+1T2026 --source-course --lang en --primary-lang -d Testing --create-depts`
2. Ensure that there's a default variant set for the course: `check_courseware_variant course-v1:UAI+UAI.99 --fix-default`.
3. Create a new variant for the course in the Django Admin that matches the settings of the re-run. (Using the example, this means adding a variant option for English, Healthcare, and Short.)
4. Import the variant run: `import_courserun --courserun course-v1:UAI+UAI.99+1T2026_en_HC_S --source-course --lang en --industry HC --length S --use-specific-course course-v1:UAI+UAI.99 --run-tag 1T2026`

This should work successfully. You should see the variant run in the Course Runs listed in the Course in Django Admin, and you should see the option if you run `check_courseware_variant course-v1:UAI+UAI.99`. 

### Additional Context

This came up when trying to import a run that matched the language, so the example also has a matching language. However, this would likely have caused an issue regardless because the CourseRun object was being created with defaults, and the default for the language field changed to `en` (from empty string).

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -1503,6 +1503,10 @@ def import_courserun_from_edx(  # noqa: C901, PLR0913
     - include_in_learn_catalog (bool): Set the "include_in_learn_catalog" flag on the new page.
     - ingest_content_files_for_ai (bool): Set the "ingest_content_files_for_ai" flag on the new page.
     - is_source_run (bool): Set the "is_source_run" flag on the course run to designate it as a B2B source course.
+    - language (str): The language to set the imported run to. Defaults to "en".
+    - is_primary_language (bool): Whether or not this is the primary run for this language. Defaults False.
+    - variant_length (str): One of the length codes. Defaults to empty string (Full).
+    - variant_industry (str): One of the industry focus codes. Defaults to empty string (Original).
     Returns:
     tuple of (CourseRun, CoursePage|None, Product|None) - relevant objects for the imported run
     """
--- a/courses/api.py
+++ b/courses/api.py
@@ -1457,6 +1457,10 @@ def import_courserun_from_edx(  # noqa: C901, PLR0913
     include_in_learn_catalog: bool = False,
     ingest_content_files_for_ai: bool = False,
     is_source_run: bool = False,
+    language: str = "en",
+    is_primary_language: bool = False,
+    variant_length: str = "",
+    variant_industry: str = "",
 ):
     """
     Import a course run from edX.
@@ -1558,6 +1562,10 @@ def import_courserun_from_edx(  # noqa: C901, PLR0913
         live=live,
         is_self_paced=edx_course_run.is_self_paced(),
         is_source_run=is_source_run,
+        language=language,
+        is_primary_language=is_primary_language,
+        variant_industry=variant_industry,
+        variant_length=variant_length,
     )
 
     course_page = None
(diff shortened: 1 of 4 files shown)


courses/api.py
```diff
@@ -1457,6 +1457,10 @@ def import_courserun_from_edx(  # noqa: C901, PLR0913
     include_in_learn_catalog: bool = False,
     ingest_content_files_for_ai: bool = False,
     is_source_run: bool = False,
+    language: str = "en",
```
dsubak, line 1460: Nit - could we add the args to the docstring?

courses/management/commands/import_courserun.py
```diff
@@ -319,22 +319,51 @@ def handle(self, *args, **kwargs):  # pylint: disable=unused-argument  # noqa: C
...
+                    "import_courserun couldn't import %s - validation error",
+                    edx_course.course_id,
+                )
+                self.stderr.write(
+                    self.style.ERROR(
+                        f"Validation error importing {edx_course.course_id}."
+                    )
+                )
+                return f"Validation error importing {edx_course.course_id}."
+            except LookupError:
```
dsubak, line 356: For my edification, where do we expect a LookupError to be thrown from? 

With a little digging I see that the `VariantOptionsModel` catches and rethrows it as a ValidationError and that we've overridden the save method to perform Django model validation before saving - wanted to make sure I understood the language/variant specific flow that leads to this block being necessary.

dsubak (approved):
This seems safe enough at first glance, but I will note that I have not tested it myself. I don't currently have a working setup of edx.

I'll spend some time later today to try and get this tested, but as I know there's some urgency behind this I've gone ahead and approved it based on review alone.

courses/management/commands/import_courserun.py
jkachel: The `LookupError` is thrown by pycountry. That specific type is being caught in the `VariantOptionsModel` mixin but catching it here in case the upstream code changes.

courses/api.py
jkachel: fixed in https://github.com/mitodl/mitxonline/pull/3626/commits/6b470ae6902ff1e602c4beeb8bc337801e66b0b9

cp-at-mit (approved):

PR #3627, opened 2026-06-02
add command to trasfer grades to another account

annagav:
N/A

Add a command to seamlessly combine accounts for user who happened to have two accounts and they would like to merge.

### How can this be tested?
Created two users with grades, enrollments and certificates in both accounts


rachellougee (approved):
LGTM

PR #3629, opened 2026-06-02
Make min and max price decimal number

annagav:
Fix https://github.com/mitodl/hq/issues/11582

Make min and max price decimal number

### How can this be tested?
Go to the CMS and try to set the min and max price with decimal points

Go to http://mitxonline.odl.local:8013/api/v2/courses/ and verify the correct values for min and max price.


daniellefrappier18 (approved):
Look good to me I tried with and without the decimal and got the expected results 

Image

Image

PR #3630, opened 2026-06-02
Remove now-unused language options field from course API responses

jkachel:

n/a

Now that the variant options are in place, we don't need to list out the language options. We get the list of supported languages for the course (or contract) as part of the payload for the course and we have other APIs that can pull the variants (including languages) for a given course.

### How can this be tested?

OpenAPI spec check should fail since this removes fields. The Learn front end should no longer be using these fields, though.

Otherwise, test the affected APIs: create some courses with some language variants, and then load the course APIs. You shouldn't see the `language_options` field anymore (and it ought to be a little bit faster, since calculating that took some effort).


annagav (approved):
LGTM 👍

PR #3631, opened 2026-06-02
Force color terminal output under GH actions

rhysyngsun:

N/A

This forces pytest to render color output so that we can see failures in diffs easier.
 / at all
### How can this be tested?

See the github action test output in the first commit which failed, it should render the diff from `assert_drf_json_equal` w/ color plus all of the pytest output generally: https://github.com/mitodl/mitxonline/actions/runs/26830041042/job/79108417304


annagav (approved):
👍

PR #3632, opened 2026-06-02
Add ETL-specific version of the course data endpoint

jkachel:

Closes mitodl/hq#11530

Adds a copy of the `CourseViewSet` that performs less filtering, so that all the course runs appear for each course returned, even if they're for B2B contracts or are variant runs that we'd otherwise put elsewhere in the dataset.

This also adds an `is_etl` flag to the `User` model. This flags grants access to the above API endpoint. The ETL process on the Learn side should use an account that has this flag set.

### How can this be tested?

For best results, make sure you have some courses with variants set up (not source runs). Then, hit the API: `v3/courses/ingestible/`. You should see all the runs, including variants and B2B runs.

Some sample output is available in this comment: https://github.com/mitodl/mitxonline/pull/3632#issuecomment-4615789212 (Note that this was pulled using an earlier revision of this code where this was part of the `v2` urlspace - the `v3` version won't pass data at all unless you have the ETL flag set.)

### Additional Context

This should include _all_ runs, including B2B runs and variants, in the `courseruns` prop in each course. For the normal endpoint, we find the canonical run and then just return that; variants are pulled from a separate endpoint and B2B runs require filtering. This does make it important to ensure that users with the `is_etl` flag set are audited occasionally.

On the Learn side, the ETL process should use the OAuth2 client credentials flow to authenticate into Keycloak, and then it should be fine to access the API. Otherwise, the ETL process code shouldn't need to change; the new API uses the same serializers and as the `v2` endpoint.

Some data isn't populated. The ingestible API doesn't pre-fetch B2B contract data or the language options. Also, some options aren't available; the ingestible API doesn't support filtering, since it is intended to be used programmatically.


pdpinch:
> You should see all the runs that are enrollable

Why are we limiting it to enrollable runs? Is that necessary / a good idea?

jkachel:
@pdpinch didn't seem like ingesting stuff from courses that people can't use was worthwhile - can change that if that assumption is wrong. But the thought was to not continue to update and ingest content from courses that people can't enroll in - whatever was there when the course ended would be sufficient for anyone who can still access the course run.

@shanbady FYI - this is a before/after of the API results from `v2/courses/` as hit by a user with the `is_etl` flag off or on:

[11530-api-no-flag.json](https://github.com/user-attachments/files/28565641/11530-api-no-flag.json)
[11530-api-with-etl-flag.json](https://github.com/user-attachments/files/28565642/11530-api-with-etl-flag.json)

This course is mostly B2B/etc. stuff but you should be able to see variants in here - these have the `variant_length` and `variant_industry` flags set to something. (The translations count too but ones that are just `en` are the "normal" course.) This should also include B2B runs. 

(I am checking tests right now - the code in this branch as of writing this is completely out of date; just wanted to get these sample outputs up while pytest does things.)

jkachel:
https://github.com/mitodl/mitxonline/pull/3632/commits/d52123b5eadf2c81638c3e0fa4bf13b214236d40 is where the new code for this starts - tests and pre-commit fail because there's a migration conflict ([pending PR approval](https://github.com/mitodl/mitxonline/pull/3639)) and the ORM check baseline file doesn't match up to the source anymore. But otherwise this should be OK to test.

(However - the baseline was excluding this part of the code anyway so I will probably fix it.)

jkachel:
> (However - the baseline was excluding this part of the code anyway so I will probably fix it.)

Nathan suggested not doing this - so now have moved the changes I made to the `v3` space.

jkachel:
FYI - following a discussion in the mitxonline planning meeting, the ETL-specific endpoint will move to `/api/internal/courses`. This should make it clear that the API is only for internal use, and will pull it out of the OpenAPI spec. Nothing else about it will change; it'll just be in a different place in the codebase and mounted to a different part of the URL-space.

courses/serializers/internal/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+
+from rest_framework import serializers
+
+from courses.models import Course
+from courses.serializers.v2.courses import (
+    CourseRunSerializer,
+    CourseSerializer,
+)
+
+
```
mbertrand, line 12: The inherited `get_availability` method (`CourseSerializer` line 215) calls `instance.courseruns` — a manager queryset — rather than `instance.prefetched_courseruns`. With `to_attr` prefetch, the manager cache is not populated, so this issues one extra DB query per course in the response.

Maybe override `get_availability` in `IngestibleCourseWithCourseRunsSerializer` to use `prefetched_courseruns`, or update `CourseSerializer` to check for the attr first (like `get_courseruns` does here).

courses/views/internal/views_test.py
```diff
@@ -0,0 +1,176 @@
...
+
+    request = rf.get("/")
+    request.user = user
+
+    perm = IsEtlUser()
+
+    assert perm.has_permission(request, {}) == (is_etl or is_superuser)
+
+
+@pytest.mark.skip_nplusone_check
```
mbertrand, line 53: This might no longer be necessary if the n+1 queries are cleared up by the suggested change  mentioned earlier.

mbertrand (changes requested):
LGTM, except for some n+1 queries.  But since this is just an internal API maybe it's okay  to accept those if they are too stubborn to remove easily, let me know and I'll approve if so.

courses/serializers/internal/__init__.py
jkachel: Updated this but it will likely be obviated by the upstream changes to the serializer. (Long story short - `get_availability` calls other things that expect a QuerySet.)

courses/views/internal/views_test.py
jkachel: There were other parts of this that fail the check.

For what it's worth, there's [work happening upstream](https://github.com/mitodl/hq/issues/11517) to the `v2/courses` API that may resolve some of these issues; this API will get most of those fixes as they get rolled out so probably don't want to reworking stuff solely for this (or causing conflicts elsewhere by fixing it upstream).

courses/views/internal/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        dated_runs_prefetch = Prefetch(
+            "courseruns",
+            queryset=CourseRun.all_objects.enrollable().filter(is_self_paced=False),
+            to_attr="prefetched_dated_courseruns",
+        )
+        queryset = queryset.prefetch_related(
+            "departments",
+            "in_programs",
+            course_runs_prefetch,
+            dated_runs_prefetch,
```
mbertrand, line 74: Seems to need a programs prefetch too, getting this error now:

> RequiredPrefetchMissingError at /api/internal/courses/
Required prefetch is missing: 'programs'

jkachel:
Last commit should fix the remaining test failures. Not sure about the bumpver issue - I just manually updated it in pyproject but it seems both like that should be happening automatically and that fixing it is out of scope for this PR.

Tests are currently failing because it looks like the version of a dependency we're using got yanked. Maybe that should be addressed separately..

mbertrand (approved):
👍

PR #3637, opened 2026-06-03
Use fixed HubSpot certificate object types

cp-at-mit:
NA

Introduce constants for course_run_certificate and program_certificate object types and update sync functions to use them instead of reading object-type names from settings. Remove related environment settings and association ID config from main/settings.py. Simplify the create_hubspot_certificate_schema management command by removing the env-var output helper and printing a message that no additional certificate env settings are required. Update tests to stop setting the removed settings. This centralizes the custom object type names and reduces required configuration.

### How can this be tested?
Tests should still pass.  If you have hubspot configured, then creating a new certificate (program or course run) through Django Admin should result in a new cert object being created in Hubspot.


hubspot_sync/api.py
```diff
@@ -970,21 +973,13 @@ def _get_custom_object_type_id_by_name(
...
-    object_type_name = getattr(
-        settings, "HUBSPOT_COURSE_RUN_CERTIFICATE_OBJECT_TYPE", None
-    )
-    if not object_type_name:
-        log.debug(
-            "Skipping course run certificate HubSpot sync: "
-            "HUBSPOT_COURSE_RUN_CERTIFICATE_OBJECT_TYPE is not configured."
-        )
-        return None
+    object_type_name = COURSE_RUN_CERTIFICATE_OBJECT_TYPE
```
dsubak: Nit - now that this is always a constant, might make sense to just inline the constant in the three places it's used.

hubspot_sync/api.py
```diff
@@ -1054,21 +1049,13 @@ def sync_course_run_certificate_with_hubspot(cert) -> SimplePublicObject | None:
...
-    object_type_name = getattr(
-        settings, "HUBSPOT_PROGRAM_CERTIFICATE_OBJECT_TYPE", None
-    )
-    if not object_type_name:
-        log.debug(
-            "Skipping program certificate HubSpot sync: "
-            "HUBSPOT_PROGRAM_CERTIFICATE_OBJECT_TYPE is not configured."
-        )
-        return None
+    object_type_name = PROGRAM_CERTIFICATE_OBJECT_TYPE
```
dsubak: Nit - same comment w/r/t inlining the constant, if you take that suggestion

hubspot_sync/management/commands/create_hubspot_certificate_schema.py
```diff
@@ -136,8 +114,6 @@ def handle(self, *_args, **_options):
                     "  contact association typeId: (will need to be set manually or auto-created on first sync)"
                 )
 
-            env_output.extend(self._build_env_output(schema_name, assoc_type_id))
-
-        self.stdout.write("\nAdd/update these environment settings:\n")
-        for env_line in env_output:
-            self.stdout.write(env_line)
+        self.stdout.write(
```
dsubak: Think you could probably just remove the message entirely if you wanted. I think seeing this would just make me confused about what env vars _could_ be set - might raise more questions than anything else!

dsubak (approved):
Image
Looking good in sandbox!

PR #3646, opened 2026-06-05
Assign, revoke, remind and bulk_assign b2b enrollment code endpoints

dsubak:

Adds 4 new endpoints:
- `assign` takes an email and optional name and creates a `DiscountContractAttachmentRedemption` record if the specified code is unredeemed, sending an email to the address containing an `/enrollmentcode` link for the code.
- `revoke` deletes a `DiscountContractAttachmentRedemption` record if one exists for that code. This is basically only useful for people working in the dash at the moment, we may want to burn the code, but this behavior will require product input.
- `remind` takes an email and optional name and resends the email with the `/enrollmentcode` link for the specified code.
- `bulk_assign` takes a list of email/name pairs and attempts to perform the same operations as assign, as long as enough codes are available for distribution.

Email with link (probably needs design, this was mostly cribbed from existing email templates in codebase)
Screenshot 2026-06-10 at 10 11 22 AM

### How can this be tested?
First, set up a b2b contract with plenty of codes. Make sure that you're a manager on the org as well. 

If you want to check the email, optionally set up `MITOL_MAIL_CONNECTION_BACKEND` with an SMTP backend and ensure something like mailpit or mailcatcher is reachable from within your web container.

- Visit `http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/` to see a list of discount codes. If you haven't done anything, these will all be unassigned.
- Visit `http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/<one_of_the_codes_from_above>/assign`. Enter an email and optionally a name and hit POST. You should see that if you visit /codes again, that the corresponding record now reflects that the code is assigned to the name and email you specified.
- Visit `http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/<one_of_the_codes_from_above>/revoke`. Enter an the email and name from before and hit POST. You should see that if you visit /codes again, that the corresponding record now reflects that the code is once again `unassigned`.
- Visit `http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/bulk_assign`. Enter several names and emails and hit POST. You should see that if you visit /codes again, that the corresponding record now reflects that several codes are assigned to the users you've specified.

If you set up an SMTP backend, during each assignment operation you should've also gotten an email with an enrollmentcode Learn URL corresponding to the code. The exercise of setting up mailpit or equivalent and getting it networked correctly is left as an exercise to the user due to the differences in local environments, but you'll probably at minimum want to override the following values:
```
MITOL_MAIL_CONNECTION_BACKEND=django.core.mail.backends.smtp.EmailBackend
MITX_ONLINE_EMAIL_HOST=host.docker.internal
MITX_ONLINE_EMAIL_PORT=1025
```
Additionally, ensure `MAILGUN_RECIPIENT_OVERRIDE` is not present in your settings. This can filter the list of recipients in ways you might not expect, potentially resulting in no emails being sent at all.

### Open Questions:
- [x] Validate API structure with Danielle
- [ ] Get email design if possible.
- [ ] Get clarity on the behavior of revoke. Implement here or as a fast follow, depending on the complexity of the feature.

Note - neither the email design nor the revoke behavior needs to be considered blocking. The frontend is behind a feature flag at the moment and it should not be an issue to come through and finalize these later if it makes sense.


b2b/views/v0/manager.py
```diff
@@ -314,18 +417,70 @@ def assign_code(self, request, **kwargs):  # noqa: ARG002
             ),
         ],
     )
-    @action(detail=True, methods=["post"], url_path="codes/(?P<code>[^/.]+)/revoke")
-    def revoke_code(self, request, **kwargs):  # noqa: ARG002
+    @action(
+        detail=True,
+        methods=["post"],
```
daniellefrappier18: would it be semantically cleaner to use a "delete" instead of a post

daniellefrappier18 (commented):
**bulk_assign** - I think we should keep the approach with email and optional name because the frontend table is keyed by email, not code ID and also managers paste a list of people, not a list of opaque UUIDs. Also another thing to keep in mind is the frontend needs to handle the errors array in the bulk_assign response body and surface those to the user for things like "no available codes left" or "this email is already assigned" would show up there.

**revoke** - If there is only ever a single DiscountContractAttachmentRedemption for each code shown in the table then the API already assumes uniqueness by the code and adding the email would be redundant. Also I see for a potential edge case. Let say the email is revoked and reassigned but because of caching the frontend holding stale data, wouldn't the frontend sees 404 and shows"code not found" or silently fail? Probably best to drop this.

b2b/views/v0/manager.py
```diff
@@ -336,20 +491,65 @@ def revoke_code(self, request, **kwargs):  # noqa: ARG002
...
+        ).first()
+        if not assignment_record:
+            return Response(
+                {"detail": "Assignment for email does not exist"},
+                status=http_status.HTTP_404_NOT_FOUND,
+            )
+
+        # Just send the email reminder, no need to do anything else.
+        send_enrollment_code_assignment_email(assignment_record, code)
+
```
daniellefrappier18, line 529: I believe this is missing an update for assignment_record.last_reminder_sent_on

b2b/views/v0/manager.py
dsubak: Works for me, esp since we're dropping the request body.

b2b/views/v0/manager.py
```diff
@@ -314,18 +415,51 @@ def assign_code(self, request, **kwargs):  # noqa: ARG002
...
+            )
+
+        assignment_record = discount.contract_redemptions.first()
+        if not assignment_record:
+            return Response(
+                {"detail": "No assignment exists for this code."},
+                status=http_status.HTTP_404_NOT_FOUND,
+            )
+
+        assignment_record.delete()
```
daniellefrappier18, line 473: `assignment_record.delete()` runs with no check on assignment_record.status

We probably want to add a guard to this like:

```
if assignment_record.redeemed_on or assignment_record.user:
    return Response(
        {"detail": "Cannot revoke a code that has already been redeemed."},
        status=http_status.HTTP_409_CONFLICT,
    )

assignment_record.delete()
```

b2b/views/v0/manager.py
```diff
@@ -336,20 +470,67 @@ def revoke_code(self, request, **kwargs):  # noqa: ARG002
...
+        code = kwargs.get("code")
+
+        serializer = AssignRevokeCodeRequestSerializer(data=request.data)
+        if not serializer.is_valid():
+            return Response(
+                {"detail": "Invalid request data."},
+                status=http_status.HTTP_400_BAD_REQUEST,
+            )
+
+        email = serializer.validated_data["email"]
```
daniellefrappier18: is passing the email actually needed here?

b2b/serializers/v0/manager.py
```diff
@@ -89,6 +89,50 @@ class Meta:
...
+class DetailErrorSerializer(serializers.Serializer):
+    """Serializer for generic detail error responses (404, etc.)."""
+
+    detail = serializers.CharField()
+
+
+class BulkAssignResultSerializer(serializers.Serializer):
+    """Serializer for the bulk_assign response body."""
+
+    assigned = serializers.ListField(help_text="Successfully assigned codes.")
```
daniellefrappier18: assigned field is untyped. Every component that renders the assigned codes list has to cast or guess the shape. Suggested Fix: child=ManagerEnrollmentCodeSerializer().

b2b/serializers/v0/manager.py
```diff
@@ -89,6 +89,50 @@ class Meta:
...
+    errors = serializers.ListField(
```
daniellefrappier18: errors items are also untyped DictField() generates additionalProperties: {} in the spec. The frontend needs to know {email, name, detail} to render error rows. 
Suggested Fix: define a named BulkAssignErrorSerializer with those three fields and use it as child

b2b/views/v0/manager.py
```diff
@@ -36,6 +44,35 @@
...
+    name: str
+    code: str
+
+
+def assign_codes_and_send_emails(
+    assignments: list[CodeAssignment], assigning_user
+) -> None:
+    # Use bulk_create for this
+    for assignment in assignments:
+        redemption = DiscountContractAttachmentRedemption.objects.create(
```
daniellefrappier18: There is no transaction on bulk assign. If the operation fails mid-loop, the frontend gets a 500 with no way to know which codes were assigned and which weren't. The response contract breaks: you get neither a success with partial results nor a clean error.

Suggested fix:  wrap the entire loop in a transaction and separate the email sending from the DB writes:

```

def assign_codes_and_send_emails(assignments, assigning_user):
    with transaction.atomic():
        redemptions = []
        for assignment in assignments:
            redemption = DiscountContractAttachmentRedemption.objects.create(
                discount=assignment.discount,
                contract=assignment.contract,
                assigned_email=assignment.email,
                assigned_name=assignment.name,
                assigned_by=assigning_user,
                last_reminder_sent_on=now_in_utc(),
            )
            assignment.discount.prefetched_redemptions = [redemption]
            redemptions.append((assignment, redemption))

    # Emails go outside the transaction — a failed email shouldn't roll back DB writes
    for assignment, redemption in redemptions:
        send_enrollment_code_assignment_email(redemption, assignment.code)
```

This ensures either all codes are assigned or none are, and the frontend always gets either a clean 200 with full results or a clean 500 with nothing committed.

jkachel: Following onto this - email sending should be done in a Celery task, so we're not blocking on an API that may or may not return quickly, and so we can queue up a retry if we need to.

b2b/views/v0/manager.py
```diff
@@ -154,6 +191,16 @@ def get_serializer_class(self):
         """Use different serializers for list vs detail views."""
         if self.action == "list":
             return BaseContractPageSerializer
+
+        # This is ugly and I'm not a fan. We might want to break the assign/revoke/remind/bulk_assign into their
+        # own viewset as long as we can keep the URL structure the same.
+        if self.action in (
```
jkachel: I hear ya. 😄 I'm not real sure what this accomplishes? The functions for these actions return the data using the proper serializers and they have `extend_schema` decorators to keep the API spec in line, which should be sufficient. (Notably - `codes` and `enrollments` also use different serializers but aren't special-cased here. `list` and `detail` are the built-ins, though; the existing mangling was to keep from having to re-define them.)

b2b/views/v0/manager.py
jkachel: I was about to tag the request serializer and then noticed this. Agree here, the data is in the record so no need to pass it back in.

b2b/mail.py
```diff
@@ -0,0 +1,45 @@
...
+
+log = logging.getLogger()
+
+
+class EnrollmentCodeAssignmentMessage(TemplatedMessage):
+    template_name = "mail/enrollment_code_assignment"
+    name = "Enrollment Code Assignment"
+
+
+def send_enrollment_code_assignment_email(assignment, code):
```
jkachel: Noted above where this gets used - this can be made into a task pretty easily; really just need it to take a list of IDs for `DiscountContractAttachmentRedemption`, and then another quick update to load them and  spin through them to send emails. (And add the decorator.)

jkachel (changes requested):
Looks pretty good so far, just a few comments.

b2b/views/v0/manager.py
dsubak: Initially when I was doing this I was trying to assign them as serializers on the `@action` decorator and hitting snags between that and this method. But I've since wired those into `extend_schema` in (I think) where it would be needed, so we can chalk this up to user error. I'll get it removed and see if there are any issues

b2b/views/v0/manager.py
dsubak: It's not strictly necessary. There's one edge case I was thinking of where if manager 1 loads the page, manager 2 revokes and reassigns an unused code to a different email and manager 1 issues the reminder, it'll go to an "unexpected" user - having the email in the payload for a code prevents that mismatch.

But I absolutely don't feel strongly that we need to handle that if we'd prefer not to - it's a pretty fringe edge case.

b2b/views/v0/manager.py
dsubak: We've got an ongoing discussion on revoke's behavior over at https://github.com/mitodl/hq/issues/11375#issuecomment-4673754197 so I'll see what we want to do here vs in follow on PRs.

At minimum, we can add the guard though!

b2b/views/v0/manager.py
dsubak: Re transactionality: can do, but I'll see about using bulk_create if I can. Bit more accounting on my end, but not really a good reason for me _not_ to implement it that way - callers shouldn't need to care either way!

re: email sending - I have no issue pulling it into a task. The only consideration I had was that if we want to indicate to managers that something went wrong sending an email (so they know not to expect that an email has actually gone out) - if we intend to do so, keeping it in process for small payload sizes may make sense. For now I'll task it out and if we decide we need to hand info back to the user, we can always change course.

b2b/views/v0/manager.py
dsubak: This has been removed, resolving

b2b/views/v0/manager.py
dsubak: Guard added, w/ comment pointing at the above discussion. We may need to revisit this behavior, but it's safe enough for now.

b2b/views/v0/manager.py
dsubak: @jkachel I remember why I ended up doing this now. The schema is correctly set by the stuff in `extend_schema`, but if you end up using the DRF-provided UI for testing, the form is incorrect and I found that annoying.

Screenshot 2026-06-12 at 2 03 13 PM

It's not a huge deal, as the swagger UI view shows the right thing and is a fine standin, but definitely a bit of a footgun.

b2b/mail.py
dsubak: Done!

b2b/views/v0/manager.py
dsubak: I've switched the implementation to use bulk_create so it's all in one query, and I've tasked out the email sending. Resolving this for now, but feel free to reopen if anything seems amiss!

dsubak:
@jkachel @daniellefrappier18 I think I've addressed everything ya'll pointed out if. Please take another look at your discretion and let me know what you think!

jkachel (approved):
LGTM 👍

daniellefrappier18 (approved):
LGTM

PR #3651, opened 2026-06-08
Add assigned name to serializer

dsubak:

Adds assigned_name to serializer. It doesn't do much yet since there's no preassignment endpoints yet, but there is some backpopulated data for older records, so this is a step towards the finalized API structure.

### How can this be tested?
With a test contract and organization set up, visit the /codes endpoint. You should observe that there's a new field.
`http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/`. 

Have redeemed an enrollment code and set `DiscountContractAttachmentRedemption.assigned_name` field on the corresponding record you should see that reflected in the API response

Screenshot 2026-06-08 at 3 08 28 PM

### Additional Context
At the moment there isn't a way to set the assigned name - it depends on forthcoming endpoints which will allow building out the "preassignment" functionality (https://github.com/mitodl/mitxonline/pull/3646). That said, you may have some entries in your database which has the field populated [via migration](https://github.com/mitodl/mitxonline/pull/3625/changes#diff-bd4eb877fef2809a34b7f5b507d081041d1d1d38d4faec49e7fce7c414e93d53R8-R25) if you had any redeemed codes prior to that migration running.

### DIFF
--- a/b2b/serializers/v0/manager.py
+++ b/b2b/serializers/v0/manager.py
@@ -96,6 +96,7 @@ class ManagerEnrollmentCodeSerializer(serializers.ModelSerializer):
     redemption_status = serializers.SerializerMethodField()
     assigned_to = serializers.SerializerMethodField()
     assigned_on = serializers.SerializerMethodField()
+    assigned_name = serializers.SerializerMethodField()
     redeemed_on = serializers.SerializerMethodField()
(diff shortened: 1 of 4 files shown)


annagav (approved):
LGTM 👍

PR #3653, opened 2026-06-08
Increase QoL for some B2B things

jkachel:

n/a

This PR just contains a few things that I wanted to make using the system a bit easier:

- OrganizationPage admin list: show contract count for the org
- OrganizationPage admin change: add links to contract inline to navigate to the contract admin pages (both Django Admin and Wagtail); add more fields to the inline
- ContractPage admin: add links to course runs to link to the Django Admin page for the run; add more fields to the course run admin (run tag, variant opts)
- Course admin: add links to the course run in the admin
- General: added a mixin for read-only admin inlines (just disables add/delete/change)

Sort of generally this also introduces a `title_linked` prop that is in a handful of models; this generates a title with a link to the relevant Django Admin and/or Wagtail admin page. It's sort of model-dependent though.

### How can this be tested?

Test the admin: specifically, the Contract, Organization, and Course admins. You should see the new fields and be able to navigate between things successfully.

### Additional Context

This will probably not be the last of this sort of thing; this was a handful of things that I thought would be nice to have and were easy wins so I went ahead with it.

### DIFF
--- a/main/admin.py
+++ b/main/admin.py
@@ -1,5 +1,7 @@
 """Django admin functionality that is relevant to the entire app"""
 
+from collections.abc import Sequence
+
 from django.contrib import admin
 from django.contrib.admin.helpers import ACTION_CHECKBOX_NAME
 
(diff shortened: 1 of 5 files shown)


main/admin.py
```diff
@@ -62,3 +64,49 @@ def changelist_view(self, request, extra_context=None):
...
+
+    extra = 0
+
+    def has_add_permission(self, request, obj):  # noqa: ARG002
+        """Determine if the user can add new ones from here (no, they cannot)"""
+
+        return False
+
+    def has_change_permission(self, request, obj=None):  # noqa: ARG002
+        """Determine if the user can add new ones from here (no, they cannot)"""
```
dsubak, line 80: Nit: I think we could remove this comment from the methods - the comment up top makes the intent pretty clear what's going on.

dsubak (approved):
Looks good!

PR #3655, opened 2026-06-09
feat: adopt mitodl/ol-python-base:3.11, BuildKit cache mounts, granian

blarghmatey:
N/A

Adopts the new shared Python base image, adds BuildKit cache mounts, and migrates from uWSGI to granian (WSGI interface).

**Changes:**
- `FROM mitodl/ol-python-base:3.11` replaces the duplicated substrate stages. The shared base is defined in https://github.com/mitodl/ol-infrastructure/tree/main/dockerfiles/ol-python-base.
- App-specific extras (libcairo2-dev, dnsutils, wget, operator tooling) installed with `--mount=type=cache` for apt.
- `--mount=type=cache,target=/opt/uv-cache,uid=1000,gid=1000` on `uv sync` for fast `uv.lock`-bump rebuilds.
- **CMD changed from `uwsgi` / `uwsgi.ini` to `granian --interface wsgi main.wsgi:application`** on port 8013. `uwsgi.ini` is retained for reference but no longer used by the `django-server`/`production` targets.
- The corresponding local-dev `deployment.yaml` is updated in https://github.com/mitodl/ol-infrastructure/pull/4750.
- `apt.txt` trimmed to extras only with comments documenting what moved to the shared base.
- Multi-stage targets (django-server, production, jupyter-notebook) preserved.

**Image size (django-server target):** 4.09 GB → 2.16 GB (−1.93 GB).

### How can this be tested?
After https://github.com/mitodl/ol-infrastructure/pull/4750 merges and the Concourse pipeline publishes the base image:
```bash
DOCKER_BUILDKIT=1 docker build --target django-server -t mitodl/mitxonline-app:test .
docker run --rm mitodl/mitxonline-app:test granian --version
```
In the local-dev Tilt environment: `tilt up` mitxonline, confirm the webapp comes up healthy under granian and the readiness probe at `/health/readiness` passes.

### Additional Context
granian was already present as a dependency (`granian>=2.5.4,<3` in pyproject.toml) but the Dockerfile and deployment were still running uWSGI. This PR completes the migration. Merge order: ol-infrastructure PR first → Concourse pipeline registered → this PR can be merged.


blarghmatey (commented):
Addressing the bot review comments:

- **`/var/media` ownership** (comment on line 78): no change needed. The base image (`mitodl/ol-python-base:3.11`) already runs `mkdir -p /src /opt/venv /var/media && chown -R mitodl:mitodl /src /opt/venv /var/media`, so the directory is created with the correct ownership before our layers run.
- **`apt.txt` two sources of truth**: fixed — `apt.txt` has been deleted; the Dockerfile is now the single source of truth for app-specific packages.
- **`ENV PORT` ignored by granian CMD**: fixed — CMD now uses shell form `exec granian ... --port ${PORT:-8013} ...` so the port can be overridden at runtime.
- **`uv sync --group prod` installs uwsgi/uwsgitop**: fixed — removed the `prod` dependency group from `pyproject.toml` (and dropped `--group prod` from the Dockerfile `uv sync` call). `uv.lock` updated.
- **`COPY certs/*.crt` fails when no `.crt` files exist**: fixed — changed to `COPY --chmod=644 certs/ /usr/local/share/ca-certificates/` (copies the whole directory; `update-ca-certificates` no-ops if there are no `.crt` files).

rhysyngsun (changes requested):
The nginx configuration needs to be updated to route to granian instead of uwsgi.

https://github.com/mitodl/mitxonline/blob/main/config/nginx.conf.erb

blarghmatey:
Is that template still used at all? My recollection is that the .erb config is a Heroku-only artifact that we should just delete.

rhysyngsun:
It is, that's the template used locally for nginx too, [I normalized this project](https://github.com/mitodl/mitxonline/pull/613) to use a single nginx configuration for heroku and localdev so we didn't have drift between the two.

rhysyngsun:
Here's my nginx logs showing that file getting compiled and the resulting 502 when it tries to connect to the upstream via uwsgi:

```
nginx-1  | 10-listen-on-ipv6-by-default.sh: info: IPv6 listen already enabled
nginx-1  | /docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
nginx-1  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-compile-nginx-conf-erb.sh
nginx-1  | 20-compile-nginx-conf-erb.sh: Running erb on /etc/nginx/templates/nginx.conf.erb to /etc/nginx/nginx.conf
nginx-1  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
nginx-1  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
nginx-1  | /docker-entrypoint.sh: Configuration complete; ready for start up
nginx-1  | 2026/06/18 20:57:07 [error] 29#29: *1 connect() failed (111: Connection refused) while connecting to upstream, client: 172.21.0.9, server: _, request: "GET / HTTP/1.1", upstream: "uwsgi://172.21.0.7:8011", host: "mitxonline.odl.local"
nginx-1  | "18/Jun/2026:20:57:07 +0000" client=172.21.0.9 method=GET request="GET / HTTP/1.1" request_length=828 status=502 bytes_sent=300 body_bytes_sent=150 referer=- user_agent="Mozilla/5.0 (X11; Linux x86_64; rv:151.0) Gecko/20100101 Firefox/151.0" upstream_addr=172.21.0.7:8011 upstream_status=502 request_time=0.000 upstream_response_time=0.001 upstream_connect_time=- upstream_header_time=-
```

blarghmatey:
Fixed in 4e828ba7. Replaced `uwsgi_pass` / `uwsgi_param` block with `proxy_pass http://...` and standard HTTP proxy headers in `config/nginx.conf.erb`, and renamed `NGINX_UWSGI_PASS` → `NGINX_UPSTREAM` in `docker-compose.yml` to match.

rhysyngsun:
I'm still getting a 502 error because the `docker-compose.yml` configuration [overrides the command](https://github.com/mitodl/mitxonline/blob/main/docker-compose.yml#L71) to use `scripts/run-django-dev.sh` which is [hardcoded to run uwsgi still](https://github.com/mitodl/mitxonline/blob/main/scripts/run-django-dev.sh#L27). I also noticed that `uwsgi.ini` is still in the root directory and that can be nuked.

rhysyngsun (approved):
LGTM

PR #3656, opened 2026-06-10
Removing n+1 warning from programs serializer

annagav:
N/A

Removing n+1 warning from programs serializer

This PR makes the v2 program serializer prefetch-aware for collections and requirements, adds regression tests to ensure those paths stay query-efficient, and updates the DRF lint baseline accordingly.

### How can this be tested?
Tests should pass.
Everything should work as expected.


courses/serializers/v2/programs.py
```diff
@@ -646,6 +624,45 @@ def _get_program_collection_ids(instance: Program) -> list[int]:
     )
 
 
+def _get_all_program_requirements(instance: Program) -> list[ProgramRequirement]:
+    """Return all requirements using prefetched data when available."""
+    prefetched_requirements = getattr(instance, "prefetched_all_requirements", None)
```
rachellougee: Does `prefetched_all_requirements` exist?

courses/serializers/v2/programs.py
rachellougee: This is unused method now, do we keep it?

rachellougee (approved):
Tests look good. I left two questions, but they are non-blocking.

PR #3659, opened 2026-06-10
Exclude variant runs from course and program certificate considerations

jkachel:

Closes mitodl/hq#11586

Excludes variant course runs from certificate considerations. This means runs that have either industry or length options set - languages are OK.

### How can this be tested?

For these tests: Make sure the program and the runs that you create are set up so that they're otherwise OK for certificates - they need the `verified` enrollment mode, appropriate start/end dates (or self-paced flag set), and a certificate available date in the past. The program and run enrollments should be verified mode as well.

Make sure everything has product pages and certificates available in Wagtail too. It won't make certificates if there's no certificate to assign.

_Course run cert generation_

Course run generation is relatively easy to test; it pulls a list of eligible runs and you can just evaluate that. Create:

- Course
   - Regular run
   - Translated run
   - Variant run - `variant_length` and/or `variant_industry` set

Test course run eligibility by dropping into a Django shell and running `courses.api.get_certificate_grade_eligible_runs`. You should see the regular and the translated run in the list but not the variant run.

If you want to test certificate generation, you will need to ensure your runs have corresponding runs in edX, and that you have users that are enrolled and have passing grades. (Or, you should also be able to create the grades in mitxonline with the `set_by_admin` flag set.) Then, you should be able to trigger certificate generation using `manage_certificates` or triggering the task directly. (Note that using the override options _also_ disables the certificate eligibility checks too.)

_Program cert generation_

The best data structure for this is:

- Program
   - Required Course
      - Regular Run w/ verified enrollment
   - Required Course
      - Regular Run w/ audit enrollment
- Program
   - Required Course
      - Translated Run w/ verified enrollment
   - Required Course
      - Translated Run w/ audit enrollment
- Program
   - Required Course
      - Variant Run w/ verified enrollment
   - Required Course
      - Variant Run w/ audit enrollment

You'll need enrollments and grades in all of these runs and matching data in edX, and a verified enrollment in the program(s). Then, trigger program certificate generation. The variant runs shouldn't get a certificate and the program they're in shouldn't generate one either, assuming the above buildout.

### Additional Context

This uses a `nonvariant` filter that was added to the `CourseRunQuerySet` - if we need to adjust the logic, we should be able to fairly easily.

I did not add the "certificate eligible" flag as I had suggested; at least for now it seems like the variants that we have (other than language) are generally more "stripped down" versions of the content so it didn't seem particularly useful. This can be revisited later.

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -931,12 +931,18 @@ def get_certificate_grade_eligible_runs(now):
     # Get all the course runs eligible for certificates generation
     # For a valid run it would be live, certificate_available_date would be in future or within a month of passing
     # the certificate_available_date.
+    # Changes for variants: if the run is a variant for anything _other_ than a
+    # language, make it ineligible for a certificate.
 
-    course_runs = CourseRun.objects.live(include_b2b=True).filter(
-        Q(certificate_available_date__isnull=True)
-        | Q(
-            certificate_available_date__gt=now
-            - timedelta(days=settings.CERTIFICATE_CREATION_WINDOW_IN_DAYS)
+    course_runs = (
+        CourseRun.objects.live(include_b2b=True)
+        .nonvariant()
+        .filter(
+            Q(certificate_available_date__isnull=True)
+            | Q(
+                certificate_available_date__gt=now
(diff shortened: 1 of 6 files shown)


annagav (approved):
LGTM 👍

PR #3498, opened 2026-04-16
Configure bumpver for Concourse release pipeline

blarghmatey:
### What are the relevant tickets?
N/A

### Description (What does it do?)
Adds `[tool.bumpver]` configuration to `pyproject.toml` so the Concourse
release pipeline can update the application version automatically on each
release.

The new version format is `YYYY.MM.DD.N` (e.g., `2026.04.16.1`), which
replaces the previous semver-style version strings. This is required by the
Concourse `release` resource workflow being rolled out in
mitodl/ol-infrastructure#4506.

The `VERSION` constant in Django settings is updated to the initial release
format version as part of this change.

### How can this be tested?
1. With `bumpver` installed (`pip install bumpver`), run `bumpver update --dry` from the repo root and confirm it shows the expected diff with no errors.
2. Check that `VERSION` in Django settings matches `current_version` in `pyproject.toml`.

### Additional Context
Part of the Concourse release pipeline modernization — migrating from the Doof
Slack bot to a Concourse-native release workflow using GitHub Issues as
production gates.

PR #3660, opened 2026-06-10
Fix pyproject version to match settings.py

rhysyngsun:

N/A

Follow up to https://github.com/mitodl/mitxonline/pull/3498 which overwrote the pyproject.toml version with an older version. Updated it to match the current version (1.153.0).

### How can this be tested?

Tests should pass.


gumaerc (approved):
👍

PR #3663, opened 2026-06-12
Fix migrate edx data future enrollment

rachellougee:
### What are the relevant tickets?

https://github.com/mitodl/hq/issues/11545

### Description (What does it do?)

updates future enrollments migration in `migrate_edx_data`  management command: 
 - Creates enrollments with edx_enrolled=False 
 - Ensures LegalAddress and UserProfile are created for all enrolled users, as required by `repair_faulty_edx_user`
 - Creates fulfilled orders for verified enrollments (same as before)

### How can this be tested?

Similar instruction as https://github.com/mitodl/mitxonline/pull/3647

For local testing, modify the query in line 779 to be something like:
```
query = (
    "SELECT user_mitxonline_id, courserun_id, courserunenrollment_enrollment_mode, "
    " product_version_id, discount_id, "
    " user_address_country, user_gender, user_birth_year "
    "FROM ("
    "  VALUES "
    "    (1, 2, 'audit', NULL, NULL, 'US','Female', NULL),"
    "    (2, 3, 'audit', NULL, NULL, NULL, NULL, NULL)"
    ") AS t(user_mitxonline_id, courserun_id, courserunenrollment_enrollment_mode, "
    "product_version_id, discount_id, user_address_country, user_gender, user_birth_year)"
)
```
Replace 1,2,3,4 with real IDs from your local, and the migration will create LegalAddress/UserProfile if they don't already exist for this user.

Run `postgres=# delete from users_legaladdress where user_id=1;` to verify the migration recreates it

Run `manage.py migrate_edx_data --type future_enrollments`

Verify

enrollments created at http://mitxonline.odl.local:8013/admin/courses/courserunenrollment/ with edx_enrolled=false
user profile http://mitxonline.odl.local:8013/admin/users/user/

PR #3668, opened 2026-06-15
Update migrate_edx_data to call repair_faulty_edx_user for all users during future enrollments migration

rachellougee:

https://github.com/mitodl/hq/issues/11545

Update migrate_edx_data to call `repair_faulty_edx_user` for all users during future enrollments migration, not just verified users.

### How can this be tested?

Same as https://github.com/mitodl/mitxonline/pull/3663

### DIFF
--- a/courses/management/commands/migrate_edx_data.py
+++ b/courses/management/commands/migrate_edx_data.py
@@ -773,8 +773,13 @@ class Command(BaseCommand):
         if user.id in repaired_user_ids:
             return
         try:
-            repair_faulty_edx_user(user)
+            created_user, created_auth_token = repair_faulty_edx_user(user)
             repaired_user_ids.add(user.id)
+            self.stdout.write(
+                self.style.SUCCESS(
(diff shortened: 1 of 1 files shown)


courses/management/commands/migrate_edx_data.py
```diff
@@ -768,6 +768,20 @@ def _upgrade_enrollment_modes(verified_rows, existing_enrollments):
                 enrollment_mode=EDX_ENROLLMENT_VERIFIED_MODE
             )
 
+    def _repair_user(self, user, repaired_user_ids):
+        """Attempt to repair a faulty edX user, skipping if already repaired."""
+        if user.id in repaired_user_ids:
+            return
+        try:
+            repair_faulty_edx_user(user)
```
annagav: Do we know here if the user got repaired successfully?

rachellougee: I logged the return values of repair_faulty_edx_user, so now we should know if the repair is successful or not.

annagav (approved):
LGTM 👍

PR #3669, opened 2026-06-16
Fix rogue serializer call that isn't prefetching

rhysyngsun:

N/A

This sets the flag that skips the prefetch checks for the CMS APIs.

### How can this be tested?

TBD


ChristopherChudzicki (approved):

dsubak (approved):

dsubak:
Screenshot 2026-06-16 at 9 50 38 AM
Looking good - no longer chucks an error on access to that property

PR #3675, opened 2026-06-16
Reassign endpoint, without fully deleting row

dsubak:

Adds a reassign endpoint, which is functionally similar to an atomic revoke + assign. 

I put together a more extensive PR [here](https://github.com/mitodl/mitxonline/pull/3674) which implemented this by fully deleting the row and recreating it, which necessitated a larger refactor. This is a slightly thinner attempt which works by swapping out individual values on the row. 

The only practical difference here is that the created_on timestamp will reflect the old values instead of the time of reassignment. It's probably up for debate as to whether or not we'd like to overwrite that as well, but it can be overwritten with this approach if needed.

#### Manual Testing
- With a b2b contract set up, assign at least one code to yourself via /assign. You can get the list of codes by visiting `http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/`, and it's easiest to assign the codes using the [swagger UI for /assign](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_codes_assign_create)
- Once executed, you should see at /codes that reflects that the code you assigned has values for `assigned_email`, `assigned_name`, `assigned_by` and `last_sent`. It should also have a `redemption_status` of `assigned`.
- Using the swagger UI for [the new reassign endpoint](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_codes_reassign_update), attempt to reassign the code you've selected.
- The response should reflect the updated email, name, and last_sent values you specified. If you have mailpit set up, you should've received an email invitation with the values specified.

API tests should also continue to pass.

### DIFF
--- a/b2b/views/v0/manager.py
+++ b/b2b/views/v0/manager.py
@@ -698,7 +698,7 @@ class ManagerContractViewSet(NestedViewSetMixin, viewsets.ReadOnlyModelViewSet):
             )
 
         assignment_record.assigned_email = email
-        assignment_record.name = name
+        assignment_record.assigned_name = name
         assignment_record.assigned_by = request.user
         assignment_record.last_reminder_sent_on = now_in_utc()
         assignment_record.save()
--- a/b2b/views/v0/manager.py
+++ b/b2b/views/v0/manager.py
@@ -421,6 +421,7 @@ class ManagerContractViewSet(NestedViewSetMixin, viewsets.ReadOnlyModelViewSet):
         responses={
             200: ManagerEnrollmentCodeSerializer,
             404: DetailErrorSerializer,
+            409: DetailErrorSerializer,
         },
         parameters=[
             OpenApiParameter(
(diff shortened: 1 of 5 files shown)


b2b/views/v0/manager.py
```diff
@@ -421,6 +421,7 @@ def assign_code(self, request, **kwargs):
         responses={
             200: ManagerEnrollmentCodeSerializer,
             404: DetailErrorSerializer,
+            409: DetailErrorSerializer,
```
dsubak, line 424: This was just correcting an oversight - revoke kicks out a 409 if a code was already assigned, now the schema reflects that.

jkachel (approved):
LGTM 👍 

Editing to add: I did try to reassign an unassigned code and got a 404. That's not part of the manual tests so did want to note that it worked as expected.

PR #3676, opened 2026-06-16
Add the prefetch to the CMS course_detail object

rhysyngsun:

N/A

This is a follow up to https://github.com/mitodl/mitxonline/pull/3669 that adds in the correct prefetch and removes the temporary hack. Note: there is still technically an N+1 here if you're logged in and fetch course pages via the list API - but that is a lot messier to fix becuse of how wagtail pages are polymorphic. The long-term fix is to provide page-type specific endpoints so that the wagtail API view types the queryset to the actual course page model instead of the generic `Page` type and we can then do prefetches on that API.

### How can this be tested?

Tests should pass and you should be able to hit http://mitxonline.odl.local/api/v2/pages/?type=cms.coursepage&fields=* and it not error and have `course_details` populated.

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -21,7 +21,6 @@ from django.template.response import TemplateResponse
 from django.urls import reverse
 from django.utils.functional import cached_property
 from django.utils.text import slugify
-from mitol.common.serializers import THIS_IS_NOT_AN_API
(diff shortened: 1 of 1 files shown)


dsubak (approved):
Image

Looks good to me!

PR #3677, opened 2026-06-16
Update to pick up fixes in mitol-django-common

rhysyngsun:

N/A

This picks up the changes from https://github.com/mitodl/ol-django/pull/483

### How can this be tested?

Tests should pass

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
     "ipython>=9.13,<10",
     "mitol-django-apigateway==2026.4.29",
     "mitol-django-authentication==2026.4.29",
-    "mitol-django-common==2026.6.10",
+    "mitol-django-common==2026.6.16.4",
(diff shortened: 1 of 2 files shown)


gumaerc (approved):
I verified locally that switching between this branch and `main` (rebuilding containers in between) stops the hard error from being thrown when you fetch without `prefetch_related`. Here is the test script I used:

```
# Confirm DEBUG is False (simulate production)
from django.conf import settings
settings.DEBUG = False

from courses.models import Course
from courses.serializers.v2.courses import CourseSerializer
import logging
logging.basicConfig(level=logging.ERROR)

# Fetch WITHOUT prefetch_related("programs") — the missing required prefetch
course = Course.objects.first()

# Old version (2026.6.10): raises RequiredPrefetchMissingError here
# New version (2026.6.16.4): logs an error and returns data
print(CourseSerializer(course).data)

```

PR #3680, opened 2026-06-17
Fix the datetimes we generate for course test data

rhysyngsun:

Fixes https://github.com/mitodl/hq/issues/10457

The issue here is that we're generating the catalog data earlier in the test run and the enrollment end date was being specified as needing to be a future one, but by random chance these tests would occasionally fail when faker decided on a datetime only a few minutes in the future. So by the time we got to the assertion we had a stale view on the value of `is_enrollable`. I _think_ this change should avoid this but since it appears to have been a rare occurrence only time will tell. Since we were using "date_time_this_month" and `before_now=False`, this meant the odds of hitting this edge case went up the closer we got to the end of the month.

### How can this be tested?

Tests should pass

### DIFF
--- a/courses/factories.py
+++ b/courses/factories.py
@@ -148,20 +148,20 @@ class CourseRunFactory(DjangoModelFactory):
     )
     run_tag = factory.Sequence("R{0}".format)
     start_date = factory.Faker(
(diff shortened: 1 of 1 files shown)


jkachel (approved):
LGTM 👍

PR #3681, opened 2026-06-17
Fix assign + redeem flow

dsubak:
Danielle pointed out that if you preassigned a code then tried to use the /enrollmentcode learn endpoint to redeem it, the operation would fail. This is because we used to filter out discounts that had DiscountContractAttachmentRedemption records (and were not unlimited seat discounts) - this was fine back when you only had DCAR records which were created while redeeming.

In order for this to function correctly now we need to make a couple changes:
- Filter out DCARs representing assignments when checking discount validity
- Change the behavior of attach so that it will conditionally create or update the DCAR based on whether or not it was preassigned.

### How can this be tested?
- With a b2b contract set up, assign at least one code to yourself via /assign. You can get the list of codes by visiting `http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/`, and it's easiest to assign the codes using the [swagger UI for /assign](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_codes_assign_create)
- Attempt to redeem that code using the [attach API endpoint](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_attach_create). It should succeed with a 201.

### DIFF
--- a/b2b/views/v0/__init__.py
+++ b/b2b/views/v0/__init__.py
@@ -315,12 +315,17 @@ class AttachContractApi(APIView):
     def _get_valid_discount(self, enrollment_code, now):
         """Return a valid discount for the given enrollment code and time."""
         return (
-            Discount.objects.annotate(Count("contract_redemptions"))
+            Discount.objects.annotate(
+                actual_redemptions=Count(
+                    "contract_redemptions",
+                    filter=Q(contract_redemptions__user__isnull=False),
+                )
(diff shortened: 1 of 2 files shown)


jkachel (approved):
LGTM 👍

PR #3683, opened 2026-06-17
Increase test samples and use months

rhysyngsun:

N/A

The test was flaky because the factory defined dates for the `complete` trait in terms of minutes instead of months, probably leading to some undefined behavior. Surprised it worked at all.

### How can this be tested?

Tests should pass


jkachel (approved):
LGTM 👍

PR #3684, opened 2026-06-17
Use CSRFExemptSessionAuthentication for ManagerContractViewSet

dsubak:
Danielle found that most of the mutative endpoints fail w/ CSRF errors (we didn't catch it because /codes is a GET which doesn't enforce CSRF checks, and my testing has been through the DRF and Swagger endpoints, which handle CSRF tokens correctly)

This takes a page from the `attach` endpoint and uses the same authenticator to skip the check for CSRF. It still enforces that you're an authenticated manager.

### How can this be tested?
It's a bit of a pain because you still need to be authenticated as a manager, but the easiest way is going to be by testing the `/remind` endpoint and modifying a request from within the browser.
- Ensure you have a b2b contract w/ at least one assigned code
- Login as a manager for the contract. Open your network tab
- For that assigned code go to `http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_codes_remind_create` and execute the request. This should return a 200.
- Modify the request to drop the `X-CSRFTOKEN` header and resend. The specifics of this step will vary browser to browser. You should observe that the request continues to succeed.

If those steps are taken on a branch without this change, the second request will result in a 403.

### Additional Info
- Chris' comment [here](https://github.com/mitodl/mitxonline/pull/3684#issuecomment-4735838488) suggests that we really ought to be enforcing CSRF. If that's the case, this PR is definitely not the right way to go, and I'll circle back with Danielle to try and figure out what the intended behavior ought to be and we'll chuck this code!

### DIFF
--- a/b2b/views/v0/manager.py
+++ b/b2b/views/v0/manager.py
@@ -100,9 +100,6 @@ class ManagerOrganizationViewSet(viewsets.ReadOnlyModelViewSet):
 
     permission_classes = [IsAuthenticated]
     serializer_class = OrganizationPageSerializer
-    authentication_classes = [
-        CsrfExemptSessionAuthentication,
-    ]
 
     def get_queryset(self):
         """Filter to organizations where the user is a manager."""
@@ -178,6 +175,9 @@ class ManagerContractViewSet(NestedViewSetMixin, viewsets.ReadOnlyModelViewSet):
     """List an organization's contracts."""
 
     permission_classes = [IsAuthenticated, IsOrganizationManager]
+    authentication_classes = [
+        CsrfExemptSessionAuthentication,
+    ]
 
     def get_queryset(self):
         """Get the queryset; add some annotations/etc for computed fields"""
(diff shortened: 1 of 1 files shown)


ChristopherChudzicki:
@dsubak Is this different from mutative enrollment / unenrollment endpoints? Those are also POST / DELETE and enforce CSRF; seems to work with `NEXT_PUBLIC_MITX_ONLINE_CSRF_COOKIE_NAME=csrf_mitxonline` (which i guess is the name that mitxonline hardcodes)

dsubak:
@ChristopherChudzicki I don't see why this API would be different from other mutative APIs, though admittedly I'm not nearly as familiar w/ how APIs are called from Learn as I'd like to be.

This was a pretty naive pattern matching approach on my part against a few of the other B2B endpoints - if consensus that we really _should_ be enforcing CSRF for this type of API, something else is going on and I'll need to figure it out.

jkachel:
We skip the CSRF check for `attach` because, at one point in time, we were generating links directly to it that were intended to be sent to end users (usually via email or some other external interface - where we wouldn't have a CSRF token to check). We don't do that anymore. Users get funneled through the Learn frontend now, and the `attach` API should really be updated to check the CSRF token.

This API won't be used in practice by anything other than Learn and the Swagger UI, so it should check CSRF tokens.

dsubak:
Got it, that all tracks. 

This is absolutely the wrong approach then - I'll close this PR and sync up w/ @daniellefrappier18 and we'll try to get to the bottom of what she's seeing! Thanks for the insight, folks

PR #3685, opened 2026-06-17
Paginate /codes endpoint

dsubak:

- Primarily this PR adds filtering and pagination to /codes. Filtering takes a single `search_term` argument which is used to search across assigned emails and names or redeemed user emails and names.
- At the moment it also paginates /contracts, /course_runs and /enrollments due to pagination being done for the whole viewset. We can avoid doing that if we'd like, but /enrollments may have a cardinality similar to that of /codes, so my thinking is that we should at least paginate those two endpoints.

### How can this be tested?
- With a b2b contract set up, assign a few codes using a few different emails and users. You can get the list of codes by visiting http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/, and it's easiest to assign the codes using the [swagger UI for /assign](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_codes_assign_create)
- You'll also want to set up a few course runs and enrollments in order to view pagination on those endpoints.
- Lastly, to fully test the filtering behavior, you'll want to redeem a few codes using the [attach API](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_attach_create).

#### Pagination

- Visit /codes, /course_runs and /enrollments and /contracts. Observe that each of them returns a paginated response now.
- If you have enough records for the endpoint, observe that you can modify the page_size and page parameters. The responses should reflect these params.

#### Filtering
- Using the [swagger UI for /codes](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_codes_list), specify a search term corresponding to the emails or names you used when assigning them. Your response should have only records which have a substring match on either their name or email.
  - For assigned but unredeemed codes, the search term should filter on the assigned name or assigned email
  - For codes which have been assigned and redeemed, the search term should filter on the assigned name, assigned email, the redeemed email or redeemed user name.

### Additional Info
- This is obviously pretty chock full of breaking changes. Danielle and I will coordinate on scheduling this breaking change to minimize downtime


b2b/views/v0/manager.py
```diff
@@ -292,49 +293,53 @@ def codes(self, request, **kwargs):  # noqa: ARG002
...
+                # I didn't want to just call all() and grab a slice because we might
+                # have _more_ redemptions that we technically allow (if, say, we
+                # adjust the limit down or manually create some redemptions or
+                # something).
+
+                if codes_for_output.count() < contract.max_learners:
+                    # We have seats available, so grab some more codes.
+                    codes_for_output = list(discounts.all()[: contract.max_learners])
+
+        paginator = PageNumberPagination()
```
dsubak: It sucks to do this in the view itself - we should probably bite the bullet and paginate the whole viewset.

b2b/views/v0/manager.py
jkachel, line 110: Thinking this should also be paginated for consistency.

jkachel (changes requested):
One small thing, but otherwise ready to go.

b2b/views/v0/manager.py
dsubak: Can do, should be straightforward - for my edification, is this viewset used anywhere outside of the admin dash to your knowledge?

dsubak:
@jkachel Think this is ready for another look!

b2b/views/v0/manager.py
```diff
@@ -292,50 +320,57 @@ def codes(self, request, **kwargs):  # noqa: ARG002
...
-
-            if codes_for_output.count() < contract.max_learners:
-                # We have seats available, so grab some more codes.
-                codes_for_output = discounts.all()[: contract.max_learners]
-
-            return Response(
-                ManagerEnrollmentCodeSerializer(codes_for_output, many=True).data
+            discounts = contract.get_discounts().prefetch_related(
+                Prefetch(
+                    "contract_redemptions",
```
daniellefrappier18: The search filter and the prefetch_related are operating on different slices of contract_redemptions. The prefetch grabs the latest redemption per code:


```
Prefetch(
    "contract_redemptions",
    queryset=...order_by("-created_on")[:1],
    to_attr="prefetched_redemptions",
)
```
But the search filter matches on any redemption:


```
discounts = discounts.filter(
    Q(contract_redemptions__assigned_email__icontains=search_term)
    ...
).distinct()
```
If a code has two redemptions, one old (assigned to alice@) and one newer (reassigned to bob@) then searching alice will return the code, but the serializer will display bob's data from `prefetched_redemptions[0]`. The search result shows a code that doesn't visibly contain the search term. Managers won't understand why "alice" returned a row showing "bob."

b2b/views/v0/manager.py
```diff
@@ -292,50 +320,57 @@ def codes(self, request, **kwargs):  # noqa: ARG002
...
+                # The point of this is to ensure we always get _all_ the redeemed
+                # codes, and then some unredeemed ones if there's space allowed.
+                # I didn't want to just call all() and grab a slice because we might
+                # have _more_ redemptions that we technically allow (if, say, we
+                # adjust the limit down or manually create some redemptions or
+                # something).
+
+                if codes_for_output.count() < contract.max_learners:
+                    # We have seats available, so grab some more codes.
+                    codes_for_output = list(discounts.all()[: contract.max_learners])
```
daniellefrappier18: Claude flagged this as a performance improvement so take this as what you will but `max_learners` rows fetched into Python before pagination. So for a contract with max_learners=200, all 200 rows are fetched even when page 1 only needs 25.

Before (current code):
```
codes_for_output = discounts.filter(num_redemptions__gt=0).all()

if codes_for_output.count() < contract.max_learners:
    codes_for_output = list(discounts.all()[: contract.max_learners])  # ← Python list

self.paginate_queryset(codes_for_output)  # ← can only slice the list in Python
```
DB fetches: SELECT ... LIMIT 200 → Python gets 200 rows → Python slices to 25 → 175 rows thrown away.

After (suggestion):


```
codes_for_output = discounts.all()[: contract.max_learners]  # ← still a queryset

self.paginate_queryset(codes_for_output)  # ← DRF adds LIMIT 25 OFFSET 0 to the DB query
```
DB fetches: SELECT ... LIMIT 25 OFFSET 0 → Python gets 25 rows → done.

`list()` forces Django to execute the query immediately and pull every row into Python memory. Once it's a Python list, DRF has no way to push pagination back to the database — it can only slice what's already in memory. Keeping it as a queryset means DRF can add LIMIT/OFFSET to the SQL itself, so the DB does the work and only 25 rows ever travel over the wire.

b2b/views/v0/manager.py
dsubak: Good catch, let me see if we can tighten this up!

That said, I was led to believe that we really only expect single use codes (and thus, a max of 1 redemption) for cases where max_learners exists. I suspect if we expect there to ever be more than 1 redemption, we'll end up with a lot of confusing/misleading/unspecified behavior across all these endpoints as most of them take a pretty naive approach to querying for those DACR records. If we think this is a real case we ought to be aware of, may be worth taking a step back and examining it a bit more holistically.

b2b/views/v0/manager.py
dsubak: I _think_ this is safe based on a bit of digging. I wasn't sure if it would compose the two limit/offsets properly (since we ultimately don't wanna return more than max_learners records, paged or otherwise) but as far as I can tell it's at least intended to do the right thing and the queryset carries around some state to keep tabs on it

I'll put it through its paces and as long as everything seems right, we'll stick with letting the eval happen in paginate_queryset

b2b/views/v0/manager.py
jkachel: No, this should all just be for the admin dash functionality.

b2b/views/v0/manager.py
dsubak: Well, if we care about this edge case, https://github.com/mitodl/mitxonline/pull/3685/commits/6b1146e3688f9569ff43a1422eeb4ea837834d9f might be as clean as it's gonna get given our data model and expectations for discount layout in these contracts. Resolving for now

b2b/views/v0/manager.py
```diff
@@ -48,6 +49,12 @@
 log = logging.getLogger(__name__)
 
 
+class ManagerContractOrgPagination(PageNumberPagination):
+    page_size = 25
+    page_size_query_param = "page_size"
+    max_page_size = 100
```
dsubak: Spoke w/ @daniellefrappier18 about using the /codes endpoint directly to support the export to csv behavior. At the moment, we think we can use that endpoint directly and reconstruct the payload as a CSV on the frontend - it'll just mean we specify a page_size large enough to pull back all rows in one shot.

I'm gonna need to remove this `max_page_size` endpoint to facilitate this approach. We may need to revisit it if this endpoint is too slow or the response too big to come back but for now we'll explore this approach and revisit it if it causes issues.

Image

jkachel (approved):
LGTM 👍

b2b/views/v0/manager.py
ChristopherChudzicki: @dsubak I don't think this is accurate anymore. Now always excludes unassigned codes, right?

dsubak: That's correct, I'll change the comment before we land this

PR #3686, opened 2026-06-17
add import and export to b2b_contract

gumaerc:
Closes https://github.com/mitodl/hq/issues/11876

This PR adds the `export` and `import` sub commands to the `b2b_contract` management command.

**Export** serializes a contract and its courseware blueprint to JSON:
- Organization, contract metadata, and variant options
- Programs (with ordering), courses, and departments
- Source runs per course (flagged `is_source_run=True` or `run_tag="SOURCE"`). For older contracts where source runs were never explicitly flagged, falls back to the most recent non-B2B run per variant combo and exports it with `is_source_run=True`. Also applies the course's default variant language to any run with a NULL language so that variant matching works correctly on import.

Contract-specific runs are **not** exported. The import regenerates them via `create_contract_run` (with `skip_edx=True`) so the standard key-generation, product-creation, and variant-filtering machinery runs on the target instance.

**Import** is fully idempotent — safe to re-run. It creates or finds the org, programs, courses, source runs, and contract, then calls `add_program_courses` for each program to generate contract runs. When an existing unflagged source run is found, it updates `is_source_run` and syncs variant fields from the export data so `_get_source_runs_for_course` can match it correctly.

#### Prerequisites
- A running MITx Online instance with at least one B2B contract configured (any contract with programs and courses works; both old-style unflagged runs and new-style `is_source_run=True` runs are handled)
- A target instance to import into (local Docker is fine)

#### Export

```bash
# By contract ID
python manage.py b2b_contract export 695

# By slug, to a file
python manage.py b2b_contract export universal-ai --output /tmp/contract.json

# Pipe-friendly: JSON on stdout, warnings on stderr
python manage.py b2b_contract export 695 > contract.json
```

Verify the output JSON contains `programs`, `courses`, and non-empty `source_runs` arrays. For older contracts (unflagged runs), you should see a warning like:
```
Course course-v1:MITxT+18.03.1x: no designated source runs; using 1 most-recent non-B2B run(s) as fallback
```
and the exported runs will have `is_source_run: true` and a populated `language` field.

#### Import — dry run first

```bash
python manage.py b2b_contract import contract.json --dry-run --slug my-local-clone
```

Confirm the output shows the expected programs, courses, source runs, contract variant options, and a "Created N contract run(s), 0 course(s) without source runs skipped" line before rolling back.

#### Import — committed, with a new slug (same-instance clone test)

```bash
python manage.py b2b_contract import contract.json --slug my-local-clone
```

Then verify:
```bash
python manage.py b2b_list courseware my-local-clone
```

You should see contract runs with keys in the format `course-v1:UAI_<org_key>+<course>+<run_tag>` for the new contract.

#### Idempotency check

Run the import a second time with the same slug:
```bash
python manage.py b2b_contract import contract.json --slug my-local-clone
```

Expected output: `Contract exists`, `Source runs already exist`, and `Created 0 contract run(s)` — nothing duplicated.

#### K8S → local Docker (primary use case, once deployed)

```bash
kubectl exec -n <namespace> <pod> -- python manage.py b2b_contract export <id> > contract.json
docker compose exec -T web python manage.py b2b_contract import - < contract.json
```

### DIFF
--- a/b2b/management/commands/b2b_contract.py
+++ b/b2b/management/commands/b2b_contract.py
@@ -1,17 +1,85 @@
 """Management command for B2B contracts."""
 
+import json
 import logging
+import sys
 from decimal import Decimal
 
+from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ObjectDoesNotExist
 from django.core.management import BaseCommand, CommandError
+from django.db import transaction
 from django.db.models import Q
+from mitol.common.utils import now_in_utc
 
 from b2b.constants import (
     CONTRACT_MEMBERSHIP_CHOICES,
 )
-from b2b.models import ContractPage, OrganizationIndexPage, OrganizationPage
+from b2b.exceptions import SourceCourseIncompleteError
+from b2b.models import (
+    ContractPage,
+    ContractProgramItem,
+    OrganizationIndexPage,
+    OrganizationPage,
+)
+from courses.models import (
+    Course,
+    CourseRun,
+    Department,
+    EnrollmentMode,
+    Program,
+    ProgramRequirement,
+    ProgramRequirementNodeType,
+)
+from variants.models import SupportedVariant
 
 log = logging.getLogger(__name__)
+
+
+def _serialize_date(d):
+    return d.isoformat() if d is not None else None
+
+
+def _serialize_run(run, *, force_is_source_run=False, language_override=None):
+    """Serialize a CourseRun to a dict suitable for the export JSON.
+
+    language_override: pass the course's default-variant language when the run
+    itself carries NULL/empty so the import has the correct value for variant
+    matching in _get_source_runs_for_course.
+    """
+    return {
+        "courseware_id": run.courseware_id,
+        "run_tag": run.run_tag,
+        "title": run.title,
+        "is_source_run": True if force_is_source_run else run.is_source_run,
+        "is_primary_language": run.is_primary_language,
+        "language": language_override
+        if language_override is not None
+        else (run.language or ""),
+        "variant_industry": run.variant_industry or "",
+        "variant_length": run.variant_length or "",
+        "start_date": _serialize_date(run.start_date),
+        "end_date": _serialize_date(run.end_date),
+        "enrollment_start": _serialize_date(run.enrollment_start),
+        "enrollment_end": _serialize_date(run.enrollment_end),
+        "live": run.live,
+        "is_self_paced": run.is_self_paced,
+        "has_courseware_url": run.has_courseware_url,
+        "enrollment_modes": [m.mode_slug for m in run.enrollment_modes.all()],
+    }
+
+
+def _serialize_variant_options(qs):
+    return [
+        {
+            "language": sv.language,
+            "variant_industry": sv.variant_industry,
+            "variant_length": sv.variant_length,
+            "b2b_only": sv.b2b_only,
+            "default_variant": sv.default_variant,
+        }
+        for sv in qs
+    ]
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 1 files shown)


pdpinch:
So... after this merges can we start exporting prod contracts and imporing them into RC? I think we should definitely have MIT, Open Learning and Upgrad contracts on RC, so we can ramp up our testing. 

cc: @cachob

gumaerc:
> So... after this merges can we start exporting prod contracts and imporing them into RC? I think we should definitely have MIT, Open Learning and Upgrad contracts on RC, so we can ramp up our testing.
> 
> cc: @cachob

That's the idea; run the command through kubectl on prod to export JSON, run the import command through kubectl targeted at RC and pipe the JSON back in.

b2b/management/commands/b2b_contract.py
```diff
@@ -150,6 +218,48 @@ def add_arguments(self, parser):
...
+        import_parser = subparsers.add_parser(
+            "import",
+            help="Import a contract from a JSON export file.",
+        )
+        import_parser.add_argument(
+            "input",
+            type=str,
+            help="Path to the JSON export file, or '-' to read from stdin.",
+        )
+        import_parser.add_argument(
```
jkachel, line 246: Bit of a nit - elsewhere (`b2b_codes validate` is one place) we default to doing a dry run. I think it'd be better to do that here too, especially since this will potentially create a ton of new records and/or kick off edX processes.

b2b/management/commands/b2b_contract.py
```diff
@@ -252,13 +362,592 @@ def handle_modify(self, *args, **kwargs):  # noqa: ARG002, C901
...
+            },
+            "programs": programs_data,
+        }
+
+        output = json.dumps(payload, indent=2)
+
+        if output_path:
+            with open(output_path, "w") as f:  # noqa: PTH123
+                f.write(output)
+            self.stderr.write(f"Contract '{contract.slug}' exported to {output_path}")
```
jkachel: Also a bit of a nit: because you're writing to `stderr` it implicitly applies `self.style.ERROR`. Can these be updated to explicitly use `self.style.SUCCESS` (or whatever's appropriate) so it doesn't look like it failed? In some cases, mostly in the import, it's writing to `stderr` when it doesn't really have to so just changing to `stdout` would be preferable. (Obvs not in export since it dumps the json to `stdout`.)

jkachel (changes requested):
Couple of small things. Otherwise works quite well. The idempotency got a workout because I ended up not seeing some runs I was expecting; it turned out that I didn't have a source run for a variant contract run, so it didn't grab it. (I don't think that's a bug - the data was bad, so the fix was fixing the data, and then it was fine.) Re-ran the export/import loop a couple of times after fixing it, and that all worked as expected.

That did bring up a scenario that did not work for me, though. There's no requirement that everything in a B2B contract be contained within a program; you can just add a single course if you want. (In practice, you can't because the Learn frontend doesn't support one-off courses. But MITx Online does support it.) Following the variant issue I threw a random course into the source contract, and the export didn't pick that up. It probably should grab it. I am a bit torn on whether or not this needs to be done _now_ though, since Learn doesn't support this; we could add this in later, when Learn is able to support this use case. (@pdpinch maybe you could weigh in on this?)

Also, for completeness, my test case contract included a bunch of contract runs that were unsupported variants but otherwise were for programs that were in the contract. It didn't pick those up either. This is what I would expect - these aren't valid runs for the contract - so no changes to be made for this; just noting that this case got tested successfully.

gumaerc:
@jkachel Thanks for the review; this is ready for another look

jkachel (approved):
LGTM 👍

PR #3687, opened 2026-06-18
get_user_b2b_organizations unit tests

cp-at-mit:
NA

Adds unit tests for get_user_b2b_organizations which previously had no unit tests.

### How can this be tested?
Tests should pass.

### DIFF
--- a/b2b/api_test.py
+++ b/b2b/api_test.py
@@ -32,6 +32,7 @@ from b2b.api import (
     get_active_contracts_from_basket_items,
     get_contract_products_with_bad_pricing,
     get_contract_runs_without_products,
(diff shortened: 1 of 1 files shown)


b2b/api_test.py
```diff
@@ -2108,3 +2110,159 @@ def test_get_source_course_runs_with_variants_lang_filter(filter_type):
         assert returned_run.id == main_source_course.id
     else:
         assert returned_run.id == variant_2.id
+
+
+# Tests for get_user_b2b_organizations
```
asadali145: Tests are readable, we don't need this comment.

asadali145 (approved):
LGTM!

PR #3688, opened 2026-06-18
Add check for redeemed code or assignments without assigned email and indicate errors if either case is detected in remind endpoint

dsubak:
Danielle and I realized we had a gap in what we should prevent in the /remind endpoint. This implements the following two checks:
- We didn't check for codes which were redeemed. The result is that you could send reminder emails to folks with codes that have already been used. Obviously pointless, so attempting to do so now results in a 409.
- We didn't check for assignments which don't have assigned_to emails. This is a bit of belt and suspenders as the only case where these records shouldn't have an assigned_to value is if they were directly redeemed without preassignment (in which case they'd now get caught by the previous check) but it is cheap to add and understand.

### How can this be tested?

API requests will be easiest to verify with the [Swagger UI](http://mitxonline.odl.local:9080/api/schema/swagger-ui/)

#### Redeemed code guard:
- With a b2b contract set up, redeem at least one code by visiting `/api/v0/b2b/attach/{enrollment_code}/` with an unredeemed code.
- Attempt to hit `/api/v0/b2b/manager/organizations/{parent_lookup_organization}/contracts/{id}/codes/{code}/remind/` with the code you just redeemed. You should get a 409 with an error informing you that you cannot send a reminder to an already claimed assignment.

#### No email guard:
- With a b2b contract set up, assign at least one code by visiting `/api/v0/b2b/manager/organizations/{parent_lookup_organization}/contracts/{id}/codes/{code}/assign/` with an unredeemed code.
- In the database, update the resulting `DiscountContractAttachmentRedemption` record and set `assigned_email` to an empty string.
- Attempt to hit `/api/v0/b2b/manager/organizations/{parent_lookup_organization}/contracts/{id}/codes/{code}/remind/` with the code you just redeemed. You should get a 409 with an error informing you that you cannot send a reminder to an assignment without an email.

### DIFF
--- a/b2b/views/v0/manager.py
+++ b/b2b/views/v0/manager.py
@@ -57,6 +57,12 @@ class CodeAssignment:
     code: str
 
 
+def is_redeemed_attachment_record(
+    assignment_record: DiscountContractAttachmentRedemption,
+) -> bool:
+    return assignment_record.user or assignment_record.redeemed_on
+
+
 def assign_codes_and_send_emails(
     assignments: list[CodeAssignment], assigning_user
 ) -> bool:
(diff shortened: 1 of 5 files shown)


jkachel (approved):
LGTM 👍

PR #3689, opened 2026-06-19
Update b2b_codes to work better with assignments; minor QoL fixes

jkachel:

n/a

Updates the `b2b_codes` management command so that it works better with code assignments:
- Where this filters on attachment redemptions, don't consider redemption records that are assignments
- Add assignments to the `b2b_codes output` output
- Add assignments to the `b2b_codes output --stats` output
- Add new `--assignments` flag to `b2b_codes output` that just shows current assignments
- Add `b2b_codes assign` subcommand
   - Can take a single name/email, or a CSV-format list (same as the frontend), to assign codes to a specified contract
   - Has a `--revoke` flag, which will clear the assignments for the specified user(s)
 
Also, this splits out the course run inlines in the Course change page in Django Admin into two - one that shows just source courses, and one that shows enrollable courses. This also adds the variant length and industry fields, the live flag, and the enrollment end date and removes the primary language toggle.

Late addition: this also updates the contract attach redemption admin to add more information to both the list view and the change view, and allows for the assigned user/email to be updated.

### How can this be tested?

Django Admin changes: load a course with source course runs and regular runs. You should see them in their appropriate sections, and you should see the field changes as well.

Output updates: Create a contract that uses codes and create some assignments (easiest way is to use the admin dashboard in the Learn frontend). Then, use `b2b_codes output --contract <contract>`: the default should include the assignment (if there is one) and should show the proper number of records; adding the `--stats` or `--usage` flags should show you correct assignment data. Use `b2b_codes output --assignments` to list just the assignments for the contract.

Assignment command: Create a contract that uses codes. Then, use `b2b_codes assign`:
- Specify the email and name for a single assignment: `--email <email> --name <name>`. This should work, and the assignment email should be sent.
- Specify a list of users in a CSV file. The format is the same as the frontend - email,name.
   - If any specified users are _already_ assigned to codes, it should list out those users.
   - If there are too many users for the number of seats in the contract, it will skip the contract.
   - If there are too many users for the number of _free_ seats in the contract, it will skip the contract.
   - Otherwise, it will create the appropriate assignments, which you should be able to verify with `b2b_codes output`.
- _For either of these,_ specifying an organization will apply assignments to all applicable contracts in the org.
   - Only seat-limited code contracts count.
   - However, the `--all` flag is invalid - it will not allow you to assign users to all contracts.
- _For either of these,_ changes will _not_ be made unless the `--commit` flag is specified. It should tell you this.

Revocation: Same tests as `b2b_codes assign`, but add the `--revoke` flag. (Again, no changes unless `--commit` is specified.)

### DIFF
--- a/b2b/admin.py
+++ b/b2b/admin.py
@@ -3,6 +3,7 @@
 from django.contrib import admin
 from django.contrib.contenttypes.admin import GenericTabularInline
 from django.db.models import Count
+from django.utils.html import format_html
 
 from b2b.models import (
     ContractPage,
@@ -124,10 +125,19 @@ class ContractPageCourseRunInline(DisplayOnlyAdminMixin, admin.TabularInline):
 
 
 @admin.register(DiscountContractAttachmentRedemption)
-class DiscountContractAttachmentRedemptionAdmin(ReadOnlyModelAdmin):
+class DiscountContractAttachmentRedemptionAdmin(admin.ModelAdmin):
     """Admin for discount attachments."""
 
-    list_display = ["user", "contract", "discount", "created_on"]
+    list_display = [
+        "discount_code",
+        "status",
+        "user",
+        "assignee",
+        "contract",
+        "created_on",
+        "redeemed_on",
+        "last_reminder_sent_on",
+    ]
     list_filter = [
         (
             "user",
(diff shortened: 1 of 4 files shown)


cp-at-mit (approved):
Looks good.  We may want to add some unit tests for the b2b_codes command just so we ensure it doesn't break with future changes to models or functions it depends on

PR #3692, opened 2026-06-22
Hubspot deal stage fix and refactor

cp-at-mit:
https://github.com/mitodl/hq/issues/11776

Update the deal stage to "processed" when any MITx Online course is purchased.  Refactoring some of the Hubspot functions.

### How can this be tested?
With Hubspot configured, purchase any non-UAI course or program.  Verify that the deal created in Hubspot has a stage of "Processed"

### DIFF
--- a/hubspot_sync/api.py
+++ b/hubspot_sync/api.py
@@ -2090,7 +2090,7 @@ def _normalize_deal_properties_for_target_account(  # noqa: C901
     current_stage = deal_properties.get("dealstage")
     mapped_legacy_stage = legacy_stage_map.get(str(current_stage))
     if mapped_legacy_stage and (
-        not allowed_stages or current_stage not in allowed_stages
+        current_stage not in allowed_stages
     ):
         deal_properties["dealstage"] = mapped_legacy_stage
         current_stage = mapped_legacy_stage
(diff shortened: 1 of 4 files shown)


hubspot_sync/api.py
```diff
@@ -2087,7 +2089,9 @@ def _normalize_deal_properties_for_target_account(  # noqa: C901
 
     current_stage = deal_properties.get("dealstage")
     mapped_legacy_stage = legacy_stage_map.get(str(current_stage))
-    if mapped_legacy_stage:
+    if mapped_legacy_stage and (
+        not allowed_stages or current_stage not in allowed_stages
+    ):
```
dsubak: Nit/suggestion - do we know how many items can be in `allowed_stages` at max? If it's small, it might be a bit simpler to read as
```suggestion
    if mapped_legacy_stage and (
        current_stage not in allowed_stages
    ):
```

Probably very marginally less efficient, but afaict we always get back a list of some type, so it should be functionally equivalent.

dsubak (approved):
Image

Took a bit of chicanery to get my local users in a state that didn't skip hubspot entirely but it does look like it'll drop it into processed when synced for an order

PR #3693, opened 2026-06-22
Fix return issues with the B2B one-click enroll api

jkachel:

n/a

Changes two things:
- If the B2B one-click enroll encounters an error, it should now return a 400 rather than a 406.
- If the user tries to hit the one-click API for something they're already in, it should now return a success message rather than the somewhat ambiguous "requires checkout" message.

### How can this be tested?

Attempt to hit the one-click API so that you trigger one of the error states - trying to enroll in a run that's in a contract you're not also in would do it. (Or, just hitting it without being logged in.) You should get an error message back, and it should use the HTTP 400 status code.

Attempt to hit the one-click API using a run that you have an active verified enrollment in already. You should see an error in the log, but the API response should be successful. You should see the order that resulted in the enrollment, if there is one.

### Additional Context

For the second change:
- You may not have an order because there may not be one. If the enrollment got changed to verified manually, there won't be a corresponding order.
- This will still log an error because it's technically still an error. The frontend should not be hitting that API if you're already enrolled in the course. But we should also not prevent you from getting to the content.


b2b/views/v0/__init__.py
```diff
@@ -174,7 +174,7 @@ def post(self, request, readable_id: str, format=None):  # noqa: A002, ARG002
             CreateB2BEnrollmentSerializer(response).data,
             status=status.HTTP_201_CREATED
             if response["result"] == USER_MSG_TYPE_B2B_ENROLL_SUCCESS
-            else status.HTTP_406_NOT_ACCEPTABLE,
```
dsubak, line 177: I think the schema isn't correct for this endpoint - looks like it only reflects that we return a 200 if I'm reading this right. Should we correct it so that it shows a 201/400 status code return while we're touching the file?

b2b/api.py
```diff
@@ -1204,6 +1205,23 @@ def _validate_b2b_enrollment_prerequisites(user, product: Product) -> Union[dict
         )
         return {"result": main_constants.USER_MSG_TYPE_B2B_ERROR_NO_CONTRACT}
 
+    if (
+        isinstance(purchasable_object, CourseRun)
+        and purchasable_object.enrollments.filter(
+            user=user,
+            change_status="",
```
dsubak: This might be a dumb question, but it looks like change_status is nullable and frequently checked against None (as opposed to an empty string) - I'm having some trouble hitting this codepath, and it looks like I have null values in my DB records. Should this be filtering on None or is my local data just messed up?

I did also check prod and we've got some of each case, but far more rows with null than with an empty string
Image

jkachel: It's either since the field is set nullable. Updated the API to explicitly exclude records with the field set to one of the allowed choices.

dsubak: Gotcha, I wasn't sure if we functionally only expected one vs the other. Thanks for clarifying that both mean the same thing.

With that change, everything is working as expected!

dsubak (approved):

PR #3695, opened 2026-06-23
Relabel certificate "Product name" field to "Certificate Title"

zamanafzal:
### What are the relevant tickets?
Closes https://github.com/mitodl/hq/issues/11907

### Description (What does it do?)
- Sets `verbose_name="Certificate Title"` on `CertificatePage.product_name` so the CMS edit form shows **"Certificate Title"** instead of "Product name".
- Updates the help text to: *"Specify the course/program name. Note: This is what will appear on the learner's certificate."*
- Includes the generated `AlterField` migration (`0065_…`) and a field-label test.

## Why relabel only — we do NOT rename the field
The DB column / API field name stays **`product_name`**; we only change its **label** (`verbose_name`) and help text. `product_name` is referenced in ~40 places across DRF serializers, the Wagtail API schema, the certificate template, factories, HubSpot sync, management commands — **and the mit-learn API client**. Renaming the actual field would require a data migration, an OpenAPI/client regeneration, and updating every one of those references. Since the field is only surfaced in the CMS via its edit-form panel, `verbose_name` gives the desired label with none of that blast radius. The migration is a DB no-op (`verbose_name`/`help_text` are metadata only) but is required so CI's `detect_missing_migrations` check passes.

### How can this be tested?

**Local**
1. Run the MITx Online stack on this branch.
2. Migration check: `./manage.py makemigrations --check --dry-run` → **"No changes detected"**.
3. Field-label test: `pytest cms/models_test.py::test_certificate_page_product_name_field_labels`.
4. In the Wagtail CMS admin, open a certificate page's edit form — Pages → a Course/Program → its **Certificate** child → **Edit** (or `/cms/pages/<id>/edit/`).

**Verify**
- The field previously labeled **"Product name"** now reads **"Certificate Title"**, with help text *"Specify the course/program name. Note: This is what will appear on the learner's certificate."*
- `makemigrations --check` is clean; `pytest` passes.
- API field name is unchanged (`product_name`) — no API spec / client change.

### Screenshot
Screenshot 2026-06-23 at 3 23 35 PM

Screenshot 2026-06-23 at 3 24 01 PM

PR #3697, opened 2026-06-24
Add backfill_program_certificate_titles management command

zamanafzal:

https://github.com/mitodl/hq/issues/11938

### Description
Issued program certificates **freeze** their "Certificate Title" (`product_name`) into the Wagtail page revision in effect at issue time (`ProgramCertificate.certificate_page_revision`), and the display path reads `certificate_page_revision.as_object().product_name`. So editing the live "Certificate Title" in the CMS does **not** update certificates already issued — they keep their old frozen title until a one-time backfill is run.

This adds a `backfill_program_certificate_titles` management command (workstream 1 of hq#11938, follow-up to hq#11907 / #3695) that rewrites each issued program certificate's frozen `product_name` to the live CMS page value.

```
python manage.py backfill_program_certificate_titles \
    (--program <readable_id> [--program <readable_id> ...] | --all-programs) \
    [--commit]
```

## Behavior & safety

- **Dry-run by default** — reports planned changes; `--commit` is required to write.
- **Mode validation** — exactly one of `--program` (repeatable) / `--all-programs`; neither or both raises `CommandError`, so a bare run touches nothing.
- **Placeholder/empty guard** — skips any program whose live title is empty or still a `"PLACEHOLDER - "` default; placeholder text is never stamped onto real certificates.
- **Idempotent** — revisions already matching the live title are reported `Will change: NO`; a second `--commit` run is a no-op.
- **Distinct-revision dedup** — certs issued at the same page revision share one `Revision` row, so each shared revision is edited once (report shows `(N cert(s))`).
- **Tolerant skips (never crash)** — unknown readable_id, missing CMS page, missing certificate page, and null-revision (revoked) certs are reported and skipped.
- **Rollback-safe output** — each program's report block is built in memory and flushed to stdout only after its writes commit, so a rollback never leaves behind output claiming a change that was undone.
- Per-program writes run inside `transaction.atomic()`; a final `Summary:` line tallies programs (will change / already correct / skipped), revisions changed, and certs affected.

## Output

Every targeted program prints an admin-readable block — `readable_id`, program title, current frozen title, target live title, and a `YES` / `NO` / `SKIPPED` change flag — so the dry-run can be shared with product/admins for sign-off before committing:

```
=== DRY RUN ===

program-v1:UAI+B2C
  Program Title    : UAI B2C Program
  Cert Title (new) : Universal AI 0909
  Cert Title (now) : Universal AI11  (1 cert(s), revision 126)
  Will change      : YES

Summary: 1 program(s), 1 will change, 0 already correct, 0 skipped (1 revision(s), 1 cert(s) affected).
```

In `--commit` mode the block is identical; only the banner changes to `=== COMMIT ===`.

## Write strategy

Rewrites the existing `Revision.content["product_name"]` (per design): simplest, fewest writes, fixes all certs sharing a revision at once. Trade-off — the page's revision-history view will retroactively show the corrected title, which is acceptable for a one-time correction.

## How to test

The command rewrites the **frozen** "Certificate Title" on already-issued program certificates to the **live** CMS value, so you need a program whose live Certificate Title differs from the one frozen on an issued cert.

### 1. Update the "Certificate Title" in Wagtail
1. Open the Wagtail CMS admin (local): `http://mitxonline.odl.local:8013/cms/pages`
2. Go to a **Program** page that already has an issued certificate (e.g. `program-v1:UAI+B2C`) → its child **Certificate** page.
3. Change the **"Certificate Title"** (`product_name`) field to a new value, distinct from the title currently frozen on the issued cert, and **Publish**.

(If the program has no issued certificate yet, issue one for your user via MITx Online Django admin first.)

### 2. Dry-run — preview what would change (no writes)
```
docker compose run --rm web python manage.py backfill_program_certificate_titles --program program-v1:UAI+B2C
```
The edited program reports `Will change: YES`, with `Cert Title (now)` (old frozen value) → `Cert Title (new)` (your new live value). Save the report to a file to share with an admin for sign-off:
```
docker compose run --rm web python manage.py backfill_program_certificate_titles --all-programs > cert_backfill_dryrun.txt 2>/dev/null
```
(The report goes to stdout; `2>/dev/null` drops container startup noise. Colors are auto-stripped when not a TTY.)

### 3. Commit — apply the change
```
docker compose run --rm web python manage.py backfill_program_certificate_titles --program program-v1:UAI+B2C --commit
```
Or for every program at once:
```
docker compose run --rm web python manage.py backfill_program_certificate_titles --all-programs --commit
```

### 4. See the change
- **Re-run the dry-run** — the program now reports `Will change: NO` (frozen title matches live), proving the backfill landed and is idempotent.
- **On the certificate** — open the issued program certificate and confirm it now displays the updated Certificate Title.

### Automated tests
```
docker compose run --rm web uv run pytest courses/management/commands/test_backfill_program_certificate_titles.py
```

## Test plan

- [x] `--commit` rewrites frozen `product_name` to the live title (asserted via the display path `revision.as_object().product_name`)
- [x] Dry-run (default) makes no DB writes and reports the planned change
- [x] Idempotent: second `--commit` reports zero changes
- [x] Placeholder / empty live title → program skipped, no writes
- [x] Null-revision (revoked) cert → not touched
- [x] Shared revision: multiple certs → single edit, all updated
- [x] Rollback-safe: a mid-commit save failure rolls back fully and emits no success output
- [x] Mode validation: neither flag → `CommandError`; both flags → `CommandError`
- [x] `--all-programs` processes clean programs and skips placeholder / page-less ones
- [x] Unknown `--program` readable_id reported and skipped; other programs still processed
- [x] 15 tests pass; `ruff format` + `ruff check` clean
- [x] Verified end-to-end locally: edited a CMS Certificate Title → dry-run showed `YES` → `--commit` applied it → re-run reported `NO`

### DIFF
--- a/courses/management/commands/backfill_program_certificate_titles.py
+++ b/courses/management/commands/backfill_program_certificate_titles.py
@@ -115,12 +115,8 @@ class Command(BaseCommand):
 
         Returns ``(changed, revisions_changed, certs_affected)``.
         """
-        # Build the whole report block in memory and decide which revisions to
-        # rewrite, but do NOT print anything yet. The writes happen first, inside
-        # transaction.atomic(); only once they commit do we flush the block. That
-        # way a rollback (e.g. a save() error on a later revision) can never leave
-        # behind output claiming a change that was undone — on failure the
-        # exception propagates before any line is written.
+        # Build the block in memory and flush it only after the writes commit, so
+        # a rollback never leaves behind output claiming a change that was undone.
         lines = [
             "",
             readable_id,
@@ -153,9 +149,8 @@ class Command(BaseCommand):
         if commit and to_change:
             with transaction.atomic():
                 for revision in to_change:
-                    # Reassign rather than mutate in place so the change is
-                    # always picked up regardless of save()/update_fields or
-                    # any dirty-tracking behavior on the model.
+                    # Reassign (don't mutate in place) so save(update_fields=...)
+                    # reliably detects the JSONField change.
                     content = revision.content
                     content["product_name"] = title
                     revision.content = content
--- a/courses/management/commands/backfill_program_certificate_titles.py
+++ b/courses/management/commands/backfill_program_certificate_titles.py
@@ -1,0 +1,152 @@
+"""Backfill the frozen "Certificate Title" on already-issued program certificates."""
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+
+from courses.models import Program, ProgramCertificate
+
+
+class Command(BaseCommand):
+    """Rewrite issued program certificates' frozen product_name to the live title."""
+
+    help = __doc__
+
+    def add_arguments(self, parser):
+        """Add the command's program-selection and --commit arguments."""
+        parser.add_argument("--program", action="append", dest="programs")
+        parser.add_argument("--all-programs", action="store_true", dest="all_programs")
+        parser.add_argument("--commit", action="store_true")
+        super().add_arguments(parser)
+
+    def handle(self, *args, **options):  # noqa: ARG002
+        """Backfill frozen program-certificate titles from the live CMS page."""
+        programs = options["programs"]
+        all_programs = options["all_programs"]
+        if bool(programs) == bool(all_programs):
+            msg = (
+                "Provide exactly one of --program <readable_id> (repeatable) "
+                "or --all-programs."
+            )
+            raise CommandError(msg)
+
+        commit = options["commit"]
+        banner = "COMMIT" if commit else "DRY RUN"
+        self.stdout.write(self.style.WARNING(f"=== {banner} ==="))
+
+        programs_processed = 0
+        programs_skipped = 0
+        revisions_changed = 0
+        certs_affected = 0
+
+        for readable_id, program in self._target_programs(programs, all_programs):
+            if program is None:
+                programs_skipped += 1
+                self.stdout.write(
+                    self.style.ERROR(
+                        f"{readable_id}: skipping — no program with that readable_id."
+                    )
+                )
+                continue
+
+            certificate_page = self._certificate_page(program)
+            if certificate_page is None:
+                programs_skipped += 1
+                self.stdout.write(
+                    self.style.WARNING(
+                        f"{readable_id}: skipping — no certificate page."
+                    )
+                )
+                continue
+
+            title = certificate_page.product_name
+
+            if not title or title.startswith("PLACEHOLDER - "):
+                programs_skipped += 1
+                self.stdout.write(
+                    self.style.WARNING(
+                        f"{readable_id}: skipping — live Certificate Title is "
+                        f'empty or a placeholder ("{title}"). Fix the CMS page first.'
+                    )
+                )
+                continue
+
+            programs_processed += 1
+            self.stdout.write(f'{readable_id}: target Certificate Title "{title}"')
+
+            with transaction.atomic():
+                for revision, certs in self._distinct_revisions(program):
+                    old = revision.content.get("product_name")
+                    if old == title:
+                        self.stdout.write(
+                            f'  {revision.id}  ({len(certs)} certs)  "{old}"  [skipped]'
+                        )
+                        continue
+                    if commit:
+                        revision.content["product_name"] = title
+                        revision.save()
+                        status = "changed"
+                    else:
+                        status = "would-change"
+                    revisions_changed += 1
+                    certs_affected += len(certs)
+                    self.stdout.write(
+                        self.style.SUCCESS(
+                            f"  {revision.id}  ({len(certs)} certs)  "
+                            f'"{old}" -> "{title}"  [{status}]'
+                        )
+                    )
+
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"Summary: {programs_processed} program(s) processed, "
+                f"{programs_skipped} skipped, "
+                f"{revisions_changed} revision(s) changed, "
+                f"{certs_affected} cert(s) affected."
+            )
+        )
+
+    @staticmethod
+    def _target_programs(programs, all_programs):
+        """Yield ``(readable_id, program_or_None)`` for each targeted program.
+
+        For ``--program`` a missing readable_id yields ``(readable_id, None)`` so
+        the caller can report and skip it. For ``--all-programs`` every program is
+        yielded (with its own readable_id).
+        """
+        if all_programs:
+            for program in Program.objects.all():
+                yield program.readable_id, program
+            return
+        by_readable_id = {
+            program.readable_id: program
+            for program in Program.objects.filter(readable_id__in=programs)
+        }
+        for readable_id in programs:
+            yield readable_id, by_readable_id.get(readable_id)
+
+    @staticmethod
+    def _certificate_page(program):
+        """Return the program's live ``CertificatePage``, or ``None`` if the
+        program has no CMS page or no certificate page.
+        """
+        try:
+            page = program.page
+        except ObjectDoesNotExist:
+            return None
+        if page is None:
+            return None
+        return page.certificate_page
+
+    @staticmethod
+    def _distinct_revisions(program):
+        """Yield ``(revision, certs)`` for each distinct non-null revision shared
+        by the program's active certificates.
+        """
+        grouped = {}
+        for cert in ProgramCertificate.objects.filter(program=program):
+            revision = cert.certificate_page_revision
+            if revision is None:
+                continue
+            grouped.setdefault(revision.id, (revision, []))[1].append(cert)
+        return list(grouped.values())
--- a/courses/management/commands/backfill_program_certificate_titles.py
+++ b/courses/management/commands/backfill_program_certificate_titles.py
@@ -1,10 +1,29 @@
-"""Backfill the frozen "Certificate Title" on already-issued program certificates."""
+"""Backfill the frozen "Certificate Title" on issued program certificates."""
+
+from collections import Counter
 
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.management.base import BaseCommand, CommandError
 from django.db import transaction
 
 from courses.models import Program, ProgramCertificate
+
+# A live Certificate Title still carrying this prefix has never been filled in.
+PLACEHOLDER_PREFIX = "PLACEHOLDER - "
+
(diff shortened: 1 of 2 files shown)


courses/management/commands/backfill_program_certificate_titles.py
```diff
@@ -0,0 +1,152 @@
...
+                for revision, certs in self._distinct_revisions(program):
+                    old = revision.content.get("product_name")
+                    if old == title:
+                        self.stdout.write(
+                            f'  {revision.id}  ({len(certs)} certs)  "{old}"  [skipped]'
+                        )
+                        continue
+                    if commit:
+                        revision.content["product_name"] = title
+                        revision.save()
```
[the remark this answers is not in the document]
zamanafzal: It's a false positive, but I applied the hardening.

courses/management/commands/backfill_program_certificate_titles.py
```diff
@@ -0,0 +1,210 @@
...
+
+        commit = options["commit"]
+        banner = "COMMIT" if commit else "DRY RUN"
+        self.stdout.write(self.style.WARNING(f"=== {banner} ==="))
+
+        programs_will_change = 0
+        programs_already_correct = 0
+        programs_skipped = 0
+        revisions_changed = 0
+        certs_affected = 0
```
asadali145: We can use `from collections import Counter`

courses/management/commands/backfill_program_certificate_titles.py
```diff
@@ -0,0 +1,211 @@
...
+                f"({revisions_changed} revision(s), {certs_affected} cert(s) affected)."
+            )
+        )
+
+    def _write_block(self, readable_id, *, program_title, new_title, reason):
+        """Print a full per-program block for a program that is being skipped."""
+        self.stdout.write("")
+        self.stdout.write(readable_id)
+        self.stdout.write(f"  Program Title    : {program_title}")
+        self.stdout.write("  Cert Title (now) : —")
```
asadali145: Instead of `now`, "old" or "existing". might be better here.

courses/management/commands/backfill_program_certificate_titles.py
```diff
@@ -0,0 +1,211 @@
...
+                programs_will_change += 1
+            else:
+                programs_already_correct += 1
+
+        total = programs_will_change + programs_already_correct + programs_skipped
+        self.stdout.write("")
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"Summary: {total} program(s), "
+                f"{programs_will_change} will change, "
```
asadali145: This should say `Changed` when in commit mode.

courses/management/commands/backfill_program_certificate_titles.py
```diff
@@ -0,0 +1,211 @@
...
+            old = revision.content.get("product_name")
+            label = f"{len(certs)} cert(s), revision {revision.id}"
+            lines.append(f"  Cert Title (now) : {old}  ({label})")
+            if old == title:
+                lines.append("  Will change      : NO")
+                continue
+            to_change.append(revision)
+            revisions_changed += 1
+            certs_affected += len(certs)
+            lines.append(self.style.SUCCESS("  Will change      : YES"))
```
asadali145: This should say `Changed: YES` in commit mode.

courses/management/commands/backfill_program_certificate_titles.py
```diff
@@ -0,0 +1,211 @@
...
+    def handle(self, *args, **options):  # noqa: ARG002
+        """Backfill frozen program-certificate titles from the live CMS page."""
+        programs = options["programs"]
+        all_programs = options["all_programs"]
+        if bool(programs) == bool(all_programs):
+            msg = (
+                "Provide exactly one of --program <readable_id> (repeatable) "
+                "or --all-programs."
+            )
+            raise CommandError(msg)
```
asadali145:
```suggestion
if not programs and not all_programs:
    raise CommandError("Provide --program <readable_id> or --all-programs.")
if programs and all_programs:
    raise CommandError("--program and --all-programs are mutually exclusive.")
```

courses/management/commands/backfill_program_certificate_titles.py
```diff
@@ -0,0 +1,211 @@
...
+        except ObjectDoesNotExist:
+            return None
+        if page is None:
+            return None
+        return page.certificate_page
+
+    @staticmethod
+    def _distinct_revisions(program):
+        """Yield ``(revision, certs)`` for each distinct non-null revision shared
+        by the program's active certificates.
```
asadali145: Docs are inconsistent with implementation.

courses/management/commands/backfill_program_certificate_titles.py
```diff
@@ -0,0 +1,211 @@
...
+    @staticmethod
+    def _target_programs(programs, all_programs):
+        """Yield ``(readable_id, program_or_None)`` for each targeted program.
+
+        For ``--program`` a missing readable_id yields ``(readable_id, None)`` so
+        the caller can report and skip it. For ``--all-programs`` every program is
+        yielded (with its own readable_id).
+        """
+        if all_programs:
+            for program in Program.objects.all().order_by("readable_id"):
```
asadali145, line 194: objects is an Active certificates manager for both course and program certificates, it will result in skipping the revoked and future certificates with bad titles.

courses/management/commands/test_backfill_program_certificate_titles.py
```diff
@@ -0,0 +1,298 @@
...
+def _run(*args):
+    out = StringIO()
+    call_command("backfill_program_certificate_titles", *args, stdout=out, stderr=out)
+    return out.getvalue()
+
+
+def _frozen_title(certificate):
+    revision = certificate.certificate_page_revision
+    revision.refresh_from_db()
+    return revision.as_object().product_name
```
asadali145: Can we do?

```suggestion
def _frozen_title(certificate):
        revision = certificate.certificate_page_revision
       revision.refresh_from_db()
       return revision.content["product_name"]
```

courses/management/commands/test_backfill_program_certificate_titles.py
```diff
@@ -0,0 +1,298 @@
...
+    cert_page.product_name = live_title
+    cert_page.save()
+
+    return program_page.program, certificate
+
+
+def _run(*args):
+    out = StringIO()
+    call_command("backfill_program_certificate_titles", *args, stdout=out, stderr=out)
+    return out.getvalue()
```
asadali145, line 55: Copilot: Merging stdout and stderr into the same buffer makes it impossible to distinguish normal output from error messages in assertions. Tests like test_dry_run_output_reports_planned_change that assert "DRY RUN" in output could inadvertently pass if the string appears in an error message. Use separate buffers:
```suggestion
def _run(*args):
    out, err = StringIO(), StringIO()
    call_command("backfill_program_certificate_titles", *args, stdout=out, stderr=err)
    return out.getvalue()
```

asadali145 (changes requested):
Looks good overall.

- I have tested the updates in revision with the certificate APIs as I don't have Learn setup.
- I would have liked this to be more generic like it should also work for course certificates. Not sure why we are not doing that. This could be helpful util in future.
- As discussed with Zaman, the updated revision product_name will be only visible in Learn and MITx Online certificate endpoints will still display the frozen names.  We should confirm this with Peter or Linda whoever is related.

I have left some inline review comments. Also, there are quite a few magic strings that could be converted to module level constants for better readability and reusability.

zamanafzal:
Thanks for the suggestions!

> I would have liked this to be more generic like it should also work for course certificates. Not sure why we are not doing that. This could be helpful util in future.

The plan is to do this initially for programs, since that's the priority, and then extend it to courses in the next phase — as mentioned [here](https://github.com/mitodl/hq/issues/11907#issuecomment-4751876948). @pdpinch, does this phased approach (programs now, courses later) work on your end, or is there urgency on the course side we should account for sooner?

> As discussed with Zaman, the updated revision product_name will be only visible in Learn and MITx Online certificate endpoints will still display the frozen names. We should confirm this with Peter or Linda whoever is related.

Yes, that's correct — the certificate title change was only made on the Learn side so far. @skurchak, do we need to make the same change on the MITx Online side alongside this?

zamanafzal:
@asadali145 I've addressed the feedback. Please take a look.

asadali145 (approved):
Looks good and works as expected to update the product_title for the program certificate revision. 

Change is only visible in the certificates API for the MIT Learn. MITx Online certificates still display the old certificate title.

zamanafzal:
@mitodl/devops Can someone please run this management command on RC and share the cert_backfill_dryrun.txt file?

`docker compose run --rm web python manage.py backfill_program_certificate_titles --all-programs > cert_backfill_dryrun.txt 2>/dev/null`

zamanafzal:
I discussed with @asadali145, and I'll be running this management command myself after completing EKS Access Setup.

zamanafzal:
I've run the management command on RC, and this is the result.
Dry Run Mode on RC:
[cert_title_dry_run.txt](https://github.com/user-attachments/files/29736388/cert_title_dry_run.txt)

Commot Mode on RC:
[cert_title_commit.txt](https://github.com/user-attachments/files/29736588/cert_title_commit.txt)

Dry Run(again) Mode on RC:
[cert_title_dry_run_after_commit.txt](https://github.com/user-attachments/files/29736718/cert_title_dry_run_after_commit.txt)

There was one certificate whose revision needs an upgrade and that is done with the management command.
  Cert Title (new) : Universal AI Certificate
  Cert Title (old) : Universal AI

zamanafzal:
This is the result of production when I run the management command in dry run mode. 
[cert_title_dry_run_PRODUCTION.txt](https://github.com/user-attachments/files/29790489/cert_title_dry_run_PRODUCTION.txt)

@pdpinch @skurchak Please take a look at the certificate titles, which we are going to update for the old revisions. Once I'll get the go ahead I'll run the management command in commit mode and apply these changes.

pdpinch:
Thanks @zamanafzal. @cachob and I need to take a look at those MicroMasters certificates before you run the script.

skurchak:
@pdpinch how can I help check old titles vs new, as the only one I have instance I have first hand knowledge of is the UAI B2B Program certificate?

pdpinch:
@skurchak @cachob I'd like to review the file @zamanafzal generated with you: [cert_title_dry_run_PRODUCTION.txt](https://github.com/user-attachments/files/29790489/cert_title_dry_run_PRODUCTION.txt) (see https://github.com/mitodl/mitxonline/pull/3697#issuecomment-4912630923)

PR #3704, opened 2026-06-25
Remove integration-specific options from contracts

jkachel:

Closes https://github.com/mitodl/hq/issues/11657

This PR:
- Migrates all contracts with `membership_type` set to `sso` or `non-sso` to `managed` and `code`, respectively.
- Drops the `integration_type` field from `ContractPage`.
- Drops the `sso` and `non-sso` type choices.
- Updates places where the code references these now-removed things.

### How can this be tested?

Automated tests should pass, and the system should work in general. The B2B system specifically should work as expected - you should be able to create contracts, add courseware, etc. and everything that is expected to happen should happen.

Additionally, the auto-attach functionality should be tested:
- Create an org in Keycloak, and a corresponding org in MITx Online. (The `sync_keycloak_orgs` command may be helpful here.)
- Create a contract for the org. Ensure it is set up properly for auto-attach: type `managed`, no price, no seat limit.
- Create a _second_ contract for the org. Ensure that it is _not_ set up for auto-attach - use membership type `code`.
- Sign in with a _new_ user. That is, go through the registration process fresh; don't use an existing Keycloak account, etc.
- Ensure the account looks correct in MITx Online - it should have no B2B contracts or orgs.
- Log out of MITx Online. (Ideally, clear cookies or just close out your incognito session to make sure you don't have any stale data hanging out.)
- In Keycloak, add the user to the org you created earlier.
- Log back into MITx Online using the new user.
- Check the user's account. You should see that they're now in the org, and that they're in the first contract, but not the second.
- Log out of MITx Online again (and clear cookies, reset incognito, etc.)
- In Keycloak, remove the user from the org.
- Log back into MITx Online for a third time with the new user.
- Check the user's account. They should now no longer be in the auto-attach contract or in the org.

If you wish you can also add them to the contract with an enrollment code, and then test adding and removing the user from the Keycloak org again; they should remain in the contract. (This functionality didn't change, so it should work as expected - if there's questions about what the behavior should be, double-check against main.)

The migration in this PR is reversible. I didn't add any new options so it does not set anything back to use `sso` or `non-sso`, but it does make sure the `integration_type` field is re-populated from `membership_type`.

### Additional Context

As noted elsewhere, the contract doesn't specify whether or not users are to log in via SSO or not. It does specify whether or not the user gets access to the contract based solely on their membership in the org, though.


b2b/api_test.py
```diff
@@ -309,12 +309,9 @@ def test_ensure_enrollment_codes(  # noqa: PLR0913
     assert_price = price if price else Decimal(0)
 
     contract = factories.ContractPageFactory(
-        integration_type=CONTRACT_MEMBERSHIP_SSO
+        membership_type=CONTRACT_MEMBERSHIP_MANAGED
         if is_sso
```
dsubak: This is an ultra-nit, but we could switch the arg names referencing sso at this point to use the term `managed` instead for consistency if we wanted to (`is_sso` -> `is_managed` and `update_sso` -> `update_managed`)

dsubak (commented):
Still owe you some manual testing before I can approve, but it looks pretty straightforward mechanically. Big fan of making the contract models a bit less crufty!

b2b/management/commands/b2b_contract.py
```diff
@@ -111,7 +111,7 @@ def add_arguments(self, parser):
             help="The name of the contract.",
         )
         create_parser.add_argument(
-            "integration_type",
+            "membership_type",
             type=str,
             help="The membership type for this contract.",
             choices=[value[0] for value in CONTRACT_MEMBERSHIP_CHOICES],
```
dsubak: Think this constant got dropped - we'll need to fix the import and this choice valueset

dsubak (approved):
I think this is looking good! After a bit of hullabaloo fighting keycloak, I was able to step through the manual testing and get my test user added to the contract on login while in the KC Org and removed from it while removed.

PR #3705, opened 2026-06-25
Add summary statistics to contract retrieve endpoint and remove unassigned codes from /codes output. Add filtering on assigned + redeemed statuses

dsubak:
Now that we've paginated the API, the frontend doesn't have a particularly good way to grab the breakdown of codes to show to the contract manager:
Screenshot 2026-06-25 at 12 10 28 PM

Previously, these were just calculated by looking at the total rowset returned from /codes. 

This PR attempts to put that info into the contract retrieve endpoint instead - that endpoint already looked like it was outputting some values like that (most notably a miscalculated `total_codes` value), so it seemed reasonable to add here.

Additionally, this PR removes unassigned codes from the response entirely. 

We no longer actually intend to show unassigned codes in the UI - contract admins shouldn't need to see them as the assign endpoints already handle distributing them automagically. The only codepath which returned them is for contracts which don't have automatic membership types AND a max_learners set (the former case never returns any codes, the latter returns the first one if it exists). We considered adding a query param to filter those out, but as this endpoint already has a bit too many edge cases, we thought this might make it easier to reason about - this makes /codes more like "used/assigned codes"

If we decide later that we do need them in this payload, we can either add a kwarg or create a new endpoint for "available codes".

Lastly, this PR adds filtering on the "status" of a code for the /codes endpoint. Since we've removed unassigned codes from the payload, we only have to worry about support for "assigned" and "redeemed" codes

Screenshot 2026-06-25 at 2 26 44 PM
Screenshot 2026-06-25 at 2 26 58 PM

### How can this be tested?
In addition to all tests passing, manual testing can be performed in the following manner.

Common setup
- With a b2b contract set up, assign a few codes using a few different emails and users. You can get the list of codes by visiting http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/, and it's easiest to assign the codes using the [swagger UI for /assign](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_codes_assign_create)

To test the filtering behavior, you'll want to redeem a few codes using the [attach API](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_attach_create).

At the end of setup, you should have at least a few values in redeemed, assigned and unassigned to make subsequent steps easier.

#### Removal of Unassigned codes from /codes
- As long as you've got at least one code that has not been assigned or redeemed, simply hit `http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/` and observe that it is not returned in the resultset.
  - If you've got a lot of codes, in assigned/redeemed state, you can specify a very large page size to get it all in one request for verification

#### Summary Statistics
- Visit http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/
- You should observe that the following fields are populated in the response and are reflective of your contract. The only one which is not self-explanatory is total_codes, and that should match the max_learners set on the contract itself.
```
total_codes
assigned_codes
unassigned_codes
redeemed_codes
```

#### Status filtering
- Visit `http://mitxonline.odl.local:9080/api/v0/b2b/manager/organizations/<org_id>/contracts/<contract_id>/codes/` and specify a status parameter of either "assigned" or "redeemed"
- You should observe that only codes in the matching status are returned.
- Additionally, if you specify a search_term AND a status parameter, both filters should be properly applied.

### Additional Information
Take note - this PR is a branch off of https://github.com/mitodl/mitxonline/pull/3685; in order to simplify the release, this branch will be merged into that PR and the whole thing will be released in one shot.

### DIFF
--- a/b2b/serializers/v0/manager.py
+++ b/b2b/serializers/v0/manager.py
@@ -4,7 +4,12 @@ from datetime import datetime
 
 from rest_framework import serializers
 
-from b2b.models import REDEMPTION_STATUS_UNASSIGNED, ContractPage
+from b2b.models import (
+    REDEMPTION_STATUS_ASSIGNED,
+    REDEMPTION_STATUS_REDEEMED,
+    REDEMPTION_STATUS_UNASSIGNED,
+    ContractPage,
+)
 from b2b.serializers.v0 import ContractPageSerializer
 from courses.models import CourseRun, CourseRunEnrollment
 from ecommerce.models import Discount
@@ -16,6 +21,9 @@ class ManagerContractDetailSerializer(ContractPageSerializer):
     attachment_percentage = serializers.SerializerMethodField()
     total_enrollments = serializers.SerializerMethodField()
     total_codes = serializers.SerializerMethodField()
+    assigned_codes = serializers.SerializerMethodField()
+    unassigned_codes = serializers.SerializerMethodField()
+    redeemed_codes = serializers.SerializerMethodField()
 
     class Meta:
         model = ContractPage
@@ -24,13 +32,35 @@ class ManagerContractDetailSerializer(ContractPageSerializer):
             "attachment_percentage",
             "total_enrollments",
             "total_codes",
+            "assigned_codes",
+            "unassigned_codes",
+            "redeemed_codes",
         ]
         read_only_fields = [
             *ContractPageSerializer.Meta.read_only_fields,
             "attachment_percentage",
             "total_enrollments",
             "total_codes",
+            "assigned_codes",
+            "unassigned_codes",
+            "redeemed_codes",
         ]
+
+    def _get_codes_breakdown(self, obj) -> dict:
+        from b2b.views.v0.manager import is_redeemed_attachment_record  # noqa: PLC0415
+
+        if not hasattr(obj, "_codes_breakdown_cache"):
+            redemptions = obj.prefetched_code_redemptions
+            redeemed = sum(1 for r in redemptions if is_redeemed_attachment_record(r))
+            assigned = len(redemptions) - redeemed
+            total = obj.get_discounts().distinct()[: obj.max_learners].count()
+            obj._codes_breakdown_cache = {  # noqa: SLF001
+                "total": total,
+                REDEMPTION_STATUS_ASSIGNED: assigned,
+                REDEMPTION_STATUS_UNASSIGNED: total - assigned - redeemed,
+                REDEMPTION_STATUS_REDEEMED: redeemed,
+            }
+        return obj._codes_breakdown_cache  # noqa: SLF001
 
     def get_attachment_percentage(self, obj) -> float | None:
         """Calculate attachment percentage if seat-limited."""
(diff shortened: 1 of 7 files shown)


b2b/views/v0/manager.py
```diff
@@ -224,6 +209,17 @@ def get_queryset(self):
                 organization__organization_users__is_manager=True,
             )
         )
+        if self.action == "retrieve":
```
dsubak, line 209: Currently only used for the contract retrieve endpoints where we calculate summary stats.

b2b/views/v0/manager.py
```diff
@@ -382,17 +378,6 @@ def codes(self, request, **kwargs):  # noqa: ARG002
 
                 codes_for_output = discounts.filter(num_redemptions__gt=0).all()
```
daniellefrappier18, line 387: I tested this out and it appears that the paginated response count is still returning the total discounts count instead of the filtered assigned/redeemed count. The frontend computes totalPages = ceil(count / page_size) which gives extra pages that render empty. Can you verify that paginate_queryset(codes_for_output) is running count() against the filtered codes_for_output queryset and not the unfiltered discounts queryset?

b2b/serializers/v0/manager.py
```diff
@@ -24,14 +32,36 @@ class Meta:
...
             "attachment_percentage",
             "total_enrollments",
             "total_codes",
+            "assigned_codes",
+            "unassigned_codes",
+            "redeemed_codes",
         ]
 
+    def _get_codes_breakdown(self, obj) -> dict:
+        from b2b.views.v0.manager import is_redeemed_attachment_record  # noqa: PLC0415
```
jkachel: This is a pretty simple function - maybe it should just move to `utils.py` to avoid this sort of import?

b2b/views/v0/manager.py
dsubak: Lemme take a look - conceptually that should be fine (if not, we'd see it for anything with a filter on it, and I am sure I tested that in the previous PR) but very possible I've missed something.

dsubak: Image

The paginated response count looks right to me - I do indeed have 5 assigned/redeemed codes on my local for the specified contract.

Here's the summary statistics on the same contract for comparison, reflecting the same 5 records in assigned/redeemed and the remainder in unassigned.
Image

b2b/serializers/v0/manager.py
dsubak: Done!

b2b/views/v0/manager.py
dsubak: Spoke w/ Danielle offline and we can't repro, think we're good to resolve this as everything looks shipshape!

dsubak:
@jkachel and @daniellefrappier18 - pulling this out of draft and into ready for review!

Ya'll obviously both already looked at it a little - the only material change since then is to add the ability to filter on status. Let me know if there are any questions!

daniellefrappier18 (approved):
Tested this out in https://github.com/mitodl/mit-learn/pull/3530 and everything is working as expected on my end. :)

jkachel (approved):
LGTM 👍

PR #3706, opened 2026-06-26
Add CyberSource Export Compliance screening to enrollment

annagav:
Closes https://github.com/mitodl/hq/issues/11991

Add CyberSource Export Compliance screening to course enrollment.

Required fields https://developer.cybersource.com/library/documentation/dev_guides/Verification_Svcs_SO_API/html/Topics/Request_Fields.htm

### How can this be tested?
```
##### Testing CyberSource export compliance
from django.core.exceptions import ValidationError

from compliance.api import verify_user_with_exports
from courses.api import create_program_enrollments
from courses.models import Program
from users.models import LegalAddress

EMAIL = <your user email>

User = get_user_model()
user = User.objects.get(email=EMAIL)

addr, _ = LegalAddress.objects.get_or_create(user=user)

program = Program.objects.filter(live=True).first()
if not program:
    raise RuntimeError("No live program found to test with.")

print("User:", user.id, user.email, user.name)
print("Legal address:", addr.country, addr.state)
print("Program:", program.id, program.title, program.readable_id)

print("\n--- Direct CyberSource check ---")
result = verify_user_with_exports(user)
print("accepted   =", result.accepted)
print("decision   =", result.decision)
print("reason_code=", result.reason_code)
print("request_id =", result.request_id)

### CODE REVIEW

File: compliance/api.py [hunk 1]
```diff
@@ -0,0 +1,284 @@
...
+    )
+except ModuleNotFoundError as exc:
+    VerificationApi = None
+    Riskv1exportcomplianceinquiriesOrderInformation = None
+    Riskv1exportcomplianceinquiriesOrderInformationBillTo = None
+    Riskv1liststypeentriesClientReferenceInformation = None
+    ValidateExportComplianceRequest = None
+    _CYBERSOURCE_IMPORT_ERROR = exc
+else:
+    _CYBERSOURCE_IMPORT_ERROR = None
```
rhysyngsun: Under what scenarios would an import fail, shouldn't the library just be installed?

compliance/api.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+def _build_export_payload(user) -> Any:
+    """Build the CyberSource export compliance REST request payload."""
+    _require_cybersource_sdk()
+
+    try:
+        legal_address = user.legal_address
+    except ObjectDoesNotExist:
+        legal_address = None
```
rhysyngsun, line 135: In this case if we have no legal address we should fail the operation and this should get bubbled up to the user as an error to either fix the data or contact support.

courses/api.py
```diff
@@ -350,6 +355,21 @@ def create_program_enrollments(
...
+    if result.accepted:
+        return
+
+    message = (
+        "Export compliance check did not accept enrollment for "
+        f"user={user.id}: "
+        f"decision={result.decision!r}, reason_code={result.reason_code!r}"
+    )
+    log.warning(message)
+    raise ValidationError(message)
```
rhysyngsun: I think this should be raising a custom error type that higher level code (views, tasks, etc) catch and log / return errors to the user.

rhysyngsun (changes requested):
Good start, but I think we need to address how code like views and any tasks that call this code handle it. So there should also be accompanying tests that verify the API responses when a user is not permitted so the frontend can display an appropriate message to the user.

There should also be a feature flag for turning this on.

compliance/api.py
annagav: With descriptive message on what is missing?

PR #3707, opened 2026-06-26
Move custom querysets/managers to be near model

rhysyngsun:

N/A

It's a pain to navigate all over models.py for querysets definitions, it makes more sense for them to be local to the model definition instead of all at the top of this file.

### How can this be tested?

Tests should pass


annagav (approved):
👍

PR #3714, opened 2026-06-30
Release 1.158.5

odlbot:
## James Kachel
  - [ ] Remove integration-specific options from contracts (#3704) ([79f46ecd](../commit/79f46ecd61cc4de5b69e569ae0ec5d159b226431))

## Dan Subak
  - [x] Use more realistic URL in test send email task (#3710) ([6b40386f](../commit/6b40386f0f15ccef47fde196b4019da83c6088ac))

PR #3715, opened 2026-06-30
Fix typo in migration that removes the non-sso type

jkachel:

n/a

Noticed after testing https://github.com/mitodl/mitxonline/pull/3714 that the non-sso contracts on RC weren't changed. The migration uses `nonsso`, which is incorrect; this fixes the migration.

### How can this be tested?

_This ended up being a bit more involved._

Before checking out this branch or updating, fake revert back to 0024: `migrate b2b 0024 --fake`. 

Then, add back the `integration_type` field to the database. Fortunately this was a simple field, so the SQL to do that is also pretty simple:

```sql
alter table b2b_contractpage add integration_type varchar(255);
```

This can be run in a `dbshell`. This does make the field slightly differently than it was originally - it should be _nullable_ at this point, which is fine.

Now, update this branch and run the migration forwards. It should work, and you should be able to reverse it too. There is new intermediary migration that makes integration_type nullable, so we can put the field back and backfill it if we do need to reverse out this migration again.


dsubak (approved):
This is a straightforward correction of a string constant, so happy to approve it on that basis.

I will note that I have not been able to test this against the noted instructions atm (see [this thread](https://mitodl.slack.com/archives/C02649X2P1V/p1782830807842839?thread_ts=1782830476.851689&cid=C02649X2P1V) for details) but that is only an issue for reverts.

dsubak (approved):

PR #3719, opened 2026-07-01
fix: issues with certificate generation

asadali145:

Fixes https://github.com/mitodl/hq/issues/11914

- Adds fixes around course and program certificate creation.
- Aligns program certificate creation with course certificate creation to use the overridden save call.
- Adds logging for missing CMS pages.
- Reports failed certificates
- Continue certificate creation if it fails for a user
- Adds a command to check missing CMS pages.

### How can this be tested?

- Test task generate_course_certificates does not fail if CMS page is missing for an eligible course run.
  - It should create a certificate with empty revision.
  - It should process next run with CMS pages for course and program and should generate the certificates.
  - There should be no business logic changes.
- Test program certificates task generate_program_certificates works as expected like course certificates task.
- Test the new command by creating all the data that makes a user eligible for a certificate and then delete the course/program CMS page or certificate page, it should report those.

### DIFF
--- a/courses/api.py
+++ b/courses/api.py
@@ -15,7 +15,7 @@ import requests
 import reversion
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.db import IntegrityError, transaction
 from django.db.models import Exists, OuterRef, Prefetch, Q
 from django_countries import countries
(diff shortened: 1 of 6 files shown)


jkachel (approved):
Seems fine - everything worked as expected with some test courses/programs I created with varying amounts of certificate eligibility. (My testing was a bit limited because my local edX instance is broken somehow - but I ran the generation functions manually and they responded as expected, which is the important part.)

PR #3720, opened 2026-07-01
add the about section to the apis

abeglova:
https://github.com/mitodl/hq/issues/12143

This is a first attempt to add cms data to the mitxonline api so that it can be added to search. Based on the comment, creating a dedicated api makes more sense


cms/serializers.py
```diff
@@ -10,13 +10,34 @@
...
+class RemovableLongFieldsMixin:
+    """
+    Serializer mixin that drops the long rich-text page fields (see
+    ``LONG_PAGE_FIELDS``) from the output when ``remove_long_page_fields`` is
+    set to a truthy value in the serializer context.
+    """
+
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        if self.context.get("remove_long_page_fields"):
```
rhysyngsun, line 27: We shouldn't be having this dynamically alter the return type on the fly - it causes us to break the type signature definied in the OpenAPI spec, which could cause bugs to sneak into frontend code that would otherwise be easily caught on type checking. I think we can either a) just leave the fields in the response or b) if these fields are needed for ingestion and ingestion _only_, write a dedicated API for them. James has already carved out a space for this in `courses/views/internal`.


Varying the return type based on `context` is also generally an antipattern that we somehow got into, I'm having to removed it as part of the performance optimization work I'm doing and it's resulting in me having to touch a massive amount of code to correct so we should avoid further usages of this.

rhysyngsun (changes requested):

cms/serializers.py
abeglova: Ok - that makes sense. Maybe it makes sense to have  CoursePage/ ProgramPage data apis then instead of trying to add the cms fields into the existing api

abeglova:
Based on Nathan's comment this is the wrong way to do this. I will create a dedicated api for the cms page data instead. 
Something like https://mitxonline.mit.edu/api/v2/programs/program id/page_data

PR #3722, opened 2026-07-01
Handle no max learners case

dsubak:
https://mit-office-of-digital-learning.sentry.io/issues/7588007411/

Contracts w/o max_learners set present an edge case for the b2b admin dash APIs. In that case /codes presents up to 1 code (if one exists), and the `contract/<id>/` endpoint blows up trying to calculate a the summary statistics of total/assigned/redeemed/unassigned codes because it can't do math against a max_learners of None.

This change makes it at least not explode. /codes is unchanged, but the summary statistics if max_learners is None or 0 will result in the following summary statistics
```
total -> 0
assigned -> number of non-redeemed DiscountContractAttachmentRecords
redeemed -> number of redeemed DiscountContractAttachmentRecords
unassigned -> 0
```

It may be desirable to do something different, like change the treatment of summary statistics entirely if we have a contract with a max_learners of 0 or None, but I think that will require some coordination with my frontend counterparts

Screenshot 2026-07-01 at 5 44 03 PM

### How can this be tested?
- Set a b2b contract to have max_learners == None.
- Visit the admin page for that contract (`http://learn.odl.local:9080/organization/<org>/contract/<contract>/admin`)
- Observe that the page doesn't throw a 500.

### Additional Context
I kind of expect to throw this code out or substantially revise both the summary stats and the codes we return - as it stands this lets the page render, but the numbers are confusing and the table is bordering on pointless.

### DIFF
--- a/b2b/serializers/v0/manager.py
+++ b/b2b/serializers/v0/manager.py
@@ -55,11 +55,17 @@ class ManagerContractDetailSerializer(ContractPageSerializer):
             redemptions = obj.prefetched_code_redemptions
             redeemed = sum(1 for r in redemptions if is_redeemed_attachment_record(r))
             assigned = len(redemptions) - redeemed
-            total = obj.max_learners
+            total = obj.max_learners if obj.max_learners else 0
+            # Unclear if this is proper treatment for no max_learners
+            # If max_learners == None, right now, this will show:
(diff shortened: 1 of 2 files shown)


daniellefrappier18 (approved):
Tested out with 0 max and no max and both work

Image

Image

PR #3723, opened 2026-07-02
fix: update the beat task timing for retry_failed_edx_enrollments

marslanabdulrauf:
https://github.com/mitodl/hq/issues/11918

This pull request updates the scheduling of the `retry-failed-edx-enrollments` Celery task to avoid conflicts with another task that shares the same edX API throttle bucket. The task will now run at :10 and :40 past each hour, ensuring a 10-minute gap from the `sync_courseruns_data` task.

**Task scheduling improvements:**

* Updated the `retry-failed-edx-enrollments` task in `main/settings.py` to run at :10 and :40 past each hour using a crontab, preventing throttle conflicts with the `sync_courseruns_data` task that runs at :00.

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -951,10 +951,7 @@ B2B_GSHEETS_UPDATE_OFFSET = get_int(
 CELERY_BEAT_SCHEDULE = {
     "retry-failed-edx-enrollments": {
         "task": "openedx.tasks.retry_failed_edx_enrollments",
-        # Run at :10 and :40 past each hour to keep a 10-minute gap from the
-        # sync_courseruns_data task (:00), which shares edX's "staff" throttle
-        # bucket via the login_service_user token.
-        "schedule": crontab(minute="10,40"),
+        "schedule": RETRY_FAILED_EDX_ENROLLMENT_FREQUENCY,
     },
     "update-currency-exchange-rates-every-24-hrs": {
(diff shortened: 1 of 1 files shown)


asadali145 (changes requested):
I think this might still not be enough to fix the throttling issue due to the limits defined and the way we are using (to enroll learners as a staff user). It will still apply 40/min for course list and 120/minute for the enrollment view. 

I see that these are hard-coded for the views. Maybe we can upstream a setting to manage these or atleast define a scope so that different throttling buckets are used.

marslanabdulrauf:
> I think this might still not be enough to fix the throttling issue due to the limits defined and the way we are using (to enroll learners as a staff user). It will still apply 40/min for course list and 120/minute for the enrollment view.
> 
> I see that these are hard-coded for the views. Maybe we can upstream a setting to manage these or atleast define a scope so that different throttling buckets are used.

yes you are on spot, that is already being discussed here https://github.com/mitodl/hq/issues/11918#issuecomment-4874962307

main/settings.py
asadali145: Instead of the current change, I would like to change this to:
```suggestion
            minute=10,
```

asadali145 (changes requested):

asadali145 (approved):
LGTM!

PR #3725, opened 2026-07-02
Indicate no max_learners total and unassigned in API response

dsubak:
https://mit-office-of-digital-learning.sentry.io/issues/7588007411/

https://github.com/mitodl/mitxonline/pull/3722 was an interim fix for cases where a contract has no learners - it prevents the admin page from erroring, but the resulting numbers and layout are misleading and confusing. This PR changes some behaviors so we can make the admin dashboard more useful in contracts without a max_learner value set.

- We no longer special case the resultset returned by /codes for SSO contracts or those without a max_learners value. Previously, they would return either no codes (in the former case) or a single code - sensible behaviors when we showed the unredeemed codes in the UI. Since we no longer show unredeemed codes and instead rely on the backend to automatically send out codes, we can remove these special cases. Going forward, /codes will always return all assigned and redeemed codes so that admins can see who is in/invited to the contract.
- In cases where max_learners is unset, we now return None for the total_codes and unassigned values returned on the contract endpoint. If you don't have a seat limit, the total is unbounded and there's no analog for unassigned codes (since that's just the total minus assigned and redeemed)

### How can this be tested?
- Set a b2b contract to have max_learners == None. Make sure you've got at least one assigned code and one redeemed code
- Visit `http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_retrieve` for the specific contract. Observe that total and unassigned are `None`
- Visit http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_codes_list for the same contract. Observe that the response contains all assigned and redeemed rows associated with the contract.

### Additional Context
Making the fields nullable constitutes a breaking change to the API - we should coordinate w/ the subsequent frontend PR to minimize downtime.

### DIFF
--- a/b2b/serializers/v0/manager.py
+++ b/b2b/serializers/v0/manager.py
@@ -55,17 +55,15 @@ class ManagerContractDetailSerializer(ContractPageSerializer):
             redemptions = obj.prefetched_code_redemptions
             redeemed = sum(1 for r in redemptions if is_redeemed_attachment_record(r))
             assigned = len(redemptions) - redeemed
-            total = obj.max_learners if obj.max_learners else 0
-            # Unclear if this is proper treatment for no max_learners
-            # If max_learners == None, right now, this will show:
-            # total == 0, assigned == # of assigned codes, unassigned == 0, redeemed == # of redeemed codes
-            # Meanwhile /codes will show the first code, if it exists.
+            total = obj.max_learners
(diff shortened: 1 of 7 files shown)


jkachel (approved):
LGTM 👍

PR #3728, opened 2026-07-02
fix(openedx): stop retry_failed_edx_enrollments from DoSing the edX enrollment API

feoh:

Fixes https://github.com/mitodl/hq/issues/12132 (Sentry `MITXONLINE-2NS`, 310k events since 2023-03-09).

`retry_failed_edx_enrollments` was iterating over every failed `CourseRunEnrollment` in a tight loop with no batching, no age cap, and no give-up. Every row cost 1–3 API calls against the shared `login_service_user`, which is `is_staff=True` in the Open edX LMS and therefore capped at **120/min** by [`EnrollmentUserThrottle`](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/enrollments/views.py#L99). A backlog of ~100 stuck rows produced a burst of 200+ requests every 30 minutes and pinned the throttle bucket, so every row 429'd and none ever cleared — a self-perpetuating loop.

Live evidence from the RC/QA celery worker on 2026-07-02:

```
Task openedx.tasks.retry_failed_edx_enrollments[…] succeeded in 44.63s: []
Response - code: 429, content: {"developer_message":"Request was throttled. Expected available in 23 seconds."}
Response - code: 429, content: {"developer_message":"Request was throttled. Expected available in 18 seconds."}
…
```

The task “succeeded in 44s” with an empty result list — 100% 429s. The oldest stuck row was from `2023-10-06` (user `sd-at-mitx`); most others are automation/test residue from `pdpinch+*`, `apptest+*`, `gumaerc@mit.edu`, etc.

This change bounds the retry task three ways, all env-tunable:

- **Age cap** — `RETRY_FAILED_EDX_ENROLLMENT_MAX_AGE_DAYS` (default `7`, `0` disables). Rows whose `created_on` is older than the cap are dropped from the retry set. A 3-year-old failed enrollment is not going to enroll now, and it is the source of the throttle storm.
- **Batch size** — `RETRY_FAILED_EDX_ENROLLMENT_BATCH_SIZE` (default `30`). Only the oldest-N eligible rows are attempted per invocation. With the 30-min beat, a healthy backlog still drains in a couple of passes but never bursts past the 120/min staff bucket.
- **POST-only retry path** — the retry loop no longer calls `enroll_in_edx_course_runs`; it calls `create_student_enrollment(force_enrollment=True)` directly. That endpoint is idempotent server-side, so we skip the pre-flight `GET /api/enrollment/v1/enrollments` and halve the API cost per row.

Also: a `429` from Open edX during retry is now logged at `INFO` (not `exception`) and lets the row roll into the next pass, so Sentry `MITXONLINE-2NS` stops re-firing on every beat.

`enroll_in_edx_course_runs` (used by checkout, admin actions, refunds, and management commands) is **intentionally untouched** — the behavior change is contained to the retry task.

### How can this be tested?

Unit tests cover each bound:

- `test_retry_failed_edx_enrollments` — POST-only path, per-row error is swallowed, successes flip `edx_enrolled=True`.
- `test_retry_failed_enroll_grace_period` — rows younger than `OPENEDX_REPAIR_GRACE_PERIOD_MINS` are still skipped.
- `test_retry_failed_enroll_age_cap` — rows older than the age cap are skipped and `edx_enrolled` stays `False`.
- `test_retry_failed_enroll_batch_size` — only `BATCH_SIZE` rows are attempted per pass, oldest first.
- `test_retry_failed_enroll_throttled_is_soft_failure` — a 429 does not raise, does not flip `edx_enrolled`, and logs at `info` instead of `exception`.

Run: `pytest openedx/api_test.py -k retry_failed`

Manual validation on RC after deploy:

1. Watch `mitxonline-celery-celery-worker` logs during the next `retry_failed_edx_enrollments` invocation — should see ≤ `BATCH_SIZE` attempts, no burst.
2. Grep for `429 Client Error` in celery logs — should drop to zero, or at most a small number that clear on subsequent passes.
3. Check Sentry `MITXONLINE-2NS` — new event rate should fall off a cliff.

### Additional Context

**Ops follow-up (not in this PR):**

- The ~100 stuck rows in RC will be silently dropped from the retry set on the next pass thanks to the age cap; they can then be triaged out-of-band. Nothing here deletes them.
- Longer term, giving the service worker its own throttle scope (`service_user`/`high_service_user`) on the LMS side, or splitting the service worker per calling app, would let us stop sharing a 120/min bucket between the retry loop and real-time checkout enrollments. Separate change.

**Diagnosis notes** for future readers:

- The throttle is *per authenticated user*, and `login_service_user` is `is_staff=True` in the LMS, so it gets the `staff: 120/minute` rate, not the higher `service_user: 800/minute` or `high_service_user: 2000/minute` scopes defined in `REST_FRAMEWORK.DEFAULT_THROTTLE_RATES`. `EnrollmentUserThrottle` ignores those; the rates are hard-coded in the view.
- `existing_edx_enrollment` was making a `GET /api/enrollment/v1/enrollments` per row *before* the POST. Under throttle pressure that GET returned 429 and the code path bailed without ever POSTing, so a row that would have succeeded via `force_enrollment=True` never got the chance.

### DIFF
--- a/main/settings.py
+++ b/main/settings.py
@@ -909,6 +909,24 @@ RETRY_FAILED_EDX_ENROLLMENT_FREQUENCY = get_int(
     default=60 * 30,
     description="How many seconds between retrying failed edX enrollments",
 )
+RETRY_FAILED_EDX_ENROLLMENT_BATCH_SIZE = get_int(
+    name="RETRY_FAILED_EDX_ENROLLMENT_BATCH_SIZE",
+    default=30,
+    description=(
+        "Max failed enrollments retried per invocation of "
+        "retry_failed_edx_enrollments. Keeps the task below Open edX's per-user "
+        "enrollment throttle (120/min for staff service workers)."
+    ),
+)
+RETRY_FAILED_EDX_ENROLLMENT_MAX_AGE_DAYS = get_int(
+    name="RETRY_FAILED_EDX_ENROLLMENT_MAX_AGE_DAYS",
+    default=7,
+    description=(
+        "Failed CourseRunEnrollments older than this are no longer retried, so a "
+        "permanently-broken row cannot keep the retry task DoSing the Open edX "
+        "enrollment API forever. Set to 0 to disable the age cap."
+    ),
+)
 REPAIR_OPENEDX_USERS_FREQUENCY = get_int(
     name="REPAIR_OPENEDX_USERS_FREQUENCY",
     default=60 * 30,
--- a/openedx/api.py
+++ b/openedx/api.py
@@ -1223,35 +1223,118 @@ def enroll_in_edx_course_runs(
 
 def retry_failed_edx_enrollments():
     """
-    Gathers all CourseRunEnrollments with edx_enrolled=False and retries them via the edX API
+    Gathers CourseRunEnrollments with edx_enrolled=False and retries them via the
+    edX API.
+
+    Retries are bounded in three ways so a permanently-broken row cannot pin
+    Open edX's per-user enrollment throttle (staff = 120/min):
+
+      * an age cap (settings.RETRY_FAILED_EDX_ENROLLMENT_MAX_AGE_DAYS) drops
+        rows that have failed for so long they are almost certainly unrecoverable;
+      * a batch size (settings.RETRY_FAILED_EDX_ENROLLMENT_BATCH_SIZE) limits
+        how many rows we hit per invocation;
+      * rows are retried with a single POST to the enrollment endpoint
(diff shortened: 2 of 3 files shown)


feoh:
Closing in favor of #3726, which addresses the same throttle-storm root cause and also cleans up the existing backlog of stuck rows. Diagnosis notes for context: https://github.com/mitodl/hq/issues/12132

feoh:
Superseded by #3726.

PR #3734, opened 2026-07-06
JIT discount provisioning for /bulk_assign

dsubak:
https://github.com/mitodl/hq/issues/12207

This changes the the behavior of bulk_assign when used on a codes contract that does not have a max_learners limit set. If that endpoint is hit with a payload that contains more than the number of available (i.e. not redeemed or assigned) discount codes, it will now provision new Discount objects on the fly using the first product associated with the contract.

### How can this be tested?
The approach I took is as follows:

With main checked out:
- Create a b2b contract with discount codes, and ensure you're a manager for the containing organization. 
- Assign all existing codes to any email addresses you like either via the learn b2b admin page or by directly using the [bulk_assign swagger endpoint](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_codes_bulk_assign_create). 
  - The codes don't need to be redeemed, so the emails can be anything valid.
- Set max_learners = None for this contract within the shell.
```
from b2b.models import *
contract = ContractPage.objects.get(id=<ID_of_contract>)
contract.max_learners = None
contract.save()
```
- Attempt to assign one or more contacts and observe that you receive `No available code.` errors.
- Switch to this branch, restart your container, reauthenticate if necessary and attempt to assign one or more codes to new emails again. This time you should see that you no longer get an error and the new assignments should be visible on the admin dashboard or via the /codes endpoint.

### Additional Context
- James has put together a ticket which would see us further invest in the idea of moving away from pregenerated codes in favor of JIT discount code generation at https://github.com/mitodl/hq/issues/12211

### DIFF
--- a/b2b/views/v0/manager.py
+++ b/b2b/views/v0/manager.py
@@ -110,18 +110,21 @@ def assign_codes_and_send_emails(
     return True
 
 
-def _create_discount_codes_for_contract(contract: ContractPage) -> list[Discount]:
+def _create_discount_codes_for_contract(
+    contract: ContractPage, count_to_provision: int
+) -> list[Discount]:
     # TODO: I'm going to need to check this behavior. # noqa: TD002, TD003, FIX002
     # It looks like in trivial cases we'll have a contract with a single product, but that's far from guaranteed
     # We don't surface the product anywhere in the UI - it's all about codes, which kinda implies that this tool works
     # sanely only for contracts w/ a single product? Not sure if that's a safe assumption,
     # but if it isn't we probably need to take a step back and figure out what to do.
     new_discounts = []
-    for product in contract.get_products():
-        discount = _create_discount_with_product(
-            product, Decimal(0), REDEMPTION_TYPE_ONE_TIME
-        )
-        new_discounts.append(discount)
+    for _ in range(count_to_provision):
+        for product in contract.get_products():
+            discount = _create_discount_with_product(
+                product, Decimal(0), REDEMPTION_TYPE_ONE_TIME
+            )
+            new_discounts.append(discount)
 
     return new_discounts
 
@@ -704,7 +707,10 @@ class ManagerContractViewSet(NestedViewSetMixin, viewsets.ReadOnlyModelViewSet):
             email_assignees
         ):
             # If we are in a contract without a seat limit, provision new discounts for users up to the number required
-            new_discounts = _create_discount_codes_for_contract(contract)
+            count_to_provision = len(email_assignees) - len(available_discounts)
+            new_discounts = _create_discount_codes_for_contract(
+                contract, count_to_provision
+            )
             available_discounts.extend(new_discounts)
 
         for i, record in enumerate(email_assignees):
--- a/b2b/views/v0/manager.py
+++ b/b2b/views/v0/manager.py
@@ -113,18 +113,16 @@ def assign_codes_and_send_emails(
 def _create_discount_codes_for_contract(
     contract: ContractPage, count_to_provision: int
 ) -> list[Discount]:
-    # TODO: I'm going to need to check this behavior. # noqa: TD002, TD003, FIX002
-    # It looks like in trivial cases we'll have a contract with a single product, but that's far from guaranteed
-    # We don't surface the product anywhere in the UI - it's all about codes, which kinda implies that this tool works
-    # sanely only for contracts w/ a single product? Not sure if that's a safe assumption,
-    # but if it isn't we probably need to take a step back and figure out what to do.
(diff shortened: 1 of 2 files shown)


b2b/views/v0/manager.py
```diff
@@ -107,6 +110,25 @@ def assign_codes_and_send_emails(
...
+    # but if it isn't we probably need to take a step back and figure out what to do.
+    new_discounts = []
+    for _ in range(count_to_provision):
+        for product in contract.get_products():
+            discount = _create_discount_with_product(
+                product, Decimal(0), REDEMPTION_TYPE_ONE_TIME
+            )
+            new_discounts.append(discount)
+
+    return new_discounts
```
dsubak, line 127: This approach "works" - I was able to get bulk_assign to function even when we had no codes available to link everything to. That said, I don't understand how we're supposed to be using products in these contexts as a result of looking into this, and this _technically_ provisions more codes than is necessary if a contract has more than one product.

When I checked RC and prod, it looked like a contract w/ many products is fairly common but when we query for codes to show or assign, we basically say "get me codes for any product associated with this contract" and make no distinction visually or during automatic assignment. Is that something we should be aware of?

I get that when we use one of these codes we add the user to the org and the contract, but does it matter what product(s) this discount is tied to?

@jkachel sorry to keep bugging you, but do you have any insight into the setup that could point me in the right direction? I'd appreciate it if so!

jkachel: I'm not sure we've gone over how the B2B system and ecommerce interact fully.. that is likely my fault but it's going to be more important given the functionality changes that have happened over time. 

I started typing out an overview of how this all works but it was becoming a book. But, the gist of it is:
- Yes, lots of products. Products exist for _every_ course run and program that the user can enroll in. (Some of the newer UAI contracts with variants have around 40.)
- _At least as far as this is concerned_, since the primary goal here is to create a code that someone can use to attach to the contract, and that contract specifically has no seat limit, it really doesn't matter too much; just grab one of the products and associate the discount with it.

Reasoning here is that, at least for the near term, there'll be a handful of unlimited-use discount codes that the system will use for enrollment across the contract. (For aforementioned UAI courses, if they were unlimited seat contracts, that would mean there'd be around _40_ discounts sitting out there.) But, when it processes an enrollment, it'll grab one of those more than likely, so it doesn't matter too much that there's some oddball ones floating around. They will still get used for enrollments - but if we run out of the  piecemeal ones it'll just go back to using the unlimited use discounts.

If the contract does have a seat limit, then we've already created a giant pile of codes, so we can just grab an unused code.

It would probably be helpful to look at what we do for created code validation. There's APIs to make sure the contract has the correct number and type of discount codes created for it. I think the most relevant thing to look at is maybe [`_create_discount_with_product`](https://github.com/mitodl/mitxonline/blob/f51c6bfa628a6fe0493ec0dd04e06ee15260f852/b2b/api.py#L838) which is a part of the [`ensure_enrollment_codes_exist`](https://github.com/mitodl/mitxonline/blob/f51c6bfa628a6fe0493ec0dd04e06ee15260f852/b2b/api.py#L1094) workflow. We also make codes in-flight if there's seats available but no available discounts; that's over here in [`_apply_available_discount`](https://github.com/mitodl/mitxonline/blob/f51c6bfa628a6fe0493ec0dd04e06ee15260f852/b2b/api.py#L1278).

dsubak: I've changed this chunk to just grab the first product every time and use that for creating discounts! Resolving

b2b/views/v0/manager.py
```diff
@@ -661,6 +682,19 @@ def bulk_assign(self, request, **kwargs):  # noqa: ARG002
         )
 
         assignments = []
+        # If we're in a contract type that uses codes, doesnt have a seat limit and we're out of available discounts
+        # Provision enough on the fly to handle the payload
+        if (
+            contract.membership_type not in CONTRACT_MEMBERSHIP_AUTOS
+            and contract.max_learners is None
```
daniellefrappier18:
```suggestion
             and not contract.max_learners
```

`contract.max_learners is None` only recognized the blank version as "no limit" but there is also a possibility of 0 which means the same thing, right?

daniellefrappier18 (approved):
With the change I suggested both no max and 0 max work as intended

b2b/views/v0/manager.py
dsubak: Yeah, it looks like it - I thought there might be spots where we make a distinction, but it doesn't seem like that's the case. I'll change this to handle 0s too!

PR #3738, opened 2026-07-07
Enable live reload for the django app

rhysyngsun:

N/A

This reenables live reload since we switched to Granian for localdev.

### How can this be tested?

Edit a python source file and you should see the app reload with your change.


asadali145 (approved):
LGTM!

PR #3741, opened 2026-07-07
Enroll b2b tests

cp-at-mit:
NA

Adds unit tests for the APIClient class in the b2b views that previously did not have any tests.

### How can this be tested?
Tests should pass.


annagav (approved):
👍

PR #3744, opened 2026-07-09
Add some viz into misconfigured contracts

dsubak:
In testing https://github.com/mitodl/mitxonline/pull/3734/ I ran into a hiccup where things don't behave as you'd think if a contract doesn't have at least one product associated with it during JIT discount provisioning. While the discount provisioning worked, the enrollment code wasn't usable, and most other aspects of the B2B admin dashboard were also incorrect/misleading

Since [contracts without products are useless](https://mitodl.slack.com/archives/C0BEW5UHPJ8/p1783609961274279?thread_ts=1783447081.511689&cid=C0BEW5UHPJ8), rather than trying to handle issues from this setup, we'll just punt on it. This now logs an error to sentry and otherwise skips JIT provisioning. This at least means users and maintainers will have some visibility into what's going on!

### How can this be tested?
- Set up a b2b code-based contract w/ no max_learners and make yourself a manager. Preassign/redeem all codes (or delete all associated Discounts)
- Unlink the all products from the contract 
- Attempt to bulk_assign a code. You should get a `No available code.` for each assignment requested in the response.

### DIFF
--- a/b2b/views/v0/manager.py
+++ b/b2b/views/v0/manager.py
@@ -118,6 +118,12 @@ def _create_discount_codes_for_contract(
     # For contract attachment JIT code generation, any product associated w/ the contract should
     # work for creating the discount code, so we just grab the first one.
     product = contract.get_products().first()
+    if not product:
+        log.exception(
+            "Contract %s has no product, cannot provision JIT discount codes", contract
+        )
+        return new_discounts
+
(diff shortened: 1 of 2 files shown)


daniellefrappier18 (approved):
Works as advertised

Image

PR #3748, opened 2026-07-10
Filter active cert issue_date in db query

rhysyngsun:

Related to https://github.com/mitodl/hq/issues/12149

This ensures it's never possible to query for certificates with a stale filter on `issue_date`.

### How can this be tested?

This is a subtle error and hard to reproduce so tests passing should be adequate.


annagav (approved):
LGTM 👍

PR #3750, opened 2026-07-13
feat: update to latest revision command

asadali145:

Closes https://github.com/mitodl/hq/issues/12298

Adds a command to update certificates to the latest revision. Provides options for course, run, and program certificates. Updates certificates with missing revisions by default and provides `--all` to update all certificates including the ones with revisions.

### How can this be tested?

- Create test data, like some certificates for courses and programs, with some pointing to the latest revision and some to older ones. Create some certificates with no revisions (This can only be done via shell or from admin in case certificate page in cms does not exist.)
- Now run different variants of the command to test different cases.

```
1. Update certificates with no revision associated, for a course
./manage.py migrate_certificate_revisions --course=<course_readable_id>

2. Update ALL certificates (including ones that already have a revision), for a course
./manage.py migrate_certificate_revisions --course=<course_readable_id> --all

3. Same operations, but for a single course run
./manage.py migrate_certificate_revisions --courserun=<course_run_courseware_id>
./manage.py migrate_certificate_revisions --courserun=<course_run_courseware_id> --all

4. Same operations, but for a program
./manage.py migrate_certificate_revisions --program=<program_readable_id>
./manage.py migrate_certificate_revisions --program=<program_readable_id> --all

```

### DIFF
--- a/courses/management/commands/migrate_certificate_revisions.py
+++ b/courses/management/commands/migrate_certificate_revisions.py
@@ -1,0 +1,166 @@
+"""
+Management command to migrate CourseRunCertificates/ProgramCertificates for a
+course, course run, or program to the latest certificate page revision.
+
+By default, only certificates that don't have a certificate_page_revision
+associated are updated. Pass --all to update every certificate for the
+course/run/program, even ones that already have a revision set.
+
+Check the usages of this command below:
+
+1. Update certificates with no revision associated, for a course
+./manage.py migrate_certificate_revisions --course=<course_readable_id>
+
+2. Update ALL certificates (including ones that already have a revision), for a course
+./manage.py migrate_certificate_revisions --course=<course_readable_id> --all
+
+3. Same operations, but for a single course run
+./manage.py migrate_certificate_revisions --courserun=<course_run_courseware_id>
+./manage.py migrate_certificate_revisions --courserun=<course_run_courseware_id> --all
+
+4. Same operations, but for a program
+./manage.py migrate_certificate_revisions --program=<program_readable_id>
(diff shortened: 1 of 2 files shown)


courses/management/commands/migrate_certificate_revisions.py
```diff
@@ -0,0 +1,166 @@
...
+        parser.add_argument(
+            "--course", type=str, help="The 'readable_id' value for a Course"
+        )
+        parser.add_argument(
+            "--courserun", type=str, help="The 'courseware_id' value for a CourseRun"
+        )
+        parser.add_argument(
+            "--program", type=str, help="The 'readable_id' value for a Program"
+        )
+        parser.add_argument(
```
zamanafzal, line 61: `--all` performs a bulk update that can touch a large number of certificates at once, could we add a --dry-run flag? It would just report the count of certificates that would be updated (and ideally which course/run/program they belong to) without actually committing the change.

zamanafzal (approved):
Overall LGTM. Left a suggestion if you think that make sense.

Image

courses/management/commands/migrate_certificate_revisions.py
asadali145: @zamanafzal we already have a user confirmation input when you do --all, it displays the courseware title and revision ID, I have updated it to display the certificate count as well. Now user can abort the operation if needed.

asadali145: https://github.com/mitodl/mitxonline/pull/3750/changes/343ecc8fa6cfd613fa0907f92820bbae2e7b2731

PR #3752, opened 2026-07-14
Guard HubSpot associations against missing IDs

cp-at-mit:
NA

There's an issue when creating deals in hubspot for the following free program/course 
program-v1:UAI+B2C.1.  This is a UAI program/course so any deal created for it should be created in the Xpro hubspot account.  My thought is that this is happening because the product record in Hubspot for program-v1:UAI+B2C.1 was only created in the MItx Online hubspot account and not in the Xpro account.  All UAI courses and program deals are created in the Xpro account.  This results in errors when trying to create a deal and associate it with the product or contact in the Xpro account if the product or contact does not exist.

This PR ensures that the product and contact exist in hubspot, and adds some retry logic if the requests to create those records fail, before trying to associate the deal with the contact and product record.

### How can this be tested?
You can create a UAI course that does not have a hubspot product record.  Make sure that course is free to enroll.  Then attempt to enroll into the course and verify that a contact and product record are created in the Xpro dev hubspot account.

### DIFF
--- a/hubspot_sync/api.py
+++ b/hubspot_sync/api.py
@@ -2,6 +2,7 @@
 
 import logging
 import re
+import time
 from decimal import Decimal
 from typing import List  # noqa: UP035
 
@@ -1573,6 +1574,52 @@ def sync_deal_line_hubspot_ids_to_db(order, hubspot_order_id) -> bool:
     return matches == expected_matches
 
 
+_ASSOCIATION_RETRY_DELAYS = (1, 2, 4)  # seconds between attempts (3 total tries)
+
+
+def _associate_objects_with_retry(
+    from_type: str,
+    from_id: str,
+    to_type: str,
+    to_id: str,
+    assoc_type: str,
+) -> None:
+    """Call associate_objects_request with exponential-backoff retry on transient errors.
+
+    Retries up to len(_ASSOCIATION_RETRY_DELAYS) times on ApiException or
+    TooManyRequestsException before re-raising the last exception.
+    """
+    last_exc: Exception | None = None
+    for attempt, delay in enumerate(
+        (None, *_ASSOCIATION_RETRY_DELAYS), start=1
+    ):
+        if delay is not None:
+            log.warning(
+                "Retrying association (%s→%s) attempt %d after %ds",
+                from_type,
+                to_type,
+                attempt,
+                delay,
+            )
+            time.sleep(delay)
+        try:
+            wait_for_hubspot_rate_limit()
+            associate_objects_request(from_type, from_id, to_type, to_id, assoc_type)
+            return
+        except (ApiException, TooManyRequestsException) as exc:
+            last_exc = exc
+            log.warning(
+                "Association (%s %s → %s %s) failed on attempt %d: %s",
+                from_type,
+                from_id,
+                to_type,
+                to_id,
+                attempt,
+                exc,
+            )
+    raise last_exc
+
+
 def get_hubspot_id_for_object(  # noqa: C901
     obj: Order or Product or Line or User,
     raise_error: bool = False,  # noqa: FBT001, FBT002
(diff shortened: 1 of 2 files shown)


hubspot_sync/api.py
```diff
@@ -1573,6 +1574,51 @@ def sync_deal_line_hubspot_ids_to_db(order, hubspot_order_id) -> bool:
...
+    to_id: str,
+    assoc_type: str,
+) -> None:
+    """Call associate_objects_request with exponential-backoff retry on transient errors.
+
+    Retries up to len(_ASSOCIATION_RETRY_DELAYS) times on ApiException or
+    TooManyRequestsException before re-raising the last exception.
+    """
+    last_exc: Exception | None = None
+    for attempt, delay in enumerate((None, *_ASSOCIATION_RETRY_DELAYS), start=1):
```
dsubak: Should we consider pulling in something like [tenacity](https://tenacity.readthedocs.io/en/latest/)? The headline feature for that is the retry decorator, but there's also a bit of support for [retries in arbitrary code blocks](https://tenacity.readthedocs.io/en/latest/#retrying-code-block) like this

Dealer's choice though - if we don't think it's worth taking on an external dep for exponential backoff, I totally understand

hubspot_sync/api.py
```diff
@@ -1668,13 +1715,29 @@ def sync_line_item_with_hubspot(line: Line) -> SimplePublicObject:
...
+            HubspotObjectType.DEALS.value,
+            deal_hubspot_id,
+            HubspotAssociationType.LINE_DEAL.value,
+        )
+    else:
+        log.warning(
+            "No HubSpot ID found for order %d after sync; skipping line-deal association for line %d",
+            line.order.id,
+            line.id,
+        )
```
dsubak, line 1744: Nit/discussion - this block and the block that associate between deals and contacts are the same from a control-flow perspective so they look really close. AFAICT they differ in the `sync_*_with_hubspot` method they call, the types they pass into `_associate_objects_with_retry` and the contents of the formatted message.

Would it be worth trying to consolidate into a parameterized `associate_hubspot_objects` method that takes in the type of object we're associating as an arg and looks up the rest based on that?

dsubak (commented):
Still gotta walk through testing, but left a few comments for your consideration!

hubspot_sync/api.py
cp-at-mit: I think this would be a good idea to implement in the future.  Right now the retry functionality is only being added to the API calls that associate records in Hubspot.  We really should add it to all API calls to Hubspot.  I'll create an issue for this.  Probably best we implement it after we consolidate the Hubspot accounts as well.

dsubak: Works for me, lemme take a look at the code in its current form then, but we can come up w/ a more robust retry strategy once at a later date!

dsubak: This line reads a bit oddly to me - nothing wrong functionally, but I think we could clean it up slightly by:
- Making `_ASSOCIATION_RETRY_DELAYS = (0, 1, 2, 4)`
- Changing `if delay is not None` -> `if delay`
- Changing the loop to be `for attempt, delay in enumerate(_ASSOCIATION_RETRY_DELAYS, start=1):`

I _think_ that keeps the functionality the same, but it's got a bit less going on that readers need to keep in mind. 

If you're feeling fancy, we don't even really need to maintain a list of delays since we are adhering to exp backoff - we can just set a number of tries and an initial delay and calculate it based on those values as `delay = (2 ** attempt-1) if attempt > 1 else 0`

dsubak (approved):
Eventually managed to get it tested; had a few hoops to jump through - does reconcile previously unseen UAI products in the xpro sandbox

Left one stylistic comment you can take or leave, but functionality seems fine.

Image

PR #3755, opened 2026-07-15
assign_codes_and_send_emails tests

cp-at-mit:
NA

Adds tests for assign_codes_and_send_emails which was not previously covered.

### How can this be tested?
Tests should pass.  Code review.


dsubak (approved):
This looks good - if you wanna merge as is, that seems fine. 

That said, if we're going w/ unit tests for this method we might also want a test for the unhappy path. I think the only salient failure mode this method has is if the bulk_create fails, in which case we kick an exception out and return False to indicate that nothing was created - if that's easy enough to tack in, I think we'd end up w/ full coverage over this method.

PR #3756, opened 2026-07-15
ManagerOrganizationViewSet tests

cp-at-mit:
NA

Adds tests for ManagerOrganizationViewSet which previously was untested.

### How can this be tested?
Tests should pass.  Code review.

### DIFF
--- a/b2b/views/v0/manager_test.py
+++ b/b2b/views/v0/manager_test.py
@@ -21,6 +21,7 @@ from b2b.serializers.v0 import (
     BaseContractPageSerializer,
 )
 from b2b.serializers.v0.manager import ManagerEnrollmentSerializer
+from b2b.views.v0.manager import CodeAssignment, assign_codes_and_send_emails
 from courses.factories import CourseRunFactory
 from courses.models import CourseRunEnrollment
 from ecommerce.constants import REDEMPTION_TYPE_ONE_TIME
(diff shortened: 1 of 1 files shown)


b2b/views/v0/manager_test.py
```diff
@@ -1989,8 +2120,45 @@ def test_assign_codes_and_send_emails_bulk_create_failure(
...
+def test_assign_codes_and_send_emails_empty_list(org_setup, mock_email_task):
+    """An empty assignments list creates no records but still queues the task."""
+    manager_user, *_ = org_setup
+
+    result = assign_codes_and_send_emails([], manager_user)
+
+    assert result is True
+    assert not DiscountContractAttachmentRedemption.objects.exists()
+    mock_email_task.delay.assert_called_once_with([])
+
```
dsubak, line 2143: This might actually be a bug (it's at least unintentional). I'll file a ticket for that and fix it - there's no need to do any further processing if we don't have any assignments to create. It doesn't hurt anything as is, but it is pointless. We could trip this functionality if we got a payload which only included users who already had assigned codes. AFAICT, the API level behavior is correct, but 

If you could add a comment noting that this is behavior that we are going to remove, that'd be great and I'll follow up and remove/adjust this test in the PR w/ the fix!

dsubak (approved):
Looks good, I requested one very minor comment be added just to make it clear that one bit of unintentional behavior is going away Soon™ - other than that, no notes!

PR #3757, opened 2026-07-16
Tag enrollment code assignment emails via mailgun tag header

dsubak:
Small part of https://github.com/mitodl/hq/issues/12114

This adds a [mailgun tag](https://mailgun-docs.redoc.ly/docs/mailgun/user-manual/tracking-messages/#tags) for the enrollment code assignment emails sent by the B2B dashboard and retrieves/generates a [message_id](https://www.mailgun.com/glossary/message-id/) for anything sent

The forthcoming functionality to store email deliverability status for display depends on one of two approaches:
- Calling the Mailgun events API and filtering down to emails we care about in a downstream task. This is what xpro does, though as email delivery is asynchronous and unpredictable, this raises questions about how close to real time the data would be available.
- Setting up webhooks and processing events for emails we care about. This has the potential to be closer to real time than pulling events on a cadence, at the cost of needing to process spiky workloads asynchronously

In either case we need to be able to store the message id we sent so that when we get an event in either manner, we are able to unambiguously tie it back to the email we sent. If we don't have that value and a user, for example, received multiple enrollment code emails, we wouldn't know for sure which one an event corresponded to.

In a subsequent PR, we'll add fields to the DiscountContractAttachmentRedemption records to persist these values for lookup when we process deliverability events, and the values we get back from those event payloads will be used to update email deliverability status fields on the corresponding record.

Screenshot 2026-07-16 at 11 49 43 AM

### How can this be tested?
- Set up mailpit and set MITOL_MAIL_CONNECTION_BACKEND='django.core.mail.backends.smtp.EmailBackend'
- Set MITX_ONLINE_EMAIL_HOST and MITX_ONLINE_EMAIL_PORT to wherever mailpit is running.
- For a b2b contract you are an admin for, assign a code.
- In mailpit you should get an invitation, and if you look at its headers there should be a `X-Mailgun-Tag: enrollment-code-assignment` header.

Mailgun can't be as easily tested, so we'll verify that in CI/RC. That said I did run the following snippet in RC to validate the behavior of how we pull the message_id from the sent message:
```
from mitol.mail.api import get_message_sender

from b2b.mail import EnrollmentCodeAssignmentMessage, get_learn_hostname

# EDIT ME before running
TEST_RECIPIENT = "dansubak+testingmessageid@mit.edu"
TEST_CODE = "TESTCODE123"
TEST_ORGANIZATION_NAME = "Test Organization"
TEST_CONTRACT_NAME = "Test Contract"

learn_hostname = get_learn_hostname()
code_url = f"https://{learn_hostname}/enrollmentcode/{TEST_CODE}"

with get_message_sender(EnrollmentCodeAssignmentMessage) as sender:
    message = sender.build_message(
        TEST_RECIPIENT,
        {
            "code": TEST_CODE,
            "code_url": code_url,
            "organization_name": TEST_ORGANIZATION_NAME,
            "contract_name": TEST_CONTRACT_NAME,
        },
    )
    message.send()

    print("anymail_status:", repr(message.anymail_status))
    recipient_status = message.anymail_status.recipients.get(TEST_RECIPIENT)
    message_id = recipient_status.message_id if recipient_status else None
    print(f"mailgun message_id for {TEST_RECIPIENT}: {message_id}")
```

### DIFF
--- a/b2b/mail.py
+++ b/b2b/mail.py
@@ -1,4 +1,5 @@
 import logging
+import uuid
 
 from django.conf import settings
 from mitol.mail.api import get_message_sender
@@ -32,7 +33,7 @@ def get_learn_hostname():
 def send_email_helper(email, code, code_url, organization_name, contract_name):
     try:
         with get_message_sender(EnrollmentCodeAssignmentMessage) as sender:
-            sender.build_and_send_message(
+            message = sender.build_message(
                 email,
                 {
                     "code": code,
@@ -41,6 +42,21 @@ def send_email_helper(email, code, code_url, organization_name, contract_name):
                     "contract_name": contract_name,
                 },
             )
+            # send_message swallows exceptions, so we'll favor direct calls to message.send()
+            message.send()
+
+            # The message_id is primarily to be used to tie back to mailgun webhook events unambiguously
+            # In the case of a local SMTP server the message_id will be none so we just generate a UUID
+            if (
+                settings.MITOL_MAIL_CONNECTION_BACKEND
+                == "anymail.backends.mailgun.EmailBackend"
+            ):
+                recipient_status = message.anymail_status.recipients.get(email)
+                message_id = recipient_status.message_id if recipient_status else None
+            else:
+                message_id = str(uuid.uuid4())
+
+            return message_id
     except:  # pylint: disable=bare-except  # noqa: E722
         log.exception("Error sending enrollment code assignment email.")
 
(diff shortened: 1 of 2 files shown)


daniellefrappier18 (approved):
Tested locally against Mailpit and sent an enrollment code assignment email via send_email_helper and confirmed the X-Mailgun-Tag: enrollment-code-assignment header is present on the outgoing message.

Image

PR #3760, opened 2026-07-17
Short circuit empty list

dsubak:
In reviewing a set of tests written for some of our b2b admin dash API endpoints and helper methods, we found that we're [dispatching pointless tasks](https://github.com/mitodl/mitxonline/pull/3756#discussion_r3603800257) if a payload has everything filtered out. This can happen due to all emails already having an assigned or redeemed code - the response is correct, but we're doing more work than we need to.

This PR simply short circuits it. Once the abovelinked PR lands, I'll rebase and update the test.

### How can this be tested?
Tests should pass.

- Visit a b2b contract you're an admin for and attempt to assign a code to an email that has an outstanding assignment or has redeemed a code in that contract.
- You should not observe a `queue_send_enrollment_code_assignment_email` task being fired.

### DIFF
--- a/b2b/views/v0/manager.py
+++ b/b2b/views/v0/manager.py
@@ -75,6 +75,12 @@ class CodeAssignment:
 def assign_codes_and_send_emails(
     assignments: list[CodeAssignment], assigning_user
 ) -> bool:
+    # If we're passed an empty list, short circuit and return true
+    # This can happen in bulk_assign if all users in the payload have already
+    # been assigned or redeemed a code for the contract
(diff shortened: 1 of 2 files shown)


daniellefrappier18 (commented):
Verified locally — no queue_send_enrollment_code_assignment_email task in the celery logs when assigning an already-assigned/redeemed email (and it still fires for a fresh assignment). Works as advertised. ✅

daniellefrappier18 (approved):

PR #3768, opened 2026-07-20
fix: add 'stay update' boolean field for course page

ahtesham-quraish:
https://github.com/mitodl/hq/issues/12419#issuecomment-5006171368

Add new boolean field for stay updated form for courses and this will be used to hide or show the `stay update` button on the learn course page

Screenshot 2026-07-20 at 3 56 19 PM

### How can this be tested?
For testing you just need to checkout the branch and make sure the migrations are added and visit the course creating url like given below `http://mitxonline.odl.local:9080/cms/pages/add/cms/coursepage/<course_page_id>/` and you would see the boolean field.


cms/migrations/0066_coursepage_show_stay_update.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("cms", "0065_alter_certificatepage_product_name"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="coursepage",
+            name="show_stay_update",
```
gumaerc: The field name should be `show_stay_updated`. I know the comment @ChristopherChudzicki left [here](https://github.com/mitodl/hq/issues/12419#issuecomment-5006171368) says `show_stay_update` but I think that's a typo because it's not grammatically correct. It's grammatically correct to say `show_stay_updated`.

cms/wagtail_api/schema/serializers.py
```diff
@@ -198,6 +198,7 @@ class Meta:
             "topic_list",
             "include_in_learn_catalog",
             "ingest_content_files_for_ai",
+            "show_stay_update",
```
gumaerc: This only adds the field to the `CoursePage` serializer. The product page templates in Learn call the Wagtail pages API, so this will work for courses but don't we also need to be able to do this for programs? The courses in the example are actually courses, but we also have "program as course" so courses can technically be programs. If we don't add the field to `ProgramPage`, then we won't be able to control it on program product pages.

gumaerc (changes requested):
I can see the field coming through in the Wagtail pages API, which should be good for course product pages, but not programs or program as course. I also have a suggestion regarding the field name.

cms/wagtail_api/schema/serializers.py
ahtesham-quraish: > This only adds the field to the `CoursePage` serializer. The product page templates in Learn call the Wagtail pages API, so this will work for courses but don't we also need to be able to do this for programs? The courses in the example are actually courses, but we also have "program as course" so courses can technically be programs. If we don't add the field to `ProgramPage`, then we won't be able to control it on program product pages.

My assumption was to keep it only for coursePage but as per your suggestion I have added it for programPage as well.

gumaerc (approved):
👍 LGTM

PR #3775, opened 2026-07-21
B2B Enrollment email deliverability indications via webhooks

dsubak:
https://github.com/mitodl/hq/issues/12114

At a high level, this PR has:
- DiscountContractAttachmentRecord model changes to support linking events to messages via mailgun's provided message ID, and fields to save relevant information from that event.
- An endpoint which can receive [Mailgun event webhooks](https://documentation.mailgun.com/docs/mailgun/user-manual/webhooks/webhooks) and enqueue processing tasks.
- A celery task which persists the event status and timestamp to a DiscountContractAttachmentRecord via mailgun's message ID.
- Tags for enrollment code emails sent through Mailgun via the B2B admin dashboard to enable filtering out events we don't care about.
- Serializer changes to display event statuses and timestamps on the /codes and related APIs
- A small number of changes to existing endpoints to handle event lifecycle (notably in the remind and reassign API endpoints)
- A script to generate synthetic webhooks for local testing

After populating a synthetic delivered event via the local webhook endpoint w/ signature validation disabled.
Screenshot 2026-07-27 at 3 58 03 PM

#### Webhook status persistence
Included in this PR is a small test script to enable rudimentary testing without the need to set up a mailgun account or webhook. This is probably the most reasonable way to approach testing this functionality.
- Set signature validation to be disabled.
- Ensure you've got a contract which can be managed via the b2b admin dash.
- Using the admin dashboard, assign a code to any email address. Get the ID of the corresponding DiscountContractAttachmentRedemption record.
  - Note, you don't need to actually send an email here if it's not set up as we're faking the webhook round trip.
- Run `./manage.py b2b_send_mailgun_webhook <record_id> <email_event_status> --host='http://<local_mitxonline_host>:<port>'`
  - You can specify any of the values in `EMAIL_STATUSES` as the status. We don't make any meaningful differentiation between them at the moment, though we may in the future.
- Observe by hitting the [/codes](http://mitxonline.odl.local:9080/api/schema/swagger-ui/#/b2b/b2b_manager_organizations_contracts_codes_list) endpoint for the corresponding contract that the code assigned now has an email_status and timestamp populated. 

#### Status reset on reassignment/resending
- Perform the above steps, making sure not to actually redeem the code.
- In the admin dash, send a reminder or reassign the pending invite.
- Observe via /codes that the email status and timestamp are cleared.

### Additional Context
I did some initial testing with a mailgun RC domain webhook by exposing the newly created endpoint via ngrok. This, frankly, sucks - exposing the endpoint alone is annoying, but even then you have to go through the rigamarole of spoofing your local data to have DCARs with matching message IDs before the webhook comes in since you're sending from RC.

To wit, I am providing a small script which sends synthetic webhooks against the endpoint for use in testing. Obviously this is a bit less representative, but the payload structure should be consistent and stable, so it should be sufficient for local testing (for now at least).

I am not a fan of how large this PR has become, but it's more or less the minimum amount for anything like an end-to-end test. If folks would prefer to review it in parts, I'm happy to break this down into more independently shippable parts - it just won't be meaningfully testable until several have landed.

### Checklist:
- [x] PR to provision mailgun webhook signing secrets in ol-infra
- [x] Figure out task delegation
- [x] Webhook endpoint
- [x] Delete email deliverability bits from DCAR at appropriate times (reassign, remind)
- [x] Determine if we want a unique constraint on message ID
- [x] Small script to punch synthetic events into webhook endpoint for specified DCAR assignment records (far too annoying to test w/o a mailgun acct otherwise)
- [x] setting to control whether or not we validate signatures for local testing.
- [x] Finalize webhook endpoint location
- [x] Find out idiomatic way to validate payload is json in DRF.
- [x] Change test message to not send mailgun tag. That way we can easily filter it out.
- [x] E2E tests for new functionality


b2b/serializers/v0/manager.py
```diff
@@ -249,6 +254,15 @@ def get_last_sent(self, obj) -> datetime | None:
         redemption = self._get_redemption(obj)
         return redemption.last_reminder_sent_on if redemption else None
 
+    @extend_schema_field(serializers.ChoiceField(choices=EMAIL_STATUSES))
+    def get_email_status(self, obj) -> str | None:
```
daniellefrappier18, line 262: This getter can return three different things but the schema only describes one of them. It returns None for an unassigned code, "" for a code that's been assigned but hasn't gotten a webhook back yet and one of the 5 enum values ("delivered" | "accepted" | "opened" | "clicked" | "failed")  once a webhook lands.

The problem is `@extend_schema_field(ChoiceField(choices=EMAIL_STATUSES))` generates a schema that marks the field as required, non-nullable, and always one of the 5 values. You can see email_status_event_timestamp below it gets nullable: true, but this one doesn't.

mit-learn generates its TS client from this spec, so the ContractAdmin dashboard will see email_status: EmailStatusEnum and won't be forced to handle `""` or `null`. Any code that maps a status to a label will get undefined for assigned-but-not-yet-delivered codes.

Could we normalize `""` to None and mark the field nullable so the contract matches what we actually return?

```
@extend_schema_field(serializers.ChoiceField(choices=EMAIL_STATUSES, allow_null=True))
def get_email_status(self, obj) -> str | None:
    redemption = self._get_redemption(obj)
    return (redemption.email_status or None) if redemption else None
```
Then regenerate the spec so the frontend has a single null for "no delivery info yet."

dsubak: Good catch, yeah, we can kick back None for any situation where it isn't one of the email statuses we're tracking. There's no practical deliniation to be made between "" and None, will get this tightened up

b2b/api.py
```diff
@@ -1724,3 +1731,75 @@ def process_remove_org_membership(user, organization):
...
+            # We don't want to save
+            return None
+
+    # At this point we know that the payload is for the email we care about, it's from mailgun, and it's one of the events we care about
+    # Save it to the row corresponding to the event we just got and move on with our lives
+    message_id = event_data["message"]["headers"]["message-id"]
+    assignment = DiscountContractAttachmentRedemption.objects.get(
+        email_message_id=message_id
+    )
+    assignment.email_status = event_type
```
daniellefrappier18, line 2011: Two potential issues here:

1. Mailgun webhooks are not ordered and multiple Celery workers process them concurrently. A late-arriving delivered will clobber an already-recorded opened/clicked, so the dashboard shows a regressed status. 

2. `email_status_event_timestamp` is set to processing time (`now_in_utc()`), not the event's own time. So if we plan to expose this on the dashboard time will be wrong/misleading.

Suggested Fix for 1: Only advance status by a defined precedence (or refuse to overwrite when the incoming event timestamp is older than the stored one).

Suggested Fix for 2: The payload carries event-data.timestamp and use that for `email_status_event_timestamp`

b2b/views/v0/manager.py
```diff
@@ -650,7 +659,7 @@ def send_reminder_for_code_assignment(self, request, **kwargs):  # noqa: ARG002
 
         # Just send the email reminder and update the last sent timestamp
         queue_send_enrollment_code_assignment_email.delay([assignment_record.id])
-        assignment_record.last_reminder_sent_on = now_in_utc()
+        clear_assignment_email_deliverability_fields(assignment_record)
```
daniellefrappier18, line 663: This used to set `last_reminder_sent_on = now_in_utc()` here, but now it just clears the deliverability fields and that timestamp only gets set later when the send task runs. So the response we return right below still has the old last_sent value (or null if it was never sent), and the comment above about updating the last sent timestamp isn't true anymore.

If the dashboard shows a "reminder sent" time after this call, it'll be stale until the task finishes and the manager refetches. Could we set `last_reminder_sent_on = now_in_utc()` here too so the response reflects the click right away?

b2b/api.py
dsubak: I think comparing the value of `email_status_event_timestamp` to the value of `event-data.timestamp` is probably the simplest fix. There might be one oddity in that users could bop between opened and clicked if they reopened old emails, but it'd still be an accurate view of the most recent user action. I think we'll go w/ this for now, and if that ends up confusing folks we can revisit building a little tiny state machine for these events.

Relatedly, this change means we won't definitively be able to tell when we processed an event based on DB state - that might have some diagnostic utility in the future, but we can always tack that in later if it becomes necessary.

b2b/api.py
```diff
@@ -1724,3 +1731,75 @@ def process_remove_org_membership(user, organization):
...
+        return False
+
+    # Possibly overbuilt, but we'll just make sure we're getting something shaped vaguely correctly
+    if not payload or not isinstance(payload, dict):
+        return False
+
+    # Check for the right message tag - if it's not there, do nothing else.
+    # We want to throw out unrelated messages as fast as possible
+    event_data = payload.get("event-data", {})
+    return ENROLLMENT_CODE_ASSINGMENT_TAG in event_data.get("tags", [])
```
daniellefrappier18: I had a security agent review this and it flagged this as a potential crash point that malformed payload can reach.  Anyone can post arbitrary JSON here `event_data.get("tags", [])` only falls back to `[]` when the key is missing. If someone sends "tags": null, this returns None and TAG in None raises a TypeError, which becomes a 500. Same kind of thing happens down in `process_mailgun_webhook_for_enrollment_code_emails` where `payload["signature"]["token"]`. assume the keys exist, so a payload with the right tag but no signature block throws a KeyError in the task.

None of this is a data risk, but it's an easy way to spam our error logs with 500s. Could we make these lookups defensive, something like `event_data.get("tags") or []` here, and read the signature keys with `.get()` and bail to None if they're missing?

b2b/views/v0/manager.py
dsubak: Hmmm, my rationale for moving it to the task is that this as populated field is currently a bit misleading - it's entirely possible that we kick out the task to send and the mailgun API call outright fails, or the task is super super delayed or something. If that occurs, right now it looks like you sent a reminder to someone but in reality the learner didn't get any email. It's "more accurate" to have it set only once the task runs to completion.

That said, the staleness thing is a straight up oversight on my part - I definitely meant for that helper to also clear `last_reminder_sent_on`. That'd result in the the response here showing None until the task runs and actually sends it.

Maybe what we could do is leave `last_reminder_sent_on` set only at the API layer, but indicate to the user that a task is "Pending" or something until we get an `accepted` webhook from mailgun? Even that's not _quite_ accurate in the event of an outright API call failure, but it's more accurate in the case of a task processing delay. I wasn't planning on configuring accepted webhooks since it's typically a short lived/uninteresting state for users, but it's easy enough to do if we think this is a decent solve. Thoughts?

b2b/api.py
dsubak: Yeah, we can tighten this up. At minimum we can make the existing checks more defensive

Initially I thought it might be overkill to do it, but I'll see if it's straightforward enough to tack in a Pydantic model to validate the incoming dictionary for the fields we care about. We've already got the dependency in build, so as long as it isn't annoying to read, that might be a more robust option long term.

b2b/api.py
dsubak: This should be addressed now and tests are written for this case - I'm gonna resolve this, but feel free to reopen if further discussion is needed!

b2b/serializers/v0/manager.py
dsubak: Done! Resolving

b2b/api.py
dsubak: I tossed in a pydantic model (technically like 5 models composed in a trenchcoat) so this should be much more resilient now. Let me know what you think!

b2b/views/v0/manager.py
daniellefrappier18: That reasoning makes sense to me and I'm fine with the field meaning "Mailgun took it" rather than "someone clicked remind." Two things though: On clearing last_reminder_sent_on in the helper, I think that's right for reassign but wrong for remind. On reassign it's a different person so the old send doesn't mean anything. On remind, if we clear it and the send fails, we don't end up showing "pending", we end up showing nothing at all and we've thrown away the fact that a reminder did go out on some earlier date. Losing that seems worse than showing a slightly stale timestamp. Could we keep the helper clearing message ID and status but only clear last_reminder_sent_on on the reassign path?

Separately I think there's an ordering bug in the remind path right now regardless of which way we go. We call .delay() before assignment_record.save(), so if a worker picks it up quickly the task writes the message ID and timestamp and then that bare save() writes the stale in-memory instance back over it. 

```
queue_send_enrollment_code_assignment_email.delay([assignment_record.id])
clear_assignment_email_deliverability_fields(assignment_record)
assignment_record.save()

```
Reassign already does save first and then delay, so I think remind just needs the same order.

On the pending state, I like it, but after researching this with Claude I don't think we need to turn on accepted webhooks for it. If we set a local "pending" marker in the view when we queue the task, that covers the delay case and it also stays visibly stuck if the Mailgun call fails outright, which the accepted webhook wouldn't catch.

dsubak: Re: remind vs assigned: that makes sense to me. We'll clear it only in the reassign path - I was thinking that if you've clicked remind you may not care about the older reminders, but it's totally reasonable to keep it around.

Re ordering - yup, good catch on the race condition, will get that flipped around. Depending on how complicated the state being managed is it may make sense to scope the save to the relevant fields, but for now I think that's unnecessary.

Re local pending marker - as a point of clarification, you just mean jamming a new, non-mailgun derived state into email_status? If so, that works for me - nothing strictly saying we have to rely on mailgun's statuses for that and I think this makes sense to add. I'll probably still add accepted to the set of webhooks we process, but we can map that and whatever this is to `pending` in the serializer since I don't think the distinction is meaningful to users, just for our diagnostic purposes. And obviously I'll update comments and the api spec accordingly!

dsubak: Confirmed offline re: local pending marker. I'll get those changes tacked in and kick it back up for review!

dsubak: @daniellefrappier18 Ordering fixed, last_reminder_sent_on set to `None` only during reassign endpoint, pending was added to any place that kicked off an email sending task and the serializer maps both pending and accepted to the same `pending` value so as not to confuse the user.

I think this is ready for another review, and I'll resolve this comment thread!

b2b/api.py
```diff
@@ -1878,3 +1888,127 @@ def process_remove_org_membership(user, organization):
...
+class MailgunWebhookEventData(BaseModel):
+    """The event-data block of a Mailgun webhook payload."""
+
+    model_config = ConfigDict(extra="allow")
+
+    event: str
+    tags: list[str] = Field(default_factory=list)
+    message: MailgunWebhookMessage
+    severity: str | None = None
+
```
daniellefrappier18, line 1939: Can you add `timestamp` as a required field here to guard against the payload missing it and crashing the task with an unhandled KeyError

dsubak: Good catch, done!

daniellefrappier18 (approved):
LGTM

PR #3776, opened 2026-07-21
Add redirect for terms, privacy, and honor code

alexfigtree:
Closes https://github.com/mitodl/hq/issues/12031

Replaces mitxonline's hardcoded links to its own Terms of Service, Privacy Policy, and Honor Code pages with links to their MIT Learn equivalents, since Learn's terms of service already cover MITx offerings:

- https://learn.mit.edu/terms (was /terms-of-service/)
- https://learn.mit.edu/privacy (was /privacy-policy/)
- https://learn.mit.edu/honor_code (was /honor-code/)

Updated in 6 places:

- main/templates/footer.html — site-wide footer (all three links)
- frontend/public/src/lib/urls.js — informationLinks route constants consumed by the React signup form (all three links)
- mail/templates/email_base.html — base transactional email footer (ToS + Privacy)
- ecommerce/templates/mail/product_order_receipt/body.html — order receipt email (ToS + Privacy)
- courses/templates/mail/partials/_uai_footer.html — B2B/partner-branded email footer (ToS + Privacy)
- cms/templates/certificate_page.html — public certificate page footer (ToS + Privacy)

No honor code link exists in any of the email templates or the certificate page, so those two files only needed the ToS/Privacy changes.

No visual/layout changes — link text and placement are unchanged, only the href destinations changed.

### How can this be tested?

1. On mitxonline, load any page and check the site footer — Terms of Service, Privacy Policy, and Honor Code links should all point to learn.mit.edu and load the correct page.
2. Go through the account registration/signup flow and confirm the same three consent links in the signup form point to Learn.
3. Trigger (or inspect a rendered copy of) the order receipt email, and any B2B/partner-branded course email, and confirm the footer's Terms of Service / Privacy Policy links point to Learn.
4. View a public certificate page and confirm its footer links point to Learn.
5. yarn lint and yarn fmt:check pass clean on frontend/public (only urls.js is JS/lintable — the Django templates have no linter configured in this repo).

### Additional Context

- Confirmed the exact Learn destination paths (/terms, /privacy, /honor_code) against Learn's actual Next.js routes rather than assuming a naming pattern — note /honor_code uses an underscore, not a hyphen, unlike mitxonline's original /honor-code/.
- No trailing slashes on the new URLs — Learn's Next.js app uses the default trailingSlash: false, so these are the canonical forms.
- RegisterEmailForm.js (the component that renders the signup-form links) needed no changes — it reads the fixed constants from urls.js directly.
- Checked for tests asserting on these link values; found none (the only similarly-named test, openedx/api_test.py, covers an unrelated Open edX enrollment API field also named honor_code).
- Scope is limited to ToS/Privacy/Honor Code links only — other mitxonline→Learn redirect work (product pages, catalog, home page) is tracked in separate tickets under the parent epic (#12027).

### Checklist:
- [ ] Confirm with @pdpinch that privacy policy should not link to honor code

### DIFF
--- a/mail/views.py
+++ b/mail/views.py
@@ -1,6 +1,5 @@
 """Mail views"""
 
-from django.conf import settings
 from django.http import JsonResponse
 from django.shortcuts import render
 from django.utils.decorators import method_decorator
@@ -42,7 +41,7 @@ class EmailDebuggerView(View):
             return JsonResponse({"error": "invalid input"})
 
         email_type = form.cleaned_data["email_type"]
-        context = {"base_url": settings.SITE_BASE_URL, "site_name": settings.SITE_NAME}
+        context = api.get_base_context()
 
         email_extra_context = EMAIL_DEBUG_EXTRA_CONTEXT.get(email_type, {})
         context.update(email_extra_context)
--- a/main/settings.py
+++ b/main/settings.py
@@ -656,6 +656,24 @@ MIT_LEARN_DASHBOARD_URL = get_string(
     description="Dashboard URL for UAI enrollment emails",
 )
 
+MIT_LEARN_TERMS_URL = get_string(
+    name="MIT_LEARN_TERMS_URL",
+    default="https://learn.mit.edu/terms",
+    description="Terms of service URL",
+)
+
+MIT_LEARN_PRIVACY_URL = get_string(
+    name="MIT_LEARN_PRIVACY_URL",
+    default="https://learn.mit.edu/privacy",
+    description="Privacy policy URL",
+)
+
+MIT_LEARN_HONOR_CODE_URL = get_string(
+    name="MIT_LEARN_HONOR_CODE_URL",
+    default="https://learn.mit.edu/honor_code",
+    description="Honor code URL",
+)
+
 # Logging configuration
 LOG_LEVEL = get_string(
     name="MITX_ONLINE_LOG_LEVEL", default="INFO", description="The log level default"
--- a/main/settings.py
+++ b/main/settings.py
@@ -33,6 +33,7 @@ from mitol.scim.settings.scim import *  # noqa: F403
 from redbeat import RedBeatScheduler
 
 from main.celery_utils import OffsettingSchedule
+from main.constants import ENV_TO_LEARN_HOSTNAME_MAP
 from main.env import get_float
 from main.sentry import init_sentry
 from openapi.settings_spectacular import open_spectacular_settings
@@ -658,19 +659,19 @@ MIT_LEARN_DASHBOARD_URL = get_string(
 
 MIT_LEARN_TERMS_URL = get_string(
     name="MIT_LEARN_TERMS_URL",
-    default="https://learn.mit.edu/terms",
+    default=f"https://{ENV_TO_LEARN_HOSTNAME_MAP.get(ENVIRONMENT, 'learn.mit.edu')}/terms",
     description="Terms of service URL",
 )
 
 MIT_LEARN_PRIVACY_URL = get_string(
     name="MIT_LEARN_PRIVACY_URL",
-    default="https://learn.mit.edu/privacy",
+    default=f"https://{ENV_TO_LEARN_HOSTNAME_MAP.get(ENVIRONMENT, 'learn.mit.edu')}/privacy",
     description="Privacy policy URL",
 )
 
 MIT_LEARN_HONOR_CODE_URL = get_string(
     name="MIT_LEARN_HONOR_CODE_URL",
-    default="https://learn.mit.edu/honor_code",
+    default=f"https://{ENV_TO_LEARN_HOSTNAME_MAP.get(ENVIRONMENT, 'learn.mit.edu')}/honor_code",
     description="Honor code URL",
 )
 
--- a/.env.example
+++ b/.env.example
@@ -39,3 +39,8 @@ MITOL_APIGATEWAY_USERINFO_UPDATE=True
 # Host for Digital Credentials issuer-coordinator service
 VERIFIABLE_CREDENTIAL_SIGNER_URL=http://dcc.odl.local:4005/instance/test/credentials/issue
 VERIFIABLE_CREDENTIAL_BEARER_TOKEN='test'
+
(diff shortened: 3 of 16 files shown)


dsubak (commented):
I've still gotta test this guy but did want to ask one question based on a quick review - I think previously many (all?) of these links would've been environment aware (i.e. RC cert pages would point at rc.mitxonline.com/terms-of-service while prod would point at mitxonline.com/terms-of-service). Now we'll always direct folks to the prod version of those pages. Should we try and retain the per-env host behavior?

dsubak (commented):
Everything looking good afaict in testing. If we don't care about [conditionalizing the host](https://github.com/mitodl/mitxonline/pull/3776#pullrequestreview-4754964951), I'm happy to approve as is - just let me know!

alexfigtree:
@dsubak I'm seeing there's a precedent for this in the codebase. Specifically, I see the frontend already injects other env-driven constants, so I'll pull the value from an env var in urls.js baked in at build time instead of using literal strings. I can add new url settings (prod defaults as-is), thread them through the same context processor for the Django templates/emails.

ecommerce/views/legacy/__init__.py
```diff
@@ -804,14 +804,13 @@ class CheckoutCallbackView(View):
     def __init__(self, *args, **kwargs):  # noqa: ARG002
         self.logger = logging.getLogger(__name__)
 
-    def post_checkout_redirect(self, order_state, order, request):
+    def post_checkout_redirect(self, order_state, order):
```
dsubak: Is this change intended to be in this PR? It looks like it's unrelated to the URL changes afaict

mail/views.py
```diff
@@ -42,7 +41,7 @@ def post(self, request):
             return JsonResponse({"error": "invalid input"})
 
         email_type = form.cleaned_data["email_type"]
-        context = {"base_url": settings.SITE_BASE_URL, "site_name": settings.SITE_NAME}
+        context = api.get_base_context()
```
dsubak, line 44: Nice!

main/settings.py
```diff
@@ -652,6 +652,24 @@
     description="Dashboard URL for UAI enrollment emails",
 )
 
+MIT_LEARN_TERMS_URL = get_string(
```
dsubak, line 660: Nit: Can we add these settings to the .env.example file too? Local dev URLs are kinda all-over the place so it could be nice to have em in there explicitly as a signpost for people who want this to point back at where they run learn.

dsubak:
Piping the settings values through looks solid, and as it stands, I think this'll work out of box for prod. Is there a corresponding PR to set the env vars in ol-infra for RC/CI?

ecommerce/views/legacy/__init__.py
alexfigtree: Not originally intended, but I took it out because ruff's unused-argument check flagged `request` as dead — it was never referenced anywhere in the method body, unlike post()'s signature which Django's View.dispatch requires as-is.

dsubak: If I'm reading this right, it doesn't appear that it was an unused variable prior to the changes to the error logging path - `PaymentGateway.get_formatted_response` took it as a parameter if I'm looking at the right spot. We're cool changing the logged message there too?

Sorry for the back and forth - I've never worked with this chunk of code, so if this is expected no problem, but I just wanted to double check that everything's expected!

alexfigtree:
> Piping the settings values through looks solid, and as it stands, I think this'll work out of box for prod. Is there a corresponding PR to set the env vars in ol-infra for RC/CI?

No separate ol-infra PR needed — in a4d1f776, I moved the URL resolution to use the same environment→hostname mapping the Verifiable Credentials code already relies on (courses/api.py), so the terms/privacy/honor-code URLs now resolve per-environment from settings.ENVIRONMENT automatically (prod/rc/ci). Operators can still override via env var if they ever need to, but nothing has to be set in ol-infra for this to work correctly out of the box.

dsubak:
> > Piping the settings values through looks solid, and as it stands, I think this'll work out of box for prod. Is there a corresponding PR to set the env vars in ol-infra for RC/CI?
> 
> No separate ol-infra PR needed — in [a4d1f77](https://github.com/mitodl/mitxonline/commit/a4d1f7764d826a667a71ae737378765105f2b7fa), I moved the URL resolution to use the same environment→hostname mapping the Verifiable Credentials code already relies on (courses/api.py), so the terms/privacy/honor-code URLs now resolve per-environment from settings.ENVIRONMENT automatically (prod/rc/ci). Operators can still override via env var if they ever need to, but nothing has to be set in ol-infra for this to work correctly out of the box.

Works for me! The verifiable creds code was something I hard coded at the time, but given that it seems to be a pattern were going to see more of going forward, having it be a reusable constant is a winner!

ecommerce/views/legacy/__init__.py
alexfigtree: Digging into it, that whole chain (the original indentation fix, the test that locked it in, and the later removal) was self-contained churn from this branch's own history interacting with a merge from main — it never actually reflected anything main itself needed changed. main's post_checkout_redirect already has the full detailed log intact and always did.

Rather than patch the logged message further, I removed the whole chain of unrelated commits from this branch (they were out of scope for the terms-of-service work anyway, as you flagged originally) — this PR's ecommerce/views/legacy/__init__.py now matches main exactly, with the detailed logging fully intact. No separate PR needed as I thought there might be; there was nothing to fix once the unrelated commits were gone.

main/settings.py
alexfigtree: Added in c0ab6f97 — MIT_LEARN_TERMS_URL, MIT_LEARN_PRIVACY_URL, and MIT_LEARN_HONOR_CODE_URL are now in .env.example with their production defaults shown, so anyone running a local Learn instance has a clear signpost to override them.

ecommerce/views/legacy/__init__.py
dsubak: Awesome - I appreciate ya digging into that! If I knew a bit more about the diagnostic needs for checkout logging I'd have been able to provide more targeted advice, but we got there in the end! Resolving!

dsubak (approved):

PR #3781, opened 2026-07-22
Single hubspot account

cp-at-mit:
NA

Removes all of the code that was interfacing with the hubspot account defined by `UAI_MITOL_HUBSPOT_API_PRIVATE_TOKEN`.  This hubspot account was previously receiving any user/course/enrollment/deal info associated with UAI courses and programs.  This PR updates our hubspot code so that any course, regardless of whether it's a UAI course/program or not is synced with the hubspot account defined by MITOL_HUBSPOT_API_PRIVATE_TOKEN.

### How can this be tested?
Use a new user and enroll and pay for a new UAI course and program.  Verify that your contact, products, deal, line items, etc are all created in a single Hubspot account linked to MITOL_HUBSPOT_API_PRIVATE_TOKEN


pdpinch:
Not sure how we would test this, but I think we also want to make sure that directing everything to one account doesn't leave us with overlapping or confusing contacts and deals in Hubspot.

PR #3785, opened 2026-07-23
Add CyberSource export compliance app and legal address profile fields

annagav:
Adds the compliance app (CyberSource export compliance verification via the REST API) along with the supporting LegalAddress fields (street address, city, postal code) needed to build the compliance check payload, plus factories/serializers/migrations/openapi updates for those new profile fields.

### What are the relevant tickets?
<!--- If it fixes an open issue, please link to the issue here. -->
<!--- Closes # --->
<!--- Fixes # --->
<!--- N/A --->

### Description (What does it do?)
 - Adds the compliance app implementing CyberSource export compliance verification via the REST API (compliance/api.py, compliance/exceptions.py), including the payload builder, client configuration,   
  and error types (ExportComplianceError, ExportComplianceDataError)                                                                                                                                       
  - Adds the LegalAddress profile fields needed to build that compliance payload: street_address_1, street_address_2, city, postal_code — with migrations, serializer updates, and factory support         
  - Backfills the new address fields from edX data in migrate_edx_data                                                                                                                                     
  - Updates the OpenAPI specs (v0–v2) to reflect the new LegalAddress fields                                                                                                                               
  - Enables the compliance app in INSTALLED_APPS                                                                                                                                                           
                                                                                                                                                                                                           
  This is a self-contained foundational change — the compliance app doesn't yet enforce anything on enrollment. It has no dependency on courses or any feature flag.                                       
                                                                                                                                                                                                           
  Split out from ag/add_export_compliance. Enrollment enforcement (wiring the export check into create_run_enrollments/create_program_enrollments behind EXPORT_COMPLIANCE_CHECK_ENABLED) is in a stacked  
  follow-up: ag/export-compliance-enrollment.                                                                                                                                                              
                                                                                                                                                                                                           
  Test plan                                                                                                                                                                                                
                                                                                                                                                                                                           
  - [ ] docker compose exec web pytest compliance/                                                                                                                                                         
  - [ ] docker compose exec web pytest users/                                                                                                                                                              
  - [ ] docker compose exec web python manage.py makemigrations --check --dry-run

PR #3786, opened 2026-07-23
Enforce export compliance verification on course/program enrollment

annagav:
Related to https://github.com/mitodl/mitxonline/pull/3706
 - Wires the compliance app's CyberSource export check into create_run_enrollments and create_program_enrollments via a new _verify_exports_compliance_for_enrollment helper in courses/api.py, gated     
  behind the EXPORT_COMPLIANCE_CHECK_ENABLED feature flag (main/features.py)                                                                                                                               
  - Enrollment is blocked and an ExportComplianceError is raised when CyberSource does not accept the check, for both course-run and program enrollments (audit and verified modes)                        
  - Surfaces ExportComplianceCheckError as a user-facing error from the v1–v3 enrollment surfaces:                                                                                                         
    - courses/serializers/v1/v2/v3/courses.py — raises a DRF ValidationError with exc.to_error_detail()                                                                                                    
    - courses/views/v1/__init__.py — redirects with a user-facing enrollment-blocked message                                                                                                               
    - courses/views/v2/__init__.py — returns a 400 response, including for the verified-program-enrollment reconciliation path                                                                             
    - courses/views/v3/__init__.py — raises a ValidationError                                                                                                                                              
  - Adds an autouse mocked_export_compliance conftest fixture so existing tests continue to pass with the check enabled by default, and updates drf_lint_baseline.json for line shifts in the touched
  serializers
  - Updates `_verify_exports_compliance_for_enrollment` to pass a courseware object (a `CourseRun` for run enrollments, a `Program` for each program enrollment) through to `verify_user_with_exports`, matching that function's `(user, courseware_object)` signature from #3793

  Stacked on nl/export-compliance-cache (#3793), which adds the `ExportComplianceLog` caching model this PR's compliance checks read/write through, and which is itself stacked on ag/export-compliance-app (#3785). This PR only contains the enrollment integration; diff against the base branch reflects just that.

  Test plan                                                                                                                                                                                              
                                                                                                                                                                                                           
  - [ ] docker compose exec web pytest courses/api_test.py                                                                                                                                                 
  - [ ] docker compose exec web pytest courses/                                                                                                                                                            
  - [ ] Manually exercise course-run and program enrollment with EXPORT_COMPLIANCE_CHECK_ENABLED=True for both an accepted and a rejected CyberSource decision


courses/views/v2/__init__.py
```diff
@@ -810,6 +814,51 @@ def _create_course_enrollment_from_program(request, courserun_id, program_enroll
...
+            # we should stop.
+            log.error(
+                "add_verified_program_course_enrollment: user %s enrolling in %s has no verified enrollment in %s",
+                request.user,
+                courserun_id,
+                root_program,
+            )
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+    except ExportComplianceCheckError as exc:
+        return Response(exc.to_error_detail(), status=status.HTTP_400_BAD_REQUEST)
```
rhysyngsun: I think it might be better here and the other places we're returning an explicit `Response` instance to instead [subclass DRF's `APIException`](https://www.django-rest-framework.org/api-guide/exceptions/#apiexception), something like this:

```python
class EnrollmentError(APIException):
    status_code = 400
    default_detail = "Unable to complete enrollment"
    default_code = "unable_to_complete_enrollment"
```

Benefits from this approach:
- Don't need to return anything or check the return
- We get a well structured JSON error instead, which we should've been doing on this API already. You probably need to add it to the `@extend_schema` call to get it in the schema, but we should be able to more or less get that for free.

I also think that we don't want to propagate detailed information on the compliance error down to the client, because historically if a compliance check or enrollments fail, we just tell the user to contact support.

rhysyngsun (changes requested):

courses/views/v2/__init__.py
```diff
@@ -831,6 +837,52 @@ def _create_course_enrollment_from_program(request, courserun_id, program_enroll
     )
 
 
+def _reconcile_verified_program_enrollments(
```
rhysyngsun: Could we move this into `api.py`?

rhysyngsun (changes requested):
Code looks good, one small separation of concerns tweaks.

Still trying to test this locally but I'm pretty close, should be able to approve tomorrow barring any other issues.

rhysyngsun (approved):
LGTM

rhysyngsun (approved):

PR #3788, opened 2026-07-24
Fix OAS comment generation

rhysyngsun:

N/A

The OAS diff sometimes fails (see https://github.com/mitodl/mitxonline/actions/runs/30037342589/job/89308503010) because the get summary step can't parse the summary. That said, the summary _always_ ends up being "## Changes for v0.yaml:" because that's always the first line and this isn't helpful and in fact misleading because the contents that are hidden include changes for _all_ the specs. So I'm just getting rid of the summary.

### How can this be tested?

Tests should pass and an OAS comment should get left on this PR.


annagav (approved):
👍

PR #3790, opened 2026-07-27
Use Redis-backed HubSpot rate limiter

cp-at-mit:
NA
https://mitolproduction.grafana.net/a/grafana-lokiexplore-app/explore/service/celery-worker/logs?patterns=%5B%5D&from=now-3h&to=now&var-ds=grafanacloud-logs&var-filters=service_name%7C%3D%7Ccelery-worker&var-fields=&var-levels=&var-metadata=&var-patterns=&var-lineFilterV2=&var-lineFilters=caseInsensitive,0%7C__gfp__%3D%7Cd371c37a-d01d-4b81-87bf-dfc6ff9fe0fe&timezone=browser&var-lineFormat=&var-jsonFields=&var-all-fields=&urlColumns=%5B%5D&visualizationType=%22logs%22&displayedFields=%5B%5D&userDisplayedFields=false&sortOrder=%22Descending%22&wrapLogMessage=false&prettifyLogMessage=false

Replaces the per-process HubSpot rate limiter with a Redis Lua-script sliding window so Celery workers enforce limits globally. Adds a thread-safe in-memory fallback for Redis outages, including warning logs and jittered sleep handling. Tests were reorganized to cover both the fallback limiter and Redis-backed behavior (script invocation, args, wait parsing, sleep/no-sleep paths, and fallback conditions).

### DIFF
--- a/hubspot_sync/rate_limiter.py
+++ b/hubspot_sync/rate_limiter.py
@@ -4,85 +4,100 @@ Rate limiting for HubSpot API calls
 
 import logging
 import random
-import threading
 import time
-from collections import deque
+import uuid
 
 from django.conf import settings
 
 log = logging.getLogger(__name__)
 
+# Atomic sliding-window rate limiter implemented as a Redis Lua script.
+#
+# The sorted set stores claimed slots: member = unique ID, score = scheduled
+# execution time. Cleanup removes slots whose window has expired. When all
+# slots in the current window are claimed, the next slot is pushed past the
+# oldest slot's window boundary, distributing load across workers automatically.
+_RATE_LIMIT_LUA = """
+local key       = KEYS[1]
+local now       = tonumber(ARGV[1])
+local window    = tonumber(ARGV[2])
+local max_req   = tonumber(ARGV[3])
+local min_delay = tonumber(ARGV[4])
+local member    = ARGV[5]
+
+redis.call('ZREMRANGEBYSCORE', key, '-inf', now - window)
+
+local count = redis.call('ZCARD', key)
+local target
+
+if count < max_req then
+    if count > 0 then
+        local last   = redis.call('ZRANGE', key, -1, -1, 'WITHSCORES')
+        local last_t = tonumber(last[2])
+        target = math.max(now, last_t + min_delay)
+    else
+        target = now
+    end
+else
+    local oldest   = redis.call('ZRANGE', key,  0,  0, 'WITHSCORES')
+    local last     = redis.call('ZRANGE', key, -1, -1, 'WITHSCORES')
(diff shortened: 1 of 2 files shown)


hubspot_sync/rate_limiter.py
```diff
@@ -4,86 +4,101 @@
...
+        )
+        wait_ms = float(result)
+
+        if wait_ms > 0:
+            sleep_time = wait_ms / 1000
             jitter = random.uniform(-0.05, 0.05) * sleep_time  # noqa: S311
             sleep_time = max(0, sleep_time + jitter)
-
             log.debug("Rate limiting: sleeping for %.3f seconds", sleep_time)
             time.sleep(sleep_time)
```
dsubak, line 100: Do we know practically how long these tasks end up taking?

This is an outside concern but since I don't really know how our queues and tasks are arranged, do the queues contain a mix of hubspot and non-hubspot tasks. If so and the value of wait_ms was long (resulting in the hubspot tasks taking a long time), we could just be busy waiting instead of processing other tasks. If that were the case, we might be better served by redispatching the task w/ an ETA/countdown as opposed to busy waiting

Let me know if any of my assumptions are wrong or if the tasks are generally not waiting long and we can move ignore my concerns!

dsubak (commented):
I left one question, but the code looks good to me! Once we've resolved that comment I'll pop and approval on it!

hubspot_sync/rate_limiter.py
cp-at-mit: I think the hubspot tasks run within their own queue called `hubspot_sync`.  No other tasks run on that queue.

dsubak (approved):
Seems good - I ran the following script to check that it did correctly limit me if I tried to make more than 19 req/s

```
import logging
import time

from hubspot_sync.rate_limiter import rate_limiter

logger = logging.getLogger("hubspot_sync.rate_limiter")
logger.setLevel(logging.DEBUG)
if not logger.handlers:
    handler = logging.StreamHandler()
    handler.setFormatter(logging.Formatter("%(asctime)s %(levelname)s %(message)s"))
    logger.addHandler(handler)
logger.propagate = False

rate_limiter.min_delay_ms = 1
print(f"max_requests_per_second={rate_limiter._max_requests_per_second}")
print(f"min_delay_ms={rate_limiter.min_delay_ms}")

N = 30
start = time.monotonic()
for i in range(N):
    call_start = time.monotonic()
    rate_limiter.wait_for_rate_limit()
    elapsed = time.monotonic() - call_start
    print(f"call {i:02d}: blocked for {elapsed:.3f}s")

print(f"total elapsed for {N} calls: {time.monotonic() - start:.3f}s")
```

PR #3793, opened 2026-07-28
Add CyberSource export compliance data model

rhysyngsun:

Part of https://github.com/mitodl/hq/issues/12528

This adds a data model to store the result of a compliance check so we don't have to make repeated compliance checks for the same user if they repeatedly try to purchase/enroll.

### How can this be tested?

This doesn't integrate with anything yet so tests passing is our check.


annagav (approved):
LGTM

PR #3794, opened 2026-07-28
fix: read Learn's Fastly service ID from MIT_LEARN_FASTLY_SERVICE_ID

ChristopherChudzicki:

Refs https://github.com/mitodl/hq/issues/12381

1. Renames the Fastly token and URL settings to read the env vars set by https://github.com/mitodl/ol-infrastructure/pull/5119 switched to setting (`MITX_ONLINE_FASTLY_*`), 
    - ☝️ renamed env vars somewhat erroneously, but the new names are more consistent with the rest of the env vars used in the repo. 
2. Introduces `MIT_LEARN_FASTLY_SERVICE_ID`, replacing `MITX_ONLINE_FASTLY_SERVICE_ID`. The value is **Learn's** service ID — Learn is what tags its product pages with the `mitxonline:*` surrogate keys, so purging MITxOnline's own service does nothing. The old name invited exactly that mistake.
    - No live code was using `MITX_ONLINE_FASTLY_SERVICE_ID` in a way that worked.
3. Makes `service_id` an explicit argument to `queue_fastly_surrogate_key_purge`, so a purge's target is visible at the call site. Defaults to `None` so messages enqueued by the previous release skip rather than raise mid-deploy.

Also reads these via `django.conf.settings` instead of importing from `main.settings` at module import time, which is what hid them from Django's `settings` fixture in tests.

### How can this be tested?

Tests should pass. Not easily testable locally otherwise.

### Additional Context

Safe to deploy alone: 
- the service ID is unset before (due to mismatch in name introduced by https://github.com/mitodl/ol-infrastructure/pull/5119)
- Even before name mismatch, the purge was non-functional because project id was wrong.

It starts working when the infra change below lands.

Soft → hard purge is deliberately excluded, so [this infra change](https://github.com/mitodl/ol-infrastructure/pull/5143) can be validated in RC while soft purge is still in place.

### Checklist:

- [ ] Add `MIT_LEARN_FASTLY_SERVICE_ID` to `mitxonline/k8s_secrets.py`, per environment. Values already exist as `learn_service_id_{ci,qa,production}` under `infrastructure/fastly` in `src/bridge/secrets/concourse/operations.production.yaml`. Not `secret-mitxonline`'s `fastly.service_id` — that's MITxOnline's own service, and using it is what made the original purge a no-op.
- [ ] Don't restore `FASTLY_SERVICE_ID`; nothing reads it now.
- [ ] Confirm the token is authorized for Learn's service — purge-by-key needs `purge_select` on the target service, and this token comes from MITxOnline's Vault path.

### DIFF
--- a/cms/signals_test.py
+++ b/cms/signals_test.py
@@ -1,0 +1,67 @@
+"""Tests for cms.signals"""
+
+from unittest.mock import patch
+
+import pytest
+
+from cms.factories import CoursePageFactory, ProgramPageFactory, ResourcePageFactory
+from cms.signals import purge_fastly_cache_on_publish
+
+pytestmark = pytest.mark.django_db
+
+LEARN_SERVICE_ID = "test-learn-service-id"
+
+
+@pytest.fixture
+def learn_service_id(settings):
+    """Point the purge at a known MIT Learn Fastly service."""
+    settings.MIT_LEARN_FASTLY_SERVICE_ID = LEARN_SERVICE_ID
+    return settings.MIT_LEARN_FASTLY_SERVICE_ID
+
+
+@patch("cms.signals.transaction.on_commit", side_effect=lambda callback: callback())
+@patch("cms.tasks.queue_fastly_surrogate_key_purge.delay")
+def test_purge_fastly_cache_on_publish_course_page(
+    mock_purge_delay, mock_on_commit, learn_service_id
+):
+    """Publishing a CoursePage purges the key for its course."""
+    course_page = CoursePageFactory.create()
+    # Building the page saves a Course, which fires its own post_save purge.
+    mock_purge_delay.reset_mock()
+
+    purge_fastly_cache_on_publish(sender=None, instance=course_page)
+
+    mock_purge_delay.assert_called_once_with(
+        f"mitxonline:course:{course_page.course.readable_id}", learn_service_id
+    )
+
+
+@patch("cms.signals.transaction.on_commit", side_effect=lambda callback: callback())
+@patch("cms.tasks.queue_fastly_surrogate_key_purge.delay")
+def test_purge_fastly_cache_on_publish_program_page(
+    mock_purge_delay, mock_on_commit, learn_service_id
+):
+    """Publishing a ProgramPage purges the key for its program."""
+    program_page = ProgramPageFactory.create()
+    # Building the page saves a Program, which fires its own post_save purge.
+    mock_purge_delay.reset_mock()
+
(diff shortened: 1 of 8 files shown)


cms/tasks.py
```diff
@@ -98,47 +96,57 @@ def queue_fastly_full_purge():
...
 
     Key format: mitxonline:course:<readable_id> or mitxonline:program:<readable_id>
 
     Args:
         surrogate_key (str): The surrogate key to purge, e.g.
             "mitxonline:course:course-v1:MITx+6.00.1x"
+        service_id (str): The Fastly service ID whose cache should be purged.
+            Callers pass settings.MIT_LEARN_FASTLY_SERVICE_ID. Defaults to None
+            so that messages enqueued by an older release, before this argument
+            existed, skip the purge rather than raising during a rolling deploy.
```
marslanabdulrauf: This fix is one sided and will always work if the celery is the one to be deployed first.
The other way is if we started passing two positional arguments and the function only accept one.

cms/signals_test.py
```diff
@@ -0,0 +1,67 @@
...
+@patch("cms.tasks.queue_fastly_surrogate_key_purge.delay")
+def test_purge_fastly_cache_on_publish_course_page(
+    mock_purge_delay, mock_on_commit, learn_service_id
+):
+    """Publishing a CoursePage purges the key for its course."""
+    course_page = CoursePageFactory.create()
+    # Building the page saves a Course, which fires its own post_save purge.
+    mock_purge_delay.reset_mock()
+
+    purge_fastly_cache_on_publish(sender=None, instance=course_page)
```
marslanabdulrauf: `signals_test.py` should test the signal call instead of directly calling the task, I would say. 
These tests are more suited for tasks.

marslanabdulrauf (changes requested):
Added a couple of findings: 
- **Rolling deploy is handled in one direction only**
- **Minor** `cms/signals_test.py` calls `purge_fastly_cache_on_publish` directly rather than firing `page_published`

cms/tasks.py
ChristopherChudzicki: Good point.

I do want to keep the service id passing explicit, since i think it's somewhat unexpected that mitxonline's `queue_fastly_surrogate_key_purge` purges a different service id. But I made the passing optional, and a [follow PR](https://github.com/mitodl/mitxonline/pull/3804), stacked on this one, makes it required.

marslanabdulrauf (approved):
LGTM

PR #3806, opened 2026-07-31
fix: dirty-check edX sync and ProductPage saves to cut Fastly purges

Anas12091101:
https://github.com/mitodl/hq/issues/12614

Fastly surrogate-key purges were firing hundreds of times a day for changes that never happened. Root cause: two code paths call `.save()` unconditionally, and every `CourseRun`/`Course` save triggers a `post_save` purge signal.

This PR adds dirty checks so those saves (and their purges) only happen when something actually changed:

- **`sync_course_runs`** (hourly edX sync): compares incoming edX values against the stored run and skips `save()` when nothing differs. This eliminates the bulk of the purges (a course with 8 live runs was emitting ~192 identical purges/day, none reflecting a change) and a pointless full-column `UPDATE`.
- **`ProductPage.save()`**: only pushes the title down to the linked `Course`/`Program` (and saves it) when the title actually changed, removing the redundant purges emitted on every page save.

### How can this be tested?
The goal is to confirm that a purge is only scheduled when data actually changes. We do this in a Django shell by watching the purge task and running the sync twice — the second run should schedule **zero** purges.

1. Make sure your local stack is up and migrations are applied:
   ```bash
   docker compose up -d
   docker compose run --rm web python manage.py migrate
   ```
2. Open a Django shell:
   ```bash
   docker compose run --rm web python manage.py shell
   ```
3. Paste the following into the shell. It creates a throwaway course run, feeds it fake edX data, and runs the sync twice. Everything is rolled back at the end, so nothing is saved to your DB:
   ```python
   import uuid
   from unittest.mock import patch
   from edx_api.course_detail import CourseDetail
   from courses.factories import CourseRunFactory
   from courses.api import sync_course_runs

   suffix = uuid.uuid4().hex[:8]
   cid = f"course-v1:Demo+Purge+{suffix}"

   with patch("courses.signals.transaction.on_commit", side_effect=lambda cb: cb()), \
        patch("cms.tasks.queue_fastly_surrogate_key_purge.delay") as purge, \
        patch("courses.api.get_edx_api_course_list_client") as client:
       try:
           with transaction.atomic():
               run = CourseRunFactory.create(
                   courseware_id=cid,
                   course__readable_id=f"course-v1:Demo+Purge{suffix}",
               )
               client.return_value.get_courses.return_value = [CourseDetail({
                   "id": cid, "start": "2015-09-15T05:00:00Z", "end": "2015-12-31T05:00:00Z",
                   "enrollment_start": "2015-09-01T00:00:00Z", "enrollment_end": None,
                   "name": "Demo Course", "pacing": "instructor",
               })]

               purge.reset_mock()
               s, f = sync_course_runs([run])
               print(f"1st sync (data changed):   saved={s}  purges={purge.call_count}  -> expect saved=1 purges=1")

               purge.reset_mock()
               s, f = sync_course_runs([run])
               print(f"2nd sync (nothing changed): saved={s}  purges={purge.call_count}  -> expect saved=0 purges=0")

               raise RuntimeError("rollback")  # discard the throwaway data
       except RuntimeError:
           pass
   ```
4. Expected output:
   ```
   1st sync (data changed):   saved=1  purges=1  -> expect saved=1 purges=1
   2nd sync (nothing changed): saved=0  purges=0  -> expect saved=0 purges=0
   ```
   The second line is the fix in action: identical data → no save → no Fastly purge. Before this PR, the second line would show `purges=1` for a change that didn't happen.

New regression tests covering both fixes are included and run automatically in CI, so no extra steps are needed for those.

### Additional Context

- Kept `.save()` in the sync path (didn't switch to `.update()`) so `CourseRun.clean()` still runs — the `expiration_date` logic depends on it.
- The dirty check stops bumping `CourseRun.updated_on` (`auto_now`). Nothing reads it (no admin `list_display`/serializer), so this is safe to lose.
- Surrogate-key dedup and narrowing the receiver to MIT-Learn-rendered fields were considered but left out — they're largely redundant once the unnecessary saves are stopped at the source (dedup also carries a rollback footgun since Django has no `on_rollback`).

### DIFF
--- a/cms/models.py
+++ b/cms/models.py
@@ -1320,7 +1320,11 @@ class ProductPage(VideoPlayerConfigMixin, MetadataPageMixin):
             courseware_object = self.course
         elif self.is_program_page:
             courseware_object = self.program
-        if courseware_object:
+        # Only push the title down (and save) when it actually changed. An
+        # unconditional save here fires the courseware object's post_save
+        # Fastly purge signal on every page save, so publishing one page
+        # would otherwise emit several redundant purges of the same key.
+        if courseware_object and courseware_object.title != self.title:
             courseware_object.title = self.title
             courseware_object.save()
 
--- a/cms/models_test.py
+++ b/cms/models_test.py
@@ -664,6 +664,21 @@ def test_courseware_title_synced_with_product_page_title(test_course):
     assert courseware.title == updated_title
 
 
+@pytest.mark.parametrize("test_course", [True, False])
+def test_product_page_save_skips_courseware_save_when_title_unchanged(test_course):
+    """
+    Saving a product page without changing its title must not re-save the
+    linked courseware object, so it fires no redundant Fastly purge signal.
+    """
+    product_page = CoursePageFactory() if test_course else ProgramPageFactory()
+    courseware = product_page.course if test_course else product_page.program
+
+    with patch.object(type(courseware), "save") as mock_courseware_save:
+        product_page.save()
+
+    mock_courseware_save.assert_not_called()
+
+
 @pytest.mark.parametrize("flex_form_for_course", [True, False])
 def test_flexible_pricing_request_form_context(flex_form_for_course):
     """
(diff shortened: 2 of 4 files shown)


ChristopherChudzicki (approved):
👍  Looks good to me!

PR #3807, opened 2026-07-31
feat(b2b): add a service-scoped org-manager check endpoint

blarghmatey:
Related: https://github.com/mitodl/hq/issues/10594 (this endpoint is deleted once that lands)

Adds a small, service-scoped endpoint that answers one question: **does user X manage organization Y?**

```
GET /api/v0/b2b/service/organization-manager-check/
    ?sso_organization_id=<keycloak org uuid>&user_global_id=<keycloak sub>
→ 200 {"is_manager": true}
```

`OAuth2Authentication` + `TokenHasScope`, requiring a new `b2b:manager-check` scope that is only ever granted to a service Application. Backed by the existing `is_organization_manager()` in `b2b/models.py`.

#### Why this is needed

`ol-analytics-api` serves the B2B analytics dashboard and gates every endpoint on org-manager status. It currently gets that status by calling `/api/v0/b2b/manager/organizations/` and forwarding the end user's `X-Userinfo` header. **That has never worked** — in QA every request 502s, because MITx Online sees the call as anonymous and returns 403.

Two independent reasons a lighter fix isn't available:

1. **It can't reuse `ManagerOrganizationViewSet`.** `get_queryset` filters on `self.request.user`, so a service-authenticated call answers "manages nothing" and every user is denied.
2. **It can't forward the user's identity.** Our APISIX build's `openid-connect` plugin strips client-supplied identity headers before they reach an upstream, by design:

   ```lua
   -- Snapshot the client-supplied X-Access-Token (it doubles as a bearer
   -- input via get_bearer_access_token) and clear the four headers this
   -- plugin advertises as outputs so client-supplied values cannot bleed
   -- through to the upstream.
   ctx.openid_connect_client_x_access_token = core.request.header(ctx, "X-Access-Token")
   core.request.set_header(ctx, "X-Access-Token", nil)
   core.request.set_header(ctx, "X-Userinfo", nil)
   ```

   Bearer tokens aren't an alternative either: the plugin's bearer path is gated on `bearer_only or introspection_endpoint or public_key or use_jwks`, and the mitxonline route sets none of them, so a bearer token is never inspected at all. (Verified by probing the live RC endpoint — a bogus bearer token returns the identical 403 to sending no header.)

So the only remaining option is a service credential plus an explicit subject parameter.

#### Design notes

- **Its own module** (`b2b/views/v0/service.py`), not folded into `manager.py`. The user-facing dashboard path is completely untouched — no risk of regressing the live manager UI — and the eventual removal is a file deletion rather than unpicking a conditional out of a live authorization path.
- **Fails closed.** Unknown user, unknown org, or no membership row all return `is_manager: false` rather than 404. Besides being the safe default, a 404 would let a caller enumerate which org UUIDs and user IDs exist here.
- **A malformed `sso_organization_id` is a 400,** not a 500 from the ORM rejecting a non-UUID on a `UUIDField` lookup.
- **Marked temporary in three places** — module docstring, URL comment, and the scope entry in settings — each pointing at mitodl/hq#10594.

### How can this be tested?

```bash
pytest b2b/views/v0/service_test.py
```

13 tests covering: manager → true; member-but-not-manager → false (the case that motivates the whole endpoint, since the Keycloak token carries membership but not the manager flag); manager of a *different* org → false; unknown user/org/both → false; unauthenticated → rejected; **valid token without the scope → 403**; expired token → rejected; missing params → 400; malformed UUID → 400.

Verified locally against Postgres 15.16:
- `pytest b2b/views/v0/service_test.py` → 13 passed
- `pytest b2b/` → **510 passed, 1 skipped** (the skip is pre-existing), so nothing existing regressed
- `ruff check` and `ruff format --check` clean; pre-commit hooks pass
- `./manage.py generate_openapi_spec --fail-on-warn` regenerated; the spec diff is included

### Additional Context

**This is one of three PRs.** It can merge independently and is inert until a service Application is granted the scope:
- this PR — the endpoint
- mitodl/ol-analytics-api — consumes it via client-credentials (stacked on https://github.com/mitodl/ol-analytics-api/pull/21)
- mitodl/ol-infrastructure — Vault secret + Pulumi env wiring

**Pre-merge step:** an OAuth2 `Application` (confidential, `client-credentials`) needs creating per environment with the `b2b:manager-check` scope, and its client id/secret stored in Vault for ol-analytics-api. No migration or fixture is included — say the word if you'd prefer one over doing it through the admin.

The scope appears in all three of `openapi/specs/v{0,1,2}.yaml` because the generator emits every path into each version's spec; `v1.yaml` already contained 20 other `/api/v0/b2b/` paths before this change, so that's the existing convention rather than something introduced here.

### Checklist:
- [ ] Create the OAuth2 Application + `b2b:manager-check` scope grant in each environment, and store its credentials in Vault, before the ol-analytics-api side deploys

### DIFF
--- a/b2b/views/v0/service.py
+++ b/b2b/views/v0/service.py
@@ -33,7 +33,10 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from b2b.models import OrganizationPage, is_organization_manager
-from b2b.serializers.v0.service import OrganizationManagerCheckSerializer
+from b2b.serializers.v0.service import (
+    OrganizationManagerCheckSerializer,
+    ServiceDetailErrorSerializer,
+)
 from users.models import User
 
 # Distinct from the user-facing scopes in OAUTH2_PROVIDER["SCOPES"]: this one
@@ -81,7 +84,10 @@ class OrganizationManagerCheckView(APIView):
                 description="The user's Keycloak subject (User.global_id).",
             ),
         ],
-        responses=OrganizationManagerCheckSerializer,
+        responses={
+            200: OrganizationManagerCheckSerializer,
+            400: ServiceDetailErrorSerializer,
+        },
     )
     def get(self, request, *args, **kwargs):  # noqa: ARG002
         """Return the manager status for the (user, organization) pair."""
--- a/b2b/views/v0/service.py
+++ b/b2b/views/v0/service.py
@@ -1,0 +1,131 @@
+"""Service-to-service B2B views.
+
+TEMPORARY -- delete this module when org-manager status becomes visible in
+Keycloak (mitodl/hq#10594).
+
+Unlike everything in manager.py, these endpoints answer questions *about* a
+named user rather than about the caller. They exist because `is_manager`
+(b2b.models.UserOrganization) is curated only here, in the Django admin, and
+never reaches the Keycloak token -- so a downstream service that needs the
+flag has no way to learn it except by asking MITx Online.
+
+The one consumer today is ol-analytics-api, which gates its B2B analytics
+endpoints on org-manager status. It cannot reuse
+ManagerOrganizationViewSet: that viewset scopes its queryset to
+`self.request.user`, so a service-authenticated call would always answer
+"manages nothing". Nor can it forward the end user's identity -- the APISIX
+openid-connect plugin strips client-supplied X-Userinfo/X-Access-Token
+headers before they reach an upstream, by design, so a forwarded identity
+never survives the gateway.
+
+Hence a service credential plus an explicit subject parameter. Kept in its
+own module, off the user-facing manager surface, so that the deletion in
+mitodl/hq#10594 is a file removal rather than an unpicking of a live
+authorization path.
+"""
+
+import uuid
+
+from drf_spectacular.utils import OpenApiParameter, OpenApiTypes, extend_schema
+from oauth2_provider.contrib.rest_framework import OAuth2Authentication, TokenHasScope
+from rest_framework import status
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from b2b.models import OrganizationPage, is_organization_manager
+from b2b.serializers.v0.service import OrganizationManagerCheckSerializer
+from users.models import User
+
+# Distinct from the user-facing scopes in OAUTH2_PROVIDER["SCOPES"]: this one
+# is only ever granted to a service Application, never to a user-facing client.
+MANAGER_CHECK_SCOPE = "b2b:manager-check"
+
+
+class OrganizationManagerCheckView(APIView):
+    """Answer whether a given user manages a given organization.
+
+    Deliberately fails closed: an unknown user, an unknown organization, or a
+    user with no membership row all return `is_manager: false` rather than a
+    404. A 404 would let a caller enumerate which Keycloak organization UUIDs
+    and user IDs exist here, which is more than this endpoint needs to reveal
+    to answer its one question.
+    """
+
+    authentication_classes = [OAuth2Authentication]
+    permission_classes = [TokenHasScope]
+    required_scopes = [MANAGER_CHECK_SCOPE]
+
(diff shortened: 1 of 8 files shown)


b2b/views/v0/service.py
```diff
@@ -0,0 +1,131 @@
...
+        missing = [
+            name
+            for name, value in (
+                ("sso_organization_id", sso_organization_id),
+                ("user_global_id", user_global_id),
+            )
+            if not value
+        ]
+        if missing:
+            return Response(
```
dsubak, line 107: This might be overkill for a temporary fix but it might be good to indicate the 4xx responses in the generated schema. `DetailErrorSerializer` looks like it already matches the structure of the 400 responses here

b2b/views/v0/service_test.py
```diff
@@ -0,0 +1,265 @@
...
+        {
+            "sso_organization_id": str(org.sso_organization_id),
+            "user_global_id": "anything",
+        },
+    )
+
+    assert response.status_code in (
+        status.HTTP_401_UNAUTHORIZED,
+        status.HTTP_403_FORBIDDEN,
+    )
```
dsubak: Presumably this'll only ever be one of these values for this test (guessing 401?) - can we assert that directly vs asserting that it's one of two response codes?

b2b/views/v0/service_test.py
```diff
@@ -0,0 +1,265 @@
...
```
dsubak: Same question as https://github.com/mitodl/mitxonline/pull/3807/changes#r3706467210

dsubak (changes requested):
Few small questions, but nothing show stopping - think we're in business once addressed!

b2b/views/v0/service.py
blarghmatey: Done in 08712e21 — added a ServiceDetailErrorSerializer (same shape as DetailErrorSerializer in manager.py, kept local to this module so its deletion stays a plain file removal) and documented the 400 response via responses={200: ..., 400: ServiceDetailErrorSerializer}. Regenerated the v0/v1/v2 specs and confirmed via scripts/test/openapi_spec_check.sh (against a real Postgres) that they match the generator's output exactly.

b2b/views/v0/service_test.py
blarghmatey: Good catch, and yes it's deterministic — traced it through DRF and django-oauth-toolkit: OAuth2Authentication.authenticate() returns None (not an exception) for both no-token and invalid/expired-token cases, so request.successful_authenticator stays unset. APIView.permission_denied() checks exactly that condition (request.authenticators and not request.successful_authenticator) and raises NotAuthenticated, i.e. 401, before TokenHasScope is ever consulted -- 403 only happens when authentication succeeds but the scope check fails (test_requires_the_manager_check_scope below). Tightened to assert response.status_code == status.HTTP_401_UNAUTHORIZED in 08712e21 and confirmed by actually running the test against a real Postgres+Redis stack, not just reasoning about it.

b2b/views/v0/service_test.py
blarghmatey: Same fix and same reasoning as the sibling comment on test_requires_authentication -- an expired token also leaves OAuth2Authentication.authenticate() returning None, so it hits the identical NotAuthenticated (401) path rather than PermissionDenied (403). Tightened to assert response.status_code == status.HTTP_401_UNAUTHORIZED in 08712e21, verified by running the test.

blarghmatey: Same fix and same reasoning as the sibling comment on test_requires_authentication -- an expired token also leaves OAuth2Authentication.authenticate() returning None, so it hits the identical NotAuthenticated (401) path rather than PermissionDenied (403). Tightened to assert response.status_code == status.HTTP_401_UNAUTHORIZED in 08712e21, verified by running the test.

dsubak (approved):
LGTM

PR #3810, opened 2026-08-03
Add learner refund request workflow

cp-at-mit:
https://github.com/mitodl/hq/issues/12553

Introduce a RefundRequest model, serializer, migration, and v0 API endpoint so learners can submit refund requests for their own fulfilled non-B2B orders. This also exposes order refund eligibility in order serializers and sends customer service an email notification using a configurable refund request address.

- New endpoint: POST /api/v0/orders/refund-requests/
- New field: refund_eligible (boolean) on all Order API responses
- Email notification: sent to MITX_ONLINE_REFUND_REQUEST_EMAIL (defaults to EMAIL_SUPPORT) via Celery when a request is submitted

### How can this be tested?
- Fetch any fulfilled order and confirm the new field appears, `refund_eligible`.
- Submit a refund request to the new API endpoint: `http://mitxonline.odl.local:8013/api/v0/orders/refund-requests/`.
- Sample POST body:
```
{
    "order": <fulfilled_order_id>,
    "refund_reason": "course_not_as_expected",
    "refund_reason_text": "The course content was not at the level I expected.",
    "consent_given": true
  }
```
- A RefundRequest row should exist in the database and a Celery task should have been queued to send the notification email.


ecommerce/models.py
```diff
@@ -1202,3 +1223,39 @@ class DiscountRedemption(TimestampedModel):
...
+        "enrolled_in_another_course",
+        "I enrolled in another course",
+    )
+    COURSE_NOT_AS_EXPECTED = "course_not_as_expected", "Course is not what I expected"
+    TECHNICAL_DIFFICULTIES = "technical_difficulties", "Technical difficulties"
+    FINANCIAL_REASONS = "financial_reasons", "Financial reasons"
+    OTHER = "other", "Other"
+
+
+class RefundRequest(TimestampedModel):
```
jkachel, line 1343: I think a status field would be useful - this can just be `Pending`, `Approved`, `Denied` at this stage.

ecommerce/serializers/__init__.py
```diff
@@ -899,3 +912,38 @@ def get_purchaser(self, instance):
     class Meta:
         fields = ["purchaser", "lines", "coupon", "order", "receipt"]
         model = models.Order
+
+
+class RefundRequestSerializer(serializers.ModelSerializer):
```
jkachel: Why here and not in `v0`? The APIs that use this serializer are in `views/v0` and this is new work so seems a more appropriate place for this serializer.

jkachel (changes requested):
Works OK. Requesting a handful of nit-picky changes since this will need a bit of work for conflict resolution (def will need at least some migration renumbering):

- Would be good to have an admin set up for the refund requests; can't see them in the app as it is. This can be a super simple ReadOnly one at this stage.
- Would also be good to have a couple basic tests for the API too. 
- Maybe less of a nit but I don't see the `refund_eligible` field on the order history API endpoints. I do see it on the receipt API, though.

jkachel (approved):
LGTM 👍

PR #3558, opened 2026-05-07
Adding the Listed Price field to ProductPage

annagav:
### What are the relevant tickets?
Fix partially https://github.com/mitodl/hq/issues/11119
### Description (What does it do?)
Add a new Listed Price field
Populate with default value product.price

### How can this be tested?
Checkout the branch 
Run migrations
go to cms
check the new Listed Price field and verify it was populated correctly.

Screenshot 2026-05-11 at 1 36 20 PM

PR #3577, opened 2026-05-13
remove price on ProductPage

annagav:
### What are the relevant tickets?
Related to https://github.com/mitodl/hq/issues/11119

### Description (What does it do?)

remove price on ProductPage

### How can this be tested?
Nothing should break

PR #3813, opened 2026-08-04
Remove the non-functional CMS price field from product pages

AhtishamShahid:

Closes https://github.com/mitodl/hq/issues/6123

Related context, worth skimming to see why this is safe:

- **hq#2844** (the Unito mirror of the linked `mitxonline-issues#471` discussion) — course teams hand-maintained this price text and it drifted from what pages actually displayed. The Core meeting decided non-program course price info comes from BMS/Quickbase, set by Engineering.
- **hq#9910** — closed with one acceptance criterion left undone: *"it seems like CoursePage.price is unused now, so it could be removed."* That's this PR.
- **hq#11119** → #3558 — added `ProgramPage.list_price`, the numeric field that replaces the hand-typed program price.

---

The CMS product-page editor had **three** price inputs stacked on top of each other. Only two of them did anything.

| Field | Who reads it | Kept? |
| --- | --- | --- |
| `price` (StreamField, **required**) | nothing on course pages; one line of the legacy program info box | ❌ **removed** |
| `min_price` / `max_price` | MIT Learn catalog + `/api/v2/courses`, `/api/v2/programs` | ✅ kept |
| `list_price` (programs) | MIT Learn product pages, program certificate price | ✅ kept |

So editors were **forced** to fill a field (it's been `blank=False` since migration `0014`) whose value no course page ever showed. This PR deletes it and moves the one remaining reader onto `list_price`.

---

### Demo — watch this first, then reproduce it below

23s, silent. It shows both halves of the change with real captures from a local instance, and ends on the exact three steps to check it yourself.

https://github.com/user-attachments/assets/2d46688b-cb8c-413c-832d-e0eabc30d783

| Timestamp | What you're looking at |
| --- | --- |
| 0:03 | **CMS editor, before** — `Price *` (required) sitting directly above `Min price *` &nbsp;·&nbsp; `/cms/pages/15/edit/` |
| 0:08 | **CMS editor, after** — `Max weeks` runs straight into `Min price`; the field is gone |
| 0:11 | **Program about page, before vs after** — `Certificate track: $1,000 - $4,000 for the program certificate` (typed by hand into the CMS) becomes `Certificate track: $1,000.00` (from `list_price`) &nbsp;·&nbsp; `/programs/program-v1:MITx+DEDP/` |
| 0:16 | **What stays behind** — the two price sources that survive |
| 0:19 | **Reproduce it in three steps** — pause here; the commands and URLs are on screen |

---

### How can this be tested?

The three steps from the end of the video, spelled out.

**1 — Get the branch onto a running stack**

```bash
git checkout remove-cms-price-field
docker compose exec web python manage.py migrate
```

**2 — Open the CMS editor: no Price field, and it still saves**

http://mitxonline.odl.local:8013/cms/pages/15/edit/ (any product page works — substitute your own page id)

- there is **no** `Price` / `Price details` block
- `Min price *`, `Max price *` and (on programs) `List price` are still there
- click **Save draft** → *"Page … has been updated"*

That last click is the real check: the removed field was **required**, so a clean save proves nothing is still validating against it.

**3 — Open the program about page: the price still renders**

http://mitxonline.odl.local:8013/programs/program-v1:MITx+DEDP/

The right-hand info box reads `Certificate track: $1,000.00`, now sourced from `list_price`. Needs a `list_price` on the page and a purchasable run — otherwise the certificate block hides itself.

**API check**

```bash
curl -s http://mitxonline.odl.local:8013/api/v2/programs/ | jq '.results[0].page | keys'
```

`price` is absent; `list_price` is present.

**Automated**

```bash
docker compose exec web pytest cms courses ecommerce -n logical   # 1610 passed
docker compose exec web python manage.py makemigrations --check --dry-run
./scripts/test/openapi_spec_check.sh
```

---

### What changed

**Data loss?** None that matters — migrations `0047` and `0055` already mined this text into `min_price`/`max_price` with a regex. Wagtail page revisions keep the old JSON harmlessly.

---

### Notes for reviewers

**This is a breaking API change**, deliberately. `price` disappears from the `ProgramPage`, `CoursePageItem` and `ProgramPageItem` schemas, which affects `/api/v1/programs`, `/api/v2/programs`, `program_enrollments`, the v0 ecommerce basket/order responses and the Wagtail `/api/v2/pages/` endpoints. `oasdiff breaking --fail-on ERR` passes with the ignore entries added here; I checked each of the 17 against the removal rather than blanket-ignoring.

**Downstream follow-up (not blocking):** mit-learn's `frontends/api/src/mitxonline/test-utils/factories/pages.ts` imports `PriceItem` and sets `price: [priceItem()]`. That breaks the next time they regenerate `@mitodl/mitxonline-api-axios`. Its ETL (`learning_resources/etl/mitxonline.py`) only reads `min_price`/`max_price`, so ingestion is unaffected. (Work in progress on that )


AhtishamShahid:
Downstream fix is now open as a draft: mitodl/mit-learn#3725 — it removes the `price` references from the two mitxonline page factories in that repo.

It is deliberately a draft: `price` is currently a **required** property, so the fix fails typecheck against mit-learn's pinned client (`2026.7.22`) and against the already-published `2026.8.3`, which still ships `PriceItem` because this PR isn't merged yet. Merge order is: this PR → regenerate + publish the client → bump it in mit-learn → merge that one (or just fold its 20-line diff into the bump PR).

Validated locally by generating a client from this branch with `mitxonline-api-clients/scripts/local-generate.sh` (nothing pushed) and running `tsc --noEmit` in mit-learn's `frontends/api`: 0 errors on today's baseline, 5 with this branch's client, 1 after the fix — and that last 1 is `sso_organization_id` from #3789, unrelated to this PR.

AhtishamShahid:
Closed in favour of https://github.com/mitodl/mitxonline/pull/3577

PR #3818, opened 2026-08-05
Adding missing profile fields to profile api

annagav:
Related to https://github.com/mitodl/hq/issues/12528

 - Expose a compliance_missing_fields read-only field on the users/me (profile) API, listing which profile fields still need to be filled in before a CyberSource export compliance check can run for the 
  user (e.g. first_name, country, state, postal_code).                                                                                                                                                     
  - Refactor compliance/api.py to share the missing-field logic between the existing validation path (_validate_bill_to_fields, which raises ExportComplianceDataError) and a new public                   
  get_missing_export_compliance_fields(user) helper, so both stay in sync with a single source of truth.                                                                                                   
  - Map CyberSource's internal bill-to field names (address1, locality, administrative_area) back to the corresponding LegalAddress/User field names (street_address_1, city, state) via                   
  BILL_TO_FIELD_TO_PROFILE_FIELD, so the API returns names the frontend can map directly to profile form fields.                                                                                           
  - Regenerate openapi/specs/{v0,v1,v2}.yaml to include the new field.       
  

### How can this be tested?
Go to  http://mitxonline.odl.local:8013/api/users/me

### DIFF
--- a/compliance/api.py
+++ b/compliance/api.py
@@ -95,8 +95,22 @@ def _normalize_administrative_area(
     return normalized_state
 
 
-def _validate_bill_to_fields(user, bill_to: dict[str, str]) -> None:
-    """Raise a clear error when required CyberSource bill-to fields are missing."""
+# Maps CyberSource bill-to field names to the profile field names an API
+# consumer should show the user as missing.
+BILL_TO_FIELD_TO_PROFILE_FIELD = {
+    "first_name": "first_name",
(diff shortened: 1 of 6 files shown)


rhysyngsun (commented):
Code looks good, I'll do a functional testing pass first thing tomorrow.

rhysyngsun (approved):
LGTM

PR #3820, opened 2026-08-05
Remove unused fetch_enrollment helper

cp-at-mit:
NA

Deleted `EnrollmentChangeCommand.fetch_enrollment` from `courses/management/utils.py` and removed its dedicated test. This cleans up dead enrollment lookup logic that is no longer part of the command flow and keeps the test suite aligned with current behavior.

### How can this be tested?
Tests should pass.


annagav (approved):
👍

PR #3823, opened 2026-08-05
fix issues with orders api

gumaerc:
https://github.com/mitodl/hq/issues/12684

Fixes three things that were fundamentally broken about the v0 Orders API, so we
don't have to work around them in the new receipt page.

1. **`purchaser` was typed as an array but returns an object.** `get_purchaser` was
   annotated `many=True`, so generated clients declared
   `Array<ExtendedLegalAddress>` for a single object — unusable without casting
   around the type.

2. **`orders/history` returned a different shape depending on whether `limit` was
   passed.** `OrderHistoryViewSet` had no `default_limit`, so DRF skipped pagination
   entirely and returned a bare list; with any `limit` it returned the documented
   `{count, next, previous, results}` envelope.

3. **The receipt endpoint 500s when a product's courseware object has been deleted.**
   `TransactionLineSerializer` dereferenced `content_object.start_date` without
   checking that the `GenericForeignKey` resolved.

Scoped to v0, which is what the generated clients target. The unversioned
`/api/orders/...` duplicate isn't in the spec and produces no TypeScript, so it's
left alone.

Only (1) changes the OpenAPI spec (`array` → `allOf`). Response bodies are
unchanged, so the oasdiff findings are recorded as a spec correction.

### How can this be tested?

**Data you need:** one fulfilled order on your own account. If you don't have one,
make a 100%-off code and use it to create a verified enrollment, which fulfills an
order as a side effect:

```
./manage.py generate_discount_code --count 1 --prefix RECEIPT_TEST_ \
    --amount 100 --discount-type percent-off --payment-type staff

./manage.py create_verified_enrollment --user=<your email> \
    --run=<course_run_courseware_id> --code=<the generated code>
```

Note the order id from `/api/v0/orders/history/` for the checks below.

**(2) — order history shape.** Hit the endpoint both ways:

```
GET /api/v0/orders/history/            # on main: bare JSON array
                                       # here:    {count, next, previous, results}
GET /api/v0/orders/history/?limit=10   # envelope either way
```

**(1) — purchaser type.** `GET /api/v0/orders/receipt/<order id>/` and confirm
`purchaser` is a single object, not a list. That was always true at runtime; the fix
is that `openapi/specs/v0.yaml` now agrees, so a regenerated client types it as an
object. Worth a glance at the spec diff to confirm.

**(3) — dangling courseware object.** Delete the course run behind the order's
product, then reload the receipt:

```
./manage.py shell -c "
from courses.models import CourseRun
CourseRun.objects.filter(courseware_id='<course_run_courseware_id>').delete()
"
```

`GET /api/v0/orders/receipt/<order id>/` 500s on main and returns 200 here, with
null `start_date`/`end_date` on the line. (Do this on a throwaway run — it also
removes the enrollment.)


jkachel (approved):
LGTM 👍

PR #3824, opened 2026-08-05
Fix OpenAPI spec partitioning

rhysyngsun:

N/A

This is a long-standing issue I tried to fix 6 months ago and it lead to some whack-a-mole regressions. Things seem to be better now so trying to get this addressed again. Up until now our openapi specs contain all the APIs because the namespacing strategy was configured with an incorrect key and so ignored. This fixes that and also patches around a few oddities of how some versioned APIs are defined by using explicit version classes.

### How can this be tested?

The APIs defined in each of the versioned schemas should be limited to APIs for that version and no extraneous non-versioned APIs should appear either.


cp-at-mit (approved):

rhysyngsun:
Will rebase and merge on Monday.

PR #3836, opened 2026-08-10
fix(scim): send name.givenName/familyName in outbound SCIM payload

shaidar:
N/A

Users migrated into Keycloak via SCIM ended up with blank `firstName`/`lastName` ("General" section in the Keycloak admin console) — and, downstream, blank names in Learn too, since Keycloak's own outbound sync just propagates whatever it has.

Root cause: `LearnUserAdapter.to_dict()` (`users/adapters.py`) never included a `name` object in the SCIM payload — only `displayName`, `userName`, `emails`, `active`. Confirmed by decompiling the installed `scim-for-keycloak` Keycloak plugin jar that:
- `name.givenName`/`name.familyName` on an inbound SCIM request map directly (hardcoded) to Keycloak's core `firstName`/`lastName` attributes.
- `displayName` is never read by the plugin for that purpose at all — sending it has zero effect on Keycloak's stored name.

So the only fix is sending the structured `name.givenName`/`name.familyName` fields, which requires splitting whatever combined name data we have (Keycloak has no "full name" field to hand a single string to).

**Fix**: `LearnUserAdapter._resolve_name()` resolves `(given_name, family_name)` with three tiers of confidence:
1. `legal_address.first_name`/`last_name`, if both are set — real structured data.
2. `User.name`, split heuristically (last whitespace token = family name, remainder = given name) — a fallback for users who only came through the edX migration and never had a `legal_address` name recorded. No naive split is correct for every name (breaks on single-name accounts, multi-word surnames, non-Western conventions), but it's the best data available.
3. Neither available — blank, same as today.

The tier-2 derived split is **never persisted back** onto `legal_address` — that model is used for SDN/denied-persons compliance screening, and writing a heuristic guess into a field compliance may rely on for an accurate legal name is a real risk, not just a data-quality nitpick. It's computed fresh in `to_dict()` for the outbound payload only.

`ATTR_MAP` also gained entries for `name.givenName`/`name.familyName` (and flat `givenName`/`familyName`, matching upstream `django_scim`'s own default `UserFilterQuery.attr_map` convention) so SCIM filter/PATCH path resolution works for the new fields, and `from_dict()` writes inbound `name.givenName`/`name.familyName` directly to `legal_address` (real data from an external SCIM client, never a derived guess).

**Also included**: `remediate_keycloak_user_names`, a standalone management command for accounts that were already migrated before this fix and are stuck with blank/stale names in Keycloak. `sync_users_to_scim_remote` only ever *creates* users that don't already exist remotely — it has no update/PATCH path — so re-running the sync does nothing for already-existing accounts. This command talks to Keycloak's Admin API directly instead (reusing the existing `b2b/keycloak_admin_api.py` `KeycloakAdminClient`, paginating users 100 at a time rather than one request per user):

- Default mode is dry-run: reports every candidate and what would change, writes nothing.
- `--apply` performs the actual `PUT` of `firstName`/`lastName`, using the same `_resolve_name()` tiers as the adapter so a backfill and a fresh migration always agree on what a user's name should be.
- `--limit` caps how many get patched in a single `--apply` run, for a cautious first pass against production.
- After each patch, re-fetches the user to confirm the stored value now matches, rather than trusting a 2xx.
- Users with no name data anywhere in mitxonline are reported separately as unpatchable, left untouched.

### How can this be tested?
`users/adapters_test.py` was rewritten — the existing test imported the base `mitol.scim.adapters.UserAdapter` (a re-export), not the actual `LearnUserAdapter` subclass used in production, so it never exercised the real adapter at all. New tests cover all three `_resolve_name()` tiers in `to_dict()`, and `from_dict()` writing tier-1 data back to `legal_address`.

`users/management/tests/remediate_keycloak_user_names_test.py` covers: dry-run reporting without patching, `--apply` patching + re-verification, already-up-to-date users being skipped, unpatchable (no name data) users being reported separately, `--limit` capping writes, unmatched Keycloak users being ignored, and pagination across multiple pages.

Ran the full `users/` suite locally with a fresh test DB (`uv run pytest users/ --create-db`): 138 passed, no regressions. Ran `ruff check` on all changed files with no findings (aside from one pre-existing, unrelated missing-module-docstring lint on `users/adapters.py` that predates this change).

### Additional Context
This is the independent, mergeable half of a larger fix. A companion PR in `mitodl/ol-django` (https://github.com/mitodl/ol-django/pull/544) makes `sync_users_to_scim_remote` return its results instead of `None`, which a follow-up mitxonline PR (a new `migrate_and_sync_users` orchestrator command, replacing the old manual `migrate_edx_data` + ad hoc sync script process) depends on for post-sync verification. That follow-up isn't included here since it would call code that doesn't exist yet in the currently-published `mitol-django-scim` — it'll be opened as a draft PR blocked on the ol-django release.

### DIFF
--- a/users/adapters.py
+++ b/users/adapters.py
@@ -140,13 +140,16 @@ class LearnUserAdapter(UserAdapter):
         # Inbound name.givenName/familyName always writes to legal_address
         # directly - this is real data from an external SCIM client, never
         # a derived guess (see _resolve_name's tier 2, which is outbound-only).
+        # An explicit JSON null (as opposed to an absent key) is treated the
+        # same as absent - legal_address.first_name/last_name are non-nullable
+        # CharFields, so assigning None would raise an IntegrityError on save.
         name = d.get("name") or {}
-        self.legal_address.first_name = name.get(
-            "givenName", self.legal_address.first_name
-        )
-        self.legal_address.last_name = name.get(
-            "familyName", self.legal_address.last_name
-        )
+        given_name = name.get("givenName")
+        if given_name is not None:
+            self.legal_address.first_name = given_name
+        family_name = name.get("familyName")
+        if family_name is not None:
+            self.legal_address.last_name = family_name
 
     def _save_related(self):
         self.user_profile.user = self.obj
--- a/users/adapters.py
+++ b/users/adapters.py
@@ -140,15 +140,17 @@ class LearnUserAdapter(UserAdapter):
         # Inbound name.givenName/familyName always writes to legal_address
         # directly - this is real data from an external SCIM client, never
         # a derived guess (see _resolve_name's tier 2, which is outbound-only).
-        # An explicit JSON null (as opposed to an absent key) is treated the
-        # same as absent - legal_address.first_name/last_name are non-nullable
-        # CharFields, so assigning None would raise an IntegrityError on save.
+        # An absent key, an explicit JSON null, or a blank/whitespace-only
+        # string are all treated as "no value provided" and leave the
+        # existing value alone - legal_address feeds SDN compliance
+        # screening, so a client silently sending an empty name should not
+        # be able to blank out a previously-valid one.
         name = d.get("name") or {}
-        given_name = name.get("givenName")
-        if given_name is not None:
+        given_name = (name.get("givenName") or "").strip()
+        if given_name:
             self.legal_address.first_name = given_name
-        family_name = name.get("familyName")
-        if family_name is not None:
+        family_name = (name.get("familyName") or "").strip()
+        if family_name:
             self.legal_address.last_name = family_name
 
     def _save_related(self):
--- a/users/adapters.py
+++ b/users/adapters.py
@@ -21,6 +21,10 @@ class LearnUserAdapter(UserAdapter):
         ("active", None, None): "is_active",
         ("userName", None, None): "username",
         ("fullName", None, None): "name",
+        ("name", "givenName", None): "legal_address__first_name",
+        ("givenName", None, None): "legal_address__first_name",
+        ("name", "familyName", None): "legal_address__last_name",
+        ("familyName", None, None): "legal_address__last_name",
     }
 
     obj: "User"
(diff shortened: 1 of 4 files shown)


users/adapters.py
```diff
@@ -52,16 +56,59 @@ def display_name(self):
...
+        if self.legal_address.first_name and self.legal_address.last_name:
+            return self.legal_address.first_name, self.legal_address.last_name
+
+        given_and_family_name_parts = 2
+        full_name = (self.obj.name or "").strip()
+        if full_name:
+            parts = full_name.rsplit(None, 1)
+            if len(parts) == given_and_family_name_parts:
+                return parts[0], parts[1]
+            return parts[0], ""
```
rhysyngsun: This can be wrong for a lot of names, if we don't already have the values in legal_address we should just return `("", "")`

users/adapters.py
```diff
@@ -52,16 +56,59 @@ def display_name(self):
...
+
+        return "", ""
+
     def to_dict(self):
         """
         Return a ``dict`` conforming to the SCIM User Schema,
         ready for conversion to a JSON object.
         """
+        given_name, family_name = self._resolve_name()
         return {
```
rhysyngsun, line 84: My understanding is that for users imported from edxorg we don't have first/last name, we only have full name. I think the real root of the missing names is that we aren't populating `fullName` into keycloak from `User.name`. You can see the reverse of this happening [here](https://github.com/mitodl/mitxonline/pull/3836/changes#diff-303f39fee02d9fcf3d90398703a506d24e431a072a0c3ca9ee60449ce88598e1R138)

users/management/commands/remediate_keycloak_user_names.py
```diff
@@ -0,0 +1,181 @@
...
+        patched, would_patch, unpatchable, up_to_date = [], [], [], []
+        patch_count = 0
+
+        for keycloak_user in self._paginate_keycloak_users(client):
+            user = mitxonline_users_by_scim_id.get(keycloak_user.id)
+            if user is None:
+                continue  # not a user we can trace back to mitxonline
+
+            adapter = LearnUserAdapter(user)
+            given_name, family_name = adapter._resolve_name()  # noqa: SLF001
```
rhysyngsun, line 90: I mentioned in my other comment, I don't think we have first/last name for edxorg migrated users, we only have full name, so I think this script needs to populate that too. No harm in copying over first/last if we happen to have it, but I think for most users we won't.

rhysyngsun (changes requested):

users/adapters.py
shaidar: Fixed in 1c126107 — `_resolve_name()` no longer guesses a given/family split from `self.obj.name`. It now only returns `legal_address.first_name/last_name` when both are on file; otherwise it returns `("", "")`.

users/adapters.py
shaidar: Good catch — you're right that the root cause was `fullName` never being sent outbound. Fixed in 1c126107: `to_dict()` now sends `"fullName": self.obj.name` on every outbound payload, mirroring the existing inbound `fullName -> User.name` mapping in `from_dict()`. So even when we can't produce a reliable given/family split (e.g. edxorg-migrated users with only a full name), Keycloak still gets the correct full name via that attribute.

users/management/commands/remediate_keycloak_user_names.py
shaidar: Agreed, and fixed in 1c126107. The command now separately resolves and patches the `fullName` Keycloak attribute from `User.name`, merging it into the existing `attributes` map (the admin API PUT does not deep-merge, so unrelated custom attributes are preserved). firstName/lastName are still only patched when `legal_address` has both parts — for users where we only have a full name, only `fullName` gets patched and the existing firstName/lastName are left alone rather than guessed at or blanked.

shaidar:
Addressed all three review threads in 1c126107: `_resolve_name()` no longer heuristically splits `User.name` into given/family (returns blank instead of guessing when `legal_address` doesn't have both parts on file), and the outbound SCIM payload now sends `fullName` directly from `User.name` — the actual root cause of missing names for edxorg-migrated users. `remediate_keycloak_user_names` follows the same rule: patches `fullName` separately from firstName/lastName, merging into Keycloak's existing attributes map so other custom attributes aren't clobbered. Tests updated/added and passing (21/21) via `docker compose run --rm web pytest`.

users/adapters.py
```diff
@@ -52,16 +56,41 @@ def display_name(self):
...
+        string can't be reliably split into given/family (breaks on
+        single-name accounts, multi-word surnames, non-Western conventions),
+        so when legal_address doesn't have both parts this returns
+        ("", "") rather than guessing. The full name itself is still sent
+        outbound on its own via the "fullName" attribute (see to_dict), so
+        nothing is lost for users - most commonly those who only came
+        through the edX migration - who have a full name but no
+        legal_address split on file.
+        """
+        if self.legal_address.first_name and self.legal_address.last_name:
```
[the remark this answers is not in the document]
shaidar: Verified — not a real issue. `LegalAddress.user` is a `OneToOneField(User, related_name="legal_address")`, so `User.legal_address` is a `ReverseOneToOneDescriptor`. Per Django's implementation, when there's no related row — whether freshly queried or pre-cached to `None` by `select_related()`'s LEFT OUTER JOIN finding no match — the descriptor's `__get__` explicitly raises `RelatedObjectDoesNotExist` rather than returning `None`:

```python
if rel_obj is None:
    raise self.RelatedObjectDoesNotExist(...)
```

That exception is dynamically created as a subclass of both `LegalAddress.DoesNotExist` **and** `AttributeError`, which is exactly what `getattr(self.obj, "legal_address", LegalAddress())` is designed to catch — so the default is applied correctly, not bypassed.

Confirmed empirically (created a user, deleted its `legal_address`, refetched with `select_related("legal_address")`):
- `user.legal_address` raises `User.legal_address.RelatedObjectDoesNotExist` (MRO includes `AttributeError`)
- `getattr(user, "legal_address", LegalAddress())` returns a fresh unsaved `LegalAddress()` as intended
- `LearnUserAdapter(user)._resolve_name()` runs cleanly and returns `("", "")`, no crash

No change needed here.

rhysyngsun (approved):
LGTM

PR #3840, opened 2026-08-11
Honor the enrollment mode in the Open edX enrollment webhook

Anas12091101:
## What are the relevant tickets?

mitodl/hq#12748

## Description (What does it do?)

`edx_enrollment_webhook` now forwards the payload's optional `mode` to `create_local_enrollment` instead of always enrolling at the default audit mode. Absent or empty values fall back to the default, so the existing course access role callers are unaffected.

Pairs with the `ol_openedx_events_handler` change in mitodl/open-edx-plugins, which starts sending `mode`.

## How can this be tested?

`pytest openedx/views_test.py` — 33 pass.

Or POST to `/api/openedx_webhook/enrollment/` with `{"email": ..., "course_id": ..., "mode": "verified"}` and confirm the resulting `CourseRunEnrollment.enrollment_mode` is `verified`. Existing enrollments keep their mode, since `create_local_enrollment` uses `get_or_create`.

### DIFF
--- a/openedx/views.py
+++ b/openedx/views.py
@@ -16,6 +16,7 @@ from rest_framework.response import Response
 
 from courses.api import create_local_enrollment, generate_course_run_certificates
 from courses.models import CourseRun, CourseRunCertificate
+from openedx.constants import EDX_DEFAULT_ENROLLMENT_MODE
 from users.models import User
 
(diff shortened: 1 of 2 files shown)


asadali145 (approved):
LGTM!

PR #3844, opened 2026-08-11
Skip HubSpot deal sync for program-enrolled courses

cp-at-mit:
NA

Add a check in sync_hubspot_deal to skip syncing orders where the purchaser is already enrolled in a program containing the purchased course run. This prevents duplicate/misleading deals in HubSpot when a user buys an individual course that's part of a program they're already enrolled in.

### How can this be tested?
- Enroll into a program and pay for it
- Enroll into one of the courses in the program
- Verify that no deal is created for the course enrollment but a deal is created for the program purchase.

### DIFF
--- a/hubspot_sync/task_helpers.py
+++ b/hubspot_sync/task_helpers.py
@@ -48,10 +48,13 @@ def _order_is_for_program_enrolled_course(order: Order) -> bool:
         purchased_object = line.purchased_object
         if not purchased_object:
             continue
-        if isinstance(purchased_object, CourseRun) and ProgramEnrollment.objects.filter(
-            user=order.purchaser,
-            program__all_requirements__course=purchased_object.course,
-        ).exists():
+        if (
+            isinstance(purchased_object, CourseRun)
+            and ProgramEnrollment.objects.filter(
+                user=order.purchaser,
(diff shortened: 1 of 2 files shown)


hubspot_sync/task_helpers.py
```diff
@@ -41,6 +42,23 @@ def sync_hubspot_user(user: User):
             )
 
 
+def _order_is_for_program_enrolled_course(order: Order) -> bool:
+    """Return True if any line is a course run belonging to a program the purchaser is already enrolled in."""
+    for line in order.lines.all():
```
dsubak: I'm a bit hazy on the data model so forgive me if this is a bad question - generally speaking, how many line items do we expect to have per order?

As written, this could execute up to one query per course run line item. If we do have orders w/ large numbers of line items I'm wondering it's worth iterating through line items to gather the list of CourseRuns from the order and then running a single ProgramEnrollment query for all courses on the order. That'd mean we only do one query regardless of the number of course run line items.

cp-at-mit: So I think right now we only support a single line item being added by the user, but we might as well make this efficient like you're saying in case that requirement changes.

dsubak (approved):

PR #3847, opened 2026-08-12
Enforce one product per purchasable object

cp-at-mit:
https://github.com/mitodl/hq/issues/11729

Updates `Product` uniqueness to always enforce a single product per `(object_id, content_type)` pair, instead of allowing duplicates when one is inactive. Adds a migration to replace the old conditional constraint and updates the model test to cover both active and inactive duplicate creation attempts.

### How can this be tested?
Create a product, then attempt to create a second active product for the same object.  This should not work even if the first product is deactivated.

### DIFF
--- a/ecommerce/migrations/0044_alter_product_unique_purchasable_object.py
+++ b/ecommerce/migrations/0044_alter_product_unique_purchasable_object.py
@@ -1,0 +1,21 @@
+from django.db import migrations, models
+
+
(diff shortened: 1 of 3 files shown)


annagav (commented):
Should we run a migration to resolve any duplicates first before setting the constraint?

annagav (approved):
👍

PR #3852, opened 2026-08-13
Regen dataclass

dsubak:
https://mit-office-of-digital-learning.sentry.io/issues/7462913757/?environment=production&project=5864687&query=is%3Aunresolved%20RealmRepresentation&referrer=issue-stream

Regenerate dataclass to fix abovelinked sentry. This follows the same steps as [an earlier PR](https://github.com/mitodl/mitxonline/pull/3711) but with a slightly modified command

### How can this be tested?
ONGOING - at the moment I've only reproduced the issue in prod and taken steps to cut the PR; may need to change the below as they're hypothetical.

- Make sure you're running against the latest version of keycloak according to the instructions [here](https://github.com/mitodl/mitxonline/pull/3711)
- From a b2b admin dashboard that you've got access to, issue an enrollment code to a new user
- Sign into learn as that user, redeem the code to enroll in the contract.
- It should not throw an error.

Performing the same sets of steps on main should result in an error the first time you attempt to redeem the code.

### Additional Context
https://mitodl.slack.com/archives/C03K5HYGPT9/p1786644985018899


jkachel (approved):
LGTM 👍 

I think your command is slightly more correct than mine was, I probably had a downloaded copy of the spec when I ran it last. Should be either `--input` or `--url`, not both.

